PR #685, opened 2014-04-28
TTT: Fixed not being able to shoot a weapon when equipped while being held by a magneto stick

Bo98:
When a weapon is equipped while it's being held by a magneto stick, errors occur. The player is unable to shoot their weapon and the ammo count is potentially incorrect or not visible at all. The shotgun also spits errors when shot.

The magneto stick calls its Reset function after a weapon is equipped. Because there was no owner when the weapon was first picked up with the magneto stick, the owner is set to nil. Reequipping the weapon worked around this issue.

This fix uses the WeaponEquip hook which is called before the owner it set. This gives us time to call SWEP:Drop() and reset the owner before the new owner is set.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
@@ -123,6 +123,10 @@ function SWEP:Reset(keep_velocity)
       if (not keep_velocity) and (no_throw:GetBool() or self.EntHolding:GetClass() == "prop_ragdoll") then
          KillVelocity(self.EntHolding)
       end
+
+      if self.EntHolding:IsWeapon() then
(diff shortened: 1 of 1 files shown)


Kefta:
@svdm 9 months later...

YourStalker:
12 months later... This has been on my server for a year and hasn't caused any problems. A year of testing :P

svdm:
Sorry for leaving this hanging for so long, I've been meaning to spend some time on my own testing and then never actually had that time available.

It's tempting to just merge this if evidently it doesn't cause problems. However I still feel this is a hacky solution that could have obscure side effects (e.g. it seems like it would leak hook entries in certain scenario's).

It doesn't seem like this PR will be improved, so I will close it. If someone is willing to investigate a cleaner solution (for example using better checks in Reset() as suggested above) then I would be happy to merge a new PR.

PR #720, opened 2014-06-06
Font fixes and changes

Bo98:
* ~~Removed OS X specific Derma font. This now uses Tahoma like Windows.~~ Already done by @UnderscoreKilburn.
* Added Linux specific Derma font. Linux distros are more likely to have the close-looking DejaVu Sans than Tahoma. _This MAY fix the italic font some people had in Derma but will NOT fix surface.CreateFont using an italic font as a default if the font cannot be found. I was unable to reproduce this so I cannot confirm whether it even fixes Derma._
* Added Windows and Linux specific fonts to the Workshop download VGUI. Before this only had Helvetica which these systems may not have.

### DIFF
--- a/garrysmod/lua/derma/init.lua
+++ b/garrysmod/lua/derma/init.lua
@@ -10,50 +10,44 @@
 -- The default font used by everything Derma
 --
 
---[[if ( system.IsOSX() ) then
+if ( system.IsLinux() ) then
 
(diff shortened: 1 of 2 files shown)


robotboy655:
Drop the new line changes and rebase.

Bo98:
Done.

PR #732, opened 2014-06-21
DListView update

Donkie:
Added Set/GetSortValue methods to DListView_Line, this lets you set a specific value which will be used during  columnsorting, other than being forced to have the text value, which isn't always nice to sort with.

Added position parameter to DListView:AddColumn

Removed DListView_Column:SetMaterial, it was unused (also removed the parameter from AddColumn)


tgp1994:
After reading today's update changelog, I just wanted to write in and say: Thank you. It's nice to see people working on these Derma controls.

PR #786, opened 2014-08-17
Right clicking on menu version opens console

Python1320:
This could be made so that it only works if "developer 1" is enabled.


zerfgog:
I agree with this pr for those times when you need to quickly check console and mashing alt+tilde just doesn't cut it,
and this:
> [A button would be handy for those times where someone requesting help really needs to type something in their console but can't open it. :v - Willox](http://facepunch.com/showthread.php?t=1413999&p=45716184&viewfull=1#post45716184)

robotboy655:
Shift+Escape is made to open the console now.

PR #796, opened 2014-08-29
few non atomic changes, see description.

SimonSchick:
Revamped dragging/sizing system to work on all borders and corners
Removed empty function descriptors
Changed a few internal variables(m_ syntax is retarded)
Added SetBorderThreshold(number) to set the size of the sizing area
Added SetMaxWidth/Height(num) which limits the panels size
Minor optimisations

### DIFF
--- a/garrysmod/lua/vgui/dframe.lua
+++ b/garrysmod/lua/vgui/dframe.lua
@@ -14,9 +14,9 @@
 local PANEL = {}
 
 AccessorFunc( PANEL, "m_bIsMenuComponent", 		"IsMenu", 			FORCE_BOOL )
-AccessorFunc( PANEL, "draggable", 			"Draggable", 		FORCE_BOOL )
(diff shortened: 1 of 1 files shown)


willox:
Changing the names of the member variables could potentially break some scripts and the drag code is pretty hard to read.

SimonSchick:
The whole point of accessors is that people don't use the member variables.
If they access internal things it's their own fault if it breaks.

And regarding the "hard to read part", how would you suggest to make it more readable?

willox:
There's no point making changes that'll break something just because the changes can be made. Try to keep compatibility.

The hard to read part is mostly just the pyramids in PANEL:Think. Space it out a bit and it'll probably be fine. I was playing with DFrames the other day and agree that being able to resize them from every side would be pretty cool.

SimonSchick:
I can't be arsed to install gmod and edit the PR.

If you want the functionalities, merge it and edit the bits you don't like.

PR #797, opened 2014-08-29
Made Color a callable table allowing static functions.

SimonSchick:
Added various predefined colors such as Color.Red
Localized various function in order to compensate _minor_ speed loss from callable table.
Optimized call overhead of Color:ToHSV()

Added static functions:
Color.FromHSV alias of HSVToColor
Color.FromHex parses a 6 digit hex string
Color.FromVector turns a vector into a color

methods:

Color:Copy() returns an equivalent color object
Color:GetRGB() returns r g b as seperate arguments
Color:GetRGBA() see above plus alpha
Color:CloneNewAlpha(a) method version of ColorAlpha

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -1,21 +1,194 @@
-local COLOR = {}
-COLOR.__index = COLOR
+local colorMeta = {}
+colorMeta.__index = colorMeta
 
---[[---------------------------------------------------------
-	Register our metatable to make it accessible using FindMetaTable
------------------------------------------------------------]]
+local mathmin = math.min
+local tonumber = tonumber
+local setmetatable = setmetatable
+local getmetatable = getmetatable
+local Vector = Vector
 
-debug.getregistry().Color = COLOR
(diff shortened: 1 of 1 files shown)


zerfgog:
I agree with the other commentors stating that the large list of color definitions should be up to the user to create. It's simply unnecessary to have this large array when users will only use a few, and is it worth it to have this large table over people creating their own?
I do like the color "library" set up in this pr, but really it's better to have ```vec:ToColor()``` and ```string.ToColor(hexorsomeotherorformat)``` as we already have instead of said libraries/classes containing these functions.
I don't mind ```GetRGB```/```GetRGBA``` and ```Clone``` much, but most people will just index the color instead of a useless function call. However, ```col:CloneNewAlpha``` is pretty oddly named. I do like that you've added it to the color meta, however.
There is also the problem of coding standards with this commit: an UpperCamelCase (essentially) library name (```Color.whatever```), and the lowerCamelCase ```colorMeta```.

SimonSchick:
2 1/2 and this is still open, lol.

Kefta:
I swear I'm going to destroy that Color file once and for all

PR #800, opened 2014-08-30
Changed all all string.* calls to method calls where appropriate

SimonSchick:
Some exceptions, all functions in string that do not take a string as first argument(mostly garry's)


robotboy655:
You'll have to rebase/merge with latest changes and split TTT changes into another PR.

PR #803, opened 2014-09-03
added recursive include functions

CallMePyro:
AddCSLuaDir( path, recursive ) will AddCSLuaFile every .lua file in the directory. If recursive is true, it will run on all sub directories, and their sub directories.

IncludeDir( path, recursive ) will do the same, except it uses include() instead of using AddCSLuaFile().

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -135,10 +135,58 @@ function IncludeCS( filename )
 	
 end
 
+--[[---------------------------------------------------------
+   Adds all Lua files in the directory to the CS file list
+   and all sub directories if enabled
+-----------------------------------------------------------]]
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/util.lua
```diff
@@ -135,10 +135,45 @@ function IncludeCS( filename )
...
+        for _, file in pairs( files ) do
+            if file:EndsWith( ".lua" ) then
+                func( path .. "/" .. file )
+            end
+        end
+
+        if not recursive then return end
+
+        for _, dir in pairs( dirs ) do
+            fileTreeTraverse( path .. "/" .. dir, func, true )
```
AbigailBuccaneer: Arguments here are seemingly in the wrong order than everywhere else.

PR #811, opened 2014-09-18
Bug Fix at DAjustableModelPanel

HunterFP:
Fixed the bug when the panel is created the model does not show up


robotboy655:
https://github.com/garrynewman/garrysmod/commit/fbe169b68a19bd73b57fc387d1384e9c52095a44

PR #815, opened 2014-09-28
Made gamemode browser sort based on server fullness, not player counts

DaaOtt:
To promote new gamemodes that *don't* have hoards of servers.

What the new gamemode list looks like


robotboy655:
This may work as an optional thing, but it will not be a replacement of the current system.

PR #821, opened 2014-09-29
net.ReadType no longer requires an argument

wiox:
This makes it actually make sense to use, but also keeps backwards compatibility to help any poor soul who has used it anyway.

Also updates net.WriteTable and net.ReadTable to use it, which makes them a bit easier to understand.


wiox:
Updated to make this merge-able after #798.

PR #825, opened 2014-10-02
Improved map counter when filtering maps

Bo98:
The previous CountFiltered function used regex instead of AngularJS's filter function, unlike the actual filterer. This lead to mismatches between the number of actual filtered maps and counter's number. This also fixes some errors that were being spammed in the case of invalid regex expressions being used in the search box.

### DIFF
--- a/garrysmod/html/js/menu/control.NewGame.js
+++ b/garrysmod/html/js/menu/control.NewGame.js
@@ -3,20 +3,20 @@ var ServerSettings = []
 var scope = null;
 var rootScope = null;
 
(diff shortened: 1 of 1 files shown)


robotboy655:
I think I accidentally fixed this with https://github.com/garrynewman/garrysmod/commit/72fe957ed00b3d7fd6296ca30bf6c19edb9a8c77

Do you still think we should change CountFiltered to your version?

Bo98:
I believe the change is still required.

Try searching !de in the Counter-Strike section.

I'll fix the merge conflict tomorrow.

Bo98:
Merge conflict is fixed now.

PR #826, opened 2014-10-02
Deprecation

willox:
__Not being merged until after next update, if at all.__

I'd like to consider this a step in the right direction for cleaning up a lot of inconsistencies in Garry's Mod's function / method naming. Calls to deprecated functions will cause a non-aborting error message once per location.

There are still a few things to be done to the DEPRECATE function itself.
If you have any suggestions for this / functions that should be deprecated, post them. If you have any arguments against this / functions this deprecates, post them.

Most of the deprecated functions are from http://wiki.garrysmod.com/page/Category:Deprecated_Functions
Two exceptions for this are math.Min and math.Max, which aren't on the wiki at all.

```
> print(math.Distance(1, 2, 3, 4))
	2.8284271247462
	
> print(math.Dist(1, 2, 3, 4))
	[ERROR] lua_run:1: calling deprecated function 'math.Dist' (use 'math.Distance')
	2.8284271247462
```

I consider this a good idea because, if functions are to be removed in the future, why not convince developers to stop using them in the present?

View the beef at https://github.com/wiox/garrysmod/blob/deprecate/garrysmod/lua/includes/deprecation.lua

__Opinions very much needed__

### DIFF
--- a/garrysmod/lua/includes/deprecation.lua
+++ b/garrysmod/lua/includes/deprecation.lua
@@ -1,0 +1,74 @@
+local function DEPRECATE( f, name, alternative )
+
+	local error_locations = {}
+	local format = "[ERROR] %s:%d: calling deprecated function '%s'\n"
+
+	if ( alternative ) then
+		format = "[ERROR] %s:%d: calling deprecated function '%s' (use '%s')\n"
+	end
+
+	return function( ... )
+
+		local info = debug.getinfo( 2, "Sl" )
+
+		if info then
+
+			local source, currentline = info.short_src, info.currentline
+
+			if ( not error_locations[source .. currentline] ) then
+
+				error_locations[source .. currentline] = true
+				ErrorNoHalt( string.format( format, source, currentline, name, alternative ) )
+
+			end
+
+		end
+
+		return f( ... )
+
+	end
+
+end
+
+
+local ENTITY = FindMetaTable( "Entity" )
+local PLAYER = FindMetaTable( "Player" )
+
+IncludeCS 									= DEPRECATE( IncludeCS, "IncludeCS", "include & AddCSLuaFile" )
+
+math.Dist 									= DEPRECATE( math.Dist, "math.Dist", "math.Distance" )
+math.Max 									= DEPRECATE( math.Max, "math.Max", "math.max" )
+math.Min 									= DEPRECATE( math.Min, "math.Min", "math.min" )
+math.mod 									= DEPRECATE( math.mod, "math.mod", "math.fmod" )
+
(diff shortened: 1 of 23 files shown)


Nyaaaa:
This is the most stupidest thing ever!

Bo98:
Thanks for the valuable feedback.

TheFreeman193:
@Nyaaaa This isn't actually a bad idea, given how many developers don't check if their code is up-to-date. Whether or not it would make a difference is another question entirely.

If you're going to criticise something, it's generally a good idea to give your reasoning.

Nyaaaa:
@TheFreeman193 I just did give my reason, It's a stupid idea, Instead of deprecation it they should just redirect the code to the new one, So the lazys don't have to do anything..

zerfgog:
I believe you misunderstand this pull request; have you read the code?
The functions set to be deprecated are CURRENTLY aliases. They were either created accidentally or by one of the devs who wanted a shortcut/didn't want to remember casing.
This PR is probably the best way to go about it: it only alerts people that the functions are outdated, hopefully giving some people opportunities to update their addons or addons that are currently not maintained to find a solution. Once this adjustment period has gone on long enough, the functions can then be safely removed outright.
(However, this is the ideal scenario. More than likely, the non-maintained addons out there will pose a problem.)

Nyaaaa:
@zerfgog I know what it means, I don't need you telling me lol..

garrysmod/lua/includes/deprecation.lua
```diff
@@ -0,0 +1,74 @@
...
+
+IncludeCS 									= DEPRECATE( IncludeCS, "IncludeCS", "include & AddCSLuaFile" )
+
+math.Dist 									= DEPRECATE( math.Dist, "math.Dist", "math.Distance" )
+math.Max 									= DEPRECATE( math.Max, "math.Max", "math.max" )
+math.Min 									= DEPRECATE( math.Min, "math.Min", "math.min" )
+math.mod 									= DEPRECATE( math.mod, "math.mod", "math.fmod" )
+
+util.tobool 								= DEPRECATE( util.tobool, "util.tobool", "tobool" )
+util.TraceEntityHull 						= DEPRECATE( util.TraceEntityHull, "util.TraceEntityHull" )
```
GamerGambit: util.TraceEntityHull has no alternative?

Nyaaaa: Just looked on the wiki i wonder if they could fix util.TraceEntityHull it could become useful!

meepdarknessmeep:
Can we get the argument to `net.ReadVar` to be deprecated? All it does is let more and more people make their code able to broken if `net.ReadVar` or `net.WriteVar` gets changed.

PR #829, opened 2014-10-04
Removed repeat statement

Kefta:
Also localised xside


Kefta:
Bump, delete it or not?

PR #833, opened 2014-10-26
Margin for displacement

mavisdracula:
A margin for the tabscroller to fix the ugly derma hook coming out of the DTab.


robotboy655:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/1515, but I will try for a different approach first.

Kefta:
@robotboy655 ?

PR #834, opened 2014-11-07
Update obj_player_extend.lua to prevent a physics crash

Cre8or:
The function Player:UnfreezePhysicsObject() will crash the game/server when attempting to unfreeze entities that don't have proper physics. The fix consists of checking whether the physics are valid, and simply not unfreezing when they're not.

As entities with custom physics (for example spherical collisions) would get filtered out by this fix, the current code simply unfreezes those without verifying anything. Hasn't caused any problems in the last weeks of testing.

For more information, see https://github.com/wiremod/advdupe2/issues/28

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua
+++ b/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua
@@ -115,11 +115,24 @@ function meta:UnfreezePhysicsObjects()
 		-- Make sure the entity to which the physics object
 		-- is attached is still valid (still exists)
 		if ( isentity( v.ent ) && IsValid( v.ent ) ) then
-		
+
+			--Before unfreezing the entities, we need to make sure that the physics metadata is valid,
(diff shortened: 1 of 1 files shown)


thegrb93:
This fix works, thanks.

robotboy655:
I don't get why this part is needed:
```
if (Phys == SOLID_BBOX) then
	v.ent:GetPhysicsObject():EnableMotion( true )
end
```
Custom physics objects are still considered valid ( PhysicsInitSphere ) for example on bouncy ball.

KateAdams:
Maybe they meant to test against `~= nil`?

Jack37:
It's support for Make Spherical tool. If you use the tool on frozen prop, the new physics object isn't considered valid until you touch the prop. However it doesn't seem to do anything, simple ````IsValid(v.phys)```` here https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua#L122 should do it.

PR #835, opened 2014-11-13
DHorizontalScroller: Added draggable elements support

KrakenZ:
* Changed the canvas panel into a **DDragBase** to support draggable elements.
* Made left and right buttons aligned on bottom.

```lua
-- If set to true, will automatically move the dragged element.
DHorizontalScroller:SetUseLiveDrag( bool )

-- By default, the DHorizontalScroller doesn't allow dragging elements.
-- If you call this with a string, all childrens will be draggable.
DHorizontalScroller:MakeDroppable( string )

-- Show / Hide the pink drop targets ( default is false )
DHorizontalScroller:[Set/Get]ShowDropTargets( bool )

-- To be overridden. Called when an internal draggable element got modified (i.e moved).
DHorizontalScroller:OnDragModified()
```

Here is what it looks like on the tabs of a simple DPropertySheet with the following used functions
```lua
MyDPropertySheet.tabScroller:MakeDroppable( "Tabs" )
MyDPropertySheet.tabScroller:SetUseLiveDrag( true )
```
Example

### DIFF
--- a/garrysmod/lua/vgui/dhorizontalscroller.lua
+++ b/garrysmod/lua/vgui/dhorizontalscroller.lua
@@ -11,24 +11,21 @@
 	
 --]]
 
-PANEL = {}
+local PANEL = {}
 
 AccessorFunc( PANEL, "m_iOverlap", 			"Overlap" )
 AccessorFunc( PANEL, "m_bShowDropTargets", 	"ShowDropTargets", FORCE_BOOL )
 
---[[---------------------------------------------------------
-
------------------------------------------------------------]]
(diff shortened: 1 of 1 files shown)


Kefta:
Why did you close this?

KrakenZ:
If you could have uses for this I can re-open it.
Some lines needs to be changed btw

Kefta:
I would use this personally, just to allow for draggable elements. I think this is a useful feature

KoJI6ac9H:
Oh that's awesome! :)

PR #839, opened 2014-11-25
Make DLabelURL functional again

Exho1:
Changed the PANEL's base from URLLabel to a DButton, that will make URLLabel useless (It wasn't being used anyways) but this element will work again. I also fixed the fonts and added accessor functions for the URL variable!

### DIFF
--- a/garrysmod/lua/vgui/dlabelurl.lua
+++ b/garrysmod/lua/vgui/dlabelurl.lua
@@ -4,7 +4,6 @@
  /'_` | /'__`\( '__)/' _ ` _ `\ /'_` )
 ( (_| |(  ___/| |   | ( ) ( ) |( (_| |
 `\__,_)`\____)(_)   (_) (_) (_)`\__,_) 
-
 	DLabelURL
 --]]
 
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dlabelurl.lua
```diff
@@ -56,14 +65,34 @@ function PANEL:GetColor()
 
 end
 
+--[[---------------------------------------------------------
+	SetFont
+-----------------------------------------------------------]]
+function PANEL:SetFont( strFont )
```
robotboy655: There's no need for this, this method is inherited from DLabel ( Which is parent of DButton ).

garrysmod/lua/vgui/dlabelurl.lua
```diff
@@ -32,6 +30,17 @@ function PANEL:Init()
 end
 
 --[[---------------------------------------------------------
+	Paint
+-----------------------------------------------------------]]
+function PANEL:Paint()
+
+	-- Some trickery to hide the DButton
+	surface.SetDrawColor(0,0,0,0)
+	surface.DrawRect( 0, 0, self:GetWide(), self:GetTall() )
```
robotboy655: Something tells me that leaving this function completely empty will achieve the same effect as drawing an invisible rectangle.

robotboy655:
If you wish to get this in, fix the issues pointed out.

PR #840, opened 2014-11-25
ENT:NetworkVar ( and ENT:DTVar  ) uses SetNW*/GetNW* instead of SetDT*/GetDT*

Jvs34:
As far as my tests go, this breaks pretty much nothing, and most people won't even notice.

Added an argument to ENT:GetNetworkVars to return entities, duplicator behaviour is still the same since it passes no arguments.

Also made the notify stuff work for both states.

Thirdperson mode and camera distance for vehicles now use NWVars instead of dtvars.

Vinh wanted to do this but I beat him to it, suck it vinh.
http://facepunch.com/showthread.php?t=1374457&p=46371734&viewfull=1#post46371734

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -190,7 +190,7 @@ function meta:InstallDataTable()
 		local dt = datatable[ key ]
 		if ( dt == nil ) then return end	
 		
-		return dt.GetFunc( self, dt.index, key )
+		return dt.GetFunc( self, dt.index )
 		
 	end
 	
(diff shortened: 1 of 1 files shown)


Bo98:
I realise that pull requests are no longer being merged until after the update but shouldn't there be an exception for this one since it's something _Kilburn would've done anyway?

robotboy655:
@Bo98 , I don't think this is a good idea for now, NWVars2 have some problems right now which need fixing before this can be even considered.

Bo98:
I assumed those issues were being fixed before the update too.

Jvs34:
Yeah I'm aware of the conflicts, I'll fix them once @UnderscoreKilburn gets off his ass and fixes the NW2 problems, when he does he might as well reimplement this stuff on his own.

Jvs34:
I can't be arsed to squash commits and rebase, I'll remake this at some point.

PR #843, opened 2014-12-15
Color to integer

EstevanTH:
Hey,

This is a new feature that allows a 32-bit Color object to be converted into a 32-bit signed or unsigned integer.
The opposite feature is placed next to the previous one.

I think it's very useful for networking and saving to disk. I'm also planning to set 2 new methods in the Entity datatable then : **Entity:SetNWColor()** and **Entity:GetNWColor()**.


Kefta:
You should put this in the util color file instead

Kefta:
?

EstevanTH:
It's a mistake, I'm trying to make a new pull request for something else...

PR #846, opened 2014-12-26
GetPlayerByNick

Exho1:
A convenience function which returns a player entity from a string rather than having to write a for loop every single time, the second argument checks for a player with the exact case as the string. Acts as the opposite of ply:Nick()

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -105,6 +105,23 @@ function Model( name )
 	return name
 end
 
+--[[---------------------------------------------------------
+	Returns a player object from a string
+-----------------------------------------------------------]]
+function GetPlayerByNick( nick, bExact )
+	for k, v in pairs( player.GetAll() ) do
+		if ( bExact ) then 
+			if ( v:Nick() == nick )  then
+				return v
+			end
+		else
+			if ( v:Nick():lower() == nick:lower() ) then
+				return v
+			end
+		end
+	end
+end
+
 
 -- Some nice globals so we don't keep creating objects for no reason
 


Lexicality:
@LennyPenny How is a function that returns the first player with the name passed in subjective?
Sure, someone could extend it with fuzzy matching and so on, but that's no reason not to have a basic function in there to start off with.

KateAdams:
I agree with @Lexicality; I do think that it should return an array of players that matches the name though.

robotboy655:
Exact and lowercase comparisons are useless.
* Exact name comparisons are rarely ever used. And this will just allow more people to do dumb shit like store players by their nickname.
* Lowercase comparisons may give the wrong player in case there are two or more players with same name but different capitalization.

And while adding player.GetByName that will return a table solves some that and supports ```blah*```, it is still not useful because this function will mostly be used by admin mods, which usually have additional targeting systems such as by SteamID and symbol replacements, etc, and thus this function will be unused.

My main problem is that the function is pretty much worthless either way, but if a lot of people ( more than 2 or 3 ) will support player.GetByNick() that returns a table, I will add it.

neico:
Count me in on that table approach, though I'd also welcome optional arguments to control the outcome a bit more.

some args I could think of:
- boolean deciding if regex or wildcard match (could help a little bit with some performance junkies)
- how many results should it contain (can make the function return only one player if 1 is used, NULL if 0 and a table otherwise, including -1 for all matches) or just let it return all in a table (though the specific result could be slightly more useful)

Also: one more use case for such a function would be debugging, you want to check something on a user but don't want to bother with looking for the userid, steamid or whatever so you just write: `MsgN( player.GetByNick( "Minge", false, 1 ):GetActiveWeapon() )`

KateAdams:
@neico That's pretty silly, you aren't even checking for a nil, and it's equivalent to `player.GetByNick("Minge", false)[1]:GetActiveWeapon()`.

UselessGhost:
I'm totally for this. Even though many admin mods will reimplement this, having a "neutral" implementation will be beneficial for simple scripts.

PR #847, opened 2015-01-01
Added util.TraceOBB and util.TraceEntityOBB

ottworks:
This is my workaround for the lack of TraceEntity and HullTrace rotation.

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -6,6 +6,183 @@ if ( CLIENT ) then
 end
 
 --[[---------------------------------------------------------
+	Name:	TraceEntityOBB
+	Params:	<table tracedata> <ent entity> <bool quick> <bool verbose>
+	Returns: <table result> <table verboseresults>
+	Desc:	Traces using an entity's bounding box offset along a line. 
+	"Quick" parameter skips tracing through every corner and only traces between the start corner and end corner.
+	"Verbose" parameter returns every trace performed in a table
+	WARNING: This traces lines between an ent's bounding box corners. Small entities may "slip through".
+	TraceData structure:
+		Vector mins: Offset mins (lower back left corner). Optional.
+		Vector maxs: Offset maxs (upper front right corner). Optional.
+		Vector start: Traces from this position, if you /really/ want to. Optional.
+		Vector endpos: Traces from the entity's position (or the starting position) to this position. Optional.
+		Entity filter: Things the trace should not hit. Can also be a table of entities or a function with one argument.
+		boolean ignoreworld: Should the trace ignore world or not
+		table output: If set, the trace result will be written to the supplied table instead of returning a new table
+-----------------------------------------------------------]]
+function util.TraceEntityOBB(tracedata, ent, quick, verbose)
+	local mins, maxs = ent:GetModelBounds()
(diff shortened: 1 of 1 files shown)


zerfgog:
:+1:+1

thegrb93:
This is just 8 line traces which won't work in many cases. I think it would be a better approximation to calculate a hull trace which is inscribes by your desired rotated hull. Still not perfect, but better than a volume-less trace.

DaaOtt:
A single hull trace defeats the point, but a combination could work. It traces through the middle by default, BTW.

DaaOtt:
Also it ends up being 8 traces on quick mode and 16 by default.

thegrb93:
That's a lot of traces. I wonder how performance compares.

DaaOtt:
Okay.

thegrb93:
Okay, you don't care then. Well these functions names are misleading. Anybody would expect the trace to actually fill a volume, and for large entities this won't be the case and people will wonder why this function doesn't work half the time.

DaaOtt:
Well I mean _technically_ "trace" means "draw along the outline of-"
Yeah, I get your point. It worked for my purposes though, why keep it to myself?

thegrb93:
It just needs some improvement or a name that is more clear about its functionality. util.TraceLineOBB would be a better name, but considering you are doing a trace from the center I'm thinking it was intended to be a hull trace improvement. I'm thinking if you inscribe a trace hull in the OBB, and do 6 line traces at the projected edge points of the OBB then it would be perfect. Requires a bit of linear algebra, but I think it's worth it. I can give it a go if you need help with the math.

Here's an example. The black box is the rotated OBB, the orange box is the hull trace, and the white dots are the trace lines, and the trace direction is along the view axis.
[Imgur](http://i.imgur.com/cauMx8g.png)

DaaOtt:
Yeah, that's what I was thinking. Have no idea how to do the math though.

thegrb93:
I'll derive a solution after final exams this week.

thegrb93:
Never mind, I forgot that these hulls can be rectangular rather than cubic and that complicates things. My next suggestion is to add a "spacing" parameter which can control the density of traces over the area of the hull. 
Traces

DaaOtt:
That in conjunction with a trace along the edges of the OBB would work nicely. Don't forget to use octrees to optimize large traces.

FPtje:
The tracer returns the first hit it finds. It disregards all other traces that may or may not have hit a different entity.
Imagine the "back left bottom" corner trace hitting something at 1000 units distance, and the "front right top" hitting an entity at 200 units away. Obviously you'd want to return the closest entity, but because of the arbitrary looping order, it returns the first hit it finds instead of the closest one.

FPtje:
Also, the "default" mode runs 56 traces, not 16:
```
For each corner i do 
   for each corner j do 
        trace from i at start to j at end unless I == j
```
You have 8 corners, for every corner you make a trace for all 7 other corners. 7 * 8 = 56

thegrb93:
Ahh, so you're crisscrossing the traces. That's a good idea.

DaaOtt:
Thanks for pointing that out.

robotboy655:
Since you can implement this in a number of different ways, I think this belongs in an addon that needs it rather than the base game.

PR #850, opened 2015-01-06
Fixed main menu text shadow transparency on OS X

Bo98:
Somewhat addresses Facepunch/garrysmod-issues#90. This could be happening to other websites loaded through a HTML/DHTML control but it won't affect most and it is fixed for the main menu.

### DIFF
--- a/garrysmod/html/css/menu/Menu.css
+++ b/garrysmod/html/css/menu/Menu.css
@@ -1,4 +1,3 @@
-
 BODY
 {
 	overflow: hidden;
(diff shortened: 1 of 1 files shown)


KateAdams:
What's with the different newlines?

Bo98:
There were a few trailing ones which were trimmed.

robotboy655:
Before/After pic? I have tried that on Windows and couldn't see any noticeable differences.

Bo98:
Note the text shadow

Before:
Before

After:
After

Never was a bug on Windows but was one on OS X.

PR #853, opened 2015-01-09
Regression: Fixed missing arrow in player count dropdown menu

Bo98:
The Tranny module previously replaced the whole inner HTML of an element when translating text. This would replace any child HTML node instead of just text ones. This wasn't an issue in the main menu until 9d71f76 when we started dealing with child HTML nodes.

https://github.com/garrynewman/garrysmod/blob/72fe957ed00b3d7fd6296ca30bf6c19edb9a8c77/garrysmod/html/template/newgame.html#L40-L43

The module now checks for child text nodes, replaces one and removes the rest while leaving all other child nodes intact. If no child nodes were found in the first place, the inner HTML is simply replaced like before.

### DIFF
--- a/garrysmod/html/js/svc.Tranny.js
+++ b/garrysmod/html/js/svc.Tranny.js
@@ -10,23 +10,44 @@ angular.module( 'tranny', [] )
 		var update = function()
 		{
 			if ( IN_ENGINE == false ) return;
-			
-			$(element).html( language.Update( strName ) );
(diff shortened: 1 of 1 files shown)


robotboy655:
https://github.com/garrynewman/garrysmod/commit/e656a79d3cd440fe4eb5b12ca2a40155a498c5ed

The only reason I am not merging this is because I am afraid it will make the already slow main menu even slower.

PR #857, opened 2015-01-26
TTTPlayerColor now retains previous colors

Kefta:


wiox:
Is model not a string?

TheFreeman193:
Looking at [this call](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/init.lua#L479) and the line above, wiox is right; _model_ is obtained from GetRandomPlayerModel(), which returns a string. You might want to remove some of the semicolons floating around, too.

Kefta:
Nevermind, just realised that this doesn't matter since there's not a way to access the default colors for that model. It would be preferable not to color the model at all, but that would require a major change that I'll need to ask Bad King about. I'll just open another PR to change the default convar value to 0, which will resolve the innumerable help threads about why their custom models are turning green.

wiox:
With additive color wouldn't white always be default?

Kefta:
I don't know; I remember testing this with the old CS:S models and white would change the default color, but I'm not sure how this has changed with colorable models. Also, is there a function to see if the model is colorable?

PR #858, opened 2015-01-26
DFrame's description typo fix

Luabee:
It said "A simpe window"

Now it says "A simple window"

For the love of all that is the English language, I **demand** this change to be made.


TheFreeman193:
This might be a good point at which to rectify/clean up any other similar issues in the VGUI elements. It'll make this pull a lot more useful. I'll let you know if I find anything like this.

Luabee:
You've made the right choice today, friend. England thanks you.

PR #859, opened 2015-01-26
Fix DPropertySheet padding on bottom edge.

AbigailBuccaneer:
Previously the bottom padding on the panel in a DPropertySheet was exactly 8px, because the active tab is 8px taller than the other tabs[¹](https://github.com/AbigailBuccaneer/garrysmod/blob/patch-1/garrysmod/lua/vgui/dpropertysheet.lua#L117). However, the default padding is 8px, so this is only revealed when changing the padding on a property sheet.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_c4.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_c4.lua
@@ -16,6 +16,7 @@ if CLIENT then
    };
 
    SWEP.Icon = "vgui/ttt/icon_c4"
+   SWEP.IconLetter = "I"
 end
 
 SWEP.Base = "weapon_tttbase"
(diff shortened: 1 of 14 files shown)


KateAdams:
It'd be better if you didn't use magic numbers.

AbigailBuccaneer:
@KateAdams the 8px that's added by the active tab is a magic number. I went for the most minimal fix instead of also modifying the tab class to not use a magic number. I'll change it. (In fact, given the `y`-position is hardcoded to `20 + self:GetPadding()` regardless of the height of the tabs, it probably makes more sense to use that as the top offset.)

AbigailBuccaneer:
Whoops - I managed to automatically close this and make it unopenable again by updating my `AbigailBuccaneer:patch-1` branch in line with `garrynewman:master`. Apologies, I'll open a new pull request now.

PR #862, opened 2015-01-31
Add min/max functionality to DFrame

TheFreeman193:
**Adds Get/Set methods for minimise, maximise and min/max animations, and integrates this functionality in the control box buttons.**
- Uses panel's native animations (optionally disabled).
- Adds OnMinimized, OnMaximized and OnRestored hooks.
- Updates maxmise button paint function in default skin for min/max restore.
- Uses Americanised keywords to match Garry's usage.
- Max/min state hierarchy and button icons matched to standard GUI windows:
  - Maximise button restores minimised window
  - Minimise button minimises even when maximised
  - Window is maximised on restore if was so before minimised
  - Restore icon shown in maximise button if minimised or maximised

### DIFF
--- a/garrysmod/lua/vgui/dframe.lua
+++ b/garrysmod/lua/vgui/dframe.lua
@@ -13,25 +13,27 @@
 
 local PANEL = {}
 
-AccessorFunc( PANEL, "m_bIsMenuComponent", 		"IsMenu", 			FORCE_BOOL )
-AccessorFunc( PANEL, "m_bDraggable", 			"Draggable", 		FORCE_BOOL )
-AccessorFunc( PANEL, "m_bSizable", 				"Sizable", 			FORCE_BOOL )
-AccessorFunc( PANEL, "m_bScreenLock", 			"ScreenLock", 		FORCE_BOOL )
-AccessorFunc( PANEL, "m_bDeleteOnClose", 		"DeleteOnClose", 	FORCE_BOOL )
-AccessorFunc( PANEL, "m_bPaintShadow", 			"PaintShadow", 		FORCE_BOOL )
+AccessorFunc( PANEL, "m_bIsMenuComponent", "IsMenu",         FORCE_BOOL )
+AccessorFunc( PANEL, "m_bDraggable",       "Draggable",      FORCE_BOOL )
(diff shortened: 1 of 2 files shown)


garrysmod/lua/vgui/dframe.lua
```diff
@@ -16,12 +16,15 @@ local PANEL = {}
 AccessorFunc( PANEL, "m_bIsMenuComponent", 		"IsMenu", 			FORCE_BOOL )
 AccessorFunc( PANEL, "m_bDraggable", 			"Draggable", 		FORCE_BOOL )
 AccessorFunc( PANEL, "m_bSizable", 				"Sizable", 			FORCE_BOOL )
+AccessorFunc( PANEL, "m_bMinMaxAnim", 			"MinMaxAnim", 		FORCE_BOOL )
```
KateAdams: How to align

Jvs34: Ok? He's just following the same name convention as the file.

KateAdams: You mean indentation style?  But yes, you should change them to use spaces for alignment rather than tabs.
The reason this isn't done altogether is because it can break merges, but if you're going to edit the file and its alignments, then you should fix it and kill two birds with one stone.

TheFreeman193: She's talking about the HTs that pad out the accessors.
@KateAdams I'm planning to clean the files up in the next commit.

TheFreeman193:
(Added 1 squashed commit)
* Separates show funcs for close and min/max buttons.
* Cleans up files

zerfgog:
:+1:+1

robotboy655:
I don't think this is a good idea, it serves absolutely no purpose other than being a gimmick.

I guess the only use for this is when you have many DFrames on screen, but at the same time, maybe you should consider the fact that your UI is not very well done then.

zerfgog:
The buttons **do** exist on the current DFrame title bar serving no purpose

TheFreeman193:
I agree that this a fair bit of extra source for something that will have relatively few uses. An alternative might be to just make the skin changes, which would permit individuals to implement it themselves should they desire.

PR #863, opened 2015-02-04
Made OnUndo get called by hook.Call

meharryp:
Previously, if a player undid something in a gamemode that wasn't sandbox or derived from sandbox, there would be an error on the client, using hook.Call however prevents that.


TheFreeman193:
It'd probably be better to put this clientside in [GM:OnUndo](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/sandbox/gamemode/cl_init.lua#L39-50), otherwise the same data is being sent to clients twice. 
```lua
hook.Call( 'OnUndo', GAMEMODE, name, strCustomString )
```

meharryp:
Github messed up, going to recreate the pull.

PR #864, opened 2015-02-07
SetWorldTip

NeatNit:
SetWorldTip, a more descriptive replacement for AddWorldTip without the unused arguments

Ignore this one, look at https://github.com/garrynewman/garrysmod/pull/865 instead!

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/cl_worldtips.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/cl_worldtips.lua
@@ -12,18 +12,19 @@ local WorldTip = nil
 
 local TipColor = Color( 250, 250, 200, 255 )
 
---
--- Adds a hint to the queue
---
-function AddWorldTip( unused1, text, unused2, pos, ent )
 
+function SetWorldTip( text, pos, ent )
 	WorldTip = {}
 	
 	WorldTip.dietime 	= SysTime() + 0.05
 	WorldTip.text 		= text
 	WorldTip.pos 		= pos
 	WorldTip.ent 		= ent
-	
+end
+
+--Backwards compatibility
+function AddWorldTip( unused1, text, unused2, pos, ent )
+	SetWorldTip( text, pos, ent )
 end
 
 


wiox:
Why not just make the arguments work? (Except the ent id one that's shit)

NeatNit:
They're currently unused, I'm guessing they've been intentionally made non-functional at some point. My guess is the ent id one was used for uniquely identifying more than one tooltip, and the time to live was, obviously, customizable. My further guess is that the former was removed because there was a design choice to allow only one world tip at a time, and the latter locked to 0.05 to prevent making tooltips that last longer than they are relevant. (0.05 essentially means it disappears immediately after not being called anymore)

I don't see reason to doubt these design choices, I just think it's silly that we only have this old function with two unused arguments. This function was also undocumented, mind you, until today (I'm the one to add it to the wiki).

wiox:
The arguments can still easily be more intuitive than text, pos, ent.

NeatNit:
How so?

wiox:
Most functions that take an entity take that entity as the first parameter. Allowing for that just gives similarity across gmod.

I'd still suggest the first argument being a position or entity and the second being the text.

NeatNit:
I don't like having a parameter that can take two different types, partly because it's difficult to explain in simple terms and partly because there's no right way to document it on the wiki.

Also, I prefer to make the smallest changes possible to make the pull request as attractive as possible - this small change can be approved and merged without a second thought, but when you start doing fancy things (like an argument that takes different types) it starts being a little off-putting... To me at least.

TheFreeman193:
I reckon most of this should be rewritten anyway; it's inefficient, and doesn't do what it suggests. There's potential here for a pretty good system allowing multiple position-based tooltips.

NeatNit:
I agree, but I don't feel confident enough to go and do that. You can just
make another pull request! :)

TheFreeman193:
No worries, I'll have a play around with the code tomorrow and let you know if I come up with anything worth including. Biggest issue is going to be backwards compatibility since the legacy calls will use it every frame, but some form of identification will sort that out.

Bo98:
I'd just create a new function and deprecate the old one. It's been undocumented until now so it can probably be removed an update or two after deprecation.

TheFreeman193:
That's a good point, I've only seen this used a couple of times. I'll just make sure the original still works if I make any changes to the draw func or gamemode hook.

TheFreeman193:
#865

PR #865, opened 2015-02-08
Add worldtip library to replace AddWorldTip

TheFreeman193:
Replaces AddWorldTip
- Maintains functionality of AddWorldTip to be removed later
- Can be tied to an entity or abs vector
- Can be removed early by passing ent or vector
- All tips can be cleared
- Custom colour and display time can be specified
- Also updates draw function to operate more efficiently


TheFreeman193:
garrynewman/garrysmod#864

Bo98:
It is unfortunate that the old function is called AddWorldTip and the new one is called SetWorldTip when their functionalities are really reversed. A few ideas:

* Perhaps call them AddWorldTooltip, RemoveWorldTooltip and ClearWorldTooltips. This however breaks consistency with the the PaintWorldTips GM function. It is only a GM function though and not called as a hook. It is also undocumented so could maybe be renamed.
* Another option could be to keep the "worldtip" name but to move it to a module/library so that the functions are worldtip.Add, worldtip.Remove and worldtip.ClearAll/worldtip.RemoveAll. You could also take this opportunity to make it support base gamemodes too. The PaintWorldTips GM function could also be converted to a proper hook and can be documented. I prefer this option.

Also, you should probably convert existing calls to AddWorldTip in the repo.

NeatNit:
There is also a problem with having two World Tips in the same position. Say, for example, that I have a Weight tool which shows, via World Tip, the looked-at entity's current weight. What happens if I point it at a thruster? That thruster would draw its own World Tip (Force and, in multiplayer, owner) and the weight tool would draw a different World Tip (weight), but both would be assigned to the same entity and would have the same position. One of them will inevitably be hidden behind the other.

I assume this is why the current code allows only one at a time. There are better ways to overcome this.

For example: every frame, if there's at least one World Tip to draw, run a hook called, say, HUDDrawWorldTips. It would have a parameter, the worldtips table or a copy thereof, which the hook functions can manipulate. If the hook returns true, use the modified worldtips table.

This way my weight tool can prevent any other World Tip from drawing while it tries to draw its own, without succumbing into timing-related problems.

TheFreeman193:
@Bo98 I had thought about modularising this; it'd certainly be a lot neater.
@NeatNit That would preserve existing tips, but it might be better to stack them. If you were displaying an alternate tooltip and further calls were made, you'd have to handle those accordingly. Of course, stacking them presents the issue of whether a limit is needed.

TheFreeman193:
I will address the issue with multiple instances on the same entity in a later commit, and squash them up at the same time.

I'm leaning towards stacking them as offsets from the local screen pos.

garrysmod/lua/includes/modules/worldtip.lua
```diff
@@ -0,0 +1,56 @@
...
+
+local WorldTips = {}
+
+function GetTable()
+	return WorldTips
+end
+
+function Add( name, tgt, text, len, col )
+
+	local ent, pos = IsValid( tgt ) && tgt
```
NeatNit: How does this line work? :confused:

TheFreeman193: (You selected the line correctly)
*ent* either becomes *tgt* or *false* depending on the validity of *tgt*. *pos* is localised as *nil* to be initialised later. It's effectively the same as:
```lua
local ent = IsValid(tgt) && tgt || false
local pos
```
Just to note, I've written a system for spacing tooltips out and for per-frame calls so that AddWorldTip can be deprecated; I'll push it later this week once I've checked it over for issues.

garrysmod/lua/includes/modules/worldtip.lua
```diff
@@ -0,0 +1,56 @@
...
+
+	if ent then
+		if ent:IsWorld() then return end
+		ent:CallOnRemove( "RemoveEntityTooltip", function() WorldTips[name] = nil end )
+	else
+		if !isvector(pos) then return end
```
wiox: Will this always evaluate to true?

garrysmod/lua/includes/modules/worldtip.lua
NeatNit: I blame github for a terribad UI that doesn't convey to the user what is happening.

Anyway, I recommend changing it to:
```lua
local pos
local ent = IsValid( tgt ) && tgt
```
Simply because it's a lot more readable. That trick is confusing as hell. Even if you disagree, please remember that most gmod coders are going to be Lua newbies and some _might_ read over this code. Best not to confuse them as it confused me.

garrysmod/lua/includes/modules/worldtip.lua
Bo98: Looks like it should be `!isvector(tgt)`.

TheFreeman193: This is rectified in my working tree.

Bo98: Should probably commit that soon along with fixing the merge conflict.

TheFreeman193: Yeah, I've been too busy to look at this recently. I'll dust it off later if I get an hour.

TheFreeman193:
Have a bit of time, now. It looks like I wrote a system for handling multiple tips and never pushed it. Looks a little complex but it seems to work. I'll resolve the conflicts and push shortly.

PR #867, opened 2015-02-10
Removed useless require calls

Kefta:
Timer lua implementation was removed, and net is not a module (should be moved to extensions)


Bo98:
Probably should move net above undo in init.lua as this probably breaks the net.Receive calls.

https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/includes/init.lua#L33

Kefta:
@Bo98 #798 Already takes care of this; just cleaning up the calls he didn't already do

PR #869, opened 2015-02-12
Make OnUndo get called by hook.Call

meharryp:
Re-make of #863 since I messed up the pull.

### DIFF
--- a/garrysmod/lua/includes/modules/undo.lua
+++ b/garrysmod/lua/includes/modules/undo.lua
@@ -370,9 +370,9 @@ function Do_Undo( undo )
 	
 	if (count > 0) then
 		if ( undo.CustomUndoText ) then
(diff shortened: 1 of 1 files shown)


KateAdams:
Why is it so many people have their branches named "patch-#n", rather than a descriptive branch name (i.e. "onundo-use-hook")?

robotboy655:
I think that is the standard name given when you use this website to submit PRs.

wiox:
We need more imagination. I expect to see more "patch-penis-saggs" in the future.

KateAdams:
So, they're using the website to make changes, and then not testing them?  That's crazy!

Bo98:
Common in GMod repos sadly.

TheFreeman193:
This is now included in a bigger pull for all SendLua GM calls #878

meharryp:
#878 makes this obsolete.

PR #870, opened 2015-02-12
Merged menu.lua and init_menu.lua

Kefta:


wiox:
This breaks menu_reload.

Kefta:
How does menu_reload work?

wiox:
It sets up the Lua state, includes includes/init_menu.lua, derma/init.lua, includes/vgui_base.lua, skins/default.lua, includes/menu.lua and then calls the MenuStart hook.

Kefta:
Can't this be changed to not call menu.lua anymore? I see no reason to call a file with a single include when another file already calls them all.

wiox:
Does it matter?

Kefta:
Just seems pointless and messy. Guess it doesn't really matter

PR #871, opened 2015-02-13
Adds util.IsIndexable and improves table.SortByMember.

willox:
util.IsIndexable returns true if a value can be indexed. i.e. Lua will actually attempt to index it either as a table or via metamethods.

table.SortByMember can now sort non-table yet indexable values (such as players) as well as sort by methods. It also returns the table it has sorted for those wacky one-liners.

```lua
-- Prints a list of players in ascending order by name
for k, v in pairs( table.SortByMember( player.GetAll(), "Name", true ) ) do

	print( v )

end

-- Prints a list of players in order of most score
for k, v in pairs( table.SortByMember( player.GetAll(), "Frags" ) ) do

	print( v, v:Frags() )

end
```

Crap like that. Any code that manages to use this already without error will continue to work 100% the same.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -356,41 +356,56 @@ function table.ForceInsert( t, v )
 end
 
 --[[---------------------------------------------------------
-	Name: table.SortByMember( table )
-	Desc: Sorts table by named member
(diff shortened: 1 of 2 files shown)


zerfgog:
:+1:

wiox:
This has been the most uneventful set of GitHub notifications I've ever read.

PR #872, opened 2015-02-15
Add SWEP.IconLetter to TTT SWEPs

zerfgog:
This variable is used in the font for kill icons.
(See: http://wiki.garrysmod.com/page/CS:S_Kill_Icons)


wiox:
Are these used anywhere?

zerfgog:
No; these are only for support for other addons that make use of them.

wiox:
Fair enough

PR #874, opened 2015-02-18
Properties can only be used on weapons when unheld

wiox:
This stops people joining innocent prop-protection-less sandbox servers and
removing everybody's weapons.
Instead they can only instantly remove every prop on the map! Oh well.

The only alternative is saving right-click state server-side and that's quite gross. 
Install prop-protection, kids.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/shared.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/shared.lua
@@ -235,6 +235,13 @@ function GM:CanProperty( pl, property, ent )
 
 	end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
I agree. Properties should be changed so that the server does a trace on the calling player instead of simply the player sending a net message with an entity.

robotboy655:
@thegrb93, I tried the trace, didn't work. I guess adding a distance check can make it a little bit better

wiox:
Either way you don't want people removing other people's physguns while they are using them :v

PR #875, opened 2015-02-23
Update and rename garrysmod/gamemodes/terrortown/gamemode/lang to garrys...

pupu2314:
...mod/gamemodes/terrortown/gamemode/lang/TradChinese


TheFreeman193:
Nice work! You'll need to set the extension to **.lua** however, else TTT won't include it. It's probably also a good idea to have the filename in lowercase.

PR #877, opened 2015-02-23
Update scripted_ents.lua

NeatNit:
Why on earth would you call Get twice here?

### DIFF
--- a/garrysmod/lua/includes/modules/scripted_ents.lua
+++ b/garrysmod/lua/includes/modules/scripted_ents.lua
@@ -149,14 +149,16 @@ function Get( name )
 	
 	-- Derive from base class
 	if ( name != SEntList[ name ].Base ) then
-	
-		if (!Get( SEntList[ name ].Base )) then
+		
+		local base = Get( SEntList[ name ].Base )
+		
+		if (!base) then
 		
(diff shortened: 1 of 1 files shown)


wiox:
I'm happy to acknowledge that when my scripts load they will be free of a few unnecessary operations.

NeatNit:
That sounds like sarcasm ಠ_ಠ

While I agree that this change is very minor - you don't know how people might be using this function. If you have a sent based on a sent based on a sent, You'll be calling Get 7 times instead of 3.

wiox:
That still wouldn't be an issue. It's not a bad change, though. I'm all for good code.

garrysmod/lua/includes/modules/scripted_ents.lua
```diff
@@ -149,14 +149,16 @@ function Get( name )
 	
 	-- Derive from base class
 	if ( name != SEntList[ name ].Base ) then
```
comedinha: SEntList[ name ].Base

wiox: If you're suggesting the if statement here isn't neccesary, I assume it's so that we don't get any infinite loops on the base entities (base_anim, base_point and base_brush).

TheFreeman193:
I'm surprised, but this actually seems to be about 32-35% quicker.

PR #878, opened 2015-02-24
GM hook calls

TheFreeman193:
Cleans up gamemode hook calls using SendLua. Prevents errors in non-sandbox gamemodes.
* All GM method calls via SendLua now use hook.Run

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/player_extension.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/player_extension.lua
@@ -81,7 +81,7 @@ end
 
 function meta:LimitHit( str )
 
-	self:SendLua( "GAMEMODE:LimitHit( '".. str .."' )" )
+	self:SendLua( 'hook.Run("LimitHit","' .. str .. '")' )
 
 end
 
@@ -110,7 +110,7 @@ if (SERVER) then
 		self.Hints = self.Hints or {}
 		if (self.Hints[ str ]) then return end
 		
-		self:SendLua( "GAMEMODE:AddHint( '"..str.."', "..delay.." )" )
+		self:SendLua( 'hook.Run("AddHint","' .. str .. '","' .. delay .. '")' )
 		self.Hints[ str ] = true
 
 	end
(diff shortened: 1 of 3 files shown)


wiox:
What's wrong with hook.Run?

TheFreeman193:
Nothing as such, but it only gets the gamemode with gmod.GetGamemode and then passes everything onto hook.Call. It's good for module environments where the GAMEMODE global isn't directly accessible, but for anything in the standard environment, it's quicker and simpler just to use hook.Call.

wiox:
The GAMEMODE global is completely useless for gamemodes. Use self.

TheFreeman193:
As far as I am aware, both work inside GM methods. You are right in that I should be using self, however; I'll swap it in a sec.

wiox:
If I sound like an asshole it's because I don't like your change. I just don't understand why people favour worse code for such a small optimization.

The SendLua changes are fine.

TheFreeman193:
Is there any particular benefit, other than that it looks a tad nicer? It's the sum of small differences that produces the biggest effects, and I doubt the interpreter (or users) give a shit how pretty the code is.

_Saving the world, one line of ugly code at a time_

wiox:
Pretty code is important. If you want a faster game go optimize the widget system's PlayerTick hook. That's a real issue.

TheFreeman193:
_Readable_ code is important. An extra arg is worthwhile to avoid two needless calls, in my opinion.

(Left GAMEMODE on line 365 in *cl_init.lua* to remain compatible with #827)

TheFreeman193:
Since I'm making changes to cl_init on the base GM in #865, I'm going to remove the changes from this one. While I'm at it, what are your thoughts on this, @robotboy655? hook.Run or hook.Call? There's no point in changes if noone approves.

robotboy655:
Before I look into this, remove all the whitespace changes from all your commits. It clogs up the things you actually want to change. I said numerous times that you shouldn't bother removing whitespace from entire file. I will do it myself once I deal with most of the PRs.

robotboy655:
Don't change hook.Run to hook.Call, but you may change it other way around.

TheFreeman193:
I will do. Okay, that settles it. 

Oh, and no hard feelings @wiox, I don't take things personally.

TheFreeman193:
(Sorted & rebased)

thegrb93:
I hate the SendLua functions being used for hints. Slow and potentially exploitable.

TheFreeman193:
I'm not especially fond of it either. I reckon it'd be nice to have a global function serverside, using net to call hooks on clients, but this is something to think about after the update and when all these pulls have been sorted out.

PR #879, opened 2015-02-24
Entity:HookOutput( ), Entity:UnhookOutput( )

Kogitsune:
Creates Entity:HookOutput( name, uid, func ) and Entity:UnhookOutput( name, uid ) to catch entity outputs, especially engine ones. Conforms to the behavior of hooks, where passing the entity as the uid causes it to be passed to the function.

Could obviously use some improvement, but it is currently the only vanilla way to do this that I know.

### DIFF
--- a/garrysmod/lua/entities/info_output_hook.lua
+++ b/garrysmod/lua/entities/info_output_hook.lua
@@ -1,7 +1,5 @@
 DEFINE_BASECLASS( "base_point" )
 
-ENT.Hooks = { }
-
 local this, meta, hooks, fmt
 
 hooks = { }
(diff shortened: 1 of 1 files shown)


wiox:
I'm not sure about your EntityRemoved hook. The hook will not be called if the "self" parameter isn't valid, because that is when it'll be removed.

Should you be using ENT:OnRemove instead? If so, removing that hook would make this code good in my books.

Kogitsune:
A month later, you are right. on most parts - the k, v is force of habit so I removed it. The extra hook was removed and replaced with OnRemove.

The argument shifting is to match the expected behavior of hook.Add - if you pass an entity as the second parameter, it is passed to the function as the first argument.

`ent1:HookOutput( "OnDamaged", ent2, ent2.OnEnt1Damaged )`

versus

`ent1:HookOutput( "OnDamaged", "When My Thing Gets Hurt", function( ) ... end )`

wiox:
Fair enough :v:

robotboy655:
This is a good idea, but I do not like usage of dummy entity for this purpose. 
Perhaps this needs to be done on the C++ side.

PR #880, opened 2015-03-01
Add lua_findhooks

TheFreeman193:
Adds new ccmds **lua_findhooks** and **lua_findhooks_cl** to behave similarly to *lua_find*.
* Allows quick searching of all registered hooks by hook name & identifier
* Integrated into current ccmd callback
* Usage is: *lua_findhooks(_cl) \<hookname> \<hookid>*

Additional changes:
* Changes use of type() to isstring()
* Dev funcs now work in SP regardless of usergroup NWString

**Examples**
``` lua_findhooks render frame ``` - Finds all hooks with *render* in the hook name and *frame* in the identifier.
``` lua_findhooks_cl "" player ``` - Finds all clientside hooks of any name with *player* in the identifier.
``` lua_findhooks think ``` - Finds all think hooks.

### DIFF
--- a/garrysmod/lua/autorun/developer_functions.lua
+++ b/garrysmod/lua/autorun/developer_functions.lua
@@ -10,7 +10,7 @@ local function FindInTable( tab, find, parents, depth )
 	
 	for k, v in pairs ( tab ) do
 	
-		if ( type(k) == "string" ) then
+		if ( isstring(k) ) then
 		
 			if ( k && k:lower():find( find:lower() ) ) then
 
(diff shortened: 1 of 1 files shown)


neico:
looking good imo,
saves some time with crawling trough the gamemode table via PrintTable

thegrb93:
Does this work regardless of sv_allowcslua?

TheFreeman193:
This works in the same way as lua_find; it works only for admins in MP and always in SP. Both are independent of sv_allowcslua.

DrVrej:
Support

TheFreeman193:
@robotboy655 Do you think it's worth merging this now?

PR #881, opened 2015-03-01
Cleaned BBcode tags in addons list

EstevanTH:
I added some code that removes the very annoying BBcode tags from the descriptions of addons.
You are welcome if you can improve the pattern that escapes **`[img]`** (the end test is **`[`** instead of **`[/img]`**).


wiox:
You need to use multiple branches to separate your PRs

robotboy655:
1) You can and should use 2 regexps for this ( BB and HTML ), not a predefined list of bb code tags. 
2) 5 commits is a bit too excessive

EstevanTH:
1) Many thanks *wiox*! I was totally lost when I saw my updates going to an existing PR.
2) *robotboy655* I do not understand what you mean but okay... Anyway a fixed list of tags has to be defined somewhere otherwise tags that are not BB would be hidden while they shouldn't.
3) Only 1 commit is about this pull request, others are for another one.
4) It's a pleasure that my work was sent to the dustbin within an hour. :disappointed_relieved:

TheFreeman193:
You need to push commits for significantly different things to different branches and have a pull request for each. Robot is talking about regular expressions, have a look at [this](http://wiki.garrysmod.com/page/Patterns).

EstevanTH:
Okay, that was not easy to understand. :unamused:
Oh, was he talking about using 1 single pattern instead of my big list??

Anyway my modification has been denied. Should I work and then post a new (clean) request?

TheFreeman193:
Yeah, one for bbcode tags, and another for html tags. And yes, once you've made a new branch with just the commits for this change, you can submit a pull comparing that branch.

EstevanTH:
As far as I know, there are no supported HTML tags in addon descriptions... Why would I do anything with HTML tags? :confused: Is there a required HTML escape that has not been done??? (I just tried: HTML tags do not need any escape.)

What do you think guys? Should I give up with pull requests because I do not understand the requirements that allow to bring some work? :sob: *robotboy655* is soooo rude with me on the Wiki and on GitHub as well. :broken_heart:

robotboy655:
It doesn't matter if they are supported on Steam or not though, does it? They shouldn't appear in the description, since the space in-game is so limited.

You should not give up, you should do it properly. I am telling how to do it properly. I don't want to merge bad coding practices. I am sorry if I offended you in any way, I didn't mean to. I am just trying to do my job.

EstevanTH:
Thank you, I will do the work properly in a few days.

HTML tags are ignored by both the game and Steam, so nobody uses them. To me, trying to remove them is just a waste of time. For this reason, if HTML tags are present then they are most probably supposed to be displayed as plain text... :weary:

EstevanTH:
I could find no way to do an OR operation in capture parentheses or something similar. Those operations only exist for single characters.
I remind that Lua patterns are simplified patterns unlike those from POSIX or REGEX. They do not support syntaxes like `(foo|bar)` because the `|` is not defined as a magic character as far as I know.

References:
[http://www.lua.org/manual/5.2/manual.html#6.4.1](Patterns - Lua 5.2 manual)
[http://lua-users.org/wiki/PatternsTutorial](Patterns Tutorial - lua-users.org)

So I will keep going with a clean solution that only uses the native function `string.gsub()`. But I cannot make it with less function calls than I had in my previous solution (the tags list is **required**). If you find a solution that can reduce the number of function calls then you are welcome, give an example please.

TheFreeman193:
This would be so much easier with regex, I grant you. It's perfectly possible, though, with captures. This seems to do the trick:
```lua
result.description = string.gsub(result.description, "%[(%w+)[^%]]*](.-)%[/%1]", function(cap, int) return int end)
```

EstevanTH:
Ahhh yes the 3rd argument of `string.gsub()` can be a function!! :grimacing:
Many thanks. :heart:

TheFreeman193:
Yeah, each capture gets passed as an argument. Just make sure for the HTML pattern that you return an empty string as there are no captures in that one. No problem :smiley:

garrysmod/lua/includes/util/workshop_files.lua
```diff
@@ -4,6 +4,56 @@ local InfoCache			= {}
 local VoteCache			= {}
 local ListCache			= {}
 
+local function ClearDescription( desc )
```
thegrb93: CleanDescription would be a better name.

EstevanTH:
I couldn't do it in one single pattern. I think that *TheFreeman193*'s example may be unsafe if tags are nested into other ones, for example `[url=https://github.com/][b]GitHib[/b][img]https://avatars2.githubusercontent.com/u/928843?v=3&s=60[/img][/url]`.
By using `string.gsub()` the way I did, I get sure that such conflicts are avoided.
```lua
local BB_remove_tags = { "h1", "/h1", "b", "/b", "u", "/u", "i", "/i", "strike", "/strike", "spoiler", "/spoiler", "noparse", "/noparse", "url", "/url", "code", "/code" }
local function BB_remover_tags( cap )
	cap = string.lower( cap )
	for _,tag in ipairs( BB_remove_tags ) do
		if cap == tag then
			return ""
		end
	end
	if string.sub( cap, 1, 4 ) == "url=" then
		return ""
	end
end
local function BB_remover_img( cap )
	cap = string.lower( cap )
	if cap == "img" then
		return ""
	end
end
local function CleanDescription( desc )
	-- Uncomplete list: http://steamcommunity.com/comment/Announcement/formattinghelp
	desc = string.gsub( desc, "%[([^%]]*)%]", BB_remover_tags )
	desc = string.gsub( desc, "%[([%a]+)%]([^%[]*)%[(/[%a]+)%]", BB_remover_img )
	return desc
end
```

**Note:** I will not touch HTML tags because I do not know what HTML tags to remove because none are supported anywhere.

TheFreeman193:
A simpler way would be to use the second return from string.gsub. It reports how many matches were made- you can just repeat until there are no more replacements, like this:
```lua
local function CleanDescription( desc )
	repeat
		desc, rep = string.gsub(desc,"%[(%w+)[^%]]*](.-)%[/%1]", function(_, inner)
			return inner
		end)
	until rep == 0
	return desc
end
```

EstevanTH:
Either I'm dumb or your script will be unhappy if I put custom non-BB tags like `[MR]`, which have no closing tag. :anguished: This makes me unable to correctly understand what your script can capture.

Otherwise it's a good source of inspiration. Thanks for the lesson! :heart:

Who thinks that my last attempt is correct, clean, efficient, well made, ready? :disappointed: I'm not experienced enough to make it perfect but I think that it is rather good.

wiox:
```
local tags = {
	b = true,
	i = true,
	u = true,
	h1 = true,
	img = true,
	url = true,
	code = true,
	strike = true,
	noparse = true,
	spoiler = true
}

print( string.gsub( test, "%[/?([^%[%]]-)%]", function( bb_inner )
	local tag = bb_inner:lower():match("[^=]*")

	if tags[tag] then
		return ""
	end
end ) )
```

Test: http://i.imgur.com/bXtXVgF.png
The contents are all tags are - purposefully - stored.

PR #882, opened 2015-03-01
Radar improvements

Kefta:
Saves drawing target if it isn't visible


TheFreeman193:
Good idea, but the first change (L74) needs to go below L78, else it'll break the shrink functionality.

Kefta:
Good catch, thanks

PR #883, opened 2015-03-01
Efficiency in DmgToText and DmgToMat

Kefta:
How is DMG_DIRECT and DMG_BURN different?


wiox:
I believe DMG_DIRECT is when you are on fire and taking damage, while DMG_BURN is taking burn damage from another source.

PR #885, opened 2015-03-04
Improve error message when indexing string badly (with key not within the string library)

wiox:
The current error message is very difficult for new (and old) developers to understand and also doesn't resemble any error messages in stock Lua.

```lua
local str = "Hello!"; print( str:sab( 1, 5 ) )

--[[
    New Error: [ERROR] lua_run:1: attempt to index a string value with bad key ('sab' is not part of the string library)
    Old Error: [ERROR] lua_run:1: bad key to string index (number expected, got string)
]]
```

I'd almost be in favour of following the way stock Lua handles this and returning nil, but it'd a better idea to stick to current functionality of gmod.

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -307,7 +307,7 @@ function meta:__index( key )
 	elseif ( tonumber( key ) ) then
 		return self:sub( key, key )
 	else
-		error( "bad key to string index (number expected, got " .. type( key ) .. ")", 2 )
+		error( "attempt to index a string value with bad key ('" .. tostring( key ) .. "' is not part of the string library)", 2 )
 	end
 end
 


KateAdams:
>I'd almost be in favour of following the way stock Lua handles this and returning nil, but it'd a better idea to stick to current functionality of gmod.

I do wonder if this would be better.

Changing the definition to:

```lua
function meta:__index( key )
	local val = string[ key ]
	if ( val ) then
		return val
	elseif ( tonumber( key ) ) then
		return self:sub( key, key )
	else
		return nil
	end
end
```

The example error message would now become:

	attempt to call method 'sab' (a nil value)

As a side-effect the following code now also works as expected (will crash under gmod, but not typical Lua):

```lua
local x = "asd"
if x.field then
	print("exists")
else
	print("does not exist")
end
```

Currently, this will crash with:

	attempt to index a string value with bad key ('field' is not part of the string library)

PR #886, opened 2015-03-05
Added unpack functions

Kefta:
For angles, colors, and vectors

### DIFF
--- a/garrysmod/lua/includes/extensions/angle.lua
+++ b/garrysmod/lua/includes/extensions/angle.lua
@@ -1,11 +1,11 @@
-local meta = FindMetaTable( "Angle" )
+local ANGLE = FindMetaTable( "Angle" )
 
 -- Nothing in here, still leaving this file here just in case
 
 --[[---------------------------------------------------------
 	Angle Snap to nearest interval of degrees
 -----------------------------------------------------------]]
-function meta:SnapTo( component, degrees )
+function ANGLE:SnapTo( component, degrees )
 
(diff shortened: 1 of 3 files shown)


notcake:
Angle:Unpack () and Vector:Unpack () should be implemented in C++ for speed, since the use of .p,.y,.r/.x,.y,.z would call Angle/Vector:__index () 3 times. Also Matrix:Unpack () and corresponding :SetUnpacked (...) functions would be nice.

Kefta:
Yeah, I agree.

KateAdams:
Eh, it depends upon where it's used.  Crossing the C boundary is expensive--like crossing into kernelspace from userspace.  @garrynewman found this out when he implemented the hook system in C++.

notcake:
Ignore this comment if it was already obvious: Angle / Vector:__index() is a Lua -> C call. The Lua implementation of :Unpack() would cause 3 transitions, 1 for each component. The C++ implementation would cause only 1.

KateAdams:
@notcake I wasn't aware vector, angle, and colours overload the `__index` metamethod; if that's the case, then yes, you're correct, it would pass the C boundary 3 times minimum.

KateAdams:
Oh, also, you should set n to 3 in the return value to match the behavior of `table.pack()`:

```lua
print(table.pack(1, 2, 3).n) -- 3
print(table.pack(nil, 3).n)  -- 2
```

Kefta:
@KateAdams Colours don't; they're pure-lua implementation (for some reason)

Kefta:
Going to close this and only make COLOR:Unpack since the rest can be made in C++.

PR #887, opened 2015-03-05
Removed useless TTT util functions

Kefta:
I never knew UnbridledHateForULX existed, made me laugh a bit :V:

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang_shd.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang_shd.lua
@@ -15,7 +15,7 @@ for _, fname in pairs(files) do
    local path = "lang/" .. fname
    -- filter out directories and temp files (like .lua~)
    if string.Right(fname, 3) == "lua" then
-      util.IncludeClientFile(path)
+      if SERVER then AddCSLuaFile(path) else include(path) end
       MsgN("Included TTT language file: " .. fname)
    end
 end
--- a/garrysmod/gamemodes/terrortown/gamemode/lang_shd.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang_shd.lua
@@ -6,7 +6,7 @@
 if LANG then return end
 LANG = {}
 
-util.IncludeClientFile("cl_lang.lua")
+if SERVER then AddCSLuaFile("cl_lang.lua") else include("cl_lang.lua") end
 
 -- Add all lua files in our /lang/ dir
 local dir = GM.FolderName or "terrortown"
--- a/garrysmod/gamemodes/terrortown/gamemode/util.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/util.lua
@@ -1,11 +1,12 @@
--- Random stuff
-
 if not util then return end
 
+local util = util
 local math = math
 local string = string
 local table = table
 local pairs = pairs
+local table = table
+local band = bit.band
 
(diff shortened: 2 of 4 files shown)


garrysmod/gamemodes/terrortown/gamemode/lang_shd.lua
```diff
@@ -6,7 +6,7 @@
 if LANG then return end
 LANG = {}
 
-util.IncludeClientFile("cl_lang.lua")
+if SERVER then AddCSLuaFile("cl_lang.lua") else include("cl_lang.lua") end
```
Bo98: You forgot about line 18.

Kefta:
Is there a way to squash commits from the website? I'm short of a computer that I can use git on for the time being

Bo98:
Not that I'm aware of.

garrysmod/gamemodes/terrortown/gamemode/util.lua
```diff
@@ -1,11 +1,12 @@
...
-
 if not util then return end
 
+local util = util
 local math = math
 local string = string
 local table = table
 local pairs = pairs
+local table = table
+local band = bit.band
```
TheFreeman193: Couldn't you just modify L305 to use bit.band?

Kefta: I'm not sure if it actually has any efficiency benefits, but they're all over TTT, so I won't touch them unless all of them are changed

garrysmod/gamemodes/terrortown/gamemode/lang_shd.lua
Kefta: Fixed, thanks

Kefta:
I'm sorry for the amount of commits; I modified my master branch on my fork by accident, so I don't have the ability to squash them

svdm:
You've mixed in a lot of changes that are opinion rather than clear improvements.

* Replacing a bunch of IncludeClientFile calls with oneline if statements, just to remove 7 lines of code? That's replacing code with more code.
* Similarly, your change of GetRTranslation just ends up duplicating the role_strings lookup. What's the gain exactly?
* The cases of a local being set for a module function (e.g., `local exp = math.exp`) save a dictionary lookup when that function is called, which has benefits if that function is called very often. They have a purpose and are not exactly harmful, so why remove them?

Really the only useless function this PR removes is the ULX one.

Kefta:
1 + 2. It's still better than creating a global function to only be used in only file. In passthrough's case, it is only used one time. I could see the benefit of putting IncludeClientFile as a local function in that file since it's not used anywhere else, but passthrough is definitely unnecessary.

3. That's my fault, I thought I fixed but it didn't appear that the commit went through. 

I'll redo the PR if you tell me what to do about IncludeClientFile and passthrough

Kefta:
1+2. But they're only used in one file and you made them global functions. I see the benefit of adding IncludeClientFile as a local function to lang_shd, but passthrough has no use whatsoever.
3. I thought I fixed that; I guess my commit didn't go through. I'll revert those changes once you tell me what to do about IncludeClientFile and passthrough.

svdm:
Global? They're in a module, specifically util, where they are quite harmless. Both of them should be left alone, simply because removing them only achieves making the code more verbose.

util.passthrough is the identity function, ie. it returns its argument. This is a common concept in functional programming languages where you are often passing functions around into other functions. It's useful in the same way that, sometimes, multiplying a number by 1 is useful.

PR #888, opened 2015-03-06
Add player.GetBySteamID and player.GetBySteamID64

zerfgog:
We already have player.GetByUniqueID; having these functions is nice.
Unlike #846, the implementation of these functions is not subjective.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -31,9 +31,9 @@ function meta:__index( key )
 		local val = tab[ key ]
 		if ( val != nil ) then return val end
 	end
-	
+
 	return nil
-	
+
 end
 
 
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/player.lua
```diff
@@ -52,7 +52,43 @@ function player.GetByUniqueID( ID )
...
+
+	return false
+
+end
+
+function player.GetBySteamID( ID )
+
+	ID = string.upper( ID )
+
+	assert( string.find( ID, "STEAM_" ), "player.GetBySteamID: passed string '" .. ID .. "' is not a valid steamid!" )
```
TheFreeman193: I recommend you use ID:StartWith("STEAM_") here; it's more relevant to SteamIDs, and it's also considerably quicker.

zerfgog:
ID:StartWith("STEAM_") or string.find(ID, "^STEAM_")?

TheFreeman193:
The former; I see no need for patterns/regex here since we're only looking for absolute text values.

wiox:
If GetByUniqueID doesn't emit errors I believe GetBySteamID and GetBySteamID64 shouldn't either.

zerfgog:
Squashed commits.

PR #890, opened 2015-03-10
Allow spectators to talk in all chat -- redirects to team

Kefta:
Also, allows innos to use team chat. Both redirect to team chat and all chat respectively, but it's beneficial in situations such as a person getting killed while typing and they lose their message after they send it; or someone trying to quickly type who the killer is before they are killed, but they hit the wrong key.

This still supports the mumbling and spec all chat systems

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_keys.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_keys.lua
@@ -82,11 +82,6 @@ function GM:PlayerBindPress(ply, bind, pressed)
             GAMEMODE.ForcedMouse = true
          end
       end
-   elseif bind == "messagemode" and pressed and ply:IsSpec() then
-      if GAMEMODE.round_state == ROUND_ACTIVE and DetectiveMode() then
-         LANG.Msg("spec_teamchat_hint")
-         return true
-      end
    elseif bind == "noclip" and pressed then
       if not GetConVar("sv_cheats"):GetBool() then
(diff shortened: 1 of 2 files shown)


svdm:
I like the idea. One thing though: what happens if I teamchat something like `'"bla" ; kill'`, resulting in `ply:ConCommand('say "bla" ; kill')` being executed here?

Jvs34:
If code_gs replaces the ConCommand "say" and "say_team" with http://wiki.garrysmod.com/page/Player/Say then nothing would happen

robotboy655:
Wouldn't it be a better idea to just use http://wiki.garrysmod.com/page/GM/PlayerCanSeePlayersChat for this, and not some weird hacky workarounds?

KateAdams:
@robotboy655 I agree; this shouldn't be incorporated using this method.  If re-done properly--using the hook--then yes, this would be good.

Kefta:
@Robotboy655 Can this override teamchat restrictions? Like if I do something this:

```
function GM:PlayerCanSeePlayersChat(text, team, listen, ply)
  if teamchat then
    if ply:GetRole() == "Traitor" and listen:GetRole() == "Tratior" then return true end
    if ply:GetRole() == "Detective" and listen:GetRole() == "Detective" then return true end
  end
  .....
  return false
end
```

Kefta:
@svdm Ready to be merged

Kefta:
@svdm Hello?

PR #892, opened 2015-03-10
Update dform.lua

Azarym:
self.m_strConVar is nil.
Cause [ERROR] lua/vgui/dform.lua:132: bad argument #1 to 'RunConsoleCommand' (string expected, got nil)

### DIFF
--- a/garrysmod/lua/vgui/dform.lua
+++ b/garrysmod/lua/vgui/dform.lua
@@ -128,9 +128,10 @@ function PANEL:ComboBox( strLabel, strConVar )
 	local right = vgui.Create( "DComboBox", self )
 	right:SetConVar( strConVar )
 	right:Dock( FILL )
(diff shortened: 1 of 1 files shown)


wiox:
You don't fix an error by blindly removing code

Kefta:
https://github.com/garrynewman/garrysmod/blob/e7111a4ab8318b69c3a29e2cbea6cc57ec0bcf5a/garrysmod/lua/derma/init.lua#L97

It's not nil; your code is sending in an invalid convar

Azarym:
The problem is when you create the combobox without concommand
Example: DForm:ComboBox("Test")

Its working with other panel because the "OnSelect" isn't overrided by DForm

robotboy655:
You are not really supposed to use that function without a second argument, but oh well, I will fix it.

PR #893, opened 2015-03-14
removed unnecessary calls

Leystryku:
PANEL:Setup already handles that


robotboy655:
It is necessary in case a player changes his name during speaking.

PR #894, opened 2015-03-15
Prevent invalid fonts from being sent into SimpleText

Kefta:
Yes, it's hacky. If there was a way to cache already created fonts in some table, or if there was some kind of IsValidFont function, this would be a lot simpler.

This prevents the "Invalid font sent into CreateFont" error being spammed every frame, and instead just changes the font

### DIFF
--- a/garrysmod/lua/includes/modules/draw.lua
+++ b/garrysmod/lua/includes/modules/draw.lua
@@ -13,6 +13,7 @@ local ScrH = ScrH
 local Color = Color
 local tostring = tostring
 local color_white = color_white
+local pcall = pcall
 
 module( "draw" )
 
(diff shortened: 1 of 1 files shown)


wiox:
i'd rather have an error

Kefta:
The problem is that it errors every frame you look at the font

KateAdams:
Failing silently really isn't the best way to go.

Kefta:
I'll put an ErrorNoHalt call in the pcall failure then

robotboy655:
This is a completely pointless change. First of all, your PR doesn't even work:
```

[ERROR] lua/includes/modules/draw.lua:75: attempt to call global 'ErrorNoHalt' (a nil value)
  1. SimpleText - lua/includes/modules/draw.lua:75
   2. v - LuaCmd:1
    3. unknown - lua/includes/modules/hook.lua:84
```

If I fix that, using an invalid font results in this:
image

And there are a few problems with this:
1) You don't know where the invalid font is used
2) It doesn't notify the developer that there's an error, causing confusion

Stop trying to hide errors or warnings. They must be fixed, not hidden.

PR #895, opened 2015-03-15
Add concommand.Exists

zerfgog:
I was surprised to find that concommand.Exists didn't exist :V


KateAdams:
Why do you have brackets around the return statement?

zerfgog:
Habit from more complex logic/isolating one return value; I can remove them if it's a problem

zerfgog:
I've come to decide that this is both misleading and useless. Misleading due to the fact that non-lua concommand would return false here, and useless as I can't think of a reason where you'd need to check the existence of a concommand.

wiox:
Good decision

PR #896, opened 2015-03-18
Change TTT's silenced pistol's model to the one with a silencer

zerfgog:
It makes sense, right?
Changed model from w_pist_usp to w_pist_usp_silencer


svdm:
Pretty sure it used to have the silencer, guess something changed about the world model.

DaaOtt:
I always thought it was a stealth thing.

PR #897, opened 2015-03-18
Remove the need for DEFINE_BASECLASS in player classes

GamerGambit:
Very small and simple change that removes the need for DEFINE_BASECLASS in player classes. I find it convenient however this change is a tad costly. It is fully backwards compatible, but if developers remove the DEFINE_BASECLASS from their class files, they will have to replace BaseClass with self.BaseClass.

### DIFF
--- a/garrysmod/lua/includes/modules/player_manager.lua
+++ b/garrysmod/lua/includes/modules/player_manager.lua
@@ -313,7 +313,7 @@ function RegisterClass( name, table, base )
 
 		if ( !Type[ name ] ) then ErrorNoHalt( "RegisterClass - deriving "..name.." from unknown class "..base.."!\n" ) end
 		setmetatable( Type[ name ], { __index = Type[ base ] } )
-
+      table.BaseClass = baseclass.Get(base);
 	end
 
 	if ( SERVER ) then


wiox:
BaseClass upvalues are used over BaseClass properties to stop potential infinite loops from occuring. The following code shows how a derivation base of more than 2 classes can result in infinite loops when using the BaseClass property.

```lua
-- Base
    function GM:Initialize()
        -- Do things
    end

-- Sandbox
    function GM:Initialize()
        self.BaseClass.Initialize( self ) -- Also calls Sandbox.Initialize
    end
-- DarkRP
    function GM:Initialize()
        self.BaseClass.Initialize( self ) -- Calls Sandbox.Initialize
    end
```

The following code shows how to do this properly without inciting infinite loops.

```lua
-- Base
    function GM:Initialize()
        -- Do things
    end

-- Sandbox
    DEFINE_BASECLASS "Base"

    function GM:Initialize()
        BaseClass.Initialize( self ) -- Also calls Sandbox.Initialize
    end
-- DarkRP
    DEFINE_BASECLASS "Sandbox"

    function GM:Initialize()
        BaseClass.Initialize( self ) -- Calls Sandbox.Initialize
    end
```

GamerGambit:
For sandbox, shouldn't the BaseClass member be the base gamemode table and not sandbox? The reason I did this is because in player classes, I dislike the fact that I have to specify the base class twice which is honestly stupid even though its trivial.

neico:
I'd rather change RegisterClass to accept the baseclass itself intstead of a string as third argument so that one can do:

```lua
DEFINE_BASECLASS( "player_default" )

...

player_manager.RegisterClass( "player_sandbox", PLAYER, BaseClass )
```
which is basically:

```lua
// Why didn't anyone else notice yet that a Type[name] is totally out of place here? Type[name] just got set some lines above so we should rather check for the base class's existence instead...
if ( !Type[ base.ThisClass ] ) then ErrorNoHalt( "RegisterClass - deriving "..name.." from unknown class "..base.ThisClass.."!\n" ) end

if isstring( base ) then
	setmetatable( Type[ name ], { __index = Type[ base ] } )
elseif istable( base ) || isentity( base ) then // Just making sure here because I'm not sure if baseclass is an entity here :x
	setmetatable( Type[ name ], { __index = base } )
end
```

Jvs34:
It's the same way in derma, it's so loading order of the files doesn't fuck up inheritance of the baseclass system, just leave the whole thing alone.

Loonesbury:
> tl;dr pls don't use BaseClass members.

Only if you're using them wrong.
```lua
function GM:Initialize()
    -- bad
    self.BaseClass.Initialize(self)
    -- good
    self.BaseClass:Initialize() -- equivalent to self.BaseClass.Initialize(self.BaseClass)
end
```

wiox:
It'd be neat if you didn't spout misinformation.

self.BaseClass is the class definition of the parent, __not__ an instance of that parent. You're calling the base function and giving it an invalid object to use as the self parameter.

Loonesbury:
Yeah, you're right, though "spouting misinformation" isn't really the right word for simple ignorance. I don't work with derivation often and my example appears to work if you don't look too hard and the base doesn't change state.

Still, check your own code first before going off on people.
> ```lua
> DEFINE_BASECLASS "gamemode_base"
> ```

PR #898, opened 2015-03-18
Fix prediction on jumpboost

wiox:


Jvs34:
prediction approved :+1:

PR #900, opened 2015-03-22
net.ReadType no longer has typeid arg

Partixel:
net.WriteType writes the typeid as a UInt at the same time as writing the object

Didn't really make sense to have net.ReadType require you to manually net.ReadUInt()


Partixel:
When I first saw net.WriteType I thought I would have to manually write the type until I looked at the code, confused the heck out of me.

Bo98:
Could break compatibility though.

wiox:
#821 already handles this and doesn't break compatibility.

Partixel:
Gosh darnit, missed that wiox. Would be nice to have that implemented.

PR #902, opened 2015-03-23
table.ToString handles Color() correctly

Partixel:
Used to write it as a table, now writes it as value = Color( r, g, b, a )

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -233,10 +233,16 @@ function table.ToString(t,n,nice)
 
 			if istable(value) and not done [value] then
 
-				done [value] = true
-				str = str .. key .. tab .. '{' .. nl
-				.. MakeTable (value, nice, indent + 1, done)
-				str = str .. idt .. tab .. tab ..tab .. tab .."},".. nl
+				if 	IsColor(value) then
+					done [value] = true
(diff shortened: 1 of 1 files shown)


thegrb93:
Can't we just add a __tostring metamethod to the color metatable?

wiox:
__tostring metamethod for Color doesn't output valid Lua code (or valid varlist).

Partixel:
^

thegrb93:
Can't make a valid __tostring for color?

wiox:
tostring isn't meant for serialization, though? no other type except numbers/booleans/nil work this way.

thegrb93:
```
function ColorMeta.__tostring(t) return 'Color('..t.r..','..t.g..','..t.b..','..t.a..')' end
```
This isn't how tostring should be used?

wiox:
It's not special treatment when every other type is handled the same, have you looked at the code?

thegrb93:
Just the change. Looks like you're right.

bmwalters:
@Partixel Merge conflicts. Also, this should be changed to follow the cleaned up coding style before it is merged.

robotboy655:
Had to manually merge: https://github.com/Facepunch/garrysmod/commit/21eeef48a86edbf11b569ad056ee7885e00bcecd

PR #903, opened 2015-03-23
TTT : Fix the color of the credits transfer button

Tommy228:
You now shouldn't directly override Think on DButton


robotboy655:
You don't need this. DLabel is fixed in Dev.

PR #907, opened 2015-03-27
Added gm_showmenuerror

Kefta:


robotboy655:
Sounds like a bad idea. Fix your errors, don't hide them.

Kefta:
It's still enabled by default for normal users, and it still prints in console. Just an option for people who are aware of an error and are trying to test other things.

wiox:
It'd be good if addons couldn't mess with it.

Kefta:
Is there any way to check if a player changed it via console vs. a lua file?

wiox:
I don't think so. Is this tested? Bool isn't a valid method for ConVar objects. You should be using gm_showmenuerror:GetBool() and also move the check in to the OnLuaError hook so that changes to the ConVar take effect immediately.

Kefta:
Oops, bad habit from custom functions. Also, I'll put a check in the hook as well

wiox:
I should note the only reason I'm for this PR is that the error message position sucks for people using the developer ConVar. I'd take some ConVar that repositions it as an alternative too.

Kefta:
I'll make a seperate PR with a convar to move the message to a different corner.

Kefta:
I'll make a seperate PR with a convar to move the message to a different corner.

Kefta:
Damn

PR #910, opened 2015-03-31
Improve net.Read/WriteTable

meepdarknessmeep:
I accidentally deleted other repository and couldn't update it and since i found a few bugs, i remade it.

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -148,9 +148,8 @@ end
 local reading, writing
 
 
-local _type = type
-local function type(x)
-	local t = _type(x)
+local function send_type(x)
+	local t = type(x)
 	if(t == "table" and IsColor(x)) then
 		return "Color"
 	end
@@ -339,7 +338,7 @@ writing = {
 					net.WriteUInt(headers.reference, MAX_BIT)
 					net.WriteUInt(indices[v], REFERENCE_BIT)
 				else
-					local t = type(v)
+					local t = send_type(v)
 					net.WriteUInt(headers[t], MAX_BIT)
 					local _num = writing[t](v, rs, num)
 					if(t ~= "table") then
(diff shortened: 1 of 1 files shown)


thegrb93:
You need 16 bits for Garrysmod networked entities. I don't think 2048 limit is the same for Garrysmod. Even if it was, you should use 11 bits instead of 12.

thegrb93:
Its sort of nice that you made it not send an entity if it is invalid, but that means every entity that is valid sends an extra bit. I think the number of valid entities networked far outweighs invalid entities so I don't think that is an improvement.

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -179,3 +144,251 @@ function net.ReadType( typeid )
 
 	error( "net.ReadType: Couldn't read type " .. typeid )
 end
+
+local reading, writing
+
+
+local _type = type
+local function type(x)
```
thegrb93: Why

meepdarknessmeep:
you say it's 16 bits, but CClientEntityList::GetMaxEntities() returns 8096 (1/2 is always non networkable, thus leading to 12 bits
http://nl.cm/xsf

thegrb93:
I'll support that if the devs can verify that. Still I don't think sending that extra boolean is worth it.

meepdarknessmeep:
So you'd rather have invalid/non-networkable entities sent as the world entity instead of NULL entities?
That seems like a bad idea imo. Imagine what people would encounter and have no idea why it's happening. It's not about having the least amount of networking possible, it's more about have the best combination of network load/accuracy

thegrb93:
Doesn't EntIndex throw an error if it tries to use an invalid entity? Just do that.

meepdarknessmeep:
That's a terrible idea.
NULL Entities exist and we should acknowledge that. there's even a global variable called NULL that is a null entity.

thegrb93:
Okay that's fine and all, but you shouldn't send them to net.WriteEntity

meepdarknessmeep:
Closing this for now. I might come back at one point to improve it on default gmod again.

PR #913, opened 2015-04-03
Optimize net.Write/ReadTable

meepdarknessmeep:
Also changed net.ReadEntity and net.WriteEntity to use the proper amount of bits

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -1,152 +1,3 @@
-TYPE_COLOR = 255
-
-net.Receivers = {}
-
---
--- Set up a function to receive network messages
---
-function net.Receive( name, func )
-		
-	net.Receivers[ name:lower() ] = func
-
-end
-
---
--- A message has been received from the network..
---
-function net.Incoming( len, client )
-
-	local i = net.ReadHeader()
-	local strName = util.NetworkIDToString( i )
-	
-	if ( !strName ) then return end
-	
-	local func = net.Receivers[ strName:lower() ]
-	if ( !func ) then return end
-
-	--
-	-- len includes the 16 bit int which told us the message name
-	--
-	len = len - 16
-	
-	func( len, client )
-
-end
-
---
--- Read/Write a boolean to the stream
---
-net.WriteBool = net.WriteBit
-
-function net.ReadBool()
-
-	return net.ReadBit() == 1
-	
-end
-
---
--- Read/Write an entity to the stream
--- CClientEntityList::GetMaxEntityIndex( ) returns 8096
--- therefore we are going to use 13 bits to send the index
---
-function net.WriteEntity( e )
-
-	if( IsValid( e ) or game.GetWorld( ) == e ) then
-
-		net.WriteBool( true )
-		net.WriteUInt( e:EntIndex( ), 13 )
-
-		return
-
-	end
-
-	net.WriteBool( false ) -- NULL
-
-end
-
-function net.ReadEntity()
-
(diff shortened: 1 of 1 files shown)


meepdarknessmeep:
writing table:

meepdarknessmeep:
Writing table:
```
BIG_TABLE = {
	1,
	"trolled",
	0/0,
	"omfg",
	123456,
	123456.789,
	9,
}
BIG_TABLE.me = BIG_TABLE;
```

new code:
```
BYTES:
	41.875
```
old code:
```
overflow
```

Writing table:
```
BIG_TABLE = {
	1,
	"trolled",
	0/0,
	"omfg",
	123456,
	123456.789,
	9,
}
```

old: 
```
BYTES:
	124
```

new: 
```
BYTES:
	37.375
```

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -179,3 +148,620 @@ function net.ReadType( typeid )
...
+	
+end
+
+--
+-- Returns true if the string can be represented in 7-Bit ASCII
+-- Null bytes are not allowed as they are used for termination
+--
+local function Is7BitString( str )
+
+	return str:find( "[ \x80-\xFF%z ]" ) == nil
```
thegrb93: Might try and see if this is faster.
```
for I=1, #str do
    if bit.band( string.byte(str, I), 0x80 ) ~= 0 then return false end
end
return true
```

thegrb93: Your regex doesn't seem to work anyway.

thegrb93: Should also bias it so that it checks the first bit instead of the last and then bitshift each character down because commonly used characters are at the end of the ascii table.

thegrb93: Here's the correct regex str:find( "[%80-%FF%z]" )

thegrb93: After benchmarking, the regex seems to be faster.

thegrb93: Can't use regex if you want to check the first bit.

You can do string.byte(str, I) % 2 == 0 though.

thegrb93: Why's %z in the regex?

wiox: > Null bytes are not allowed as they are used for termination

meepdarknessmeep:
```
] lua_run_mn print(("\x80"):find("[\x80-\xFF%z]"))
1	1
] lua_run_mn print(("\x7F"):find("[\x80-\xFF%z]"))
nil
```

thegrb93: Might be because lua 5.1

thegrb93:
Can anyone confirm that 16 bits isn't required for an entity index? Kind of weird that it was 16 bits before when it didn't need to be (Maybe there were plans for increasing the entity limit further?).

thegrb93:
I'll test this against some of my mods to make sure it works. I'll mark them as I test them.
Expadv2
Nadmod
Playx
Pac3

meepdarknessmeep:
@robotboy655 pls

meepdarknessmeep:
For the entity limit thing, 
http://facepunch.com/showthread.php?t=1459242
Read this post and it will show the limit of gmod in the errors.

meepdarknessmeep:
remaking since i failed again!

PR #915, opened 2015-04-05
Optimize widgets

meepdarknessmeep:
Made widgets not allocated a table for traceresult/trace each frame
Made widgets only run when widgets are created

### DIFF
--- a/garrysmod/lua/includes/modules/widget.lua
+++ b/garrysmod/lua/includes/modules/widget.lua
@@ -5,6 +5,7 @@
 --
 
 widgets = {}
+widgets.entities = {}
 
 --
 -- Holds the currently hovered widget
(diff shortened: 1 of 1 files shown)


willox:
It'd probably be better to just remove the filter function here (and add something such as MASK_WIDGET for traces, moving the implementation to C++) seeing as that's the part that is actually slow. That way we don't have all this horrifying OnEntityCreated/EntityRemoved logic.

It's just the amount of C -> Lua calls created by the engine.

ZehM4tt:
Dot product would also help avoiding traces if you are not facing it.

garrysmod/lua/includes/modules/widget.lua
```diff
@@ -145,6 +141,24 @@ hook.Add( "PostDrawEffects", "RenderWidgets", function()
 
 end )
 
+hook.Add( "OnEntityCreated", "CreateWidgets", function( ent )
+	timer.Simple( 1, function( ) -- why is this needed? no one may ever know!
```
uRandomAlex: How about not using any timer at all?

Anyway, timer with 0 may not actually help because it can be called at this frame, not at the next frame. This has been discussed somewhere as far as I can remember. The solution was just using a very little delay (like 0.001) for the timer instead of 0.

robotboy655: timer with 0 delay should only be called on the same frame if you are adding the timer from another timer.

EstevanTH:
I would add a parameter such as
```lua
ENT.WidgetServerProcessing = false
```
to tell that a given widget does not need to do traces on the server.

In case there is no widget at all, the traces generation should be shut down. Adding or removing the hook during `InitPostEntity`, `OnEntityCreated`, `EntityRemoved`, `PostCleanupMap` seems to be a nice and clean idea.

Also it may be optimized by default with a restriction: no traces for players outside of the widget's PVS. I'm not sure about that though.

Kefta:
@willox Any update on this?

ExtReMLapin:
merge it already

SpiffyJUNIOR:
any update on getting this merged?

GAMLET228:
Как отключить виджеты ?

Astralcircle:
10 years..., close it already

i-hate-trustedinstaller:
can someone please let this request rest in peace

bloodycop6385:
R.I.P o7

YohSambre:
never forget dude

PR #918, opened 2015-04-06
[DGrid] Use table.remove instead of setting to nil

wyozi:
This fixes SortByMember not working after item removal, because there are gaps in the Items table.


KateAdams:
Just for further explanation, `table.remove` also shifts all higher elements down to remove the "gap" in the table, and thus, is still continuous.

PR #920, opened 2015-04-09
Unlink games' incorrect usage of __tostring as string serializer and create meta:Serialize() function to serialize data for console commands, etc...

Acecool:
As https://github.com/garrynewman/garrysmod/pull/919 was closed I couldn't merge the latest changes. This unlinks the games incorrect usage of __tostring as a data serializer in favor of adding a meta:Serialize( delimiter, postfix ) function for the purpose of serializing Angles, Colors, and Vectors.

As it currently stands, if you alter __tostring for Angles, Colors or Vectors, the game breaks... as __tostring is an integral part of the print function, we shouldn't need to rewrite print in order to have clean data-type output which can't be confused with other data-types ( Angles and Vectors both use identical output "0 0 0" which can confuse new and old developers alike; example, a developer is debugging their code and uses print [ a developer function ] to output angle_zero:Forward( ) which returns "0 0 0". The developer can easily be confused into thinking the data-type which results from the above sequence is an Angle when, in fact, it is a normalized Vector.. ).

With this merged, everything is in its' place ( even DarkRP merged the pull-request to correct the same issue which  I posted prior to this one ) and __tostring can be altered into any output that you want without causing the game to break ( think balloons spawning incorrectly, or with incorrect force because __tostring was used instead of manually serializing the data, or now simply using Serialize(), and other tools and elements within the game going haywire ).

print is a developers best friend when it comes down to it in terms of debugging. We use it to check if logic is properly firing, to output data-type values, etc.. We shouldn't be required to rewrite print to have proper data-type output because that would completely defeat the purpose of Lua meta-tables and meta-table functions.


garrysmod/lua/includes/util/color.lua
```diff
@@ -36,13 +36,33 @@ function IsColor( obj )
...
+
 --[[---------------------------------------------------------
 	Returns color as a string
 -----------------------------------------------------------]]
 function COLOR:__tostring()
-	
-	return string.format( "%d %d %d %d", self.r, self.g, self.b, self.a )
-	
+
+	return self:ToString( )
```
neico: So many commits and yet there's still this little one

Acecool: You're right. Updating / fixing.

Lexicality:
You should squash d4f1754 - f30816d and force push, they really don't need to be separate commits. In fact, the entire branch should really be squashed - a single commit per file is silly.
[This stack overflow question][1] has a variety of solutions but [this one][2] seems the easiest.
[1]: http://stackoverflow.com/q/5189560/823542
[2]: http://stackoverflow.com/a/5201642/823542

Acecool:
Apparently on git using Source-Tree you can select rebase children of * then drag and drop, so I'll try that. If it fails, I now have cloned the latest gmod git to my local machine so I can create a new-branch and make the changes without needing to use the website editor although this pull may need to be closed and a new one created.

I'll try rebasing first to see if that fixes it.

TheFreeman193:
@Acecool You can squash your commits with *git reset* or *git rebase*. You then need to push it with the *-f* switch to your remote. Have a look at [this](http://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git).

Acecool:
Alright, it has all been squashed into one commit; I also removed 3 files from the commit which removed a useless line of code each, at the top: local tostring = tostring....

Thanks to all and thanks to Willox for step-by-step on how to "undo" and re-process the pull.

thegrb93:
Why is __tostring incorrect for serialization? It doesn't make sense to change from tostring() to serialize(). I've always been happy with print(myvector) why are you forcing it to become print(myvector:Serialize())?

GitSparTV:
What's the point of delimiter if you don't use it anywhere in the codebase?
Following the first question, why do you need a separate function that does the same as __tostring?

robotboy655:
The difference is this:
tostring( Vector ) => `Vector( x, y, z )` i.e. human readable
Vector.Serialize => `x y z` => save to file => read from file => Vector( whatever you read )

GitSparTV:
Alright, would it be backwards compatible tho.
env_sun vector value saved in ISave for example, would it work?
Also not sure about the idea with delimiters and postfix. At least `isstring` check. I think it just check if it's not nil/false.

robotboy655:
It is if __tostring methods are not changed.

GitSparTV:
Yeah but he changed them to Serialize()

Lexicality:
Unless @Acecool is going to appear from nowhere and rebase his 5 year old PR to not be horrendously out of date, it doesn't really matter does it?

PR #922, opened 2015-04-11
TTT: Replaced use of COLLISION_GROUP_DEBRIS_TRIGGER

Kefta:
http://facepunch.com/showthread.php?p=47498190

Tested this on a dev server running this script: https://github.com/Kefta/Ragdoll-Remover
Simulated high ragdoll velocity and overlapping collision groups. Without the change, the script triggered 7/10 times. With the change, it triggered 1/10 times (from a C4 blast, which sends it at a high velocity anyway).

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
@@ -358,6 +358,8 @@ local function GetSceneData(victim, attacker, dmginfo)
 end
 
 local rag_collide = CreateConVar("ttt_ragdoll_collide", "0")
+-- 0 = Don't collide with players and props
+-- 1 = Don't collide with players
 
 -- Creates client or server ragdoll depending on settings
 function CORPSE.Create(ply, attacker, dmginfo)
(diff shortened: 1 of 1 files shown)


svdm:
I'm all for a workaround for that physics bug, but I have two gameplay-related concerns:

- Do ragdolls clip through all non-static props with this collision group? If so, players could start hiding corpses inside props, which would be a significant downside (and it might call for a separate convar so servers can choose). 
- Do ragdolls still hit triggers? I know some maps rely on detecting them.

Kefta:
http://wiki.garrysmod.com/page/Enums/COLLISION

Debris also doesn't interact with non-static props. I think this can be resolved by making ttt_ragdoll_collide true by default.

YourStalker:
With ttt_ragdoll_collide I've had players smashing bodies together and when two ragdolls clip into each other the server goes wacko and the game crashes. This also can happen on maps where walls come in and crush you, they can crush two ragdolls into each other.

Bo98:
The issue is much more complex than changing a collision group.

http://facepunch.com/showthread.php?t=1460189&p=47511040&viewfull=1#post47511040

A couple of functions are coming in the next GMod update to help fix this properly.

Kefta:
Yes, but changing the group helps reduce the amount of bad collisions.

Bo98:
Any TTT update is only going be included in GMod in the next update so you might as well try the new functions, either now or closer to the release (in case people take it directly from GitHub).

Bo98:
Now that the GMod update is released, how about looking into the new functions?

Kefta:
Tested RecheckCollisionFilter and CollisionRulesChanged. Neither -- in their undocumented goodness -- seem to have made no difference with DEBRIS_TRIGGER. I assume they're supposed to be used when you are constantly changing COLLISION flags and you want to regenerate their collision assignment, but I'm still not sure.

PR #924, opened 2015-04-17
Expanded door state functions

Exho1:
Adds the majority of the C++ functions from CBasePropDoor( https://github.com/ValveSoftware/source-sdk-2013/blob/56accfdb9c4abd32ae1dc26b2e4cc87898cf4dc1/sp/src/game/server/BasePropDoor.h#L215 ) into Lua as well as a few activator related functions.

I believe these can be very helpful and expand what developers can do with doors. The activator functions could be used if you wanted to tinker with input/output in Lua.

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -481,7 +481,79 @@ end
 if ( SERVER ) then
 
 	AccessorFunc( meta, "m_bUnFreezable", "UnFreezable" )
+	
+	--
+	-- Door state
+	--
+	
+	DOOR_STATE_CLOSED = 0
+	DOOR_STATE_OPENING = 1
+	DOOR_STATE_OPEN = 2
+	DOOR_STATE_CLOSING = 3
+	DOOR_STATE_AJAR = 4
+	
+	function meta:GetDoorState()
+		return self:GetSaveTable().m_eDoorState
+	end
+	
+	function meta:IsDoorClosed()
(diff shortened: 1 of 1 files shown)


robotboy655:
No. Especially for IsDoor() return true for prop_dynamic.
This is very much pointless, they are essentially aliases of already existing functions. We do not need more of these. Especially since they litter the Entity metatable.

Exho1:
They are not aliases of existing functions, there is not a meta:IsClosed or anything that I am trying to create another function for. There are not functions for these and accessing them through the save table is not something the ordinary person knows how to do as demonstrated here (http://facepunch.com/showthread.php?t=1450721&p=47544507&viewfull=1#post47544507).

At the very least, what about dropping the IsClosed, IsClosing, etc functions in favor of just GetDoorState, DOOR_STATE enums, and the activator functions? Is that reasonable?

robotboy655:
By aliases I mean the functions that only call one function. This is pointless. You'd be better off by creating a page on the wiki listing these things instead. I can make you a template as a starting point if you are interested.

Exho1:
So all of these are pointless? I mean I certainly understand IsDoorClosing but I think GetDoorState and the enumerations are useful. Thats what my focus has shifted to s

thegrb93:
It's like creating an NPC:IsDead() function that returns NPC:Health()<=0

Exho1:
Okay but I'm not worried about those anymore. I outlined above the few functions I would like

PR #925, opened 2015-04-18
Door and activator functions

Exho1:
Condensed from my previous PR as to not be useless. It serves to add a few extra helper functions, a couple of which might be unknown to most people as GetSaveTable is not widely used and people might not know that they could check these values


wiox:
This is cool. Maybe it'd be easier to remember usage if the enums were shorter, such as DOOR_OPEN and DOOR_AJAR.

Exho1:
Perhaps, I just left them named as they were in the CBasePropDoor file and because it makes sense for GetDooorStatus to compare to DOOR_STATUS_AJAR.

PR #926, opened 2015-04-19
Add ColorRand

Kefta:
Now with whitespace


Kefta:
Apparently, line endings get fucked up in the default Github editor; sorry about that

PR #932, opened 2015-04-22
Added Pre and PostDraw functions for DModelPanel

Exho1:
Used just like Pre/PostPlayerDraw except for the DModelPanel entity. Return false on the Pre function to stop the rendering of the model and then you can modify the entity after its drawn in the Post function without having to override the DrawModel function. This can be used to set the material of the entity or take a pointshop-esque approach and render clientside models around the entity

I use this to draw clientside models onto a DModelPanel with pretty much the same code I have in PostPlayerDraw to draw the models onto the player.
https://github.com/Exho1/ScrapCombat/blob/master/lua/scraparmor/cl_inventory.lua#L130

### DIFF
--- a/garrysmod/lua/vgui/dmodelpanel.lua
+++ b/garrysmod/lua/vgui/dmodelpanel.lua
@@ -89,6 +89,10 @@ end
 	Name: DrawModel
 -----------------------------------------------------------]]
 function PANEL:DrawModel()
+	if hook.Run("PreDrawModelPanel", self, self.Entity) == false then 
+		return 
+	end
+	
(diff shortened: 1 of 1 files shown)


zerfgog:
Why not make it a PANEL hook (function for overriding on panel's table)?

Exho1:
Good idea, I changed it. That way you could override those functions instead of having to override DrawModel itself

robotboy655:
I seems to me that these should be called inside the scissor rect.

Exho1:
Alright, done

PR #933, opened 2015-04-25
Added hooks TTTSelectTraitors and TTTSelectDetectives

Mechami:
Two parameters are passed to each hook:

1. The first parameter is the table containing the players
2. The second parameter is the table containing the previous role for each player


svdm:
Thanks for the PR. However, I intentionally don't have hooks in place for traitor and detective selection, because I do not want to facilitate/encourage servers selling improved selection chances to players.

PR #937, opened 2015-04-26
Door and activator functions 2.0

Exho1:
Reposted because I'm terrible with Github branches and I needed to make a change

Edit: Now its kinda silly because func_doors don't have m_eDoorState but instead have m_toggle_state which use different enumerations. For example: an open func_door_rotating has a m_toggle_state of 0 which for a prop_door_rotating is 2... Why Valve


zerfgog:
IMO DOOR_* is more intuitive than DOOR_STATE_*. Other than that, :+1:+1

PR #939, opened 2015-04-29
Use PlayerModel colouring for TTT players

DarthTealc:
Removes the need for checking if a playermodel is colourable as it uses GMod 13's matproxy player colouring.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/player.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player.lua
@@ -259,14 +259,13 @@ end
 
 function GM:TTTPlayerSetColor(ply)
    local clr = COLOR_WHITE
-   local should_color = hook.Call("TTTShouldColorModel", GAMEMODE, ply:GetModel())
-   if GAMEMODE.playercolor and should_color then
+   if GAMEMODE.playercolor then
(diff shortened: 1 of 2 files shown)


robotboy655:
It never did? Wow.

svdm:
So just to make sure: playermodels without any sort of support for this will simply not get a colour?

robotboy655:
Yes.

DarthTealc:
I've been using this (via a hook) for a few months on a live TTT server with additional playermodels (some colourable, some not, such as the arctic one that TTT uses already) and telling it that they're all colourable regardless of if they actually are or not (via TTTShouldColorModel) and haven't seen any issues.

DetectDefective:
Fair heads up; the corpse doesn't retain the coloring, it just reverts to default blue clothing.

Still, I'm super happy to see this being added.

DarthTealc:
Corpses were not retaining colour anyway (at least on the server I work on), but I'm looking into a solution for that. Trying to avoid having to send a net message but it's looking like that's the way it needs to be done. I'll do a pull request soon if nobody beats me to it.

ElektroPanda:
Does this deal with the problem that causes traitor team glows to lag people ONLY when traitor? (the glow outline around your t buddies through walls)

PR #940, opened 2015-04-29
Added alpha support to DrawMaterialOverlay

JamesxX:


zerfgog:
:+1:+1, but the variable name may need to be changed. Additionally you could use ```alpha = alpha or 1```

PR #941, opened 2015-05-02
Added net.WriteChunk and net.ReadChunk

thegrb93:
One painful thing about the net library is the data limit and the overflows that can occur. Lots of mods have to write their own file sending library which is tedious. So I wrote this, but this is my first contribution so let me know if anything needs to be improved.

Features:
- It can handle data transfers between the server and different clients across different net.Start calls simultaneously.
- Unlimited transfer size (aside from RAM limit).
- Won't overflow reliable channel (on its own).

I don't know if you wanted to do this in c++ or not, but here it is in lua and it works pretty well.

Here's a test program. The server gets an image online, sends it to all clients, the client sends it back, and then the server compares the sent data to the original data.

```
if SERVER then
    util.AddNetworkString("TestChunks")

    local test
    concommand.Add("TestChunks", function() 
        http.Fetch("http://4.bp.blogspot.com/-9B5oDjm1wgM/T3gFww7jqlI/AAAAAAAABoY/c79rs-PrfPM/s1600/ORLY+Wallpapers.png",function(d)
            test = d
            net.Start("TestChunks") 
            net.WriteChunk(d) 
            net.Broadcast() 
        end)
    end)


    net.Receive("TestChunks", function(len, ply)
        net.ReadChunk(ply, function(d)
            print(d == test)
        end)
    end)
else

    net.Receive("TestChunks", function(len, ply)
        net.ReadChunk(ply, function(d)
            net.Start("TestChunks")
            net.WriteChunk(d)
            net.SendToServer()
        end)
    end)

end
```

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -192,14 +192,14 @@ function net.Chunk:Request( ply )
 
 	net.Start( "ChunkRequest" )
 	net.WriteBit( false )
-	net.WriteString( self.crc )
+	net.WriteUInt( self.identifier, 32 )
 	net.WriteUInt( #self.data, 32 )
 	
-	--print("Requesting",self.crc,#self.data)
+	--print("Requesting",self.identifier,#self.data)
 	
 	if CLIENT then net.SendToServer() else net.Send( ply ) end
 	
-	timer.Create( "ChunkDlTimeout" .. self.crc, 1, net.Chunk.Timeout, function() self:Remove() end )
+	timer.Create( "ChunkDlTimeout" .. self.identifier, 1, net.Chunk.Timeout, function() self:Remove() end )
 	
 end
 
@@ -208,7 +208,7 @@ function net.Chunk:Start( ply )
 
 	if not self.active then
 	
-		timer.Remove( "ChunkKeepAlive" .. self.crc )
+		timer.Remove( "ChunkKeepAlive" .. self.identifier )
 		self.active = true
 		self:Request( ply )
 		
@@ -239,12 +239,12 @@ function net.Chunk:Remove()
 	
 	pcall( self.callback, self.returndata )
 	
-	timer.Remove( "ChunkDlTimeout" .. self.crc )
+	timer.Remove( "ChunkDlTimeout" .. self.identifier )
 	table.remove( self.queue, 1 )
 	if self.queue[ 1 ] then
 		self.queue[ 1 ]:Start()
 	else
-		net.Chunk.Queues[ self.crc ] = nil
+		net.Chunk.Queues[ self.identifier ] = nil
 	end
 	
 end
@@ -259,13 +259,17 @@ function net.WriteChunk( data )
 	end
 		
 	local compressed = util.Compress( data )
-	local crc = util.CRC( compressed )
+	local identifier = 1
 	
-	net.Chunk.Data[ crc ] = compressed
-	timer.Create( "ChunkUlTimeout" .. crc, 1, net.Chunk.Timeout, function() net.Chunk.Data[ crc ] = nil end )
+	while net.Chunk.Data[ identifier ] do
+		identifier = identifier + 1
+	end
+	
+	net.Chunk.Data[ identifier ] = compressed
+	timer.Create( "ChunkUlTimeout" .. identifier, 1, net.Chunk.Timeout, function() net.Chunk.Data[ identifier ] = nil end )
 	
 	net.WriteUInt( math.ceil( #compressed / net.Chunk.MaxSendSize ), 32 )
-	net.WriteString( crc )
+	net.WriteUInt( identifier, 32 )
 	
 end
 
@@ -290,13 +294,13 @@ function net.ReadChunk( ply, callback )
 	if not queue then queue = {} net.Chunk.Queues[ ply ] = queue end
 	
 	local numchunks = net.ReadUInt( 32 )
-	local crc = net.ReadString()
+	local identifier = net.ReadUInt( 32 )
 	
-	--print("Got info", numchunks, crc)
-	
+	--print("Got info", numchunks, identifier)
+		
 	local chunk = {
 		numchunks = numchunks,
-		crc = crc,
+		identifier = identifier,
 		data = {},
 		active = false,
 		callback = callback,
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/net.lua
```diff
@@ -179,3 +179,156 @@ function net.ReadType( typeid )
...
+
+--Pop the queue and start the next task
+function net.Chunk:Remove()
+	
+	pcall( self.callback, self.returndata )
+	
+	timer.Remove( "ChunkDlTimeout" .. self.crc )
+	table.remove( self.queue, 1 )
+	if self.queue[ 1 ] then
+		self.queue[ 1 ]:Start()
```
thegrb93: Need to add
```
else
    net.Chunk.Queues[ self.crc ] = nil
```
when I a chance

wiox:
Why is it so complicated? Surely two simple calls like below would be more reasonable.

```lua
-- On the recipient
net.ReceiveLotsaData( "Message ID", function( data, id )
    -- use data
end )


-- On the non-recipient
util.AddNetworkString "Message ID"

net.SendLotsaData( "Message ID", data, optional ply )
```

thegrb93:
I wanted to make it match the style of the net.Write/net.Read functions but I think you're right. I can do it like that.

thegrb93:
I think I remember exactly why now. I wanted to make it so people could include data along with the chunk for use in their callback. For example,

```
net.Start("SendFile")
net.WriteString("Filename")
net.WriteChunk(filedata)
net.SendToServer()

--------------------
net.Receive("SendFile", function(len, ply)
local filename = net.ReadString()
net.ReadChunk(ply, function(d)
     file.Write(filename, d, "DATA")
end)
end)

**Comment by thegrb93** [comment]
Just added keep-alive so queued data won't time out.

**Comment by meepdarknessmeep** [comment]
this is useless it will never be added

**Comment by thegrb93** [comment]
I know three mods that can use this.

Wire
Starfall
Pac3

Why is it useless?

**Comment by meepdarknessmeep** [comment]
because it's garbage
pac3 could just use a better implementation of net.WriteTable and it would work 200% better

the others i don't know about

**Comment by thegrb93** [comment]
Pac3 will send resources such as images, models, etc. to the client in the future. Things like images or large files can't be sent over the net channel conventionally because of their size.

**Comment by meepdarknessmeep** [comment]
still this will never get added. the syntax is horrid.

just make net.Receive check if it needs multiple packets - if it does wait until all packets are received then call the function

of course this will need to be done in c++ unless you feel like rewriting the whole net library

or of course they could up the limit of net data

**Comment by thegrb93** [comment]
It's not that simple. This is capable of handling uploads/downloads between different players and different net.Start instances of the same name simultaneously.

**Comment by thegrb93** [comment]
Here's an example of some seriously improved code using this library.
https://github.com/thegrb93/Starfall/compare/master...thegrb93:net_improvement?diff=split&name=net_improvement

I just need to write a stress test now to make sure everything is working. If there are any aesthetic changes that need to be made, let me know.

**Comment by Mechtecs** [comment]
+1

**Comment by Grocel** [comment]
@meepdarknessmeep Well, it's not like you are deciding what is going to be added to GMod and what not.

**Comment by thegrb93** [comment]
https://github.com/garrynewman/garrysmod/pull/917
>Literally no one would ever use your code

Every addon that sends files through the net library would benefit from my code. Here's just one examples of an addon ported to use my code. https://github.com/thegrb93/Starfall/compare/master...thegrb93:net_improvement?diff=split&name=net_improvement How is that no one?

**Comment by meepdarknessmeep** [comment]
That code using your api looks like shit. That is the main reason I am against your pr. I assumed because of the horrid syntax that no one would ever use it. Callbacks or whatever your readchunk does with the function makes it ugly and hard to use.

---- thegrb93 wrote ----

**Comment by thegrb93** [comment]
You HAVE to use a callback because the data doesn't get received immediately.

**Comment by TylerB260** [comment]
holy fucking shit you guys are ANNOYING.

my outlook is going fucking insane.

**Comment by thegrb93** [comment]
There's a mute button. Use it.

**Comment by meepdarknessmeep** [comment]
Don't put a callback inside a callback like that, make your own receiver or
an api for others to read chunks themselves.

**Comment by TylerB260** [comment]
:shit:

**Comment by thegrb93** [comment]
It is done this way so additional information can be sent along with the file in the same net.Start. That way additional work doesn't have to be done to match net data with received files.

**Comment by wiox** [comment]
Am I wrong to think that after calling net.WriteChunk, if the queue was empty you'll no longer be in your original net message?

```lua
net.Start( "Test" )
net.WriteChunk( ply, data )
net.WriteString()
net.Send( ply )

--

net.Receive( "Test", function( _, ply )
    net.ReadChunk( ply, func )
    net.ReadString() -- Wrong data
end )
```

meepdarknessmeep:
I would love to see some code on the receiving end do this:

net.Receive("stuff", function(len, client)

    if(net.ReadBool()) then
        local isend, data = net.ReadChunkPart()

        if(isend) then
            -- data contains the stuffz
        end
    end

end)

thegrb93:
Then anything sent with the chunk will need to be sent with each time a part is sent.

thegrb93:
You forgot the example for the sender.

thegrb93:
Like this
```
net.Start("SendFile")
net.WriteString("Filename.txt")
net.WriteChunk(filedata)
net.SendToServer()

--------------------
net.Receive("SendFile", function(len, ply)
     local filename = net.ReadString()
     net.ReadChunk(ply, function(d)
          file.Write(filename, d, "DATA")
     end)
end)
```

thegrb93:
Also your version can't handle multiple net.Sends

meepdarknessmeep:
I never forgot. 

As for your question "how is it better": 
- it's using only one net receiver, making it less cluttered
- the dev has full control of the code, how they want to implement a queue if they wanted to
- it works closely to the original net functions

thegrb93:
You didn't even show your idea so I have no clue what you're blabbering about. Give an example of the sender.

immibis:
Why not this?

    if SERVER then
        util.AddNetworkString("TestChunks")
        net.RegisterLargeMessage("TestChunks")

        local test
        concommand.Add("TestChunks", function() 
            http.Fetch("http://4.bp.blogspot.com/-9B5oDjm1wgM/T3gFww7jqlI/AAAAAAAABoY/c79rs-PrfPM/s1600/ORLY+Wallpapers.png",function(d)
                test = d
                net.Start("TestChunks") 
                net.WriteString(d) 
                net.Broadcast() 
            end)
        end)


        net.Receive("TestChunks", function(len, ply)
            local d = net.ReadString()
            print(d == test)
        end)
    else

        net.Receive("TestChunks", function(len, ply)
            local d = net.ReadString()
            net.Start("TestChunks")
            net.WriteString(d)
            net.SendToServer()
        end)

    end

thegrb93:
That would have to be implemented in c++

thegrb93:
I'm thinking about renaming this to net.WriteStream and net.ReadStream. Would that be an appropriate name?

AbigailBuccaneer:
I like the idea of this, it would be very useful functionality to have (and is something we've had issues with in multiple places in Wiremod, which usually leads to people implementing their own).

I don't like the idea of having a callback-within-callback. I can entirely understand why it's implemented like that, but I'd rather `net.Receive` only be called when all the data is available (which, as you say, would have to be implemented in C++ or by replacing `net.*` functions).

The ideal solution (for my use cases at least) would be to have all net messages able to internally split into multiple messages, and then everything would Just Work™. This could be done in Lua by replacing all the relevant functions but I'd rather see it implemented natively.

PR #942, opened 2015-05-05
Improved Defuser compatibility.

MyHatStinks:
This will allow the defuser to be used in third-party addons without modifying gamemode files.

A minor change, but it will make life easier for coders, especially if a server wants to use multiple defusable items from different coders.


Exho1:
Yeah this is really needed. Addons that have a defusing capability ship their own version of the TTT defuser which can be outdated and also override any custom ones which isn't cool.

zerfgog:
:thumbsup:

PR #944, opened 2015-05-14
Optimized Panel:IsChildHovered()

KrakenZ:
Removed the iDepth parameter, no longer needed.


LennyPenny:
Might want to also remove all instances of something calling `IsChildHovered` with the parameter

PR #946, opened 2015-05-14
Optimized Panel:IsChildHovered()

KrakenZ:
Removed the iDepth parameter, no longer needed.
All instances using the parameter are corrected.


TheFreeman193:
Looks good and works, panel.lua needs an EOF newline though. Only issue I can see is that some might want control over the number of levels this method looks at, such as looking only for immediate children.

KrakenZ:
Why would it need an EOF newline?
And if you want to look at an immediate children, you simply use <child>:IsHovered().

TheFreeman193:
It's just good practice- it comes from a C/Unix standard for line processing. [This](https://robots.thoughtbot.com/no-newline-at-end-of-file) explains it.

PR #947, opened 2015-05-14
Fixed Magneto Stick setting nil Owner on active SWEP

MisterDot:
Prevents the Magneto Stick from resetting a SWEP's Owner after being dropped.

It is possible to equip a weapon while holding it with the stick. Previously this prevented the weapon from being used properly and caused script errors. The player would need to equip the weapon again after dropping it to solve the issue.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
@@ -103,10 +103,12 @@ function SWEP:Reset(keep_velocity)
    end
 
    if IsValid(self.EntHolding) then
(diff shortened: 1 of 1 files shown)


svdm:
Looks good, but I suspect the change should also be made in the Pickup() logic, so that we don't set the carrying player as the weapon's owner and then never reset that when they drop it.

It should suffice to do the same IsWeapon check here:
https://github.com/MisterDot/garrysmod/blob/patch-1/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua#L375

MisterDot:
Updated.

MisterDot:
Added something else. Hope that's alright.

PR #948, opened 2015-05-15
TTT: Fixed spectators having footstep sounds

MisterDot:
Dead players sometimes play footstep sounds. Not sure exactly why but this solves that.

Perhaps it has something to do with some spectators being reported as technically 'alive' while in spectator mode if I remember correctly.


MisterDot:
Oops. Seems I've misunderstood how pull requests work. I'll make two new requests and close this one.

PR #952, opened 2015-05-16
Clamp the ball size

thegrb93:
So players can't make weird physics with ball duplications with bad ballsize keyvalues.

### DIFF
--- a/garrysmod/lua/entities/sent_ball.lua
+++ b/garrysmod/lua/entities/sent_ball.lua
@@ -9,6 +9,8 @@ ENT.PrintName		= "Bouncy Ball"
 ENT.Author			= "Garry Newman"
 ENT.Information		= "An edible bouncy ball"
 ENT.Category		= "Fun + Games"
+ENT.MinSize			= 4
+ENT.MaxSize		= 128
 
 ENT.Editable		= true
 ENT.Spawnable		= true
@@ -17,7 +19,7 @@ ENT.RenderGroup		= RENDERGROUP_TRANSLUCENT
 
 function ENT:SetupDataTables()
 
-	self:NetworkVar( "Float", 0, "BallSize", { KeyName = "ballsize", Edit = { type = "Float", min = 4, max = 128, order = 1 } } )
+	self:NetworkVar( "Float", 0, "BallSize", { KeyName = "ballsize", Edit = { type = "Float", min = self.MinSize, max = self.MaxSize, order = 1 } } )
 	self:NetworkVar( "Vector", 0, "BallColor", { KeyName = "ballcolor", Edit = { type = "VectorColor", order = 2 } } )
 
 end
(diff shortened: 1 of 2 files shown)


wiox:
This sounds like an issue that should be resolved in the duplicator library.

thegrb93:
The dupe files contain keyvalues used by DTVars that get set when the entity pasted. I don't think there's any way to clamp the Keyvalues/DTVar values internally.

wiox:
The issue is that ent:RestoreNetworkVars does not perform the minmax checks. That's a massive flaw and needs resolving.

thegrb93:
Oh okay. That's beyond my knowledge. Should I leave this open or close it and let you guys open an issue on that?

thegrb93:
Oh wait, I just need to make an ENT:RestoreNetworkVars function right?

wiox:
> Should I leave this open or close it and let you guys open an issue on that?

I guess that depends on whether or not people think the edit panel's min/max values should effect duplication.

thegrb93:
>Oh wait, I just need to make an ENT:RestoreNetworkVars function right? I think I can do that.

Nevermind, this function doesn't seem to get called when I duplicate the ball.

TheFreeman193:
@thegrb93 RestoreNetworkVars() is defined when InstallDataTable() is called and hence will override the function you're probably declaring with ``` function ENT:RestoreNetworkVars(dt) ... end ```.

I guess the real problem is that the DTVar system offers no simple means of controlling values when restoring them. One way around this, assuming the restraint is desired, would be to use the OnDuplicated() method to check the DT value.

thegrb93:
I'll change it to that if that's what people want, but I think a clamp is the best way.

TheFreeman193:
Following on from what @wiox said, it'd be nice to have a PreRestoreNetworkVars() hook called [here](../blob/master/garrysmod/lua/includes/extensions/entity.lua#L394) allowing access to the datatable before it is applied. It would permit something like this:
```lua
function ENT:PreRestoreNetworkVars( dt )
	dt.BallSize = math.Clamp( dt.BallSize, 4, 128 )
end
```
Also, if we're not going to be controlling the size, there'd need to be some defensive coding on [this](../blob/master/garrysmod/lua/entities/sent_ball.lua#L141) line to prevent negative pitch values generating out-of-bounds errors.

thegrb93:
Would it possible to just clamp it in the OnBallSizeChanged function?

Bo98:
It is called before the DT var is set so probably not.

thegrb93:
I fixed it properly now. Pls discuss.

robotboy655:
Decided to properly fix it, as well as fix pasting duplicated entities with NetworkVars not calling the NetworkVar proxies/callbacks on first spawn.

This falls into the same category as user input, you must sanitize the values before using them to prevent crazy stuff from happening, the min/max stuff in self:NetworkVar is only for the Editing GUI.

robotboy655:
https://github.com/garrynewman/garrysmod/commit/3e138636eb1b0ad6ed785dedf350020755cff5f1?w=1

thegrb93:
Seems a bit silly to sanitize the variable every time it's used, but I'll roll with it.

PR #953, opened 2015-05-17
TTT: Sped up weapon scrolling

MisterDot:
The default speed feels very sluggish in my opinion. I can understand a slight delay to prevent multiple scrolls at once but the previous value felt more like an arbitrary limitation that did not improve the quality of the game. I haven't heard any complaints about the length of delay being proposed here even with at least a year of testing from different communities.

If there's a problem with this change such as something technical I would not mind creating a convar and adding a menu option to allow a custom setting so that the few negatively impacted by this new delay won't suffer. I haven't encountered anyone like that however.


robotboy655:
Congratulations! You just won an award:

The Pull Request of the month!

zerfgog:
Even though this is a pretty minor change (and doesn't affect me much as fast weapon switch is god), I tested this change myself and I enjoy being able to select from crowbar to holstered in one scroll, while maintaining accuracy.

Kefta:
Nice string of PRs you've had. I set it to .25 on my server and it's suited me well over the past six months I've been doing testing on it.

wiox:
0.25 must be insanely slow

Kefta:
Oops

PR #954, opened 2015-05-17
More gamemode map types

Donkie:
Added missing maptypes for Flood and Jailbreak.

### DIFF
--- a/garrysmod/lua/menu/getmaps.lua
+++ b/garrysmod/lua/menu/getmaps.lua
@@ -161,7 +161,6 @@ local function UpdateMapPatterns()
 	MapPatterns[ "^gmt_" ] = "GMod Tower"
 	MapPatterns[ "^jb_" ] = "Jailbreak"
 	MapPatterns[ "^ba_jail_" ] = "Jailbreak"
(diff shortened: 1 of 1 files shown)


Kefta:
I thought ba_ was used by something else as well, maybe I'm wrong

Donkie:
Half the maps I use for my servers are prefixed with ba_, I've never seen it being used anywhere else.

Kefta:
Found it: http://css.gamebanana.com/maps/cats/45

Battle arena was a CS:S knifing gamemode that was around for ~5-6 years. Jailbreak was originally a modification of that gamemode. That's why I made the jailbreak filter ba_jail_ since no jailbreak maps I knew at the time had just the prefix ba_

Donkie:
Ah, makes sense.
I believe Battle arena is more of a CSS thing though, compared to Jailbreak which works both in CSS and Garrysmod, thus IMO ba_ should stay.

Kefta:
Yeah, but CS:S maps are widely used in Garry's Mod. I'll bet that most of your JB maps are from CS:S. ba_jail_ should suffice imo

Donkie:
Alright, removed ba_

PR #955, opened 2015-05-17
TTT: Fixed Rifle scope, zoom spread, unscope delay

MisterDot:
There were two gaps when zoomed in with the rifle on the top and bottom of the screen respectively. Fixed by drawing two lines on the top and bottom of the screen.

It was also possible to miss a well aimed shot with the sniper rifle while scoped previously, though it would be a rare occurrence surely. Fixed by removing all spread while scoped.

In the default weapon base, when the weapon is fired it will delay both its primary and secondary attack by the value of the primary attack. This is just fine for weapons that fire relatively quickly but having to wait a second and a half each time you fire the rifle to unzoom seems like an oversight instead of a conscious design choice especially considering the fact that you can switch weapons freely while this is happening. Feels very clunky. Gave it an 0.1 secondary fire delay after shooting. Should be friendly with weapon base updates/customizations too.

Corrected some tabbing inconsistencies.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_rifle.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_rifle.lua
@@ -44,53 +44,63 @@ SWEP.IronSightsPos      = Vector( 5, -15, -2 )
 SWEP.IronSightsAng      = Vector( 2.6, 1.37, 3.5 )
 
 function SWEP:SetZoom(state)
-    if CLIENT then
-       return
-    elseif IsValid(self.Owner) and self.Owner:IsPlayer() then
-       if state then
-          self.Owner:SetFOV(20, 0.3)
-       else
-          self.Owner:SetFOV(0, 0.2)
-       end
-    end
+   if CLIENT then
+      return
+   elseif IsValid(self.Owner) and self.Owner:IsPlayer() then
+      if state then
+         self.Owner:SetFOV(20, 0.3)
+      else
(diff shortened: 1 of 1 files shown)


MyHatStinks:
> Fixed by removing all spread while scoped.

"Fixed"? Why would you want a weapon with perfect accuracy at any range? I'd say that's the opposite of a fix.

MisterDot:
The Rifle scope was fixed, the rest of the items in the topic were a suggestion. I can understand how that might be misleading but I was working with a character limit.

I'm not sure why you'd disagree with a weapon meant for sniping being perfectly accurate while scoped, though. It's a game where people can teleport and buy equipment out of thin air using points earned by killing people, so if you're talking about realism that's out of the window. And if you're talking about game balance, the weapon is already extremely likely to hit what you aim at to begin with. If you miss in that case it would be due to bad luck.

This probably isn't the best place for this discussion, but if it's seen as an issue I can revert that part. My focus is not to change the game balance but polish the experience by removing unnecessary frustration in this case.

wiox:
I think I agree with @MisterDot. There's spread that exists in a game for balance, stopping you from spraying or having overpowered shotguns, and then there is spread that exists just because it can.
I'd say it's a decision down to @svdm.

svdm:
The downside of removing the spread completely is that it is still perfectly accurate at very long range. With the current cone, the shooter gets a little bit of uncertainty the further they are from their target, which discourages players from just sniping from across the map all game. So I'd prefer to keep it in.

I'm all for the other changes/fixes in this PR.

PR #956, opened 2015-05-17
Fixed nil value error when no command is passed

FastDeath:
Fixes the following serverside error when executing the `` _u `` command without any arguments:

````
[ERROR] lua/includes/modules/concommand.lua:51: bad argument #1 to 'lower' (string expected, got nil)
  1. lower - [C]:-1
   2. Run - lua/includes/modules/concommand.lua:51
    3. unknown - addons/ulib/lua/ulib/shared/commands.lua:1310
     4. unknown - lua/includes/modules/concommand.lua:54
````

### DIFF
--- a/garrysmod/lua/includes/modules/concommand.lua
+++ b/garrysmod/lua/includes/modules/concommand.lua
@@ -47,7 +47,11 @@ end
    Desc: Called by the engine when an unknown console command is run
 -----------------------------------------------------------]]
 function Run( player, command, arguments, args )
-
+	
+	if ( command == nil ) then
+		return false
+	end
+	
 	local LowerCommand = string.lower( command )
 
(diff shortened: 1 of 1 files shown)


wiox:
Sounds like an issue with ulib.

Kefta:
Pass a command then. I'd rather have to error to tell me one isn't being passed

neico:
Well instead of silencing the error I'd use ErrorNoHalt to actually notify the developer / user with a meaningful message...

Kefta:
ErrorNoHalt would just create two errors. I'll make a PR to send an error about a missing command

FastDeath:
Im sorry, I didn't realize the `` _u `` command was a ulib thing.
Probably going to post an issue as neico said. :)

PR #957, opened 2015-05-18
TTT: Fixed C4 beeps not playing through some walls 

MisterDot:
I assume the reason this happened before is the culling done to prevent unnecessary network information from being sent. Unfortunately, hearing something on the other side of a wall that's about to kill you is necessary in my opinion. It's seems that this isn't an intended feature as there are many times that the C4 is well out of reach but still within your PVS.

To test the difference plant a C4 in that infamous spot underground between the two houses on dm_richland. Before you wouldn't be able to hear it but now you will.

I'm not very familiar with using ints/uints in networking so if someone has advice to make this more efficient I'd appreciate it. Already did a bit of optimization which should be enough.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
@@ -330,7 +330,6 @@ function ENT:IsDetectiveNear()
    return false
 end
 
-//local beep = Sound("weapons/c4/c4_beep1.wav")
 local MAX_MOVE_RANGE = 1000000 -- sq of 1000
 function ENT:Think()
    if not self:GetArmed() then return end
@@ -382,7 +381,6 @@ function ENT:Think()
 
       if SERVER then
          -- send the sound manually to prevent culling
-         //sound.Play(beep, self:GetPos(), amp, 100)
          self:SendBeepSound(self:GetPos(), amp)
       end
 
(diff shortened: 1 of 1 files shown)


wyozi:
This might be better way to do it.

http://wiki.garrysmod.com/page/ENTITY/UpdateTransmitState

MisterDot:
"Entity:EmitSound" did not seem to work on the C4. Plain old "EmitSound" works but suffers from the same issue even with the entity being flagged to always transmit. "sound.Play" is of course not the solution either.

Someone else should try.

svdm:
I could see why EmitSound wouldn't work if an entity was not in PVS, but sound.Play should not be affected by that, assuming the server is still running the entity's Think at that point.

MisterDot:
I've given it some thought and I believe the original implementation to be better. The entity itself does not need to be considered outside of PVS. It is only the sound that must always be noticed. The type of variables sent could be lighter but it's already fairly light, possibly lighter even than what sound.Play and alternatives have to send because of how specialized it is.

garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
```diff
@@ -350,7 +381,9 @@ function ENT:Think()
       end
 
       if SERVER then
-         sound.Play(beep, self:GetPos(), amp, 100)
+         -- send the sound manually to prevent culling
+         //sound.Play(beep, self:GetPos(), amp, 100)
```
samuelmaddock: You should probably just remove this commented out line.

PR #960, opened 2015-05-20
PrintTable no longer creates 2 global variables

wiox:


TheFreeman193:
This has been unnoticed for almost a year?! Good spot.

wiox:
I was just playing around with https://github.com/wiox/gmod-global-finder again and noticed it

TheFreeman193:
Ah, fair dues. I wonder how many more fundamental issues like that are lying around in the repo :pensive:

PR #961, opened 2015-05-20
TTT: Weapon base adds view punch to bullet dir

MisterDot:
Bullets fired did not previously consider the view punch added from firing the weapon, meaning that the center of the cone did not line up with the camera. This was fairly noticeable on weapons with a lot of recoil and/or a fast fire rate.

This edit better reflects the direction that the player is looking.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -305,7 +305,7 @@ function SWEP:ShootBullet( dmg, recoil, numbul, cone )
    local bullet = {}
    bullet.Num    = numbul
    bullet.Src    = self.Owner:GetShootPos()
-   bullet.Dir    = self.Owner:GetAimVector()
+   bullet.Dir    = (self.Owner:GetAimVector():Angle() + self.Owner:GetPunchAngle()):Forward()
    bullet.Spread = Vector( cone, cone, 0 )
    bullet.Tracer = 4
    bullet.TracerName = self.Tracer or "Tracer"
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -246,6 +246,7 @@ function SWEP:PrimaryAttack(worldsnd)
    local owner = self.Owner
    if not IsValid(owner) or owner:IsNPC() or (not owner.ViewPunch) then return end
 
+   math.randomseed( self:Clip1() * self.Primary.Recoil * self.Primary.ClipSize )
    owner:ViewPunch( Angle( math.Rand(-0.2,-0.1) * self.Primary.Recoil, math.Rand(-0.1,0.1) *self.Primary.Recoil, 0 ) )
 end
 


wiox:
It'd be sweet if somebody fixed these weapons to be properly predicted too (pls)

MisterDot:
Sure, I wouldn't mind trying. What should I be looking for in particular?

wiox:
At this point, everything. If I remember correctly there's some code like "if not IsFirstTimePredicted() then return end". That shit gotta go.

http://wiki.garrysmod.com/page/Prediction

xaviergmail:
This is completely horrible. Have you not tested this? ViewPunch fucks up, it is FPS dependent. Therefore your client's ViewPunch will decay much faster than the server and thus, cause prediction issues. I'm voting no for this until ViewPunch gets fixed.

wiox:
ViewPunch isn't broken, but given that the weapon isn't predicted properly it'll probably act like it is broken.

xaviergmail:
It is. I had to rewrite ViewPunch in Lua entirely to counter this.

wiox:
Prove it.

MisterDot:
I have tested this for a considerable amount time actually and I can say for sure that it is an improvement. However it is not perfect. I wasn't sure why exactly sure why. Your explanation gives some insight. I would certainly not call it horrible xavier unless you've got some convincing documentation, but depending on the level of desync I would agree that this change should wait.

xaviergmail:
There is a slight de-sync but nothing to be worried about I guess. I'm  just anal about prediction that this bothered me for my server. I don't see it being such a problem for others.

Kefta:
@wiox The TTT base uses IsFirstTimePredicted correctly: he sets the predicted values, then returns before the trace/shot is done. Only thing that needs to be done is added lag compensation to the bullet trace.

wiox:
That explains why shooting with a high ping gives such jittery recoil

Jvs34:
There's desync because there's a non predicted random in the viewpunch code in PrimaryAttack.

```
owner:ViewPunch( Angle( math.Rand(-0.2,-0.1) * self.Primary.Recoil, math.Rand(-0.1,0.1) *self.Primary.Recoil, 0 ) )
```

And also in ShootBullet there's a SetEyeAngles which is only called clientside in multiplayer for some reason?

wiox:
LuaJIT's RNG giving the same results cross-platform isn't exactly something I'd trust to always be around. I'd suggest using util.SharedRandom rather than math.randomseed and math.random.

Faerachnidendro:
Sorry to necro here, but willox, thank you for that little gem of info right there about util.SharedRandom, I'm definitely going to roadtest it on my weapons.

RE: weapon_tttbase and it's concept of recoil/viewpunch and accuracy penalties... (excuse my attitude here) give up, write your own base or modify a better quality existing one, to be based off of weapon_tttbase and otherwise suit the needs of the gamemode.
The vanilla ttt weapon base is just deplorable in so many ways and rewards utterly illogical use of weapons(nevermind the absolute bollocks damage on some), compared to the standard of weapons available on workshop and etc... you're far better of with them as a starting point, either literally or as reference material to write your own, especially if you're after more 'realistic' weapons that behave a little more like you might expect them to, with greater control of the handling/stability/accuracy of each weapon.

(Please note that my frustration with the ttt weapon base is directed at BadKing, who I'm guessing did the brunt of the work there and then just left it as-is, I mean as little disrespect as possible :P)

PR #962, opened 2015-05-22
TTT: Allow detectives to reequip their dropped hat

MisterDot:
A vital, long awaited solution that should finally solve this horse shit bug. How are we supposed to get immersed if we can't even pick our own hats back up? Who thought that was a good idea? No wonder TTT is so unpopular.

Comes with a cvar that is enabled by default. I'd send the player a text notification but I'm not sure I want to give translators any more hassle.

A slight sound is made when equipping the hat ~~which only detectives may do~~ if they are missing one for whatever reason. It doesn't have to be your own hat.

Edit: added a cvar disabled by default that allows innocents to equip hats as well.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
@@ -5,6 +5,8 @@ ENT.Type = "anim"
 ENT.Base = "base_anim"
 
 ENT.Model = Model("models/ttt/deerstalker.mdl")
+ENT.CanHavePrints = false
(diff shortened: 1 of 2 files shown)


Kefta:
What new things need to be translated? I can do Spanish and Swedish, I can get Russian and Chinese done by friends, and someone is actively maintaining Portuese. You could go ahead and out the text in

MisterDot:
Ah, I'm glad it's not a problem then. Thank you.

"hat_retrieve" must be translated into something appropriate.
"You picked up a detective's hat." is generic yet accurate and effective.

wyozi:
Players should totally be able to pick up the hat as well. Just think about the RP possibilities.

PR #963, opened 2015-05-23
Improved duplicator and properties support for prop_effect

NO-LOAFING:
- Effects now save their AttachedEntity's entity modifiers, bodygroups, and bone manips when copied, and reapply them all to the AttachedEntity when pasted. Before, all of these were lost when saving.

- Fixed an issue where using the duplicator on an effect would copy the original effect's AttachedEntity value, causing anything that looks for ent.AttachedEntity to get the wrong entity. For example, if you spawned an effect, copy-pasted it with the duplicator, and then used the color tool on the new effect, it would recolor the first effect instead of the one you clicked on.

- The Skin, Bodygroup, and Bone Editor properties (context menu right-click options) now all work properly with effects - if the ent you've right-clicked on has an AttachedEntity value, the properties will use that instead.

- An effect's AttachedEntity value is now available clientside. This was necessary to get the properties working, but I'm sure other things could have a use for it too.


robotboy655:
:+1:

PR #969, opened 2015-05-29
Apply playermodel colour to corpses/ragdolls

DarthTealc:
Applies the playermodel colour to corpses/ragdolls in TTT. This isn't the optimal way to do it but it's the best I can do without messing too much with the corpse code.
Current TTT: Ragdolls all take the default blue colour regardless of the colour of the living players
With this change: Ragdolls maintain the colour used for playermodels this round.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
@@ -5,6 +5,8 @@ CORPSE = {}
 
 include("corpse_shd.lua")
 
+util.AddNetworkString("TTTCorpseCreated")
+
 --- networked data abstraction layer
 local dti = CORPSE.dti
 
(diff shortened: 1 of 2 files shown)


zerfgog:
that code isn't fun

DarthTealc:
Yeah I know. It's the best I'm able to do without rewriting how corpses work. :/
If you (or anyone else) have a better solution I'll close this one. Anyone out there who has worked a bit with TTT ragdolls?

I think the best method would be if Gmod could set the colour serverside and have that network to clients, as it does when you do it on a player. I don't think that's something we can add though.

thegrb93:
I would send the entity index immediately rather than 100ms after creation and once the client receives the index, create a think hook which waits for the entity of that index to become valid (or removes the hook after a couple seconds if the entity never appears).

thegrb93:
Added some pseudo code

wiox:
I think if you use a DTString it'll be readable in the OnEntityCreated hook clientside.

robotboy655:
Using a DTString or DTVector seems like more elegant solution, then you just have to apply the
```e.GetPlayerColor = function() return e:GetDTVector( 1 ):ToColor() end``` to ragdolls on client.

**Comment by comedinha** [comment]
Or change
```
rag:SetColor(ply:GetColor())
```
to
```
rag:SetColor(ply:GetPlayerColor())
```
?

comedinha:
No fixes?

DarthTealc:
I no longer have a TTT server to test this on so I'll pull the request. If someone wants to use the code to make their own working version, feel free to.

PR #970, opened 2015-05-30
Fix hat_retrieve German translation

doctorluk:


Kefta:
Sorry about that; asked a Swedish friend who was semi-fluent for help on the German translation

PR #973, opened 2015-06-01
Fix weapon_zm_rifle's IconLetter

Fuzzik:
The wiki has r and n swapped. Whereas r should be the AWP and n should be the Scout.


zerfgog:
Does the wiki have them mixed up?e

Fuzzik:
Yeah, I left a comment on the original commit that added them.

PR #978, opened 2015-06-03
Error/Crash Prev.

shadowmaruz:
By having validity check before we call a non-global function (instead of during) prevents errors.


wiox:
Lua has short-circuit evaluation. In the original code, if IsValid( attacker ) returns false, attacker:IsPlayer() won't be called.

robotboy655:
The crash issue with TTT was already fixed.

PR #979, opened 2015-06-04
Consistency for number of return values

JamesxX:
Gamemode hooks can return any number of values where as hooks created through hook.Add() could only return at most 6. This pull creates consistency between the two by allowing hooks created through hook.Add() to also return any number of values.

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -1,8 +1,8 @@
-local gmod                        = gmod
-local pairs                        = pairs
-local isfunction        = isfunction
-local isstring                = isstring
-local IsValid                = IsValid
+local gmo			= gmod
+local pair			= pairs
+local isfunction		= isfunction
+local isstrin			= isstring
+local IsVali			= IsValid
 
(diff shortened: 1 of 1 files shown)


notcake:
Are there any hooks which actually return 6 or more values?

I don't think this is worth the performance hit from the garbage created. Some hooks are many times per frame (~30 per frame for HUDShouldDraw and 900 times per frame for PlayerCanHearPlayersVoice for 30 players)

garrysmod/lua/includes/modules/hook.lua
```diff
@@ -93,7 +93,7 @@ function Call( name, gm, ... )
 					--
 					-- If the object is valid - pass it as the first argument (self)
 					--
-					a, b, c, d, e, f = v( k, ... )
+					a = { v( ... ) }
```
notcake: This should be v( k, ... )

JamesxX: Thanks

robotboy655:
This was already discussed, it's not going to happen.

You can always return a table from your custom hooks.

PR #982, opened 2015-06-06
Addon compatibility

shadowmaruz:
After the recent update returning nil on a player say hook will cause the following error :"Error: hook->PlayerSay returned a non-string!".

Many addons return nil during their hooks.  

This will stop those errors.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/player.lua
+++ b/garrysmod/gamemodes/base/gamemode/player.lua
@@ -97,7 +97,7 @@ end
 -----------------------------------------------------------]]
 function GM:PlayerSay( player, text, teamonly )
 
-	return text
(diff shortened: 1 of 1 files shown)


Kefta:
That's not the problem. I already fixed the problem in my latest update; if an addon is sending nil into PlayerSay, then they aren't using the hook correctly

shadowmaruz:
returning, not sending.

Kefta:
They can just return then, not return nil. 

Anyway, there's no point to this since text will never be nil unless nil is specifically sent in, which then there would be no point of using the hook.

KrakenZ:
The game should not adapt to addons; Addons must adapt to the game.

PR #983, opened 2015-06-06
TTT: Disable weapon picking up

Kefta:
This crash has been in TTT forever and now it's being widely abused. Until the magneto stick is entirely redone or the cause of the crashes are found, this should be disabled by default


wiox:
If https://github.com/Facepunch/garrysmod-requests/issues/424 were implemeneted, the magneto stick could be properly fixed.

nudge nudge rubat

PR #987, opened 2015-06-08
TTT: Fixed console chat behaviour

Kefta:
Last one hopefully :rainbow:

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
@@ -79,7 +79,7 @@ local function AddDetectiveText(ply, text)
 end
 
 function GM:OnPlayerChat(ply, text, teamchat, dead)
-   if not IsValid(ply) then return true end 
+   if not IsValid(ply) then return self.BaseClass:OnPlayerChat(ply, text, teamchat, dead) end 
    
    if ply:IsActiveDetective() then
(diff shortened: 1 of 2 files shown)


wiox:
This entire file is actually calling base methods wrong, I'm surprised it works. Even if not using DEFINE_BASECLASS and a BaseClass upvalue, the self parameter to each method should be the gamemode itself.

Kefta:
So should I be calling baseclass.Get/DEFINTE_BASECLASS or replace the self calls with GAMEMODE?

wiox:
It doesn't matter that much, but only because of pure luck.

```lua
self.BaseClass:Method() -- Wrong, self parameter passed to Method is not an instance of the gamemode
self.BaseClass.Method( self ) -- Wrong if the gamemode is ever derived from, but I don't know anybody that does this with TTT.
baseclass.Get( "gamemode_base" ) . Method( self ) -- Right
```

Instead of baseclass.Get you'd ideally use the DEFINE_BASECLASS macro.

Kefta:
@wiox @svdm Baseclass fixed. I'll fix it in the other TTT files in a separate PR

wiox:
You need to pass the gamemode's instance as the first parameter, not the base class itself.

    BaseClass.Method( self, ... )

Kefta:
@wiox Thanks; never called the baseclass this way before

PR #991, opened 2015-06-13
Make light and lamp entities editable

SuperHighVoltage:
This allows the light and lamp entities to be edited from the context menu, allowing for users to change its properties and see the results in real time


robotboy655:
I wanted to do this for all (most) the entities, but now that I think about it, this would allow people make rainbow lights and shit since you can spam the net messages with Lua.

SpencerSharkey:
Sounds like a really useful addition

+1 for rave lights

robotboy655:
Rave lights are the least of my concerns.

robotboy655:
I have committed a better version of this, including proper support for direct method calls & editable support for more tools.

PR #992, opened 2015-06-14
Add server list "All gamemodes" mode, make default

DarthTealc:
Screenshot: https://i.imgur.com/nWxzr6z.png

Additional related changes/notes:
- Gamemode column added.
- Fake "all" gamemode created (displays as "All Gamemodes") as a server gamemode group. All servers are added to their real gamemode plus this one.
  - Any gamemode which actually calls itself "all" (no legit ones should) will be filtered into an "all_" category as their real gamemode but should display as "all" or whatever the gamemode's display name is.
- Added two new language strings. Will need translations for non-English languages.
  - "server_gamemode" (English: "Gamemode") - for the Gamemode column title
  - "filter_by_gamemode" (English: "Filter by Gamemode") - button to show you a list of gamemodes that you can filter by.
- Servers displayed increased to 256 (from 128).


Donkie:
Wouldn't an infinite-scroll system work?

Bo98:
> Wouldn't an infinite-scroll system work?

Lags when scrolling back up.

If people are interested, I can finish up my WIP server browser performance boosts I worked on a few months back.

DarthTealc:
Could you please skim over my changes and see if there's anything that could be improved for performance too? Fork (or copypaste) my branch if you want.

Bo98:
Don't see any obvious performance issues specific to the changes you made.

DarthTealc:
Any chance of this being added? Are there changes I need to make?

BlackCetha:
+1 Good idea, nicely done

DarthTealc:
The filtering by gamemode by default has killed too many smaller and unique gamemodes. Please consider making this change if going back to the 'legacy' server list isn't happening.

DarthTealc:
Anything I can do to improve this to get it added?

uRandomAlex:
Looks good. But then it will kill some special and popular gamemodes with a few servers running it, lol...
I think you need to add a real FILTER to gamemodes, with ability to blacklist gamemodes. Many people do not play sandbox, or just don't want to play it this time, but there is MANY, MANY servers with it, and because of a limit user will not see other servers. Also, fake gamemode and other hacks isn't so good.
When you will done with blacklist thing, you need to create pull request again, because Rubat, Willox or who-is-working-on-GMOD-too will not be able to see it.

robotboy655:
I don't think this is a good idea at all.
1) This will further incentivize servers to fake player counts/ping/anything because they are not just competing against serves in the same gamemode any longer, but also against servers with other gamemodes
2) The 256 server limit is definitely unacceptable for this sort of listing ( or any limit tbh )
3) This is essentially a crappier version of the legacy browser

PR #997, opened 2015-06-25
Add COLOR:ToInt and IntToColor

Kefta:
It has been a good nine months since I made my original PR to add this( #820 ), but after relearning the entire English language, Lua, and the bit library, I now see what Willox meant. It was an integer the whole time :ok_hand:

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -86,6 +86,6 @@ function IntToColor( int )
         return Color( 
         	bit.band( bit.rshift( int,16 ), 0xFF ), 
         	bit.band( bit.rshift( int,8 ), 0xFF ), 
-        	bit.band( bit.rshift( hex,0 ), 0xFF ) 
+        	bit.band( bit.rshift( int,0 ), 0xFF ) 
         )
 end
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -71,3 +71,21 @@ function COLOR:ToVector( )
 	return Vector( self.r / 255, self.g / 255, self.b / 255 )
 
 end
+
+--[[---------------------------------------------------------
+	Converts color to an integer/hex - loss of precision / alpha lost
+-----------------------------------------------------------]]
+function COLOR:ToInt( c )
+	return bit.lshift( c.r,16 ) + bit.lshift( c.g, 8 ) + bit.lshift( c.b, 0 )
+end
+
+--[[---------------------------------------------------------
+	Converts integer to color - no alpha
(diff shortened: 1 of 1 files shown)


thelastpenguin:
+1 code_gs worked so hard on this

wiox:
I still like the idea of the Color function being able to take a single argument for this rather than a new IntToColor function.

Kefta:
How should Color's tostring act in this case if people want the integer?

wiox:
It's unlikely you'll want the integer for debugging. It's pretty useless compared to each separate channel.

Kefta:
Is it okay if I keep COLOR:ToInt and remove IntToColor in favour of letting Color accept integers?

notcake:
I don't like the function names. It's not clear whether :ToInt packs the alpha channel as well and that IntToColor doesn't handle alpha.
:ToRGB888 and RGB888ToColor (as opposed to ToARGB8888 and RGB8888ToColor) would be ok but feels a bit verbose.

Kefta:
ToRGB888 still feels a bit verbose; ToInt seemed fine since it implied standard RGB implementation that everyone who is aware of color hexes are familiar with

thelastpenguin:
@kefta why not simply include the alpha component and eliminate all ambiguities?

Kefta:
@thelastpenguin The original intent when I made these last year was to use standard #RRGGBB hex/int color implementation that could be used with webpages and storage, however, including the alpha wouldn't be an issue. If everyone agrees ToInt is descriptive enough of a name for an RRGGBBAA color, I'll make the change. I'll also allow Color() to take integers instead of having IntToColor.

wiox:
RRGGBBAA means that Color( 0xFF0000 ) won't be red, which completely invalidates the only good reason this PR exists for.

notcake:
@wiox I think @kefta meant AARRGGBB.

Kefta:
http://pastebin.com/FgXy1fDL This look good? Has support for using integers in Color() and 0xFF0000 is still red

wiox:
```lua
string.len( r ) < 10
```

Wat?

```lua
r < 0xFFFFFF
```

Kefta:
0x000000?

wiox:
You've got to decide if your function takes a string or an integer.

Kefta:
Color had tonumber's before in it, so I'd gladly make it only take integers if you wanted to break that support.

Kefta:
I'll have to take the string since 0x00000000 will cause issues with integer length.

Kefta:
Revised code: http://pastebin.com/Uh4xGVgG

Kefta:
Refer to https://github.com/garrynewman/garrysmod/pull/1050

PR #998, opened 2015-06-25
Add Tooltip accessors

Kefta:
Solves https://github.com/Facepunch/garrysmod-requests/issues/514

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel.lua
+++ b/garrysmod/lua/includes/extensions/client/panel.lua
@@ -256,6 +256,22 @@ function meta:PositionLabel( labelWidth, x, y, lbl, ctrl )
 end
 
 --[[---------------------------------------------------------
+	Name: GetTooltip
+-----------------------------------------------------------]]
+function meta:GetTooltip()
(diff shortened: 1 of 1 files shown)


zerfgog:
please don't introduce aliases

Kefta:
I have to since the modifier has one. I'll deprecate both when this gets accepted.

Bo98:
I understand why you added it but I honestly don't see the point of adding an alias if we are just going to tell everyone not to use it. That's my opinion anyway.

Kefta:
If I can remove the SetToolTip aliases, I won't add these. However, I would rather remove them at the same time instead of having a discrepancy.

Kefta:
@zerfgog @Bo98 Which one is deprecated -- Tooltip or ToolTip?

Bo98:
http://wiki.garrysmod.com/page/Panel/SetToolTip

Kefta:
@Bo98 Thanks!

Kefta:
@robotboy655 Can this be merged?

PR #999, opened 2015-06-26
Improve map filtering display

DarthTealc:
Currently when searching for a map (singleplayer) it shows all the categories and it can be hard to see which ones have matches.

This change makes it only show categories with matches: https://puu.sh/iAONC/119f84b5a4.jpg


Kefta:
Great change :+1:

PR #1000, opened 2015-06-26
Fixed PrintTable erroring when it fails tostring

DaaOtt:


wiox:
tostring shouldn't have any failure conditions

wyozi:
tostringing panels can fail iirc

wiox:
tostring isn't supposed to fail unless given 0 arguments. That should be fixed.

robotboy655:
tostringing physobjects and panels has been already fixed

PR #1003, opened 2015-06-30
Add Particle convenience function

Kefta:


wiox:
Why shared?

Kefta:
Yes. You can use the particle string in shared functions while still caching it clientside in one nice function

PR #1004, opened 2015-07-02
TTT: Removed all traces of teamchat/globalchat hints

Kefta:
TTT likes to call cl_lang instead of the actual lang file for language strings in some cases. This will get rid of the removed language strings for good.


Bo98:
Before anybody asks, this doesn't fix any error AFAIK. That is a problem with outdated addons.

Kefta:
Yeah, but this will at least error now through the addon and not TTT, letting the owner know to update their content.

PR #1005, opened 2015-07-03
Added math.IsBetween and math.IsEven

Mysterypancake1:
I've been in many situations where I needed to check whether a number was between two other numbers, for example when using convars with limits. I just thought this might be helpful, even though a function returning booleans may not be classified as 'math'.

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -70,7 +70,7 @@ math.Min = math.min
     Name: EaseInOut(fProgress, fEaseIn, fEaseOut)
     Desc: Provided by garry from the facewound source and converted
           to Lua by me :p
-   Usage: math.EaseInOut(0.1, 0.5, 0.5) - all parameters shoule be between 0 and 1
+   Usage: math.EaseInOut(0.1, 0.5, 0.5) - all parameters should be between 0 and 1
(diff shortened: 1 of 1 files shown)


wiox:
This is math.IsBetweenOrEqual. You can also remove the "== true" from your code.

Mysterypancake1:
Saying 5 isn't in between 5 and 7 for example doesn't sound very logical?

wiox:
Normally I'd agree. In a programming language that has separate "greater than" and "greater than or equal to" operators, I wouldn't. How about math.InRange?

TheFreeman193:
An _inclusive_ parameter might be nice in something like this.

bmwalters:
Not sure how this change is meant to help developers. 
`if math.InRange(5, minheight, maxheight) then`
is actually four more keystrokes than
`if 5 >= minheight and 5 <= maxheight then`
It isn't even hard to remember how to write an "is in range" check, as opposed to other math functions like math.NormalizeAngle (which is also short, but is nice not to have to remember).

It's a similar case with `math.IsEven`. Additionally, the introduction of `math.IsEven` into the codebase might result in people using the following to check if a number is odd:
`if not math.IsEven(num) then` (as opposed to `num % 2 == 1`
*shudder*

Seeing as #1229 (a similar PR) was also closed (:+1:), I don't see this being merged.

PR #1006, opened 2015-07-05
Add util.Tracer

Kefta:


Kefta:
This doesn't work correctly without having a way to get ParticleIDs: https://github.com/Facepunch/garrysmod-requests/issues/632

PR #1008, opened 2015-07-07
Cleaned up weapons

Kefta:
Added SWEP.HoldType to base variables for easy setting without overwriting Initialize, removed useless variables, fixed up a lot of syntax, removed repeat code, and replaced all useless self.Weapon instances with self.

### DIFF
--- a/garrysmod/gamemodes/base/entities/weapons/weapon_base/cl_init.lua
+++ b/garrysmod/gamemodes/base/entities/weapons/weapon_base/cl_init.lua
@@ -1,23 +1,28 @@
 
-include('shared.lua')
+include( "shared.lua" )
 
-SWEP.PrintName			= "Scripted Weapon"		-- 'Nice' Weapon name (Shown on HUD)	
-SWEP.Slot				= 0						-- Slot in the weapon selection menu
-SWEP.SlotPos			= 10					-- Position in the slot
-SWEP.DrawAmmo			= true					-- Should draw the default HL2 ammo counter
-SWEP.DrawCrosshair		= true 					-- Should draw the default crosshair
-SWEP.DrawWeaponInfoBox	= true					-- Should draw the weapon info box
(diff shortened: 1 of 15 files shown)


wiox:
You should remove the instances of 'self.Weapon'.

Kefta:
Hold on, this isn't finished. GitHub decided to push this from the client for some reason

Kefta:
Ignore this; git isn't playing nice

Kefta:
Done @wiox

robotboy655:
I have a few problems with this PR:
1) This is ugly, and completely unnecessary:
```
SWEP.Secondary = 
 {
 	ClipSize = 8,					-- Size of a clip
 	DefaultClip = 32,					-- Default number of bullets in a clip
 	Automatic = false,				-- Automatic/Semi Auto
 	Ammo = "Pistol"
 }
```

2) I really don't like the idea of adding SWEP.HoldType by default, it will just add confusion, especially for new comers when they start overriding SWEP:Initialize and complain why SWEP.HoldType does not work.

3) And I really hate the fact that you changed some SWEP.* fields into locals. There's absolutely no need for that, you are just fucking up "modding" support of those SWEPs.
```
 SWEP.HealAmount				= 20	-- Maximum heal amount per use
 SWEP.MaxAmmo				= 100	-- Maxumum ammo
```

```
local HealAmount	= 20	-- Maximum heal amount per use
```

Kefta:
I agree with 3; I made it a while ago and don't remember why I made them locals. As for 1, I can agree to the preference and I will change it. On 2, I assume that's what was intended since your NP++ highlighting plugin considers SWEP.HoldType as a default variable, plus I think it's just more convenient. People can always just run SetHoldType themselves if they're starting on SWEPs, and I think it's just useful overall.

robotboy655:
If you wish to get this in, recreate it with fixes for stuff I pointed out, and move the TTT stuff into a separate PR.

PR #1009, opened 2015-07-07
Added DComboBox:SetSortItems

disseminate:
Allows coders to disable alphabetical sorting of DComboBox items and instead use the order in which options were inserted


robotboy655:
You gotta rebase

robotboy655:
Recreate it if you want it in.

PR #1011, opened 2015-07-07
Removed arbitrary undo limit

ottworks:


mcd1992:
Commenting out code isn't how you remove it on a revision control platform like git.

Svix:
There is actually a sane reason for doing this, believe it or not. Forgive me if I have forgotten some of the semantics here but the short version was that huge amounts of objects in the undo list would cause a memory overflow/buffer overflow due to the way they're stored -- However I'm sure increasing the limit to some huge number like 1012 would hardly upset anyone.

robotboy655:
Not to mention the limit only applies to UI, not the undo library as a whole.

I imagine huge numbers would still cause problems because the panel is rebuilt every time you create/remove "an undo", which might cause freezes with bigger amounts. I will increase it, but not to 1k.

PR #1012, opened 2015-07-10
Added string.Random(length, declower, decupper)

Mysterypancake1:
The declower and decupper values are the ASCII decimal codes of the characters you may want to generate random letters between (some decimal codes are viewable here : http://www.ascii-code.com)

For example, string.Random(6, 48, 57) would generate a string consisting of 6 random numbers. You can probably explain this in a better way than I can.

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -356,3 +356,17 @@ function string.Comma( number )
 	return number
 
 end
+
+--[[---------------------------------------------------------
+    Name: Random( length, declower, decupper )
+    Desc: Generates a random string of a specified length, with declower and decupper being two optional ASCII character decimal codes to choose characters between.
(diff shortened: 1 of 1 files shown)


robotboy655:
Can we stop with the pointless functions

Mysterypancake1:
But math.Random is okay? I suppose it is more useful though, I just thought something like this could be used to generate random names for something to be altered later, such as the automatically generated password for new Facepunch users.

disseminate:
The reason he's not adding this is that you can do this yourself in four lines without the need of a function (a for loop with math.random(), string.char() and an empty string).

zerfgog:
@Mysterypancake1 probably should think of some example usages before making a pr next time; what would the random names to be altered later be used for? etc.

Mysterypancake1:
Yeah, I have a bad habit of coming up with some useless thing and then requesting it for some reason.

PR #1013, opened 2015-07-10
Added 4 new functions

Mysterypancake1:
I always thought it was odd that the math library didn't include many basic functions such as ways to find the average, mean and mode of a group of numbers. I was originally going to only make these functions work with a variable number of arguments that were only numbers, but then I thought of how hard it'd be to not only get all your numbers together, but to separate them by commas so the function actually worked. Most would just use a table, so that's why I made these functions work with both numbers and tables. The function math.GetNumbersFromTbl would make more sense in the table library, but since it'd be pretty pointless adding additional functions to the table library such as GetBoolsFromTbl to make that function seem fit, I decided to keep it in the math library just because it has something to do with numbers. The reason these functions don't check subtables for numbers is because it'd get a whole lot more complicated when you could just loop through the keys of the table or specify the key containing a group of tables you may want.

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -264,3 +264,82 @@ end
 function math.Remap( value, inMin, inMax, outMin, outMax )
 	return outMin + ( ( ( value - inMin ) / ( inMax - inMin ) ) * ( outMax - outMin ) )
 end
+
+--[[---------------------------------------------------------
+    Name: GetNumbersFromTbl( table, subtablecount )
+    Desc: Loops through a table to retrieve any values that have the type of a number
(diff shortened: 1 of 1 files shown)


notcake:
I don't mean to discourage you, but I don't see myself using these functions in their current form and would probably end up rolling my own if I needed statistics functions.

- math.GetNumbersFromTbl should be called table.Flatten or something similar and should handle any non-table data type.
- I'm not a fan of how math.Mean, Mode and Median support recursive nested associative tables, I don't see that particular feature being used very often. Numeric arrays or vararg numbers are main ways they would be called.
- math.Mean, Mode and Median create garbage when called (the {...} table). This is bad for performance if called very many times to process large numbers of data sets.
- Statistics functions don't feel like they belong in the math library, maybe a new statistics library table would be better.
- You're missing a function to calculate the variance of a data set.

PR #1014, opened 2015-07-10
Added two new functions

Mysterypancake1:


Kefta:
I personally think these functions are useless, but either way, remove the local function in flatten. Just run the loop straight through.

zerfgog:
in the current implementation it's required for nested flattening

Mysterypancake1:
Kefta, if I ran the loop straight through how would I return a single table?
Also, I've found the table.IsEmpty function quite handy when attempting to loop through a table that may or may not have values

PR #1016, opened 2015-07-13
Removed clamping in light stool

besogon:
Seriously though, I can still spawn bright lights through duplicator. What's the point in making lights such dimmed?

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/light.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/light.lua
@@ -36,12 +36,6 @@ function TOOL:LeftClick( trace, attach )
 	
 	local key = self:GetClientNumber( "key" )
 	
-	-- Clamp for multiplayer
-	if ( !game.SinglePlayer() ) then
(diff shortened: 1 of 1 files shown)


robotboy655:
The point is to stop people from spawning lights that light up the entire map and lag everyone's PCs into oblivion. But now that you mention the duplicator, I better take care of it too.

thegrb93:
Related https://github.com/garrynewman/garrysmod/pull/952

besogon:
What's the point? Isn't it server admin's business to overwatch stuff happening on server?

besogon:
What if player gets trapped in 100 thrusters with smoke effect (which is additive)? This surely will lag their videocard. But we don't limit thrusters in lua code, are we?

besogon:
I'd understand clamping those values to max values of sliders in tool tab, but this... This is just random.

thegrb93:
sbox_maxthrusters?

besogon:
Yeah, server admins control thrusters amount, but not clamping.

besogon:
I've commited proper clamping, please review.

thegrb93:
Don't see it.

Bo98:
Will only appear when the pull request is reopened.

PR #1022, opened 2015-07-19
string.Comma efficiency

Kefta:
Also added optional second argument for dots instead of commas

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -344,15 +344,9 @@ function string.ToColor( str )
 
 end
 
-function string.Comma( number )
-
-	local number, k = tostring( number ), nil
-
-	while true do  
-		number, k = string.gsub( number, "^(-?%d+)(%d%d%d)", '%1,%2')
-		if ( k == 0 ) then break end
-	end
-
-	return number
-
+function string.Comma( number, period )
+	number = tostring( number )
+	local symbol = period and "." or ","
+	-- We reverse so that we can start in patterns of three and not have to deal with 1-2 num precedes( 25,000,000 )
+	return number:reverse():gsub( "(...)", "%1" .. symbol ):gsub( "%" .. symbol .. "$", "" ):reverse()
 end


wiox:
This doesn't work for values with decimal points.

thegrb93:
>Don't try to fix what isn't broken?

@wiox It's called improvement. Even though it doesn't work in this case, there are many others that could use it.

wiox:
The second argument is a perfect example of improvement. The unnecessary attempt to make it more efficient which causes bad output for a common use-case isn't.

robotboy655:
I agree with @wiox, an improvement is when there are no regressions.

thegrb93:
Like I said, this isn't a case of improvement, but you shouldn't dismiss attempts of improvement with "don't fix what isn't broken"

wiox:
What happens if optimization PRs begin being accepted and nobody notices the tiny flaws that they introduce?

thegrb93:
That can be said about any pr.

wiox:
Hell no. If this PR had been as simple as allowing for "." rather than "," it'd be obvious if it introduced any bugs or not.

thegrb93:
So what? He wanted to improve the algorithm. Yes a new algorithm requires more testing, but that is okay as long as it is tested.

TheFreeman193:
```lua
local function addCommas( str, period )

	local symbol = period and "." or ","

	str = str:reverse():gsub( "(...)", "%1" .. symbol ):reverse()

	if ( str:StartWith( symbol ) ) then return str:sub( 2 ) else return str end

end

function string.Comma( number, period )

	number = tostring( number )

	local dpos = number:find( "." , 0, true )

	if ( dpos ) then
		return addCommas( number:sub( 1, dpos - 1 ), period ) .. number:sub( dpos )
	else
		return addCommas( number, period )
	end

end
```
Handles decimals and is quicker than both; much quicker with longer strings.

Kefta:
I didn't even think to test floats with this; sorry for the oversight. 

@TheFreeman193 Nice job; please make a PR to my branch for this

SimonSchick:
This kind of commit is why we should have unit tests (for basic functions).

zerfgog:
Please write a unit test lua file for every library function. Please include three different tests per function. Your time starts now.

TheFreeman193:
@Kefta Will do. Going to tweak it a little in order to accommodate negative values, but this won't be an issue.

PR #1024, opened 2015-07-20
Added PlayerRunCommand hook

aStonedPenguin:
In many addons people have to detour concommand.Run so it'd be better to just have a hook.

### DIFF
--- a/garrysmod/lua/includes/modules/concommand.lua
+++ b/garrysmod/lua/includes/modules/concommand.lua
@@ -51,6 +51,7 @@ function Run( player, command, arguments, args )
 	local LowerCommand = string.lower( command )
 
 	if ( CommandList[ LowerCommand ] != nil ) then
+		hook.Run( "PlayerRunCommand", player, command, arguments, args )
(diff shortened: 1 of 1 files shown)


thelastpenguin:
:+1:

crazyscouter:
+1

SimonSchick:
If you add the hook, why not make it able to prevent the actual execution by return false for instance?

SimonSchick:
Also, this also runs when a command is executed by console so "PlayerRunCommand" is kinda misleading, how about "RunConsoleCommand"?

Kefta:
PlayerSay also runs when console talks. If it's changed, which I don't think it should, it should be OnCommandRun; RunConsoleCommand can be confused with the function to run commands on the player.

thegrb93:
OnConcommand

aStonedPenguin:
Yeah OnCommandRun is a better fit.

uRandomAlex:
This should be done in engine itself, because engine calls concommand.Run only if command is not registered in engine.

PR #1025, opened 2015-07-21
Fix bug in string.JavascriptSafe

meepdarknessmeep:
for example ("\x0034"):JavascriptSafe() would result in \034 in javascript land, which is wrong


neico:
I don't see what this has to do with the zero string delimiter...

\0 != 0

wiox:
string.JavascriptSafe ( "\0" .. "1" ) == "\\01" which JavaScript will interpret as a single byte. This is a good PR.

neico:
If you use it like that then this makes more sense, his example though doesn't as much, \x would be \\x and therefore change anything, and if I'm not totally wrong then \x0034 == \x34 == 52, but since I rarely use hex values in strings at all I wouldn't know for sure :/

wiox:
"\x0034" = "\x00" .. "34"

neico:
oh right, 2 numbers build one hex value. I really don't do this stuff enough...

nevermind then, go ahead~

PR #1026, opened 2015-07-24
Typo in Primary Fire Sound

YourStalker:
The real question is, why did the sound work in the first place and how was this silly typo never noticed?


zerfgog:
That could be one of those special modifier strings placed before sound names, or it could be a typo. Not sure why it would only be placed on that sound, so probably a typo.

Kefta:
None of their are typos; they're all soundscape modifiers interpreted by the Source engine

zerfgog:
@Kefta :+1:

YourStalker:
Interesting, can anyone link me to more info on these modifiers? I can't seem to find any.

Bo98:
https://developer.valvesoftware.com/wiki/Soundscripts#Sound_Characters

PR #1027, opened 2015-07-24
Add LerpColor

Kefta:
Since we have LerpAngle and LerpVector

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -66,8 +66,17 @@ end
 --[[---------------------------------------------------------
 	Converts Color To Vector - loss of precision / alpha lost
 -----------------------------------------------------------]]
-function COLOR:ToVector( )
+function COLOR:ToVector()
 
 	return Vector( self.r / 255, self.g / 255, self.b / 255 )
+
+end
+
+--[[---------------------------------------------------------
+	Linear interpolates between two colors
+-----------------------------------------------------------]]
+function LerpColor( fraction, from, to )
+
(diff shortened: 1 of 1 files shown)


SimonSchick:
Imo, this should be a method rather than yet another global function that pollutes the global scope.

Donkie:
@SimonSchick 
The inconsistency that brings is far worse than polluting the global scope this does.

Kefta:
@simonschick I'm all for making it a method if LerpAngle and LerpVector were changed in the C++ code to reflect this. However, for now, I'd rather match the format we already have.

disseminate:
They should probably be moved to the math library if anything

Kefta:
@disseminate I will copy the format of whatever is done with LerpAngle and LerpVector

Acecool:
Actually, I'd support it in the meta-table so Color:Lerp( ... ) ( with proper arguments to match existing system ) that way the color itself can modify itself within lerp so as long as it isn't defined in the render / draw function, it'd update properly instead of needing additional calls. That's how I wrote Angle:SnapTo( ... ) and a lot of other functions are written that way.

I personally feel that a lot of things are in the math library which could easily exist in a specific meta-table so that it's all modular, ie if Vector didn't exist then the functions wouldn't be there instead of needing to remove them from global, math and vector... But, the issue with that is that many people have written code for the global / math functions already and getting them to rewrite something that "works" isn't easy.

Kefta:
@Acecool I fully support them being methods, so long as the angle and vector lerps are changed in the C++ scope since they are already implemented there as global funcs. I rather not create duplicates or implement any vector/angle functions lua side for efficiency sake.

Kefta:
I really like the idea of having method instead, and although I'd like to have consistency, I won't make duplicate functions.

PR #1030, opened 2015-07-29
Help prevent rigid rope from making crazy physics.

thegrb93:
I can't think of a reason why a rigid rope would need an add length, but I left it a little bit just in case.


thegrb93:
Probably doesn't matter after the new crazy physics stuff

PR #1031, opened 2015-07-29
Make TTT only precache models once

meepdarknessmeep:
This will help a lot with lag on startup

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -483,6 +483,7 @@ function PrepareRound()
    -- Schedule round start
    local ptime = GetConVar("ttt_preptime_seconds"):GetInt()
    if GAMEMODE.FirstRound then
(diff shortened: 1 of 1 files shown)


sirsavary:
+1, will make TTT much better

comedinha:
:+1:

svdm:
Forcing a precache serverside will also tell all clients to precache (perhaps that's the part causing lag). This is actually necessary, because a client that does not have a traitor weapon cached when a traitor buys one can see the precache happening in their console, hence they know it has been bought. It's been a while but if I recall correctly precaches are shown if you set `developer 1`, which is not flagged as a cheat.

This change will reintroduce that problem for clients that connect after the first round. So for this to work an alternate fix should be found first for the problem above.

BlackCetha:
Maybe you could move the check away from `GAMEMODE.FirstRound` and hook it into the client setup on connection?

meepdarknessmeep:
Alright @svdm, I will look for a more convenient and less laggy method for this.

PR #1032, opened 2015-07-29
Add math.IsFinite

meepdarknessmeep:
http://facepunch.com/showthread.php?t=1478507


JamesxX:
While adding these functions as a utility is cool, it doesn't guarantee that they would be used, so we would still see poorly coded addons.

wiox:
Maybe we should just get tonumber patched to work like newer versions of Lua.

meepdarknessmeep:
we could, but that's mostly up to the developers. make a request on garrysmod-requests or garrysmod-issues

thegrb93:
tonumber patch is a very good idea. I can't think of any reason for float exceptions to be parsed.

robotboy655:
tonumber no longer returns nan of inf

meepdarknessmeep:
Nan and inf still exist...

PR #1033, opened 2015-07-30
Make LANG.ProcessMsg send the same way RecvMsg Receives

meepdarknessmeep:
This way it can't get confused


meepdarknessmeep:
@robotboy655 ttt tag pls

PR #1037, opened 2015-08-13
Made camera correctly track different player sizes

DaaOtt:


Kefta:
:+1: Tested with CS:S and default hull sizes; works perfectly.

DaaOtt:
@robotboy655

PR #1043, opened 2015-08-16
Unrestrict thin ropes.

thegrb93:
Will ignore ropes with 0 width, but should allow ropes with width < 1.

### DIFF
--- a/garrysmod/lua/includes/modules/constraint.lua
+++ b/garrysmod/lua/includes/modules/constraint.lua
@@ -327,7 +327,7 @@ function CreateKeyframeRope( Pos, width, material, Constraint, Ent1, LPos1, Bone
 	if ( width <= 0 ) then return nil end
 	
 	-- Clamp the rope to a sensible width
-	width = math.Clamp( width, 1, 100 )
+	width = math.Clamp( width, 0, 100 )
 
 	local rope = ents.Create( "keyframe_rope" )
 	rope:SetPos( Pos )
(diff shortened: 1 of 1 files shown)


thegrb93:
Bump

Svix:
You don't want to do this -- Ropes can cause tremendous framerate loss when set to ludicrously low values like 0.00000000000000001.

0.01 would be a pleasant sanitycheck methinks.

thegrb93:
Why? I thought ropes were just rendered using beams. Is this a problem with beams too?

Svix:
I've always chalked it up to something to do with Source disliking extremely tiny values generally. I can't tell you WHY it harms framerates -- only that when dealing with sophistocated contraptions ropes with exceedingly tiny width tend to decimate framerates.

thegrb93:
I can't reproduce what you're talking about.

Svix:
Apologies for the consistently long delays in my replies -- Life happened.

I experimented further and discovered my memory of rope behavior was wrong -- Moreso ropes trash framerates when they stretch accross the map and are not 0 width. Something to do with shadows -- The way to reproduce this is to go on a map with many brush details that protrude upwards and purposefully catch the visible rope on the map geometry repeatedly until it's very excessively stretched accross the map. Not sure if the framerate loss can be fixed -- But at least we have rope_collide 0.

thegrb93:
Still can't use my thin roped contraptions elsewhere due to this

robotboy655:
I feel like 1 is already thin enough, maybe we could go down to 0.5 or 0.4 but not lower than that for sure.

thegrb93:
Just because you feel like it or is there some actual reason? Perhaps a convar would be better.

thegrb93:
Feel free to change it however you think is good.

robotboy655:
If I remember correctly too thin of a width would cause the rope to just not render in some cases, but I can't reproduce it anymore.

Going to settle for 0.2 as a minimum I guess if 1 is too thick for you. I don't want to allow for infinitely small ropes in multiplayer just in case people can somehow abuse that like shown above making it harder for server admins to detect the offending ropes.

PR #1046, opened 2015-08-19
Update weapon_tttbasegrenade.lua

MrUnknownGamer:
Grenade update :P

Changed some velocity and aim things.


iSnipeu:
you should mention what this fixes or improves on, not everyone can tell from looking at the code

Kefta:
All he did was copy the SDK grenade throw code, which just is different than the other throwing method. Not really better or worse.

svdm:
This is a revert of this PR'd commit: https://github.com/garrynewman/garrysmod/commit/f5c60ce37c7abbfd9ddd410d58477b35a19f5cc9
I don't necessarily have a problem with going back to the original SDK-based version, but is there any reason to do so?

PR #1047, opened 2015-08-23
Indentation fix

xaviergmail:
This is a change that most of everybody has been waiting for.
======

###### It replaces all of the 3-space indentation with single tab characters.

The original conversion was done with a shell script. I then verified each file manually while making sure that anonymous functions as well as multi-line argument / operator chaining still line up nicely.

I also changed every function call that had spaces inside the parenthesis to match the more dominant calling style throughout the gamemode. `print( arg1, arg2 ) -> print(arg1, arg2)`

The few occurrences of the C-Style comparison operators (||, &&) have been replaced with their Lua counterparts.

To observers:
------
Please review my code and test it to make sure everything works fine. If you think anything needs tweaking, please create a pull request on my fork, and I will update this PR (fix-tabs branch): https://github.com/xaviergmail/garrysmod/tree/fix-tabs

To project maintainers:
------
Please take this pull request into consideration. A consistent coding style is required, not only for readability, but it also encourages the community to contribute to the project.


zerfgog:
this was already made and declined like last week

xaviergmail:
Well, shit. It was fun while it lasted. I'll still keep this up though.

zerfgog:
it's ok i still support you and this

HDG-MrDeeJay:
+1

BlackCetha:
:+1:

Exho1:
It would be really cool to have because editing anything in TTT's code turns messy fast with the inconsistancy between tabs and spaces. I use tabs and everything else in Gmod uses tabs too. But Bad King Ugrain said no

Bo98:
You really shouldn't have to be messing with TTT code outside pull requests anyway so you shouldn't need to be stressing over TTT's indentation, no? Hooks are what you should be using. If there's any missing, send a pull request for it and your never have to touch the code again.

I don't really get the whole "I use tabs so everyone else should" mindset that comes across when reading some comments. Yeah 3 spaces may be awkward but that's the life of contributing to another person's work. I can't really say for NP++ but Sublime handles his for me so it's not something I've ever had to worry about in the odd day I edited the code for a pull request.

I disagree that it needs to be consistent with GMod. They are almost completely separate. The only thing linked is that it is shipped together.

The only somewhat valid reason I see is that it discourages pull requests. The author really shouldn't have to adjust his coding style to cater for others though. If you really can't handle the indentation and your editor can't either then just open a thread or something in the TTT forums or wherever the best place to discuss TTT/report bugs is.

Anyway, one of the reasons given by Stefan for not changing it now is that it will absolutely destroy `git blame`.

Have I missed something?

HDG-MrDeeJay:
you can configure np++ to insert x amount of spaces when you hit tab

Just look up instructions from the minecraft community, cant remember what language they use but it has a hissy fit every time you use a tab

xaviergmail:
>You really shouldn't have to be messing with TTT code outside pull requests anyway

There are _many_ reasons for doing such a thing. Sure, extending the base gamemode with an addon would be the preferred approach, but the majority's instinct is to change things right from where they see it.


> Anyway, one of the reasons given by Stefan for not changing it now is that it will absolutely destroy `git blame`.

This is possibly the only valid objection I see. Even then, *there are many ways around this* that a collaborator could do to keep the history intact, as I don't have the privileges to do so.


>Just look up instructions from the minecraft community, cant remember what language they use but it has a hissy fit every time you use a tab


One of the reasons for this is that it discourages beginners that don't know what in the world is happening that makes their code not line up and leaving them heavily confused. Oh, by the way, it's YAML you're looking for.

HDG-MrDeeJay:
Ah yes, YAML. That's the one.

xaviergmail:
Thank you for leaving very valuable input before closing this.

robotboy655:
I doubt there's anything there is any valuable input to be given considering that this is his personal preference, the same way me and Garry prefer C syntax over Lua's normal syntax ( because it's shorter imo ) and the same way there are people who only use Lua's normal syntax.

PR #1048, opened 2015-08-26
Fix shitty SWEP crosshairs

UselessGhost:
Override DoDrawCrosshair in weapon_base to provide better looking and more consistent crosshairs in SWEPs


robotboy655:
I did this in C++, but you should add this example to the wiki.

PR #1050, opened 2015-08-31
Added hexadecimal/integer support to Color object

Kefta:
Supports RRGGBB and AARRGGBB hexes. Due to how Lua handles hexadecimals, setting the alpha to 0 through an int will not work, and will default to 255. Also added color to int/hex conversion.

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -4,17 +4,38 @@ COLOR.__index = COLOR
 --[[---------------------------------------------------------
 	Register our metatable to make it accessible using FindMetaTable
 -----------------------------------------------------------]]
-
 debug.getregistry().Color = COLOR
 
 --[[---------------------------------------------------------
-	To easily create a colour table
+	Creates color object
 -----------------------------------------------------------]]
 function Color( r, g, b, a )
 
-	a = a or 255
-	return setmetatable( { r = math.min( tonumber(r), 255 ), g =  math.min( tonumber(g), 255 ), b =  math.min( tonumber(b), 255 ), a =  math.min( tonumber(a), 255 ) }, COLOR )
-	
+	r = tonumber( r )
+
+	if ( g != nil ) then
+		a = tonumber( a ) or 255
+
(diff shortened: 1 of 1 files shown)


zerfgog:
can you benchmark the speed of Color(255, 255, 255, 255) in this patch vs the speed as it currently is please?

wiox:
I think using a string rather than a number might be a better solution. I've a couple of reasons for that.

1) The output of COLOR:ToInt() can be misleading. This is because the bit library treats all values as 32-bit signed integers. The value would still work when passed between Color() and COLOR:ToInt() but anything else expecting a signed value is going to fail.
```lua
bit.rshift( 0xFF000000, 24 ) == 255
bit.lshift( 255, 24 ) == -0x1000000
bit.rshift( -0x1000000, 24 ) == 255
```

2) The whole 'alpha shit doesn't work due to how lua handles hexadecimals' argument is silly because hexadecimals are just handled as numbers. It's not a good way of explaining how it works to new users & just adds complications. Strings can fix this because you can just check their length for if an alpha channel is present.
```lua
"0x00FFFFFF" != 0x00FFFFFF -- Nicer to work with
 0x00FFFFFF  == 0xFFFFFF   -- Super irritating
```

Kefta:
@wiox For number 2, I can accept strings and grab the third and fourth characters to set the alpha, then sub them out and convert the string to a number. Will that implementation satisfy you, or should I grab all arguments from the string seperatrly?

I honestly don't understand what you are saying with ToInt since it will always return a number between 0 and 4294967295 (0xFFFFFFFF). There's no misleading info that says it will return a 32-bit signed int; whatever is expecting that should check if the int is greater than the cap on its own.

Kefta:
@zerf I will do one after all changes have been made

wiox:
> bit.lshift( 255, 24 ) + bit.lshift( 0, 16 ) + bit.lshift( 0, 8 ) + bit.lshift( 0, 0 )
    = -16777216

As I said, the output of all big.* functions is a signed integer.

Kefta:
@wiox Should I just drop the alpha in the int conversion and rename the function to be more accurate?

wiox:
It's easy enough to get a logical left-shift with simple math. It'd be nice if you could achieve an RGBA encoding instead. That's how the parameters to Color() are currently laid out.

```lua
a * 2^24 + r * 2^16 + g * 2^8 + b

Kefta:
Refer to #1088

PR #1052, opened 2015-09-03
Override duplicate names in AddAmmoType

Kefta:
No more unpredictability of how the engine will choose which ammo table to use when BuildAmmoTypes is called. Now, only the last one will be sent


Kefta:
It would also be nice if it was possible to override default ammo types

UselessGhost:
This would only use the first ammo type registered, not the last. You're not setting the value in the table, you're just setting the "ammo" variable in the loop scope.
What you probably want is something like this:
```
	for id, ammo in ipairs( AmmoTypes ) do
		if ( tbl.name == ammo.name ) then
			AmmoTypes[ id ] = tbl
			return
		end
	end
```

Kefta:
@UselessGhost Ah, thought Lua loops behaved like foreach where the value had the same address as the table value. Thanks!

PR #1053, opened 2015-09-03
TTT: Recheck collision rules after group reset

Kefta:
After an immense amount of testing and reading source code, it seems that CollisionRulesChanged is already called internally by SetCollisionGroup, and then RecheckCollisionFilter is called on all of the children PhysObjects subsequently. Initially, calling CollisionRulesChanged immidiately after setting the collision group had no effect as expected, however, calling it in a timer seems to have settled any initial spasms.

Note that this is not a sure fix to all ragdoll related crashes; COLLISION_GROUP_DEBRIS_TRIGGER still seems to have some serious issues sending back collision data at high velocities/high collision rates, but this fix will at least cause less crashes to occur early on.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
@@ -376,6 +376,7 @@ function CORPSE.Create(ply, attacker, dmginfo)
 
    -- nonsolid to players, but can be picked up and shot
    rag:SetCollisionGroup(rag_collide:GetBool() and COLLISION_GROUP_WEAPON or COLLISION_GROUP_DEBRIS_TRIGGER)
+   timer.Simple( 1, rag.CollisionRulesChanged )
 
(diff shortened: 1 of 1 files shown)


svdm:
Weird, but hey it's Source. Worth a shot.

robotboy655:
@Kefta have you even tested that timer? To me it appears that it would error as you do not pass the entity object to the function call.

Kefta:
@robotboy655 I tested it with the timer having a seperate function ( function() rag:CollisionRulesChanged() end ), but I thought this would shorten it. I'll test it tonight

zerfgog:
lol

wiox:
nice quality control

robotboy655:
Nice commits by @Kefta as per usual.

zerfgog:
:zap:

Kefta:
I should probably not make last minute changes and assume they work anymore

PR #1061, opened 2015-09-20
Clamped physgun color to above 0

DaaOtt:
Fixes an exploit where beams would become invisible (minges use this). 
Alternative for legitimate uses:

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/player_class/player_sandbox.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/player_class/player_sandbox.lua
@@ -105,8 +105,11 @@ function PLAYER:Spawn()
 	local col = self.Player:GetInfo( "cl_playercolor" )
 	self.Player:SetPlayerColor( Vector( col ) )
 
-	local col = self.Player:GetInfo( "cl_weaponcolor" )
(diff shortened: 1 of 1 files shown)


Velkon:
no

Acecool:
One note, you shouldn't need self.Player just like you don't need self.Entity in the Player ( or entity ) meta-tables. Those exceptions are actually hard-coded into the meta-table __index / __newindex functions because so many people still use them, but new code should avoid using those deprecated calls.

There should probably also be an admin exception ( I personally set physgun color to black, ie invisible, for admins that are invisible to prevent an orb from appearing ) or a convar to control if admins or super-admins override that behavior.

wiox:
> One note, you shouldn't need self.Player just like you don't need self.Entity in the Player ( or entity ) meta-tables. Those exceptions are actually hard-coded into the meta-table __index / __newindex functions because so many people still use them, but new code should avoid using those deprecated calls.

This is how player classes work.

DaaOtt:
>There should probably also be an admin exception ( I personally set physgun color to black, ie invisible, for admins that are invisible to prevent an orb from appearing ) or a convar to control if admins or super-admins override that behavior.

This only affects the convar, not SetWeaponColor.

PR #1062, opened 2015-09-20
Fix game cleanup removing persistent entities

wyozi:


wyozi:
Map is cleaned up -> persistent entities disappear -> map is changed -> persist.txt is saved without entities that rightfully belong there.

robotboy655:
Sounds like a shitty fix to me, if you add a persistent prop after the props were loaded > cleanup > everything you added is lost and only the stuff that was there from the beginning is there.

wyozi:
Valid point. Ideally game.CleanUpMap would not touch persistent entities at all. Do you think that would be doable on C++ side?

PR #1064, opened 2015-09-22
Call OnRowSelected when shift-selecting

Weol:
The current version does not call OnRowSelected when shift-selecting items. This change will fix that.


SpencerSharkey:
+1

robotboy655:
This PR has the following problems:
1) It will continue to fire OnRowSelected every frame as long as you move your cursor on a selected row
2) It will not fire OnRowSelected for in between rows like multi-select would without pressing shift

That being said, the issue is fixed:
https://github.com/garrynewman/garrysmod/commit/b8f42314b7978ee29629e167a0215910f89dbe0c

PR #1065, opened 2015-09-27
math.Clamp optional values

BillyOnWiiUIn144p:
Now, doing math.Clamp(-1,0) returns 0 and therefore doesn't need a max value. This means the number can then be infinite.
Forgive me if the code is a bit messy or needs changes, I have never used Github before :)

math.Clamp(-1,0) = 0
math.Clamp(0,0,100) = 0

math.Clamp(0,nil,100) = 100
math.Clamp(101,nil,100) = 101

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -40,23 +40,11 @@ end
 
 --[[---------------------------------------------------------
    Name: Clamp( in, low, high )
-   Desc: Clamp value between 2 optional values
+   Desc: Clamp value between 2 values
 ------------------------------------------------------------]]
 function math.Clamp( _in, low, high )
-	if ( low ~= nil and high ~= nil ) then
-		if (_in < low) then return low end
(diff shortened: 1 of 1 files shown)


Bo98:
You screwed up the pull request by reverting it on a separate branch and starting the pull request on that branch.

Anyway, I don't understand the usefulness in this. Why would you want to use this functionality over `math.min` and `math.max`?

BillyOnWiiUIn144p:
Thanks, this is getting confusing :/, but I'm not really sure what to do about that at the moment. Copy + paste over the code if needed.

Anyway, what's the point in math.Clamp in the first place if math.min and math.max exist? Just to make things easier. It's just added functionality to the math.Clamp function. But, if you really don't think there's a need, it's fine if this is closed.

Kefta:
Make a new PR

zerfgog:
This is totally unnecessary as has been pointed out. math.Clamp is a utility function for when you need both math.min and math.max in one call. If you only need one, just use one!

PR #1066, opened 2015-09-27
math.Clamp optional values

BillyOnWiiUIn144p:
math.Clamp(-1,0) = 0
math.Clamp(1,0) = 1

math.Clamp(100,nil,99) = 99
math.Clamp(98,nil,99) = 98


zerfgog:
This is totally unnecessary as has been pointed out. math.Clamp is a utility function for when you need both math.min and math.max in one call. If you only need one, just use one!

robotboy655:
Couldn't agree more with @zerfgog

robotboy655:
Not to mention, math.min/max takes as many arguments as you give it, so math.max( 1,2,3,4,5,6 ) will give you 6.

PR #1067, opened 2015-09-28
Fixed: DTileLayout spaceX/Y and border not being used.

GamerGambit:
I noticed that called SetSpaceX, SetSpaceY and SetBorder of DTileLayout doesnt have an affect on the layout of children so I have created a fix for that.


robotboy655:
This is a pretty terrible PR as per usual, your code may position the elements with taking into account X, Y and Border spacing, but you completely ignored things like border on the right and bottom and the system that decides how many icon should go on each row.

Original:
image

Your "fix":
image

Actually properly working code: ( How it should've been since 2012, but wasn't )
image

The green rectangle is the bounds of DTileLayout.

No offense intended, this these half-assed PRs are really getting old and not funny at all.

PR #1068, opened 2015-10-01
Preventing right-click exploit with Hydraulic/Muscle/Slider/Winch tool

Mista-Tea:
There's been a longstanding issue where players are able to right-click using these 4 tools and create a constraint between a prop they own (according to any arbitrary prop protection system) and a prop they don't own, thus allowing them to essentially take control of the constrained props via dupe, remover, etc.

This is because there is no additional CanTool check to see if the constraint can be made between the exploiter's prop and the prop in the 2nd trace. This pull request adds the additional CanTool check.

Before patch:
1. Player right-clicks their prop.
2. Server checks CanTool to see if the player can tool the prop they are looking at and returns true (they own the prop).
3. Another trace is created pointing back toward the player
4. **A constraint is created between the entity in the 1st trace and the entity in the 2nd trace, regardless of owner.**

After patch:
1. Player right-clicks their prop.
2. Server checks CanTool to see if the player can tool the prop they are looking at and returns true (they own the prop).
3. Another trace is created pointing back toward the player
4. **Server checks CanTool to see if the player can tool the prop the 2nd trace found and returns false (they don't own the prop).**

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/hydraulic.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/hydraulic.lua
@@ -124,6 +124,12 @@ function TOOL:RightClick( trace )
 		return
 	end
 
+	-- Check to see if the player can create a hydraulic constraint with the entity in the trace
+	if ( !hook.Run( "CanTool", self:GetOwner(), tr, "hydraulic" ) ) then
+		self:ClearObjects()
+		return
+	end
+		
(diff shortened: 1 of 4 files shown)


thegrb93:
Cool. Never crossed my mind because every pp has a hack to counter this and its such an old bug, but now it looks like the hack can be removed. Good fix.

DaaOtt:
:+1:

robotboy655:
Even though I do not like the solution, I don't see any other way, so EHHHHH

PR #1071, opened 2015-10-04
Fixed TableInherit creating stackoverflow(s)

maritaria:
The TableInherit creates a stack-overflow when trying to copy table values other than those on the BaseClass key.
Now the function is limited to only inherit tables on the BaseClass key.


robotboy655:
How do your reproduce this?
The weapons library has the opposite fix for this.

Kefta:
It should be k ~= BaseClass

maritaria:
It was 5 years ago, no idea if the issue is still relevant...

robotboy655:
https://github.com/Facepunch/garrysmod/commit/1c662826bc93382c0970d1e4efde31d8cc6be9dd

PR #1072, opened 2015-10-04
Halloween dynamite

Python1320:
Give comments on whether to include or not.

Yes: Halloween!
No: Prop occasionally turns black when hovered over with context menu.
No: Can't see the dynamite if you don't own TF2.


Kefta:
Could you do a file.Find check instead so servers could send the model to players?

Python1320:
Halloween has been ruined :ghost:

PotcFdk:
Garry's Mod Halloween 2015, never forget.

robotboy655:
You pretty much gave the reasons why this shouldn't be included. I am mostly worried about the last one.

PR #1073, opened 2015-10-05
Added check if player is alive when testing players

morfin60:
Added check to ignore ghosts when testing(they could trigger false positive result)

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_traitor_check.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_traitor_check.lua
@@ -21,7 +21,7 @@ function ENT:CountTraitors()
 
    local trs = 0
    for _,ply in pairs(player.GetAll()) do
-      if IsValid(ply) and ply:IsActiveTraitor() then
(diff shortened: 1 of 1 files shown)


Bo98:
Doesn't ply:IsActiveTraitor() cover that?

svdm:
Indeed, players are moved to the spectator team as soon as they die, and `IsActiveTraitor` will return false for anyone in that team. Is there a reproducible case where the entity gives a false positive?

BlackCetha:
Because he said "ghosts" ill guess its something with the spec-dm.

Kefta:
Then that's something addon specific and should be handled within the addon

morfin60:
And there is a little problem. Guy i asked to go in tester was not in DM(he was just flying ghost)

BlackCetha:
Give us steps to reproduce the problem so we can see whats going wrong please.
Also, what is a "flying ghost"? Do you mean a dead player in spectator mode?

morfin60:
It was not flying ghost - it was walking ghost with gun in ghost DM which is like in special team

PR #1074, opened 2015-10-06
Add player argument to duplicator.IsAllowed

Weol:
Adding an argument and hook to the duplicator.IsAllowed function will allow developers to have more control over what can be spawned with the duplicator, as well as who can spawn what.

### DIFF
--- a/garrysmod/lua/includes/modules/duplicator.lua
+++ b/garrysmod/lua/includes/modules/duplicator.lua
@@ -725,6 +725,8 @@ function Paste( Player, EntityList, ConstraintList )
 	-- Create the Entities
 	--
 	for k, v in pairs( EntityList ) do
+		
+		if not IsAllowed(v.Class, Player) then continue end
 	
 		local e = nil
 		local b = ProtectedCall( function() e = CreateEntityFromTable( Player, v ) end )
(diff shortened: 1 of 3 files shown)


garrysmod/lua/includes/modules/duplicator.lua
```diff
@@ -303,8 +303,18 @@ end
...
 -- Returns true if we can copy/paste this entity
 --
-function IsAllowed( classname )
+function IsAllowed( classname, Player )
 
+	if not Player then
+		return DuplicateAllowed[ classname ]
+	end
+
+	local candupe = hook.Call( "PlayerDuplicateEntity", GAMEMODE, Player, classname )
```
Bo98: PlayerCanDuplicateEntity would be a better name IMO for consistency reasons.

Weol:
Hmm, possibly. My intention was to make it simelar to the "PlayerSpawnXXX" hooks.

Bo98:
I guess so. Hook naming is very inconsistent in GMod.

GitSparTV:
This is essentialy made with CanTool. If you can't tool someone's construction, theoretically, you won't be allowed to dupe it.

thegrb93:
No, this is a dead PR. PlayerSpawnEntity/PlayerSpawnSENT hook handles this.

PR #1077, opened 2015-10-08
Removed old ignore patterns

Kefta:
styleguide and devtest maps were removed from default HL2 packaging; reduced all background maps to one pattern.

### DIFF
--- a/garrysmod/lua/menu/getmaps.lua
+++ b/garrysmod/lua/menu/getmaps.lua
@@ -194,11 +194,7 @@ local function LoadFavourites()
 end
 
 local IgnorePatterns = {
-	"^background",
-	"^devtest",
-	"^ep1_background",
-	"^ep2_background",
-	"^styleguide",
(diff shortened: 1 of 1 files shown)


robotboy655:
I don't think there's any point in changed the background map patters, you will just create more collisions with legit maps that might use that word in their name

Kefta:
@robotboy655 Normally I would agree, but Blade Symphony also added a background map that should be excluded; and in my 40GB folder of maps, I haven't run into one conflict.

zerfgog:
@Kefta then just add that Blade Symphony map?

Kefta:
@zerf But are that many background patterns needed?

PotcFdk:
@Kefta why does it matter? I agree with @zerfgog and vote against possible collisions.

Kefta:
Ran into a collision today with a map called something like gm_greenbackground; fair point.

PR #1078, opened 2015-10-11
Add GetLastMap function to getmaps.lua

zerfgog:


Kefta:
?

zerfgog:
this isn't really useful except for my single use case of reusing the default getmaps.lua with a custom menu, but if you're making a custom menu you should be fine with editing getmaps.lua

PR #1079, opened 2015-10-11
Fixed drive system inheritance

maritaria:
Previously; if you registered a drive that was derived from a base that wasn't registered yet, then that drive wouldn't inherit its methods from the base (because Type[base] is nil at that point).


maritaria:
Thanks

PR #1080, opened 2015-10-12
net.WriteTable and net.ReadTable improvements

meepdarknessmeep:
After all the drama of the last pr, I thought I'd rewrite it to make it more clean and add some tests to my local machine.

Here's my test:
http://hastebin.com/raw/gajodakuna

All these pass should pass without any errors.

Some profiling:

```recursive1	true
recursive2	true
2=2	true
0
1	=	a
2	=	b
3	=	true
4	=	3.14
cc	=	cc
t:
		1	=	a
		2	=	b
		3	=	true
		4	=	3.14
		cc	=	cc
		t	=	table: 0x34a42b48
23.375


1
1:
		1	=	a
		2	=	b
		3	=	true
		4	=	3.14
		cc	=	cc
		t	=	table: 0x17fc64b8
24.875


2
2	=	2
3	=	3
4.14141414	=	4.14141414
16.875
```

So the lengths of the three table sent in bytes + the 4 bit "which" integer is (in order)
23.375, 24.875, 16.875

The old output is:
```
recursive1	false	stack overflow
recursive2	false	stack overflow
2=2	true
2
2	=	2
3	=	3
4.14141414	=	4.14141414
55.5


```
The first two tests fail because of recursion, and the last test is over 3x larger when sent.


zerfgog:
:+1:

thegrb93:
You should mention you're also optimizing net.ReadEntity and net.WriteEntity or put it in a different pr

meepdarknessmeep:
@willox @robotboy655 @UnderscoreKilburn is there anything in this pr that i could improve or change to make it up to standards and get pulled?

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -181,3 +146,531 @@ function net.ReadType( typeid )
...
+	-- Normal gmod types we can't really improve
+	--
+	Color       = net.ReadColor,
+	boolean     = net.ReadBool,
+	number      = net.ReadDouble,
+	bit         = net.ReadBit,
+	Entity      = net.ReadEntity,
+    VMatrix     = net.ReadMatrix,
+    Angle       = net.ReadAngle,
+    Vector      = net.ReadVector,
```
Bo98: Inconsistent indentation here btw.

meepdarknessmeep: whoops, somehow it changed. Thanks for notifying me.

Bo98:
In the places you corrected you missed the double/triple/etc. indentations. For example line 325 is one tab followed by four spaces.

meepdarknessmeep:
Is it fine now?

Bo98:
LGTM

MrUnknownGamer:
Sooooo...?

GitSparTV:
Optimizing WriteTable doesn't have sense for me. You can always find the structure of your data and send it with more optimized for you way.
Also you have some flaws in concatenation, it's better to use a table and then table.concat (talking about string7 or string6)

thegrb93:
Some data structures can be completely dynamic so that assumption is wrong.

GitSparTV:
@thegrb93 As would @Kefta say: provide an example of dynamic table that can't be structualized

thegrb93:
Ones with cyclic references. Sure you can number the table references yourself and make it net friendly, but that's what writeTable should do. (Though I'm pretty sure gmod's vanilla writeTable doesn't).

GitSparTV:
Even vanilla writetable won't optimize cyclic references

thegrb93:
Well this enhancement does.

JacksonTravisRipper:
Why isn't this pulled lmfao.

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -173,11 +139,538 @@ net.ReadVars =
...
+		local compressed = util.Compress( x )
+
+		if ( compressed and compressed:len() < x:len() + 1 ) then
+			net.WriteBool( true )
+			writing.uintv( compressed:len() )
+			net.WriteData( compressed, compressed:len() )
+		else
+			net.WriteBool( false )
+			net.WriteString( x:len() )
+		end
```
JLVerified, line 614: When sending, if a string cannot be compressed down with `util.Compress`, then it will write the length of the string instead of the uncompressed string. This can happen with strings that contain multi-byte characters. 

For example: `á` will be received as `2`

```suggestion
		if ( compressed and compressed:len() < x:len() + 1 ) then
			net.WriteBool( true )
			writing.uintv( compressed:len() )
			net.WriteData( compressed, compressed:len() )
		else
			net.WriteBool( false )
			net.WriteString( x )
		end
```

thegrb93: This doesn't sound right

JLVerified: Try it yourself by networking this table:
```
local strtbl = {
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaa!",
    "abcdefghijklmnopqrstuvwxyzaaa!",
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaz",
    "abcdefghijklmnopqrstuvwxyzaaaa",
    "aaaaaaaaaaaaaaaaaaaaaaaaaaaaa∑",
    "abcdefghijklmnopqrstuvwxyaaaa∑",
}
```

Currently:
image

After the suggested change:
image

thegrb93: Oh, I thought you meant util.Compress was returning the length if it failed to compress. I see the bug in his source now.

jleave: Ah my bad there. I wasn't very clear.

GitSparTV:
Quoting code_gs:
> I don't really have any interest in improving a function that should just be removed

Kefta:
That's why it won't be removed. Doesn't mean it can't

WilliamVenner:
Irrelevant

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -173,11 +139,538 @@ net.ReadVars =
...
+
+	--
+	-- C String
+	-- NULL terminated
+	-- NOTE: Must be NULL terminated or else will break compatibility
+	-- with some addons! Also could lead to exploits
+	--
+	string = function()
+		if ( net.ReadBool() ) then -- compressed or not
+			return util.Decompress( net.ReadData( reading.uintv() ) )
```
2048khz-gachi-rmx, line 444: add a second argument; you can send zip bombs otherwise

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -173,11 +139,538 @@ net.ReadVars =
...
+
+--
+-- We need this since #table returns undefined values
+-- by the lua spec if it doesn't have incremental keys
+-- we use pairs since it's backwards compatible
+--
+local function array_len(x)
+	local indices = {};
+    for k,v in pairs(x) do
+        indices[k] = true;
```
2048khz-gachi-rmx: is there a reason to do this rather than look into `x` directly?

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -173,11 +139,538 @@ net.ReadVars =
...
+end
+
+local ReferenceType = StringToType( "reference" )
+local TableType     = StringToType( "table" )
+
+reading = {
+	--
+	-- Normal gmod types we can't really improve
+	--
+	Color       = net.ReadColor,
```
2048khz-gachi-rmx, line 363: this invalidates detours people might rather keep (profiling, peeking, etc.)

robotboy655:
I am not inclined to merge this monstrosity, even if it does make net.Write/ReadTable use less network bandwidth.

meepen:
It's not just about bandwidth, it's also about cyclic references. I'm also not sure saying it's a monstrosity is constructive feedback. If the reason is that WriteTable shouldn't exist, fine, but closing with this reason doesn't really make sense to me when you can just provide feedback.

PR #1084, opened 2015-10-14
Make DMenu:AddSpacer use the SkinHook PaintMenuSpacer

Shendow:
Previously it was hardcoded.


robotboy655:
Not sure how good of an idea this is considering you use a method that exists on other skins and is defined differently by default, but I guess we shall see.

PR #1087, opened 2015-10-23
Disable IK for DModelPanel

Azarym:
So the model isn't fucked if the origin of the map is near plane.


Kefta:
:1: Had no idea that was causing the problem

PR #1088, opened 2015-10-24
Added hex/int support to Color constructor

Kefta:
Really hope this is the last one.

Also cleaned up spacing, typos, and syntax all over the color file. I included a local function for bitwise AND since bit.band seems to be loosing precision on odd numbers by one.

This adds clamping to RBGA color construction as well so that negative values can no longer be set.

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -78,8 +78,7 @@ end
 -----------------------------------------------------------]]
 function ColorAlpha( c, a )
 
-	c.a = a
-	return c
+	return Color( c.r, c.g, c.b, a )
 
 end
 
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -2,19 +2,75 @@ local COLOR = {}
 COLOR.__index = COLOR
 
 --[[---------------------------------------------------------
-	Register our metatable to make it accessible using FindMetaTable
+	Register our metatable
 -----------------------------------------------------------]]
-
 debug.getregistry().Color = COLOR
 
 --[[---------------------------------------------------------
-	To easily create a colour table
+	Local bitwise AND function to get precise values
+-----------------------------------------------------------]]
+local function bAND( x, y )
+
+    local b, n, rx, ry = 1, 0
+
+    while x > 0 and y > 0 do
+        rx, ry = x % 2, y % 2
+        if ( rx + ry > 1 ) then n = n + b end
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/util/color.lua
```diff
@@ -2,19 +2,75 @@ local COLOR = {}
...
+	Register our metatable
 -----------------------------------------------------------]]
-
 debug.getregistry().Color = COLOR
 
 --[[---------------------------------------------------------
-	To easily create a colour table
+	Local bitwise AND function to get precise values
+-----------------------------------------------------------]]
+local function bAND( x, y )
```
robotboy655: What's wrong with bit.band?

garrysmod/lua/includes/util/color.lua
```diff
@@ -22,51 +78,51 @@ end
 -----------------------------------------------------------]]
 function ColorAlpha( c, a )
 
-	return Color( c.r, c.g, c.b, a )
-	
+	c.a = a
```
Bo98: This change could possibly break some scripts.

garrysmod/lua/includes/util/color.lua
Kefta: Slight precision loss:

http://puu.sh/kVGd2.png
http://puu.sh/kVGel.png

garrysmod/lua/includes/util/color.lua
Kefta: How so? There's no need to create an entire new object:  this function is completely useless anyway.

willox: :palm_tree: 

https://github.com/garrynewman/garrysmod/blob/3e138636eb1b0ad6ed785dedf350020755cff5f1/garrysmod/lua/vgui/dcolormixer.lua#L35

Bo98: Here's an example in GMod itself:

https://github.com/garrynewman/garrysmod/blob/3e138636eb1b0ad6ed785dedf350020755cff5f1/garrysmod/lua/entities/widget_base.lua#L205-L208

The later call will be transparent with this change while before it would not be.

Kefta:
Talked with Willox; going to add some more things tomorrow to add more compatibility

garrysmod/lua/includes/util/color.lua
immibis: Looks like that's because bit.band rounds its first argument to the nearest integer, while you were expecting it to round down (floor it).

Kefta: @immibis Already figured it out; check the new PR

PR #1089, opened 2015-10-25
Color now supports ints and color strings

Kefta:
NOW this is the last one.

- Integer support
- String support in form of a "#hex", "number", or "HTML color name" -- this includes the 140 color names supported by CSS and HTML as well as "transparent"
- Clamp support to 0 on RGBA entry
- Whitespacing/syntax fixes across the file


willox:
It's worth saying that this is actually a pretty decent feature. The code isn't extremely pretty, but aside from that:

It makes sense: Color( myColor:ToInt() ) results in the expected output

Nobody loses: It ain't gonna break anything already created, unless it does (pls check).

It just makes development easier: Am I taking colors from a css file? I can just type them out without converting them. Do I just want to make something red for debugging? I just type red.

All of that said, there are two major issues. The hex triplet support is completely wrong and it is __not possible to do what you need with a simple string->int conversion, before you try__.

__1)__ There's two kinds of hex triplets you should probably accommodate. One is the common six-hexadecimal-digit form: #RRGGBB. The other is the three-hexadecimal-digit form (shorthand hex triplet): #RGB.

The sad truth is that a hex triplet is not just a hexadecimal number. You should probably go and parse it or at least restrict it to only valid inputs some how.

    #AABBCC == #ABC

__2)__ Hex triplets don't include an alpha byte! They should result in a color of { a = 255 } and maybe MAYBE allow a second parameter to Color() for a transparency (either 0-1 or 0-255, probably the latter).


````lua
assert( Color'#F00'      ======= Color( Color'#F00':ToInt() ) )
assert( Color'#FF0000'.r ======= Color'#F00' )
assert( Color'#FF0000'   ======= Color( 255, 0, 0, 255 ) )
```

Kefta:
@willox Fixed. The code is very dirty, however.

zerfgog:
@Kefta because 0xABC ~= 0xAABBCC, it's shorthand notation and not a "hexadecimal number"

Kefta:
@zerfgog I thought he meant that #AABBCC was not equal to 0xAABBCC. Anyway, #ABC will now automatically translate into 0xAABBCC

willox:
You just removed the alpha of every predefined color?

zerfgog:
:+1: code_gs finally pulled through

SimonSchick:
Why exactly do you need to throw all of this into the *Color* function?
Lua is pretty bad when it comes to overloading.

I recommend turning the *Color* function into a callable table and add static methods for each individual construction case, this makes everything much easier.

Kefta:
@SimonSchick That will break any code that looks up the Color metatable, and it will be useless for it to be a metatable at that point since it will just be a global. The code is cluttered and has a lot of cases, but this is the result of a language where there's little polymorphism and predicted types in terms of arguments.

Kefta:
Also, thank you for your constant critiques @willox though the seven pull requests and three different people all trying to achieve the same result. This made me learn a lot of bitwise math, and I'm very happy with the result :tada:

SimonSchick:
@Kefta I never said anything about removing the metatable, what I was suggesting was something like this:
```
local Color = setmetatable( {}, {
	__call = function(self, r, g, b, a)
		return setmetatable({
			r = math.min( tonumber(r), 255 ),
			g = math.min( tonumber(g), 255 ),
			b = math.min( tonumber(b), 255 ),
			a = math.min( tonumber(a or 255), 255 )
		}, COLOR )
	end,
	FromInt = function(r)
		return Color(
			bit.band( bit.rshift( r, 24 ), 0xFF ),
			bit.band( bit.rshift( r, 16 ), 0xFF ),
			bit.band( bit.rshift( r, 8 ), 0xFF ),
			bit.band( r, 0xFF )
		)
	end,
	FromCSSString = function(str)
		local temp = tonumber(str)
		if temp then
			return temp
		elseif Color.list[ str:lower() ] then
			return Color.list[ str:lower() ]
		elseif str:sub( 1, 1 ) == "#" then
			if str:len() == 7 then
				return Color.FromInt(bit.bor(tonumber(str:sub(2), 16), 0xFF000000)
			elseif str:len() == 4 then
				return Color(
					tonumber(str:sub(2, 2), 16) * 16,
					tonumber(str:sub(3, 3), 16) * 16,
					tonumber(str:sub(4, 4), 16) * 16,
				)
			end
		end
		return Color.FromInt(numVal)
	end,
	FromVector = function(vec)
		return Color(vec.x * 255, vec.y * 255, vec.z * 255)
	end,
	--etc
	list = {
		--humongous list of html color
	}
} )
```
_G.Color = Color

willox:
I like the idea of...

```lua
Color "#FF0000"
```

... being valid code, but can't deny that @SimonSchick's method is going to be easier to document on the wiki.

Maybe a better method name than FromCSSString can be found.

Kefta:
@willox Although it might be easier for documentation, I think that having a single Color constructor is going to be a lot simpler and match other class constructors, like Vector and Angle. At the very least, I can create some local functions for each data type to make the code more readable.

SomeSortOfDuck:
@Kefta, the hex code is a clusterfuck.

There is a super secret feature of tonumber that allows you to select the base to transform the value to.
```lua
print(0x7D7D7D == tonumber("7D7D7D", 16))
> true
```
Just sub everything after the # and tonumber it with base 16.

GGG-KILLER:
[**@SomeSortOfDuck**](https://github.com/SomeSortOfDuck) commented on [Jan 4, 2016, 6:37 AM GMT-2](https://github.com/garrynewman/garrysmod/pull/1089#issuecomment-168611661 "2016-01-04T08:37:59Z - Replied by Github Comment Enhancer"):
> [@Kefta](https://github.com/Kefta), the hex code is a clusterfuck.
> 
> There is a super secret feature of tonumber that allows you to select the base to transform the value to.
> 
> <div class="highlight highlight-source-lua"><pre><span class="pl-c1">print</span>(<span class="pl-c1">0</span>x7D7D7D <span class="pl-k">==</span> <span class="pl-c1">tonumber</span>(<span class="pl-s"><span class="pl-pds">"</span>7D7D7D<span class="pl-pds">"</span></span>, <span class="pl-c1">16</span>))
> <span class="pl-k">&gt;</span> <span class="pl-c1">true</span></pre></div>
> 
> Just sub everything after the # and tonumber it with base 16.

Not really a "super secret feature" as it's documented in the wiki, maybe just infrequently used.

SomeSortOfDuck:
@GGG-KILLER what is joke.

GGG-KILLER:
oh, ok..... I need sleep....... sorry

Kefta:
Yeah, I had no idea it existed

zerfgog:
@Kefta Dumb

Kefta:
:8ball:

SomeSortOfDuck:
I would't put it on anyone to not know. 99% of the time you never have to use it. :+1: for this commit, I really want these changes.

zerfgog:
Nah, this sucks

aStonedPenguin:
Why not just have a whole color library? color.FromRGB(), color.FromHex(), color.FromInt() ect

SomeSortOfDuck:
Why not just have a one-size-fits-all solution?

SimonSchick:
The Color function has way too many functions already.
Since you cannot do proper overloading without lots of type checks, it is recommendable to create "static" functions instead to keep the code clean.

The main problem is that a metric ton of people literally SPAM new colors hundreds of times every frame (because no-one bothered to tell them to cache their colors or pure lazyness), and adding more and more mostly unneeded functionality is just making the performance worse.

SomeSortOfDuck:
>The Color function has way too many functions already.
Since you cannot do proper overloading without lots of type checks, it is recommendable to create "static" functions instead to keep the code clean.

Arguably the worst part of the code is the hex feature, which I've provided an easy solution for already. The code doesn't look too bad otherwise. Maybe a few comments should be added.

>The main problem is that a metric ton of people literally SPAM new colors hundreds of times every frame (because no-one bothered to tell them to cache their colors or pure lazyness), and adding more and more mostly unneeded functionality is just making the performance worse.

So we're going to base development off of poor programming practices now?

aStonedPenguin:
> So we're going to base development off of poor programming practices now?

One could argue doing a bunch of type checks in a function like that is poor practice too.

Kefta:
Wouldn't be an issue if Color was in C++. But anyway, I'll fix the PR by tomorrow

ExtReMLapin:
Any performance change ?

I mean, most of the cheap-ass coders regenerate like 300 colors on each frame, even if the color don't change.

zerfgog:
@ExtReMLapin This is a GitHub PR, bumps aren't a thing.
Yes, this change introduces performance changes, especially on hex string colors (all those metatable lookups and operations themselves), making it a bad idea to call every frame unless you're using rgba or numeric colors.
It might be worth it to run SysTime benchmarks.

ExtReMLapin:
Maybe it would be a good idea to cache a color table once it's created, it would improve performances on current script (the ones recreating the color table on each frames).

Kefta:
I'll run some

Kefta:
@SimonSchick Can you explain your method further? After reading more into how GMod handles polymorphism, I see that my method for a single global constructor is pretty inconsistent

Kefta:
I swear I'll get it right one day

Bo98:
> NOW this is the last one.

 - @Kefta

Jvs34:
5pulls

PR #1090, opened 2015-10-25
Add color metamethods

Kefta:
Scalers allowed on multiplication and division; distributive on modulo and power


Kefta:
Shit this is really broken code

PR #1092, opened 2015-11-05
New function: string.Contains( str, val )

Mysterypancake1:
Just a very simple function inspired by this thread:
https://facepunch.com/showthread.php?t=1009343

Since Garry has incorporated most C# operators like != and ||, I just thought 'why not add a C#-like function as well?'


thegrb93:
Because it's the same as string.find

Kefta:
Too simple of an alias to be necessary.

PR #1093, opened 2015-11-07
Updated comment

dvdvideo1234:
This comment kinda confused me :)


zerfgog:
this is actually extremely prevalent in the GMod source. I recommend using the official garry's mod wiki as opposed to reading the source, as the majority of the functions in the source are not documented or are documented incorrectly. Additionally, the GMod wiki contains documentation for C++ functions, not just Lua ones.

PR #1095, opened 2015-11-19
Added util.URLDecode and util.URLEncode

BillyOnWiiUIn144p:
Thought these might be useful for sending data in a URL that includes "URL affecting" characters such as /, ?, &, = etc. that needs escaping or summin

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -347,3 +347,27 @@ function util.RemovePData( steamid, name )
 	sql.Query( "DELETE FROM playerpdata WHERE infoid = "..SQLStr(name) )
 	
 end
+
+--[[---------------------------------------------------------
+   Name: URLEncode( str )
+   Desc: Encodes a URL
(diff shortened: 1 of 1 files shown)


SimonSchick:
Might be worth implementing something like this too: https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options

thegrb93:
Spaces should be %20, not + also the carriage return is unnecessary.

BillyOnWiiUIn144p:
%20 and + are used in urlencode and urldecode in PHP.
Removed carriage return

willox:
URLEncode can be done with a single gsub rule by taking the space character out of the first gsub call's pattern and then removing the second gsub call.

zerfgog:
:+1:+1

thegrb93:
I also think this should be in the http library rather than util.

zerfgog:
i disagree

Kefta:
Put it in string, just like JavaScriptSafe, another string formatter for outside purposes.

PR #1096, opened 2015-11-22
New util function: LerpColor

mooshbeef:
Returns a new color that is lerped from two given colors.
Usage:
```lua
local start = Color(255, 0, 0)
local target = Color(0, 255, 0)
local lerped = LerpColor(0.5, start, target)
```


Kefta:
Duplicate of #1027

PR #1097, opened 2015-11-26
relevel concommand: reloads the map

zerfgog:
credit to @Acecool for the name?

### DIFF
--- a/garrysmod/lua/autorun/developer_functions.lua
+++ b/garrysmod/lua/autorun/developer_functions.lua
@@ -27,7 +27,7 @@ local function FindInTable( tab, find, parents, depth )
 				k != "_LOADED" &&
 				k != "__index" ) then
 				
-				local NewParents = parents .. k .. ".";
+				local NewParents = parents .. k .. "."
 				FindInTable( v, find, NewParents, depth )
 			
 			end
@@ -56,7 +56,7 @@ local function Find( ply, command, arguments )
 	Msg("\n\n")
 	
 	if ( SERVER && IsValid(ply) && ply:IsPlayer() && ply:IsListenServerHost() ) then
-		RunConsoleCommand( "lua_find_cl", tostring(arguments[1]) );
+		RunConsoleCommand( "lua_find_cl", tostring(arguments[1]) )
 	end
 	
 end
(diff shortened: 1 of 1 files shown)


thegrb93:
Some servers don't want admins to have access to map changing/reloading. No.

zerfgog:
who has access to changelevel?

thegrb93:
The server itself.

Acecool:
I have this in my dev framework, but it's set to superadmin or console; it's meant to be a developer function because there aren't many reasons to reload the same map ( mostly to ensure files load properly after a big change [ load order, new files if there isn't a method to detect new files and load them easily, etc ] but they are human reasons which have use for a shorter console command, and then there are some bad coding examples such as map-vote scripts that reload the current map instead of extending the round system or resetting it [ but these are dynamic so these don't have much use of a shorter command ] ). It's incredibly useful because you don't need to type out changelevel then the full map name ( with the risk of switching to an incorrect map )...

An alternative that might fit better in with the default game structure would be to allow changelevel to be executed without arguments and for that cas to switched to the current map, keeping in mind the current permissions needed for the command.

And to extend on this, it'd be better to rewrite concommand.Run to have its own hook system ( in my framework ) so that changes to argument behavior can easily be added using something along the lines of: concommand.AddArgsHook( "changelevel", function( _p, _cmd, _args ) return ( _p, _cmd, ( !_args || _args[ 1 ] == "" && { game.GetMap( ) } || _args ) end ); but, it should be changed in such a way so that rewriting certain hard-coded commands can actually be changed...

thegrb93:
I also have my own reload map command. I think console command hooking was discussed in the past, but I don't remember how it was concluded. Probably not in favor of it since it wasn't touched since.

Acecool:
Since a lot of us find it useful, it wouldn't hurt to add it into the default game setup... At the least allow empty-argumented changelevel command.

As for command hooking: I'll have to take a look. I haven't been in-tune with a lot lately because of continuing health issues ( reason why I haven't been on facepunch for a long time ) but this caught my eye.

willox:
I have my srcds set to do this on 'retry' server-side which I think is neat. I think it was Hex's idea.

thegrb93:
I like retry much better since its already a commonly used command.

PR #1098, opened 2015-12-05
console.log can now take multiple arguments

willox:
All the old behaviour remains. Arguments after the first are ignored if Lua is executed.

### DIFF
--- a/garrysmod/lua/vgui/dhtml.lua
+++ b/garrysmod/lua/vgui/dhtml.lua
@@ -29,7 +29,7 @@ function PANEL:Init()
 	--
 	-- Implement a console.log - because awesomium doesn't provide it for us anymore.
 	--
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dhtml.lua
```diff
@@ -92,8 +92,13 @@ function PANEL:ConsoleMessage( msg )
 
 	end
 
+	local values = { ... }
+
+	-- If the first parameter is an object/array/function/undefined, just print this useless message
+	values[ 1 ] = values[ 1 ] or "*js variable*"
```
meepdarknessmeep: Maybe do this to all values? this might also not play well with null or undefined. I know that it's gmod's default behavior already, but we could fix it!

PR #1101, opened 2015-12-14
Added table.IsEmpty()

yupi2:
Checks if there is a key/value in a table.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -716,3 +716,8 @@ function table.GetKeys( tab )
 	return keys
 
 end
+
+function table.IsEmpty( tbl )
+	return next( tbl ) == nil
+end
+


Kefta:
Uh, #tbl == 0?

robotboy655:
@Kefta, that is not a reliable way of doing it ( it doesn't count string indices, etc ).

But we do not need more pointless aliases.

yupi2:
@Kefta That only works correctly for sequential tables (tables indexed 1 through X with all numbers in between).

@robotboy655 It would be better to just add 5 lines to a file than to have people duplicate it in their addons multiple times or have people incorrectly assuming things like @Kefta

thegrb93:
next(tbl)==nil

yupi2:
@thegrb93 What's the point in typing out `type(s) == "string"` when you have `isstring(s)`?
In the case of `table.IsEmpty(tbl)`, it's much more obvious what is happening compared to `next(tbl) == nil`.

thegrb93:
I prefer using the functions I know rather than having to remember a huge list of small utility functions.

PR #1102, opened 2015-12-18
Replace deprecated function calls with non-deprecated alternatives

zerfgog:
In this PR I shift usage of deprecated functions in the codebase to their recommended alternatives.
Notes:
- This includes small (1-2 lines in ~10 files) changes to TTT code.
- I kept duplicate accessors (both Set/GetDrawBackground and Set/GetPaintBackground) in panels for compatibility. I also did not alter the internal names of the accessors.
- I did not touch spawnmenu.PopulateFromEngineTextFiles and spawnmenu.DoSaveToTextFiles because using only their base functions breaks things.
- I left the only ProtectedCall call anywhere in the codebase alone, as I have been informed that it is not just an alias. Replacing it with pcall seems to not break anything, but I left it just in case.


zerfgog:
@robotboy655 are any changes needed to make this merge-ready?

robotboy655:
You should redo this for each gamemode ( especially TTT ) and lua/ folder separately so it's easier to process.

zerfgog:
Will do.

PR #1106, opened 2016-01-01
net.ReadPlayer

Python1320:
Add net.ReadPlayer/WritePlayer, saving GMod from network bloat one byte at a time.

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -66,6 +66,31 @@ function net.ReadEntity()
 	
 end
 
+
+--
+-- Read/Write a player to the stream
+--
+function net.WritePlayer( ply )
+
+	if ( !IsValid( ply ) || !ply:IsPlayer() ) then 
+		net.WriteUInt( 0, 8 )
+	else
+		net.WriteUInt( ply:EntIndex(), 8 )
+	end
+
+end
+
+function net.ReadPlayer()
+
+	local i = net.ReadUInt( 8 )
+	if ( !i ) then return end
+	
+	local ply = Entity( i )
+	return ply
+	
+end
+
+
 --
 -- Read/Write a color to/from the stream
 --


Jvs34:
:-1:

Authorjames:
:-1:

meepdarknessmeep:
:-1:

Kefta:
:-1:

thegrb93:
You guys should at least say why it's not a good idea. If a non-player entity gets put into this function then undefined behavior will occur so there needs to be a check to keep noobs from accidentally doing so.

aStonedPenguin:
:+1:

Kefta:
I doubt the IsPlayer check is faster than writing an 16-bit unsigned int  vs an 8 bit. WriteEntity serves its purpose here

Bo98:
I doubt the idea behind this has anything to do with speed of execution.

zerfgog:
:-1:

thelastpenguin:
+1
Micro optimizations makes sense in video games.

Verideth:
:-1:

UnkN56:
:+1:

DrVrej:
+1

Acecool:
Here is an explanation why I both support this, and don't; more so towards supporting this helper-function though... 

If you want to read/write Player, you need to know a few things... You don't need 8 bits for the amount of data because the max number of players is 128... You can further reduce it depending on max players, basically you use the max player value to choose the number of bits and because both client and server can read it, you can always use the max bit value for the amount of players on the server... Basically, if the max player count is 8, you don't need 7 bits, you can use 3 bits... Now, I see you want to support 0 being sent, but then why check for ! value incoming since net messages are verified unless you force it not to be... You can ignore sending the data and the read value will get nothing, ie sending is optional in some cases, since you're not reading any other data you don't have to worry about putting the optional data last either...

Now, the way I understand networking to work, you use the SUM to know the max value able to be sent... Bit counting details here: https://dl.dropboxusercontent.com/u/26074909/tutoring/networking/bit_counting.lua.html

So with 2^8 you can send a value up to 255 ( 256 if you exclude 0 and shift, which is what we'll need to do for 128 player count to save 1 more bit )... So, for 128 max-players to send 1-128 we can use WriteUInt( EntIndex - 1, 7 ) to send, and Entity( ReadUInt + 1 ) to read... UInt reduces the bit value by 1 because we don't need to know if it is positive or negative, we already know...

If you want to discuss saving a bit each time, why not save a little bit more which the above revisions?

With this said, I agree that micro-optimizations do make sense in video games.... so I'll support this, as long as you save at least 1 more bit; if you want to add in the max player bit-slider in as an additional way to optimize, then I'd support that too....

I am still around and I'm still involved in GMod...

Here...
```lua
--
-- Better, or best... For this we'll simply inverse, ie subtract, based on division... ie default bits is 7 so 128 / 128 is 1 therefore bits - 1 is 7... 128 / 64 is 2, so we get 6, etc...
--
function net.GetPlayerCountInBits( )

	local max = game.MaxPlayers( );

	-- Max players should never be 0, but I've never tried... Prevent sending 0 bits and division by 0...
	-- max = ( max > 0 ) && max || 1;

	local bits = 7;

	-- Ceiling because we need to go to the next step with decimals... +1 because we're using we want bits to be at the max, otherwise we'd need to use 8 or shift the value to allow > 64 be 0....
	return bits - math.ceil( 128 / max ) + 1;

end


--
-- Write a player to the stream
--
function net.WritePlayer( ply )

	if ( IsValid( ply ) && ply:IsPlayer() ) then
		net.WriteUInt( ply:EntIndex() - 1, net.GetPlayerCountInBits( ) )
	end

end


--
-- Read a player from the stream
--
function net.ReadPlayer( )

	local i = net.ReadUInt( net.GetPlayerCountInBits( ) ) + 1
	if ( !i ) then return end

	local ply = Entity( i )
	return ply

end
```

You could a different method for getting the bits to send but it is longer and not as clean in my opinion...
```lua
--
-- Not exactly slow, but not the best either... You could return in each of the if / else and remove the bits var; some hate 1 way in multiple ways out though...
--
function net.GetPlayerCountInBits( )

	local max = game.MaxPlayers( );
	local bits = 7;
	if ( max > 64 ) then
		bits = 7;
	elseif ( max > 32 ) then
		bits = 6;
	elseif ( max > 16 ) then
		bits = 5;
	elseif ( max > 8 ) then
		bits = 4;
	elseif ( max > 4 ) then
		bits = 3;
	elseif ( max > 2 ) then
		bits = 2;
	else
		bits = 1;
	end

	return bits;

end
```

thegrb93:
I think i spent more time reading your post than how much cpu time would be saved over 100 years.

Luabee:
Don't calculate the bit count every time it's used. Calculate it on startup and reference it, since it can't change.

Acecool:
Actually, yeah I messed up on the math for the automated bit calculation.. I'll figure it out and update it..

thegrb93:
```
function net.GetPlayerCountInBits( )

    -- math.log actually gives us bits - 1, or value shifted by -1 ie not counting 0 which works for us...
    local PLAYER_COUNT_IN_BITS = math.floor( math.log( game.MaxPlayers( ), 2 ) );
	net.GetPlayerCountInBits = function() return PLAYER_COUNT_IN_BITS end
    
	return PLAYER_COUNT_IN_BITS;
	
end
```

Acecool:
That would be another micro-optimization which removes around 2 logical controls ( referencing the variable and checking the not ) by rewriting the function as soon as the data has been initialized...

I'd write it slightly differently to be honest, but the effect would be the same... Actually, I'd use Initialize hook to set it up then just return it if that logical operation meant much of anything, but since we're talking about micro-optimizations your post is a good one!

bigdogmat:
Can't the player count be changed while in-game? Also your net.GetPlayerCountInBits gives incorrect results, you should be ceiling the value instead of flooring it, and even then if max players is 1 it's gonna return 0 bits. Either check if it's 0 and add one, or do checks so if it's 0 it just uses local player, which in this would save us having to send the 1 bit, ie

```lua
function net.WritePlayer(ply)
    if ( IsValid(ply) && ply:IsPlayer() ) then
        local bits = net.GetPlayerCountInBits()

        if ( bits > 0 ) then
            net.WriteUInt(ply:EntIndex() - , bits)
        end
    end
end

function net.ReadPlayer(ply)
    local bits = net.GetPlayerCountInBits()

    if ( bits > 0 ) then
        local i = net.ReadUInt(bits)
        if ( !i ) then return end

        return Entity(i - 1)
    else
        return LocalPlayer()
    end
end

**Comment by robotboy655** [comment]
No, you cannot change maxplayers while the server is running.

**Comment by Acecool** [comment]
I thought I tested it for 1 player and it returned 1, I'll double check...

**Comment by MrUnknownGamer** [comment]
Nice! I feel like every update you do though always involves something being changed in the net code. 


Anyways, :+1

**Comment by Leystryku** [comment]
You can use disasm magic to change maxplayers while the server is running :v:

**Comment by uRandomAlex** [comment]
@Leystryku Share with us? I've many times been in situation, where my server is full, enough CPU is free for more players but I can't change maxplayers...

Actually, GMod devs should rewrite "maxplayers" CVar to function it like this: when you change it, it applies on map change. We are able to do this, because we can change maxplayers on a listen server without restarting whole game.

**Comment by aStonedPenguin** [comment]
Let me share the super secret method with you. It's sv_maxvisibleplayers

**Comment by uRandomAlex** [comment]
@aStonedPenguin I know about this ConVar, but players are able to join my server when it's full (and I don't want it) using "connect 13.37.02.28" this way.

**Comment by Leystryku** [comment]
hook.Add("player_connect","ok", function(idc, lol) if(GetConVar("sv_maxvisibleplayers"):GetInt() > #player.GetAll()) then return "oi get out m8" end end)

**Comment by Kefta** [comment]
@Leystryku Use CheckPassword pls; no one wants to sit there downloading everything just to be kicked

**Comment by Leystryku** [comment]
Got them confused sry
Am 04.07.2016 00:19 schrieb "code_gs" <notifications@github.com>:

**Comment by uRandomAlex** [comment]
Seems that no one can get Ley's secrets if he don't want/too lazy today. ;(

**Comment by Leystryku** [comment]
Well bytepatching the game is hacky solution. The if statement to prevent
maxplayers from changing was put there for a reason. Visiblemaxplayers  and
the checkpassword hook are perfect for this kind  of stuff while keeping
maxplayers sane.
Am 04.07.2016 9:47 vorm. schrieb "uRandomAlex" <notifications@github.com>:

**Comment by BlackCetha** [comment]
Summing up what @Leystryku is doing:
He adds a CheckPassword-Hook that is called when a player joins. If sv_maxvisibleplayers is smaller than the current player count, it means the server is virtually full and the player is disconnected.
This results in the following (untested) code:

```lua
hook.Add("CheckPassword", function (steamID, ip)
    if(GetConVar("sv_maxvisibleplayers"):GetInt() < #player.GetAll()) then
        return false
    end
    return true
end)
```

Documentation for the CheckPassword-Hook: http://wiki.garrysmod.com/page/GM/CheckPassword

Acecool:
player_connect hook works like the password check hook... but, you need to add the game-event for player_connect..

ZehM4tt:
```lua
--
-- Write a player to the stream
--
function net.WritePlayer( ply )

    if ( IsValid( ply ) && ply:IsPlayer() ) then
        net.WriteUInt( ply:EntIndex() - 1, net.GetPlayerCountInBits( ) )
    end

end
```

This is a problem not sending anything when its invalid, do it like Python1320 and send 0, you shouldnt corrupt the order of variables in the stream.

Kefta:
Also, the IsValid check here is unnecessary. The possibly invalid values include nil, NULL, and a non-player; so the two checks only need to be "if (ply == nil or not ply:IsPlayer())"

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -66,6 +66,31 @@ function net.ReadEntity()
...
+		net.WriteUInt( ply:EntIndex(), 8 )
+	end
+
+end
+
+function net.ReadPlayer()
+
+	local i = net.ReadUInt( 8 )
+	if ( !i ) then return end
+
```
SnakeSVx: Shouldn't this check if i ~= 0 instead? because 0 = true in lua.

PotcFdk: If `i == 0`, `Entity (0)` (-> `Entity [0][worldspawn]`) gets returned.

Python1320: Copied from function net.ReadEntity()

GitSparTV:
This request is useless.

Because of this:
`--
-- Read/Write an entity to the stream
--
function net.WriteEntity( ent )

	if ( !IsValid( ent ) ) then 
		net.WriteUInt( 0, 16 )
	else
		net.WriteUInt( ent:EntIndex(), 16 )
	end

end

function net.ReadEntity()

	local i = net.ReadUInt( 16 )
	if ( !i ) then return end
	
	return Entity( i )
	
end`

CornerPin:
@GitSparTV, the point of this pull request is to use 8 bits instead of 16 when writing players to the net stream.

meepdarknessmeep:
The patch is bad since it uses entindex instead of handles

On Nov 27, 2017 3:35 PM, "CornerPin" <notifications@github.com> wrote:

@GitSparTV <https://github.com/gitspartv>, the point of this pull request
is to use 8 bits instead of 16 when writing players to the net stream.

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/Facepunch/garrysmod/pull/1106#issuecomment-347319567>,
or mute the thread

Luabee:
What's wrong with using an EntIndex for networking? How else could it be
done?

meepdarknessmeep:
Entindex can be used by multiple entities over time while handles will
always try to be unique per entity

Luabee:
Not players. They reserve id's 1-128 and it doesn't change as long as
theyre connected.

The point is to network the pointer to the player, not all the data
associated with it.

meepdarknessmeep:
Yes, as long as any entity continues to exist it will have an entindex.
More than one entity can have the same entindex over time.

Kefta:
Why? How can you reproduce this? They're supposed to be unique.

sannysc:
They are unique at any given moment, but not _over time_. Add a bot to a server, kick the bot, and add another bot. Both of those bots used the same index, but are, of course, not the same entity.

Kefta:
Oh, I misunderstood. Is networking order not maintained to where this wouldn't be an issue?

meepdarknessmeep:
there's no guarantees that reliable and unreliable channels will be in order

Luabee:
Since this is just for networking and not saving, and since the player's
EntIndex is read in real time just as it is sent, it's HIGHLY unlikely that
the entity being pointed to will have changed between the time that
net.WritePlayer is called and the time net.ReadPlayer being called.

It's no less accurate than net.Read/WriteEntity

meepdarknessmeep:
just because it's highly unlikely in certain use cases doesn't mean it should be be kept the same. There's a reason Valve decided to send entities as a handle in Source Engine's original code - and that reason is because it should be that way. Don't change what isn't broken if it causes more problems than before.

As for net.*Entity, I believe that needs to be changed to. Adding this just adds to the problem, however.

meepdarknessmeep:
Also, as I said before it's not read in-order if you use two different channels of reliability.

GitSparTV:
This PR should be rewritten to account game.GetMaxPlayers().
If required, I can propose changes in separate PR.

robotboy655 (approved):

GranPC:
Damn son @Python1320 your PR finally made it!

Python1320:
2023 me thinks this may be a bad idea indeed due to race conditions, but no worse than `net.WriteEntity()`. Player entities are special in any case (Strange PVS behaviour causing temporarily NULL players, for example) so this will leave room for improvement.

In addition we need a better `net.ReadPlayer()` where we can separately handle a NULL player due to disconnection and NULL player due to ongoing fullupdate or join. On metastruct we have experimented with making net messages coroutines where you can wait for a player to become valid again before continuing processing the message, with default gmod settings this delay should be no more than about 10 seconds, except maybe during spawn. This can be done by keeping track of the players via `player_disconnect`. If they did not leave then [player_disconnect](https://wiki.facepunch.com/gmod/gameevent/player_disconnect) is not fired and the entity will become valid (unless the player leaves while being invalid, edge cases!). 
On `player_disconnect` you mark players as actually left so you can see if any given player entity is just temporarily `NULL`.


This is speculation, but as stated in previous messages, on join you may get a large buffer of net messages from the server, during which players may join and leave and therefore cause the wrong player to receive data (at least it seems that net messages can be delayed compared to game state). You should therefore network for each player even an empty data (if storing data for players), for example. This way the joining players will at least eventually get the correct data for each player. If you do not network the empty dataset for every entity id that used to have data, then wrong assignments will persist, if you are unlucky. 
Also `net.WritePlayer()` may not even be the most efficient networking format for some cases, when you could just network for each player slot a status bit for whether there is `hasdata` AND of course the actual data if the `hasdata`-bit is `true`. This way you need only `MAXPLAYER/8=16` bytes to network the fact that no player has any data and really avoiding the edge case altogether if you do it right, I think.

PR #1107, opened 2016-01-03
Replace loads of if with ternary

Kngrektor:


zerfgog:
this block would also be improved by splitting the string into a table of substrings instead of :find()ing each possible choice. the wiki already tells users to separate the terms with a space, so it shouldn't break anything.

Kngrektor:
I though of doing a map / dictionary like table but didnt because it felt
unnecessary, but i guess thats what the devs thought when they did the if
statements.

PR #1109, opened 2016-01-10
Entity.SetNetworkedString marked as deprecated

JamesxX:
On the wiki, Entity:SetNetworkedString() is marked as deprecated and "may be removed in future versions". Updated instances to Entity:SetNWString() instead.


Bo98:
#1102

PR #1110, opened 2016-01-11
Fixed excessive balloon force crash

Mista-Tea:
Similar to the [thruster force crash](https://github.com/garrynewman/garrysmod/commit/6160c8f166828a3c9488b44b02ee61727451e891#diff-d3d6727ecec503cb9f9bd8afff543124]), you can crash servers with balloon force values at 1E35 and above (1E36 on the thruster) or below -1E35.

I've clamped both places where this value is used and removed the redundant code on line 62. I've tested this by both spawning a balloon regularly and by modifying a Duplicator-duped balloon's JSON value to 1E35. Crashes occurred before this patch, and were eliminated after it.


robotboy655:
1E35 is probably an overkill anyway, but ok.

Mista-Tea:
Thanks :)

PR #1114, opened 2016-01-16
Added 'wep' field to all was_pushed tables.

DarkRedemption:
TTT currently tracks if a player was pushed with a pushing weapon (Newton Launcher, Discombobulator, crowbar alternate fire), but does not track which weapon was used. This simple change to all of the push weapons' was_pushed tables adds the "wep" field, which stores the weapon's class name, allowing for easy tracking. Useful for stats and damage log addons.

Discombobulators, unfortunately, did not have a good way to pull its class name programmatically where its was_pushed table is generated. Therefore, a string literal is used instead for it.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_confgrenade_proj.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_confgrenade_proj.lua
@@ -39,7 +39,7 @@ local function PushPullRadius(pos, pusher)
 
             target:SetVelocity(vel)
 
-            target.was_pushed = {att=pusher, t=CurTime()}
+            target.was_pushed = {att=pusher, t=CurTime(), wep="weapon_ttt_confgrenade"}
 
          elseif IsValid(phys) then
             phys:ApplyForceCenter(dir * -1 * phys_force)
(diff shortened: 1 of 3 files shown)


garrysmod/gamemodes/terrortown/entities/entities/ttt_confgrenade_proj.lua
```diff
@@ -39,7 +39,7 @@ local function PushPullRadius(pos, pusher)
 
             target:SetVelocity(vel)
 
-            target.was_pushed = {att=pusher, t=CurTime()}
+            target.was_pushed = {att=pusher, t=CurTime(), wep="ttt_weapon_confgrenade"}
```
svdm: Looks like this should be `weapon_ttt_confgrenade`

DarkRedemption: That's what I get for retyping the change instead of copying it. I'll fix it momentarily.

DarkRedemption: Fixed and rebased. Thanks for pointing that out.

PR #1116, opened 2016-01-18
Fixed DHTMLControls

Bo98:
After the Awesomium functions and hooks now work/exist, this control can now be fixed. It was pretty much broken at everything except entering URLs into it.

Side notes about the reasoning behind some of the changes:
* `self.Refreshing` was replaced with a URL check as such check was required to prevent a double history insertion on setup. If the controls were hooked before the initial page began loading (likely scenario as `:OpenURL(...)` doesn't start loading immediately) then the history would be appended on setup and again on load. Instead of removing the initial insertion, a history check was added in case someone decided to add the controls later. This also covers refreshes so that special boolean isn't required now.
* `self.NavStack` was useless and is now removed.

### DIFF
--- a/garrysmod/lua/vgui/dhtmlcontrols.lua
+++ b/garrysmod/lua/vgui/dhtmlcontrols.lua
@@ -1,6 +1,6 @@
---[[__                                       _     
- / _| __ _  ___ ___ _ __  _   _ _ __   ___| |__  
-| |_ / _` |/ __/ _ \ '_ \| | | | '_ \ / __| '_ \ 
+--[[__                                       _
+ / _| __ _  ___ ___ _ __  _   _ _ __   ___| |__
+| |_ / _` |/ __/ _ \ '_ \| | | | '_ \ / __| '_ \
 |  _| (_| | (_|  __/ |_) | |_| | | | | (__| | | |
 |_|  \__,_|\___\___| .__/ \__,_|_| |_|\___|_| |_|
                    |_| 2010 --]]
@@ -8,7 +8,7 @@
 local PANEL = {}
 
 --[[---------------------------------------------------------
-	
+
 -----------------------------------------------------------]]
 function PANEL:Init()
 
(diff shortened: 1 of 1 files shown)


zerfgog:
A+ commit :+1:+10

Bo98:
Works fine for me. In what case is it glitchy for you?

zerfgog:
1. Navigate to http://wiki.garrysmod.com
2. Click on the "Getting Started" link on the home page
3. Click on the "Main Menu" link from the "Getting Started" page
4. Hit the back arrow until you cannot go back any farther
5. Hit the forward arrow once
All forward history should be gone now.

zerfgog:
Alright, so there aren't as many bugs as I'd thought.
Some flaws though:
- Refreshing shouldn't add a new history entry
- When you refresh, instead of disabling the button, replace it with the "cancel loading" button as most browsers do. Then this button can be remoevd from the right side.

Bo98:
> * Refreshing shouldn't add a new history entry

Doesn't happen for me.

> * When you refresh, instead of disabling the button, replace it with the "cancel loading" button as most browsers do. Then this button can be remoevd from the right side.

I didn't touch this but I agree that it's retarded so I'll add a commit for that.

Bo98:
There is a history corruption issue with fragment identifiers but it is impossible to fix currently without constantly (eg. Think hook/timer) checking the URL property.

@willox Can we get `Awesomium::WebViewListener::View::OnChangeAddressBar`? Not sure whether CEF is anywhere near ready or not.

Bo98:
> * When you refresh, instead of disabling the button, replace it with the "cancel loading" button as most browsers do. Then this button can be remoevd from the right side.

Added now.

zerfgog:
This will need to be updated for http://wiki.garrysmod.com/page/Panel/StopLoading for the next update.

Bo98:
Done.

Bo98:
@zerfgog The merging of the stop and refresh buttons unfortunately won't happen (and thus is reverted) since apparently some addons relied on the internal view hierarchy of DHTMLControls and broke when the buttons were merged.

zerfgog:
@Bo98 :neutral:

Kefta:
Put a warning out next update, deprecate the behaviour, merge your PR

Bo98:
This control is eventually going to be deprecated and replaced in its entirety when CEF is implemented so we can wait until then.

PR #1117, opened 2016-01-23
Fixed player being able to noclip while dead

Kefta:


Kefta:
Ignore this, used the wrong operators

PR #1120, opened 2016-01-26
Drop in library that adds an object that presents UTF8 strings as if they were normal strings to other code.

ILOVEPIE:
Added utf8string library for containing and manipulating utf8 strings using the same method names as the string library so that it can be used in most cases as a drop in replacement to allow utf8 support in non-utf8 supporting scripts that require string input (more advanced functions like patterns have yet to be implemented).

### DIFF
--- a/garrysmod/lua/includes/modules/utf8string.lua
+++ b/garrysmod/lua/includes/modules/utf8string.lua
@@ -1,0 +1,124 @@
+local string = string
+local print = print
+local PrintTable = PrintTable
+local error = error
+local unpack = unpack
+local table = table
+local tostring = tostring
+local setmetatable = setmetatable
+local getmetatable = getmetatable
+local bit = bit
+local pairs = pairs
+module("utf8string")
+local unicode_string = {}
+unicode_string._metatable = unicode_string
+unicode_string.prototype = {}
+-------------------------------------------
+function unicode_string.__index(index)
+	return unicode_string.prototype[index]
+end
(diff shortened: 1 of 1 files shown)


zerfgog:
croikey that code

willox:
So the idea is that you pass a "utf8string" object to a function that expects a string and it will magically do stuff like making str:len() return the amount of codepoints? What about when people use string.len?

ILOVEPIE:
this is easier to use than the utf8 library, also your right that some code won't work with the drop in utf8 support that this provides but that code can be easily fixed to get it working (i.e. making sure that you use str:len() instead of string.len(str)).

garrysmod/lua/includes/modules/utf8string.lua
```diff
@@ -0,0 +1,124 @@
...
+		local str = utf8("")
+		for _,value in pairs(t._string or {}) do 
+			table.insert(str._string,value)
+		end
+		for _,value in pairs(t2._string or {}) do
+			table.insert(str._string,value)
+		end
+		return str
+	else
+		error("cannot concatinate a non utf8string with a utf8string")
```
Khubajsn: concatenate

willox:
It's a cool idea but I just don't see passing unexpected data to other people's code as a good thing to support. If there's anything in the base game that you think should support UTF-8 data or any functions you think would make it easier for addon developers to do, feel free to PR that instead.

PR #1124, opened 2016-01-30
Autorun lua_run entity's code if spawnflag is set

SuperHighVoltage:
This edit would run lua_run entities code when its created if its spawnflag is set. This will eliminate the need to trigger the entity with a logic_auto when all you want to do is run a piece of code once when the map loads. 
I'm running it in the KeyValue hook so it should get called before anything is created, instead after entities are spawned.

Also, does Garry's mod have an official fgd for hammer? I know there are some floating around but I feel that one should be included .
I could look into fixing one up and merge it if that is okay

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/lua_run.lua
+++ b/garrysmod/gamemodes/base/entities/entities/lua_run.lua
@@ -13,12 +13,17 @@ end
 
 --[[---------------------------------------------------------
    Name: KeyValue
-   Desc:
+   Desc: Sets the default code and runs it if appropriate spawnflags are set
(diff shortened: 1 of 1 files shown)


neico:
The fgd file is under steamapps/common/garrysmod/bin
next to hammer.exe, just like in any other source engine game

SuperHighVoltage:
I probably knew that but it's been so long since I needed it. I searched late at night and gave up early

robotboy655:
Merged manually in https://github.com/Facepunch/garrysmod/commit/283df2d1f626b5ec729f9e9ca1cc9aadd5f0cc07

PR #1126, opened 2016-02-03
fix bug from language strings update

andyault:


Kefta:
So it's literally that easy to bypass?

andyault:
I wouldn't call it a bypass, this is what language strings are for
the reason it was removed I'd imagine was cause either admins were using language strings to be dumb or people would set language strings in their name and mess something up, something like that

Kefta:
@andyault I thought that was a fixed a LONG time ago

andyault:
it was, but I'm guessing this last update was to fix something left over from that maybe

Jvs34:
Did you test this? language.GetPhrase is clientside while the hook is called serverside, I don't see how this would work.

andyault:
woops, was trying to get zings, didn't even check

PR #1127, opened 2016-02-03
Changed glock sound from soundscape to path

Kefta:
Stop fucking whining

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_glock.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_glock.lua
@@ -34,7 +34,7 @@ SWEP.ViewModelFOV		= 54
 SWEP.ViewModel  = "models/weapons/cstrike/c_pist_glock18.mdl"
 SWEP.WorldModel = "models/weapons/w_pist_glock18.mdl"
 
-SWEP.Primary.Sound = Sound( "Weapon_Glock.Single" )
+SWEP.Primary.Sound = Sound( "weapons/glock/glock18-1.wav" )
 SWEP.IronSightsPos = Vector( -5.79, -3.9982, 2.8289 )
 
 SWEP.HeadshotMultiplier = 1.75


comedinha:
This fix the error, but "Weapon_Glock.Single" have a error...

Kefta:
@comedinha The soundscape system really needs to be fixed when it comes to duplicate strings.

Violat0r:
Works but has anyone tried any other HL1 sounds to see if they work?

Kefta:
@Violat0r That's not the problem. It's because most people don't have HL1 mounted and the HL1 version is overwriting the CS:S version for some reason

Violat0r:
Right, I see what you mean. Another possible solution to this would be to rename the soundscape for CS:S/HL1 and in weapon_ttt_glock.lua, something to consider or try.

deinemudda32:
How about fixing this problem the right way instead just for TTT?

Kefta:
@deinemudda32 Rubat already undid the change that supposedly caused this. You want to fix it? Go right ahead

deinemudda32:
But it did work the update before/before.
I didnt want to offend you, if I did. I just don't like such methods.

willox:
Do you guys have a garrysmod/scripts directory? Delete it and let me know if it fixes the issue. Hopefully we'll have this fixed for the next patch.

deinemudda32:
Glock works for me :)

willox:
You'll only have an issue if for some reason an old version of the script is in your garrysmod/scripts directory.

Violat0r:
I have the issue and I do not have garrysmod\scripts. I do have my Garry's Mod installed on my D drive instead of C, not sure if that matters but trying to provide whatever information that may help.

robotboy655:
The problem is with the server, not the client as far as I know.

Violat0r:
I am not sure what I did but it is no longer an issue for me. Perhaps it was something with the hotfix this morning as i did update my client and local server.

Violat0r:
Alright, on my main server that was NOT updated with the hotfix yet, the error exists but I am able to hear the gun fire.

Kefta:
Okay good, we're done here

Kefta:
Thanks for the updates on the issues

Violat0r:
You're welcome. I assume that some of the files may have no been pushed out with the update the other day or something in the hotfix fixed the issue.

PR #1129, opened 2016-02-07
DModelPanels with alpha

disseminate:
Takes the DModelPanel's alpha into consideration when drawing it


Jvs34:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/448 I guess

PR #1131, opened 2016-02-14
Added missing PrintName to some scripted entities

mcd1992:


garrysmod/gamemodes/sandbox/entities/entities/gmod_winch_controller.lua
```diff
@@ -1,7 +1,8 @@
 
 if ( CLIENT ) then return end
 
-ENT.Type = "point"
+ENT.PrintName 	= "Winch Controller"
```
Bo98: `PrintName` is clientside so this won't work here because of the `if` statement above.

mcd1992: Didn't notice that, thanks.

robotboy655:
https://github.com/garrynewman/garrysmod/commit/cfee866512978609bbddbe3920dfb7e39936beea

Did a cleaner version of this so I don't have to clean up after your PR.

mcd1992:
Thanks

PR #1133, opened 2016-02-17
Temporary exploit fix

JamesxX:
As described here ( https://facepunch.com/showthread.php?t=1506647 ) there is an exploit that allows players to crash servers by picking up props without gravity using their gravity gun.


Bo98:
Won't it still crash though if something calls `PhysObj:EnableGravity(false)` on it though? `PhysObj:IsGravityEnabled()` might be better.

Though this might be something better fixed in the engine as it may crash in a specific function call that is used by other things. I've not seen the stack trace.

willox:
We have a real fix

PR #1134, opened 2016-02-23
Make CreateClientConVar support help text

Kefta:


neico:
What happens when you use CreateClientConVar( "convar", "0" ), my guess is that it'll create a error (or echo "nil" as helptext) because helptext has no fail safe, change the CreateConVar call to `( helptext or "" )` to be sure (or use a `local helptext = helptext || ""`, depending on which behavior I mentioned is observed)

zerfgog:
Seeing as nil is passed when the argument is not provided, it should not error.

neico:
Hmm might be that I'm thinking too much in C++ terms.

Kefta:
Already tested. Passing nil as latter arguments has no effect in Lua, regardless if it's crossing the C++ boundry

willox:
Depending on the function's implementation it actually can, but not in this case.

PR #1138, opened 2016-02-29
Added L4D2 weapon sounds script

Kefta:
L4D1 has one, L4D2 should also


deinemudda32:
+1

Kefta:
Actually, I rather have all the new sound scripts. I'll make a new PR with them

PR #1139, opened 2016-02-29
Added duplicator support to the Eye Poser tool

NO-LOAFING:
Now uses an entity modifier to reapply the eye pose when loading a save, so you don't have to reapply it manually any more.


deinemudda32:
+1

PR #1141, opened 2016-03-04
Fixed SKIN:SchemeTextEntry not working with Panel:SetSkin

Bo98:
My last PR worked fine for getting fonts to work - but it only worked if the default skin was changed rather than `Panel:SetSkin`. This wasn't something that broke in my PR - it never worked for text colours either. The reason why it didn't work is that `Panel:SetSkin` is typically called after `vgui.Create` which of course is after `PANEL:Init` and thus after the `SKIN:SchemeTextEntry` call.

This pull request fixes this issue by moving the call to `PANEL:ApplySchemeSettings`, just like it is in other Derma controls. The reason it wasn't there in the first place was based on how the skin's call to set the text colours could override the a call that's done when setting up the panel. This was fixed by doing `||` checks before sending the colours into `Panel:DrawTextEntryText`.

This slight change of behaviour shouldn't break anything but, if possible, **this should be merged before the next update in case someone gets the wrong idea of using `Panel:SetFont` instead of `Panel:SetFontInternal`** in `SKIN:SchemeTextEntry`.

### DIFF
--- a/garrysmod/lua/skins/default.lua
+++ b/garrysmod/lua/skins/default.lua
@@ -453,16 +453,9 @@ function SKIN:PaintTextEntry( panel, w, h )
 		end
 	
 	end
-	
-	panel:DrawTextEntryText( panel.m_colText, panel.m_colHighlight, panel.m_colCursor )
 
-end
 
-function SKIN:SchemeTextEntry( panel ) ---------------------- TODO
-
-	panel:SetTextColor( self.colTextEntryText )
-	panel:SetHighlightColor( self.colTextEntryTextHighlight )
-	panel:SetCursorColor( self.colTextEntryTextCursor )
+	panel:DrawTextEntryText( panel.m_colText || self.colTextEntryText, panel.m_colHighlight || self.colTextEntryTextHighlight, panel.m_colCursor || self.colTextEntryTextCursor )
 
 end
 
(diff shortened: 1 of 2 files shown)


robotboy655:
This PR is very terrible, but the old way is also terrible, and I have no idea what to do.

Bo98:
What's wrong with this method? It's consistent with the other controls.

robotboy655:
It leaves GetTextColor & Co returning nil.

robotboy655:
Which is why ULX is broke.

Bo98:
GetTextColor is supposed to be nil IMO since that should only have been used for overriding the SKIN (see DLabel).

Maybe something like DLabel's GetColor method could be implemented here but it won't address backwards compatibility.

SticklyMan:
I've got a workaround that should suppress the error till you guys figure this out. Sorry bout the spew of issue reports!

Kefta:
@SticklyMan They already fixed your portion with https://github.com/garrynewman/garrysmod/commit/35b990d58fd9e639e291af0e0c0d3d89b59b7063

SticklyMan:
@Kefta Excellent, thanks for the heads up!

Bo98:
How does DHTMLControls have anything to do with this issue?

Bo98:
Fore those watching this issue, see #1147.

PR #1145, opened 2016-03-14
[TTT ] Fix radio calling out dead players

end360:
Fixes the radio calling out dead or spectating players.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
@@ -259,11 +259,11 @@ function RADIO:GetTargetType()
    local ent = trace.Entity
 
    if ent:IsPlayer()and  ent:IsTerror() then
-         if ent:GetNWBool("disguised", false) then
-            return "quick_disg", true
-         else
-            return ent, false
-         end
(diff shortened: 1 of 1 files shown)


zerfgog:
might be neater if lines 261 and 262 were consolidated with
```lua
if ent:IsPlayer() and ent:IsTerror() then
```

end360:
@zerfgog I assumed that if the entity Isn't a player then an error would be thrown if you tested if IsTerror() since I believe that works only with players

bigdogmat:
The second condition is only checked if the first one is true.

bigdogmat:
You forgot to undo the indention

end360:
Thanks for telling me that, I'm a bit distracted by class right now.

garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
```diff
@@ -258,7 +258,7 @@ function RADIO:GetTargetType()
 
    local ent = trace.Entity
 
-   if ent:IsPlayer() then
+   if ent:IsPlayer()and  ent:IsTerror() then
```
svdm: Good fix and I hate to nitpick, but please make the `and` have one space on either side.

PR #1147, opened 2016-03-29
Fixed some backwards compatibility issues with #1141

Bo98:
Best fix I could think of.

### DIFF
--- a/garrysmod/lua/skins/default.lua
+++ b/garrysmod/lua/skins/default.lua
@@ -455,7 +455,7 @@ function SKIN:PaintTextEntry( panel, w, h )
 	end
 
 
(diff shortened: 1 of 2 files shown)


robotboy655:
Yeah, this is what I currently have, but it has one major downside - what if the skin does not have that variable?

Bo98:
It falls back to the default skin.

https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/derma/derma.lua#L35-L37

Bo98:
Though even if it didn't, the behaviour would be the same as before.

robotboy655:
Was not aware of that fallback stuff, I suppose this fix works then.

PR #1148, opened 2016-03-30
TTT: Fix weapon shortage on CSS maps

edgarasf123:
cvars.Number("maxplayers",16) always returns 0, causing weapon shortage on some CSS maps.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
@@ -317,7 +317,7 @@ local function PlaceWeaponsAtEnts(spots_classes)
 
    local spawnables = ents.TTT.GetSpawnableSWEPs()
 
-   local max = math.max(cvars.Number("maxplayers", 16), #player.GetAll())
+   local max = game.MaxPlayers()
    max = max + math.max(3, 0.33 * max)
 
    local num = 0


edgarasf123:
Usually the script would check "maxplayers" number of the server, but since the function they use is broken, it checks by current number of players.
So basically because of this bug: If server has players 16/32 it only gets 22 weapons(primary and secondary weapons combined), when it should be giving 42 weapons originally.

Bo98:
IIRC `GetConVar("maxplayers")` returns nil which is why this happens.

willox:
The function exists because maxplayers is actually a console command

edgarasf123:
The function to get the cvar is broken, and even the cvar itself shows non-sense. 
Why check the cvar when there is already a function that gives valid value?

bigdogmat:
If I had to guess it existed before the function did and I guess the convar existed so you wouldn't have to set max players on the command line.

comedinha:
math.max(game.MaxPlayers(), #player.GetAll())??

https://facepunch.com/showthread.php?t=1500238

Bo98:
> The function exists because maxplayers is actually a console command

`ConVarExists("maxplayers")` shouldn't be returning true.

robotboy655:
@edgarasf123, It is not broken. ```maxplayers``` is not a console **variable**, it is a console **command**. ( CreateConvar vs concommand.Add ). There was a garryfix for the GetConVarNumber functions, but that was removed when we switched those to Lua without the workaround.

PR #1149, opened 2016-04-02
Improvements to server list

DarthTealc:
Allow listing of multiple gamemodes, favouriting servers, blacklisting servers, better performance, can drag the server list around, can view server info for multiple servers at the same time, apply filters on ping, options to filter out servers that are empty or full, option to filter out passworded servers, option to select specific server regions to view.


zerfgog:
:joy:

willox:
Why did nobody say the legacy browser's region filter worked before?

Kefta:
Happy April 2nd

PR #1150, opened 2016-04-04
TTT: CVar weapon spawn amount

edgarasf123:
This feature allows for server owners to control weapon spawn amount on
non populated maps, independent of max players settings.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
@@ -318,7 +318,7 @@ local function PlaceWeaponsAtEnts(spots_classes)
 
    local spawnables = ents.TTT.GetSpawnableSWEPs()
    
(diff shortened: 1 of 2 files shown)


svdm:
Seems useful, but I think `ttt_weapon_spawn_count` would be a better name, because the cvar will set the number of weapons that are going to spawn on these maps. The "max" part of the current name only makes sense in context of the whole max players thing, which is obvious to us but not to server owners who haven't seen the code.

edgarasf123:
Hmm, I thought `max` was named because of the algorithm used. In some cases the algorithm can spawn 2 grenades instead of a weapon. And in some really rare cases the algorithm could spawn only grenades in whole map. 

`spawned_guns` == `spawn_count`
`spawned_guns` <= `spawn_max` <-- This is how algorithm works now

svdm:
You're not wrong, but the grenade-cases are relatively unlikely unless a server has some very odd set of SWEPs. In the common case guns will equal count, with some extra grenades present.

PR #1151, opened 2016-04-05
Added hook OnPropertiesMenuOpened

BillyOnWiiUIn144p:
This hook will allow for addons to create more advanced properties menus.

At the moment, we can only use properties.Add to add to the properties list and the functionality of it is quite limited.

If we add this hook, it will allow for addons to programmatically change this DermaMenu by being able to add sections and use the DermaMenu functions. This allows for trees of sections (SubMenus), spacer usage (AddSpacer) and more.

The alternative to this is to add more advanced functions to the properties library. In my scenario, I would like to add a small section on the properties list for my latest addon which I would like to have SubMenus and the like.

This cannot be solved by detouring OpenEntityMenu as you would need to access the DermaMenu which is local.

### DIFF
--- a/garrysmod/lua/includes/modules/properties.lua
+++ b/garrysmod/lua/includes/modules/properties.lua
@@ -87,6 +87,8 @@ function OpenEntityMenu( ent, tr )
 	end
 	
 	menu:Open()
+	
+	hook.Run( "OnPropertiesMenuOpened", menu )
 
 end
 


robotboy655:
But you can already do both?
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/autorun/properties/bodygroups.lua#L33-L88

https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/autorun/properties/editentity.lua#L7

https://github.com/BillyOnWiiUIn144p/garrysmod/blob/patch-1/garrysmod/lua/includes/modules/properties.lua#L85
https://github.com/BillyOnWiiUIn144p/garrysmod/blob/patch-1/garrysmod/lua/includes/modules/properties.lua#L66-L68

BillyOnWiiUIn144p:
I should learn how to read

Thanks

BillyOnWiiUIn144p:
Actually MenuOpen isn't not documented on [the wiki](http://wiki.garrysmod.com/page/properties/Add)

robotboy655:
Added a few missing things

PR #1154, opened 2016-04-12
Add net overflow checking

Python1320:
Add net.HasOverflowed() for write overflow detection.
Make net.WriteTable check for this overflow.

NOTE: This PR is more likely to provide discussion for this issue than get approved.

net.WriteTable now errors if the table is too big instead of client getting cryptic errors of failed to read tables. This can happen and too few addons check for this. If you are sending 64kb in a table you are most likely doing something wrong already. The question is should we keep graceful degradation or provide explicit error signaling. With explicit error you can still continue sending the table if you trap the WriteTable assert so this does not reduce functionality in the long run.

As further improvement net.Send\* functions could have an extra parameter to check/not check for this instead and through that provide universal signaling.

What do you think?

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -109,7 +109,9 @@ function net.WriteTable( tab )
 	
 	-- End of table
 	net.WriteType( nil )
-
+	
+	assert( !net.HasOverflowed(), "WriteTable overflowed" )
+	
 end
 
 function net.ReadTable()
@@ -152,7 +154,9 @@ function net.WriteType( v )
 	end
 
 	local wv = net.WriteVars[ typeid ]
-	if ( wv ) then return wv( typeid, v ) end
+	if ( wv ) then 
+		return wv( typeid, v ) 
+	end
 	
 	error( "net.WriteType: Couldn't write " .. type( v ) .. " (type " .. typeid .. ")" )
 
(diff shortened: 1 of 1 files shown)


Acecool:
Depends; sending over 64kb in a table could be a sign of a full server ( sending once on fully connected, ie sync data ) and a lot of other information that is stored in one place for sync... I do something similar for my networking system... Keeping all data in 2 tables, private and public which is synced on fully connected... Although even with a full server and everyone with 4 vehicles ( 512 with max info such as attachments, sirens, etc... )  plus a ton of props I don't get near 64kb.. I think the largest was around 32 and that was with my framework / networking system in dev mode which transfers full-string-keys instead of assigning numerical values when not in dev mode.... 64kb is a LOT of data..

When I read the title, I thought you were referring to the client timeout / buffer overflow which happens when you send around 2MB worth of data while the client is still connecting ( voice applies to this too )... For this I'd say yes... I got around it by not sending any net messages ( the engine still sends stuff plus voice to keep-alive though ) until the user fully connects... A lot of servers that are full, have the player spawn in crowded areas or with voice on all- instead of area- and that are either poorly optimized, etc.. have this issue... For that, I could see a glowing error preventing the client from overflowing / being kicked until their queue has been processed.... But, even better, would be them to fix their code by optimizing it...

Python1320:
Making this only a warning would work too. Making it error/warn in net.Send*/net.Broadcast could also be useful.

robotboy655:
This feels like a warning from the C side is a better idea (or an ErrorNoHalt with stack)

robotboy655:
I have just added an ErrorNoHalt to the sending functions if the net message was overflown. I have not added `net.HasOverflowed()` as I do not see a point for it.

PR #1155, opened 2016-04-13
Fixed table.IsSequential being inaccurate in some cases

Bo98:
As discussed in great detail in the Next Update thread.


robotboy655:
Why change ```i``` and the position of the ```if```? It serves no benefit or makes no difference?

Kefta:
What's the big deal if it makes no difference?

thegrb93:
It actually does make a difference by increasing number of ops. I don't like it either.

Bo98:
I just copy and pasted what we decided. I can easily switch the order you want.

Bo98:
Done.

PR #1156, opened 2016-04-16
Add ordering for hooks.

thegrb93:
Not tested enough, but putting here for suggestions and review.

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -3,10 +3,13 @@ local pairs                        = pairs
 local isfunction        = isfunction
 local isstring                = isstring
 local IsValid                = IsValid
+local tinsert                = table.insert
+local tremove               = table.remove
 
 module( "hook" )
 
 local Hooks = {}
+local HooksOrdered = {}
 
 --[[---------------------------------------------------------
     Name: GetTable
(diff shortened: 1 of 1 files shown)


neico:
I'd rather call it priority, order sounds strange in this context...

I'd probably also add a optional argument for GetTable to get the Priority table as well (or merged, whatever looks better I guess)

I feel this might add some overheat to normal hooks though, so you might want to provide some benchmarks and stress tests and make sure that the priority is only accounted for if there's actually a need for it...

Ideally I'd use the Hooks table and extend it to allow optional fields per hook (which would make adding additional features to the hooks later on easier, or even customizeable behavior, ULX comes to mind) and avoid creating another which basically doubles the memory usage as it's pretty much a copy with added priority numbers.

But I guess that should probably handled by a different PR and have a grace period as there surely are scripts that rely on the current Hook table structure...

BillyOnWiiUIn144p:
Regardless of whether this PR is good or not, this is seriously needed.

thegrb93:
Can't exclude hooks that don't need it because the ones that do need it need to be called before or after the normal ones. If you have a better idea about the structure of the tables without messing up GetTable then let me know.

thegrb93:
New commit, not tested but I uploaded it for feedback. I made normal hooks unaffected so there's no overhead for normal hooks. I'll finish with this idea if it looks good.

willox:
I prefer how ULX does it without any magic numbers. This also seems like a lot of code

aStonedPenguin:
You could modify my implementation of the hook lib here to do this and at the same time even make the hook lib faster by removing the pairs loop.
https://github.com/SuperiorServers/plib_v2/blob/master/lua/plib/libraries/hook.lua

thegrb93:
You can add enums for numbers if you want. It's much more efficient this way since there aren't 2 nested tables.

thegrb93:
And yeah, there's a lot of code because I had to basically write my own ordered sets.

thegrb93:
Should I test and polish this implementation or is there something else you have in mind? I could add an ordered set class with O(log(n)) if needed.

thegrb93:
If it was a skiplist structure, I think insertions and deletions are safe while iterating too.

willox:
If anything is merged at all, it'll have to be 100% compatible with ULX's hook library. There's probably a lot you can simplify with that in mind.

thegrb93:
I'll look at it and make sure of that.

thegrb93:
I'll need to restrict the hook order values to -2, -1, 0, 1, and 2 for that to work. Is there any chance we can get a skiplist class? would make things much cleaner.

StidOfficial:
I need priority order for hook. On my gamemode !

PR #1158, opened 2016-04-19
UV support for killicons

AwfulRanger:
As far as I know, something like this is necessary to use TF2 killicons without extracting and modifying them, [due to the way they're stored in a big image](http://i.imgur.com/fx0zlor.png).

This might not be the best way to do it though, but it does work.

### DIFF
--- a/garrysmod/lua/includes/modules/killicon.lua
+++ b/garrysmod/lua/includes/modules/killicon.lua
@@ -37,7 +37,7 @@ function Add( name, material, color )
 
 end
 
-function AddUV( name, material, color, x, y, w, h )
+function AddTexCoord( name, material, color, x, y, w, h )
 	
 	Icons[name] = {}
 	Icons[name].type		= TYPE_MATERIAL
--- a/garrysmod/lua/includes/modules/killicon.lua
+++ b/garrysmod/lua/includes/modules/killicon.lua
@@ -5,7 +5,7 @@
 local surface	= surface
 local Msg		= Msg
 local Color		= Color
-
+local math		= math
 
 --[[---------------------------------------------------------
    Name: killicon
@@ -16,6 +16,7 @@ module("killicon")
 local Icons = {}
 local TYPE_FONT 	= 0
 local TYPE_TEXTURE 	= 1
+local TYPE_MATERIAL	= 2
 
 function AddFont( name, font, character, color )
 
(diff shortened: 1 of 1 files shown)


robotboy655:
I don't see any reason for the function to take UV coordinates over pixel values. You open the texture in a image editing program, select the rectangle you want and copy/paste the pixel values.

I think the name is fine, unless you have a better name for it. As for the `surface.GetTextureSize` issue, I have added info about `.png` materials to its wiki page that should help you.

GitSparTV:
PR is awaiting approval. What is missing to be merged?

garrysmod/lua/includes/modules/killicon.lua
```diff
@@ -81,7 +95,15 @@ function GetSize( name )
...
-		local tw, th = surface.GetTextureSize( t.texture )
+		local tw, th
+		
+		if ( t.type == TYPE_TEXTURE ) then
+			tw, th = surface.GetTextureSize( t.texture )
+		elseif ( t.type == TYPE_MATERIAL ) then
+			tw = t.w
+			th = t.h
+		end
+
```
robotboy655, line 106: This code causes the resulting icon to render bigger than specified in the `AddUV` function.

Tested using this:
```
killicon.AddUV( "monster_alien_slave", "hud/dneg_images", color_white, 193, 449, 60, 30 )
killicon.AddUV( "worldspawn", "killicons.png", color_white, 193, 449, 60, 30 )
```
killicons.png is the file linked in the OP.

```
if ( t.type == TYPE_MATERIAL ) then
	w = t.w
	h = t.h
end
```
Something like this works better (instead of `if ( t.type == TYPE_TEXTURE || t.type == TYPE_MATERIAL ) then`)

The default killicon rendering in Base gamemode has its own vertical alignment issues that I will need to look into resolving.
Kinda sad that the killicons offset themselves for no reason in the `killicon.Draw` function, instead of doing that in the base gamemode.

robotboy655 (changes requested):

garrysmod/lua/includes/modules/killicon.lua
robotboy655: In hindsight a more appropriate name for this func would be `AddTexCoord`

AwfulRanger: > This code causes the resulting icon to render bigger than specified in the `AddUV` function.

This is how killicon.Add would handle a 60x30 texture too, though, and I can't imagine someone specifying a size of 512x512, for example, would actually want it drawn at that size.

`killicon.Add( "worldspawn", mat:GetName(), color_white )`
`killicon.AddTexCoord( "worldspawn", mat:GetName(), color_white, 0, 0, mat:Width(), mat:Height() )`

These two produce the same result, are you sure they shouldn't? If so I'll change it.

PR #1160, opened 2016-04-21
Prevent os.date from crashing everything

stuartpb:
As described on [the wiki](https://wiki.garrysmod.com/page/os/date), format specifiers not supported by the underlying C runtime will crash the game on Windows.

This makes it so any format string to os.date that includes a non-standard specifier will have that specifier converted to plaintext.


aStonedPenguin:
pointless.

Kefta:
Should be fixed in the Lua base, really

PR #1161, opened 2016-04-22
Removed SWEP.Precache

Kefta:
http://wiki.garrysmod.com/page/WEAPON/Precache


robotboy655:
On second though, this might cause Lua errors for people who call this manually since this existed for so long.

Kefta:
Looking through the Github instances, I can't find any where people manually call SWEP.Precache -- it used to be called automatically anyway. Also, even if they manually called it, they must have overrided it right? Or else they'd be calling an empty method.

robotboy655:
I guess we shall see. SWEP.Precache was never automatically called, the hook is a myth and never actually existed.

PR #1165, opened 2016-04-24
Server List layout (possible performance improvement)

DarthTealc:
Modify the CSS of the server list to possibly make a small performance improvement by reducing the amount of "absolute" positioned elements.

Also now trims the server name and map name with "..." instead of directly cutting it off.


MrUnknownGamer:
Kinda glitched 
cc4107098c92092adf009f8a265f1077

PR #1166, opened 2016-04-24
faster string.Explode and string.Replace

CapsAdmin:
I can't test this as gmod doesn't work here but I have tested this outside of gmod and I'm sure the method works. It's about 10 times faster.

There's no need for string.PatternSafe with this method as string.find's third argument can specify whether to do a plain or pattern based search.

I could use `while true do` but `for i = 1, math.huge do ` gives me a free `i`. Now that I'm writing this maybe it could be limited to `string.len(str)` as it would definitely not exceed that.

So I'd like some feedback before this gets merged (obviously) but especially since I can't test this in gmod myself.

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -76,28 +76,26 @@ end
 -----------------------------------------------------------]]
 local totable = string.ToTable
 local string_sub = string.sub
-local string_gsub = string.gsub
-local string_gmatch = string.gmatch
+local string_find = string.find
 function string.Explode(separator, str, withpattern)
 	if (separator == "") then return totable( str ) end
-	 
-	local ret = {}
-	local index,lastPosition = 1,1
-	 
-	-- Escape all magic characters in separator
-	if not withpattern then separator = separator:PatternSafe() end
-	 
-	-- Find the parts
-	for startPosition,endPosition in string_gmatch( str, "()" .. separator.."()" ) do
-		ret[index] = string_sub( str, lastPosition, startPosition-1)
-		index = index + 1
-		 
-		-- Keep track of the position
-		lastPosition = endPosition
+
+	if withpattern == nil then
+		withpattern = false
 	end
-	 
-	-- Add last part by using the position we stored
-	ret[index] = string_sub( str, lastPosition)
+
+	local ret = {}
+	local last_pos = 1
+
+	for i = 1, math.huge do
+		local start_pos, end_pos = string_find(str, separator, last_pos, not withpattern)
+		if not start_pos then break end
+		ret[i] = string_sub(str, last_pos, start_pos - 1)
+		last_pos = start_pos + (end_pos - start_pos) + 1
+	end
+
+	ret[#ret + 1] = string_sub(str, last_pos)
+
 	return ret
 end
 
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -237,7 +237,7 @@ function string.Replace( str, tofind, toreplace )
 		return table.concat(tbl, toreplace)
 	end
 
-	return self
+	return str
 end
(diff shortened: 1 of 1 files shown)


aStonedPenguin:
Using

```lua
local str = 'awdawd;ad;awdadawd;awdawd;adawd;awdad;awd;'
local sep = ';'

require 'xbench' -- https://github.com/SuperiorServers/plib_v2/blob/master/lua/plib/libraries/xbench.lua

xbench.Compare({
	Old = function()
		string.Explode(str, sep)
	end,
	New = function()
		string.Explode2(str, sep)
	end,
}, 100000)
```

I got:
OLD: 0.2296575782384
NEW: 0.017039076028595

CapsAdmin:
It's slightly faster now. Could you do the same test agian?

<sub>also xbench.Compare is a neat idea</sub>

Kefta:
OLDPATTERN: 0.049061595078456
OLD: 0.43176123587452
NEWPATTERN: 0.048394999406099
NEW: 0.035022127699676

CapsAdmin:
On a random generated 10000000 word string (56mb) using explode method takes 0.065 sec while the gsub method takes 0.8 sec.

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -232,9 +231,13 @@ end
...
-	tofind = tofind:PatternSafe()
-	toreplace = toreplace:gsub( "%%", "%%%1" )
-	return ( str:gsub( tofind, toreplace ) )
+	local tbl = string.Explode(tofind, str)
+
+	if tbl[1] then
+		return table.concat(tbl, toreplace)
+	end
+
+	return self
```
bigdogmat: Shouldn't this return str

CapsAdmin: ops

CapsAdmin: not sure if i should copy the code from explode or use explode. It's neater this way but idk if the performance benefits from doing that are worth it.

Rusketh:
Why has this not made it in yet. Merge Garry, Merge!

Kefta:
@Rusketh Probably because it's still on RB's and Willox's testing list. Garry doesn't handle GMod anymore.

robotboy655:
Ehh, lets see how much I will regret merging this...

aStonedPenguin:
@robotboy655 I've been running it on my servers since this PR, seems to work fine.

Divran:
Garry's original explode function was replaced by another person years ago. When that happened, we discovered that wire's adv duplicator would crash the entire server. We narrowed it down to a place in adv dupe's code when it loads the dupe file, it would run string.Explode on the file, with only 1 single explode point near the center (of a very large file). The replacement explode function couldn't handle this use case, and so the server would crash.

TomyLobo made the explode function that has been in gmod ever since (the one that this PR replaces) because his was able to handle every single use case we could throw at it with about the same performance (which was 1000s of times faster than garry's in most cases).

If yours is better, that's great and it should definitely be merged and used. I just want us to make sure that it works for every use case. Specifically the one I mentioned earlier (I'd test it right now but it's late and I'm tired af).

CapsAdmin:
I tried splitting a 50mb string in 2 and it doesn't crash. I'm not sure why the previous one would crash either. Out of memory or ran for too long? The max amount of iterations here is going to be the length of the string which should be safer than the previous method.

If you can throw your use cases at the new function and report back that'd be great. 

This version is faster because it doesn't use string.gmatch which is internally a very complex function (loops and branchy) and is not a jit compilable function. The use case for string.explode tends to be without patterns so theres the string.PatternSafe which uses string.gsub and is inserting escape patterns into the string. For larger strings this can become memory intensive and slow.

TomyLobo:
I managed to find the old (2011) thread with Divran's test cases:
https://facepunch.com/showthread.php?t=1062458

PR #1167, opened 2016-04-25
Replaced SortedPairs* and RandomPairs with better implementations

FredyH:
No deep copies of tables are made anymore
Fixed a crash that could occur when a table is self referencing


garrysmod/lua/includes/extensions/table.lua
```diff
@@ -466,135 +466,66 @@ function table.ClearKeys( Table, bSaveKey )
...
-	Desc = Desc or false
-
-	local pSortedTable = table.ClearKeys( pTable, true )
-
-	table.SortByMember( pSortedTable, pValueName, !Desc )
-
-	local SortedIndex = {}
-	for k, v in ipairs( pSortedTable ) do
-		table.insert( SortedIndex, v.__key )
+	local sortedTbl = toKeyValues(pTable, pValueName)
```
bigdogmat: `toKeyValues` doesn't take a second argument, guessing this was left over.

bigdogmat:
You need to put spaces around arguments, between arguments and parentheses, and inside brackets.

bmwalters:
Worth noting that there is discussion going on in #1273 about how to best write these functions.

syl0r:
I don't think your versions are more readable to be honest. You have to use a table sorting function in each Sorted* function. I think that adds a lot of redundancy as well as confusion (sorting functions are ugly and relatively hard to understand), while mine reuses table.SortByMember.

I fixed the code formatting as @bigdomat has said even though it really hurt (thanks garry for that great coding style that no other language uses).

Either way, I think one of the two changes should be merged soonish as the current implementation of SortedPairs can cause crashes.

robotboy655:
I have looked into this and all the methods save for SortedPairsByMemberValue are noticeably slower than their current counter parts in the live game.

FredyH:
Can you please produce your benchmark code?
I just benchmarked the simple SortedPairs replacement and got the exact opposite (new one is more than 10 times faster).

Here is my code:
```LUA
local function keyValuePairs(state)
	state.Index = state.Index + 1
	local keyValue = state.KeyValues[state.Index]
	if(!keyValue) then return end
	return keyValue.key, keyValue.val
end

local function toKeyValues(tbl)
	local result = {}
	for k,v in pairs( tbl ) do
		table.insert( result, { key = k, val = v } )
	end
	return result
end

function ImprovedSortedPairs( pTable, Desc )
	local sortedTbl = toKeyValues( pTable )
	table.SortByMember( sortedTbl, "key", !Desc )
	return keyValuePairs, { Index = 0, KeyValues = sortedTbl }
end

local testTable = {}
for i = 1, 10000 do
	testTable[math.random(1, 1000000000)] = i
end


function benchmarkImprovedPairs()
	local sum = 0
	local startTime = SysTime()
	for i = 1, 10 do
		for k, v in ImprovedSortedPairs(testTable) do
			sum = sum + v
		end
	end
	print("New: ", SysTime() - startTime, sum)
end

function benchmarkOldPairs()
	local sum = 0
	local startTime = SysTime()
	for i = 1, 10 do
		for k, v in SortedPairs(testTable) do
			sum = sum + v
		end
	end
	print("Old: ", SysTime() - startTime, sum)
end
benchmarkImprovedPairs()
benchmarkOldPairs()
```

EggrollWireson:
Рубат в гмод коммунити это как дятел долбящий фонарный столб - вроде долбит, но без толку

bmwalters:
Nice @FredyH 🎉  I do have to point out that now we're back to having table sorting functions though ;)

FredyH:
Yes I have to admit that it's quite unfortunate. The only reason I had to do that though is because table.SortByMemberValue is extremely slow, mainly due to the type() check in each comparison operation as well as the many many ifs inside each comparison.

I benchmarked it and the type call alone makes it at least ~5 times slower than without. This only matters if you have very small tables though, since at some point the O(n * log(n)) complexity of my version outshines the O(n^2) complexity of the original one by far.

PR #1168, opened 2016-05-01
Major optimization to start round and WEPS.ForcePrecache

gspetrou:
Moved WEPS.ForcePrecache to GM.InitPostEntity, switched it's loop from pairs to
ipairs because it is faster and weapons.GetList is sequential, combined two duplicate if statements, and fixed a syntactical issue that really bothered me.

Overall made BeginRound() go from taking about **0.80s** to execute to **0.27s**. This practically removed the little lag at the start of each round.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
@@ -486,12 +486,14 @@ end
 -- non-cheat developer commands can reveal precaching the first time equipment
 -- is bought, so trigger it at the start of a round instead
 function WEPS.ForcePrecache()
-   for k, w in pairs(weapons.GetList()) do
-      if w and w.WorldModel then
-         util.PrecacheModel(w.WorldModel)
-      end
-      if w and w.ViewModel then
-         util.PrecacheModel(w.ViewModel)
+   for k, w in ipairs(weapons.GetList()) do
+      if w then
+         if w.WorldModel then
+            util.PrecacheModel(w.WorldModel)
+         end
+         if w.ViewModel then
+            util.PrecacheModel(w.ViewModel)
+         end
       end
    end
 end
--- a/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
@@ -487,13 +487,11 @@ end
 -- is bought, so trigger it at the start of a round instead
 function WEPS.ForcePrecache()
    for k, w in ipairs(weapons.GetList()) do
-      if w then
-         if w.WorldModel then
-            util.PrecacheModel(w.WorldModel)
-         end
-         if w.ViewModel then
-            util.PrecacheModel(w.ViewModel)
-         end
+      if w.WorldModel then
+         util.PrecacheModel(w.WorldModel)
+      end
+      if w.ViewModel then
+         util.PrecacheModel(w.ViewModel)
       end
    end
 end
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -230,6 +230,10 @@ function GM:InitCvars()
    self.cvar_init = true
 end
 
+function GM:InitPostEntity()
+   WEPS.ForcePrecache()
+end
+
 function GM:GetGameDescription() return self.Name end
 
 -- Convar replication is broken in gmod, so we do this.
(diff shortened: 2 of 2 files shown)


Kefta:
Test with a flat loop (i = 1, #tbl). LuaJIT probably already takes care of this, but I wanna make sure

zerfgog:
@Kefta who tf calls that a flat loop

gspetrou:
@Kefta Numeric was barely faster, here are the results if you're curious, though they aren't really that special.

Profiler Results: [http://i.imgur.com/q6MdFyR.png](Profiler Results)

```Lua
-- ipairs
function WEPS.ForcePrecache()
   for k, w in ipairs(weapons.GetList()) do
      if w then
         if w.WorldModel then
            util.PrecacheModel(w.WorldModel)
         end
         if w.ViewModel then
            util.PrecacheModel(w.ViewModel)
         end
      end
   end
end

-- Numeric
function WEPS.ForcePrecache()
   local weplist = weapons.GetList()
   for i = 1, #weplist do
      local w = weplist[i]
      if w.WorldModel then
         util.PrecacheModel(w.WorldModel)
      end
      if w.ViewModel then
         util.PrecacheModel(w.ViewModel)
      end
   end
end
```

Kefta:
Thanks for testing, up to @svdm if he wants to switch or keep the style consistant

Kefta:
@zerfgog That's what I saw that loop labelled as when I first read it in code -- thanks for the correction

Bo98:
> from pairs to ipairs because it is faster

A few weeks ago I was told the opposite was true. Is it inconsistent for everyone or something?

Bo98:
Also, this was already suggested before and rejected due to an information leak issue: #1031.

bigdogmat:
> A few weeks ago I was told the opposite was true. Is it inconsistent for everyone or something?

I can't do the testing at the moment, but from what I remember pairs is only faster on arrays with hundreds of thousands of elements

gspetrou:
That doesn't really make sense though. If I recall correctly the weapons cache isn't reset at the start of every round. Once a weapon is cached it is cached till either a map change happens or the server runs the flush command.

aStonedPenguin:
> A few weeks ago I was told the opposite was true. Is it inconsistent for everyone or something?
Before lua jit it was in some cases. There was an old thread on facepunch people continually reference and spread misinformation with. It's been proven many times ipairs is faster then pairs and a for loop is as fast as ipairs and sometimes faster.

Bo98:
Yeah that's what I thought. IIRC ipairs is JIT compiled while pairs isn't.

gspetrou:
I forgot to localize ipairs in my ipairs test so here it is again
[Corrected ipairs test](http://i.imgur.com/DC085a5.png)

TheEMP:
@gspetrou  Read https://github.com/garrynewman/garrysmod/pull/1031 it outlines how this pull request could create exploitable behavior.

gspetrou:
@TheEMP I have tried using "developer 1". This command does not display when a model has been cached/precached. All it seems to do it display the console on the top left of the screen. I can't seem to find any other way than the command I mentioned previously to see what models are being cached.

robotboy655:
Precache all weapon models on first player join clientside = profit?

garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
```diff
@@ -486,12 +486,14 @@ end
...
 -- is bought, so trigger it at the start of a round instead
 function WEPS.ForcePrecache()
-   for k, w in pairs(weapons.GetList()) do
-      if w and w.WorldModel then
-         util.PrecacheModel(w.WorldModel)
-      end
-      if w and w.ViewModel then
-         util.PrecacheModel(w.ViewModel)
+   for k, w in ipairs(weapons.GetList()) do
+      if w then
```
bigdogmat: Why would the state of w even need to be checked in this case, ipairs is gonna return when it hits the first nil value anyway

Kefta:
@gspetrou There's no way, unfortunately. The engine does it by directly checking the precache list.

gspetrou:
@robotboy655 Doesn't seem like this would be necessary. The "model_list", "cl_precacheinfo", and "listmodels" commands didn't leak any info when precaching once serverside at InitPostEntity. This is after testing through multiple in-game rounds.

robotboy655:
@gspetrou I am going to assume you tested it on local mutliplayer or singleplayer, where the serverside precaching also forced your client to precache the models. Try starting an srcds instance and testing then.

gspetrou:
@Robotboy655 this was all tested on a dedicated server.

svdm:
I'll try and test this myself later this week.

svdm:
From my testing on my local dedicated server, this does indeed do the job. I'd guess that a list of models that have been precached so far is sent from server to clients when they join, meaning a single precache serverside is sufficient to cause all clients to precache even if they join later on.

Kefta:
That's good to know; I'll have to implement this into my weapons as well. Thanks @svdm @gspetrou

gspetrou:
Hip hip, hooray!

PR #1169, opened 2016-05-02
Fix indexing nil field ''ActivityTranslate'

stickz:
```
XXX|137|STEAM_0:0

[ERROR] gamemodes/base/entities/weapons/weapon_base/sh_anim.lua:81: attempt to index field 'ActivityTranslate' (a nil value)
  1. unknown - gamemodes/base/entities/weapons/weapon_base/sh_anim.lua:81
   2. TranslateWeaponActivity - [C]:-1
    3. unknown - gamemodes/base/gamemode/animations.lua:335
```


Kefta:
Sounds like an issue if SWEP.SetupWeaponHoldTypeForAI is called or not. This is not a proper fix.

robotboy655:
This would break player model animations if ```ActivityTranslateAI``` is unset, and the error shown in the description is not even related to ActivityTranslateAI.

PR #1171, opened 2016-05-09
Add SteamID32

edgarasf123:
https://github.com/Facepunch/garrysmod-issues/issues/2602
This is a potentially a replacement of Player:UnqueID() which doesn't have any collisions.

* No collisions
* Faster than Player:UniqueID
* Number type, therefor faster database queries.
* Smaller database files, since number takes less space than string version of steamid.
* Account comparison, determine which account is older. With Regression Analysis, you can determine how old profile is without querying steam API (also would work with private profiles).

Also `"U:1:" .. util.SteamIDTo32( steamid )` would give you modern steamid (used by games such as CS:GO).

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -431,3 +431,30 @@ function GetConVarString( name )
 	local c = GetConVarCached( name )
 	return ( c and c:GetString() ) or ""
 end
+
+--[[---------------------------------------------------------
+   Name: SteamIDTo32( steamid )
+   Desc: Given a STEAM_0 style Steam ID will return a 32bit Steam ID
+-----------------------------------------------------------]]
+function util.SteamIDTo32( steamid )
+
+	steamid = tostring( steamid )
+	local y, z = string.match( steamid, "STEAM_0:(%d):(%d+)" )
+
+	return z * 2 + y
+	
+end
+
+--[[---------------------------------------------------------
+   Name: SteamIDFrom32( steamid32 )
+   Desc: Given a 32bit Steam ID will return a STEAM_0 style Steam ID
+-----------------------------------------------------------]]
+function util.SteamIDTo32( steamid32 )
+	
+	steamid32 = tonumber( steamid32 )
+	local y = steamid32 % 2
+	local z = ( steamid32 - y ) / 2
+
+	return "STEAM_0:" .. y .. ":" .. z
+
+end
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -439,7 +439,8 @@ end
 function util.SteamIDTo32( steamid )
 
 	steamid = tostring( steamid )
-	local y, z = string.match( steamid, "STEAM_0:(%d):(%d+)" )
+	local y, z = 0, 0
+	y, z = string.match( steamid, "STEAM_0:(%d):(%d+)" )
 
 	return z * 2 + y
 	
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -449,7 +449,7 @@ end
    Name: SteamIDFrom32( steamid32 )
    Desc: Given a 32bit Steam ID will return a STEAM_0 style Steam ID
 -----------------------------------------------------------]]
-function util.SteamIDTo32( steamid32 )
+function util.SteamIDFrom32( steamid32 )
 	
 	steamid32 = tonumber( steamid32 )
 	local y = steamid32 % 2
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -107,7 +107,8 @@ end
 
 function meta:SteamID32()
 
-	local y,z = string.match( self:SteamID(), "STEAM_0:(%d):(%d+)" )
+	local y, z = 0, 0
+	y,z = string.match( self:SteamID(), "STEAM_0:(%d):(%d+)" )
 
 	return z * 2 + y
 
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -88,6 +88,31 @@ function player.GetBySteamID64( ID )
 
 end
 
+function player.GetBySteamID32( ID )
+
+	ID = tonumber( ID )
+
+	for _, pl in pairs( player.GetAll() ) do
+
+		if ( pl:SteamID32() == ID ) then
+			return pl
+		end
+
+	end
+
+	return false
+
+end
+
+
+function meta:SteamID32()
+
+	local y,z = string.match( self:SteamID(), "STEAM_0:(%d):(%d+)" )
+
+	return z * 2 + y
+
+end
+
 --[[---------------------------------------------------------
 	Name: DebugInfo
 	Desc: Prints debug information for the player


garrysmod/lua/includes/util.lua
```diff
@@ -431,3 +431,30 @@ function GetConVarString( name )
...
+
+	return z * 2 + y
+	
+end
+
+--[[---------------------------------------------------------
+   Name: SteamIDFrom32( steamid32 )
+   Desc: Given a 32bit Steam ID will return a STEAM_0 style Steam ID
+-----------------------------------------------------------]]
+function util.SteamIDTo32( steamid32 )
```
fruitwasp: Shouldn't this be .SteamIDFrom32?

zerfgog:
Valve developer wiki calls this "account ID" fyi

neico:
STEAM_0 is not a constant: https://developer.valvesoftware.com/wiki/SteamID#Legacy_Format

Besides, your code doesn't cover the PENDING, LAN and UNKOWN cases (which will make the function return pretty much garbage)

You might want to look trough this class for other possible conversion methods: https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/steam/steamclientpublic.h#L403 (like mentioned above, it's essentially AccountID)

I'd probably rather want to get CSteamID exposed then all this gluttered up ID slicing...

thegrb93:
Wouldn't an 4 byte string be more optimal than a 64bit lua double?

edgarasf123:
Yeah, I know its AccountID. But SteamID64 is also called CommunityID which is a representation of steam account. AccountID is a 32 bit representation of steam account.
The difference is, that The other 32bit in CommunityID for `Individual` steam accounts is constant.

Steam32 is shorter and easier memory unlike `util.SteamIDFromAccountID`

I would also prefer if this implemented in C code, without ID slicing

Kefta:
ipairs loop pls

edgarasf123:
I copied the code from `player.GetBySteamID64`. And it seems [all loops with player.GetAll()](https://github.com/garrynewman/garrysmod/search?utf8=%E2%9C%93&q=player.GetAll%28%29&type=Code) uses pairs instead of ipairs, maybe there is a reason behind (like NULL entities).

bigdogmat:
> And it seems all loops with player.GetAll() uses pairs instead of ipairs, maybe there is a reason behind (like NULL entities).

player.GetAll is updated whenever a player entity is create or deleted, there shouldn't (wouldn't?) be a
NULL entity in it. Either way pairs vs ipairs wouldn't have any effect, just keeping consistency with the code base.

thegrb93:
@Kefta What I'm saying is that there should be an internal function that does this.

edgarasf123:
There is `CSteamID::GetAccountID()`
https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/public/steam/steamclientpublic.h#L722

Kefta:
@thegrb93 ipairs is faster -- also, the indices are all numbers meaning you should do ipairs.

bigdogmat:
@Kefta I never said it shouldn't be used, I stated the only reason it would be used is for consistency. 

Also ipairs requires they be sequential as well, though player.GetAll is, so that doesn't matter.

Kefta:
@bigdogmat ipairs is a default Lua function and it's used in the code-base: https://github.com/garrynewman/garrysmod/search?utf8=%E2%9C%93&q=ipairs

I don't see where the consistency lacks.

Benchmark for proof for anyone who thinks pairs is still faster:
```
] lua_run local all = player.GetAll local pairs = pairs local ipairs = ipairs require'xbench' xbench.Compare( { pairs = function() for _,_ in pairs(all()) do end end, ipairs = function() for _,_ in ipairs(all()) do end end }, 10000)
> local all = player.GetAll local pairs = pairs local ipairs = ipairs require'xbench' xbench.Compare( { pairs = function() for _,_ in pairs(all()) do end end, ipairs = function() for _,_ in ipairs(all()) do end end }, 10000)...
PAIRS: 0.055148159961618
IPAIRS: 0.043567635149202
] lua_run local all = player.GetAll local pairs = pairs local ipairs = ipairs require'xbench' xbench.Compare( { pairs = function() for _,_ in pairs(all()) do end end, ipairs = function() for _,_ in ipairs(all()) do end end }, 1000000)
> local all = player.GetAll local pairs = pairs local ipairs = ipairs require'xbench' xbench.Compare( { pairs = function() for _,_ in pairs(all()) do end end, ipairs = function() for _,_ in ipairs(all()) do end end }, 1000000)...
PAIRS: 4.6926824435681
IPAIRS: 4.457563333475
] lua_run local all = player.GetAll local pairs = pairs local ipairs = ipairs require'xbench' xbench.Compare( { pairs = function() for _,_ in pairs(all()) do end end, ipairs = function() for _,_ in ipairs(all()) do end end }, 10000000)
> local all = player.GetAll local pairs = pairs local ipairs = ipairs require'xbench' xbench.Compare( { pairs = function() for _,_ in pairs(all()) do end end, ipairs = function() for _,_ in ipairs(all()) do end end }, 10000000)...
PAIRS: 47.992726013251
IPAIRS: 44.78491756831
```

Kefta:
@edgarasf123 I'll submit a PR to change that

Kefta:
@edgarasf123 Localise pairs, os.clock, and ipairs for scope consistency in benchmarking. Also, always use SysTime -- it's the most accurate.

willox:
It's be nice if there was a way to block keywords like 'faster' on GitHub

bigdogmat:
@Kefta When did I say pairs is faster? And I said code base consistency because all player.GetBy* functions use pairs

zerfgog:
I still say that the function names should be "AccountID" as that's what literally everything else that deals with SteamIDs uses.

Kefta:
@bigdogmat I was talking to thegrb and edgar.
```
local player_GetAll = player.GetAll
local pairs = pairs
local ipairs = ipairs
local SysTime = SysTime

local stime = SysTime()
for i = 1, 10^7 do
	for _, _ in pairs( player_GetAll() ) do
	end
end
print( "pairs", SysTime() - stime )

stime = SysTime()
for i = 1, 10^7 do
	for _, _ in ipairs( player_GetAll() ) do
	end
end
print( "ipairs", SysTime() - stime )

stime = SysTime()
for i = 1, 10^7 do
	for i = 1, #player_GetAll() do
	end
end
print( "numeric", SysTime() - stime )
```

Running script loopbench.lua...
pairs		36.646346264808
ipairs	35.587476762409
numeric	35.901508709073

10^6:
Running script loopbench.lua...
pairs	1.49295410131
ipairs	1.4181309783518
numeric	1.3778134060485

10^5:
Running script loopbench.lua...
pairs	0.15523939518522
ipairs	0.14479163093461
numeric	0.13544718827069

edgarasf123:
You're are probably running this when you're alone on the server. Try benchmarking pairs/ipairs with tables that have more than 1 element.

zerfgog:
instead of calling player_GetAll each time do
```lua
local players = {}
for i = 1, 32 do players[i] = LocalPlayer()
```
then test

Kefta:
@edgarasf123 That was with 19 bots -- so 20 elements total.

Kefta:
@zerfgog 

10^6:
] lua_openscript loopbench.lua
Running script loopbench.lua...
pairs		0.1532887430991
ipairs	0.057081660644826
numeric	0.044047618438185
] lua_openscript loopbench.lua
Running script loopbench.lua...
pairs		0.15382581240044
ipairs	0.048554341532054
numeric	0.044182269750081
] lua_openscript loopbench.lua
Running script loopbench.lua...
pairs		0.15412916183914
ipairs	0.048478056187946
numeric	0.044899556777068

thegrb93:
JIT is probably converting ipairs into a normal for loop. Normal for loop is fastest.

thegrb93:
Try it on a non-numeric indexed table

Kefta:
@thegrb93 It won't matter -- ipairs/numeric loops doesn't go through non-sequential numbered values. That's why I'm arguing ipairs to be used for player.GetAll since it always has sequential numeric keys

thegrb93:
Use a for loop. At least then it is fast without JIT.

willox:
http://wiki.garrysmod.com/page/Player/AccountID

PR #1173, opened 2016-05-11
Update animations.lua

jackson-c:
Change to velocity squared for optimization

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/animations.lua
+++ b/garrysmod/gamemodes/base/gamemode/animations.lua
@@ -14,7 +14,7 @@ function GM:HandlePlayerJumping( ply, velocity )
 
 			ply.m_fGroundTime = CurTime()
 			
-		elseif ( CurTime() - ply.m_fGroundTime ) > 0 && velocity:Length2D() < 0.5 then
+		elseif ( CurTime() - ply.m_fGroundTime ) > 0 && velocity:Length2DSqr() < 0.25 then
 
 			ply.m_bJumping = true
 			ply.m_bFirstJumpFrame = false
@@ -54,7 +54,7 @@ function GM:HandlePlayerDucking( ply, velocity )
 
 	if ( !ply:Crouching() ) then return false end
 
-	if ( velocity:Length2D() > 0.5 ) then
+	if ( velocity:Length2DSqr() > 0.25 ) then
 		ply.CalcIdeal = ACT_MP_CROUCHWALK
 	else
 		ply.CalcIdeal = ACT_MP_CROUCH_IDLE
@@ -93,7 +93,7 @@ end
 
 function GM:HandlePlayerVaulting( ply, velocity )
 
-	if ( velocity:Length() < 1000 ) then return end
+	if ( velocity:LengthSqr() < 1000000 ) then return end
 	if ( ply:IsOnGround() ) then return end
 
 	ply.CalcIdeal = ACT_MP_SWIM
@@ -184,11 +184,11 @@ end
 -----------------------------------------------------------]]
 function GM:UpdateAnimation( ply, velocity, maxseqgroundspeed )
 
-	local len = velocity:Length()
+	local lensqr = velocity:LengthSqr()
 	local movement = 1.0
 
-	if ( len > 0.2 ) then
-		movement = ( len / maxseqgroundspeed )
+	if ( lensqr > 0.04 ) then
+		movement = ( math.sqrt(lensqr) / maxseqgroundspeed )
 	end
 
 	local rate = math.min( movement, 2 )
@@ -196,7 +196,7 @@ function GM:UpdateAnimation( ply, velocity, maxseqgroundspeed )
 	-- if we're under water we want to constantly be swimming..
 	if ( ply:WaterLevel() >= 2 ) then
 		rate = math.max( rate, 0.5 )
-	elseif ( !ply:IsOnGround() && len >= 1000 ) then
+	elseif ( !ply:IsOnGround() && lensqr >= 1000000 ) then
 		rate = 0.1
 	end
 
@@ -302,8 +302,8 @@ function GM:CalcMainActivity( ply, velocity )
 
 	else
 
-		local len2d = velocity:Length2D()
-		if ( len2d > 150 ) then ply.CalcIdeal = ACT_MP_RUN elseif ( len2d > 0.5 ) then ply.CalcIdeal = ACT_MP_WALK end
+		local len2dsqr = velocity:Length2DSqr()
+		if ( len2dsqr > 22500 ) then ply.CalcIdeal = ACT_MP_RUN elseif ( len2dsqr > 0.25 ) then ply.CalcIdeal = ACT_MP_WALK end
 
 	end
 


garrysmod/gamemodes/base/gamemode/animations.lua
```diff
@@ -184,19 +184,19 @@ end
...
 function GM:UpdateAnimation( ply, velocity, maxseqgroundspeed )
 
-	local len = velocity:Length()
+	local lensqr = velocity:LengthSqr()
 	local movement = 1.0
 
-	if ( len > 0.2 ) then
-		movement = ( len / maxseqgroundspeed )
+	if ( lensqr > 0.04 ) then
+		movement = ( math.sqrt(lensqr) / maxseqgroundspeed )
```
zerfgog: how is this more optimized

aStonedPenguin: the more expensive function doesn't always get called with his change there

jackson-c:
The functions contained in this file are called every tick for every player on the server, and it's by far the most expensive LUA operation from what I can see. I'm gonna go back over it and try to find some better optimization. (this commit only improves by a few %)

willox:
The real issue in this file is GetFlexName iirc

jackson-c:
You've already isolated it? If GetFlexName is slow you could totally cache it's result in a table, then clear the table if the player changes model.

willox:
I haven't looked at the code, but I'm pretty sure it's just caused by all the strings being pushed to Lua (it's surprisingly slow). I'll probably add a method to find the index from the flex name, rather than searching through them all as we currently do.

jackson-c:
The flexes are only used in MouthMoveAnimation (in that file at least), and that's only run clientside. I'm focusing on server optimization.

(while you're at it, can you add a non-copying version of player.GetAll() and ents.GetAll()?)

thegrb93:
Why'd you change how everything is indexed? pt['m_bJumping'] should be pt.m_bJumping imo.

willox:
There's no way over-the-top optimizations like this will ever be merged

Kefta:
Waaaaay too much

jackson-c:
Whenever you reference a variable on an entity using "dot notaion" (eg ply.m_bJumping), it's actually running the __index function in this file:

[https://github.com/garrynewman/garrysmod/blob/451b4ff5d1aea7b9b06a8024ef706c248a79647e/garrysmod/lua/includes/extensions/player.lua](url)

So it checks the player metatable, the entity metatable, and finally the player's unique entity table where it finally finds the value. Since the tables are indexed with strings, it has to perform three string lookups.

The same __index function is called when you use "colon notation" (eg ply:Crouching()). In this case the function is stored in the player metatable, so it is found faster. However, it still requires calling the __index function and looking up the _string_ "Crouching" in the metatable,

My approach caches the player's unique entity table at the start of each function (really, it would be best if it were passed to all the sub-functions): `local pt = A_GetTable(ply)` Then, pt is referenced directly rather than incurring the __index function every time. Also, references to all the commonly used functions are saved locally (all the A_Functions).

The result? Sure, the style is a bit weird. But it _doubles the performance_ of CalcMainActivity! From 0.03ms per call to 0.015ms per call, in my tests. You gotta realize, this function gets called _for every player, every tick_ and is by far the most expensive lua routine on the server. For a 64 player, 33 tick server this eats up **63ms** of runtime every second. I cut that in half just by changing how the player tables are indexed.

willox:
If I start up a server with 100 bots, GM:CalcMainActivity is eating up about 0.6ms/frame. How are you measuring this?

At a frequency of 66 we'll guesstimate that's about 40ms per second, so your results sound correct. But it's not the biggest issue ever, is it? 0.6ms/frame is not so much that we need to make the code look horrific.

jackson-c:
Running FProfiler on a live server. Are you running a local server on a high performance PC? If so, you're going to get better performance than an NFO VPS which is what most server owners use.

[http://i.imgur.com/gsDaDCo.png](http://i.imgur.com/gsDaDCo.png)

This is a few seconds of profiling on my server with 33 tick, and 30 players online (_with_ my changes. This takes about twice as much time without). As player count regularly gets up around 100 this code has a choking effect on server performance. But with these optimizations it greatly reduces that impact and realistically allows me to add 10 or so player slots.

Honestly, the whole system for using "dot-notation" for adding variables to entities is badly inefficient. Indexing class-global functions on the meta table and indexing entity-specific variables should not use the same code.

aStonedPenguin:
Can't we just have the ability to make our own base gamemode?

https://github.com/Facepunch/garrysmod-requests/issues/621

jackson-c:
**Okay**

I did some tests to find out how long it takes to call functions and variables on players in different ways. Averaging over 10 trials of 1000000 tests each, here are my results:

For calling a function on a player:
ply:OnGround() = **115ns**
entity.OnGround( ply ) = **80ns**
A_OnGround( ply ) = **75ns**

(where entity is the entity metatable; A_OnGround is entity.OnGround cached)

ply.variable = **160ns**
pt['variable'] = **0.5ns**
pt.variable = **0.5ns**

(where pt is ply:GetTable() cached)

So, how functions are called on players doesn't make too much of a difference. However, caching the player's entity table results in a **300 times speedup**. Check out my latest commit - I've removed the weird function caching and cleaned up the table caching.

Kefta:
There's a certain degree that you sacrifice micro-optimisation for readability -- just a widely accepted standard since what you really mean by 300x faster is in nano-seconds. Also, if you do cache these functions in your code, I would recommend using playerMetatable.functionName for storage since the entity metatable might have a different implementation than the player.

thegrb93:
Like he said though, this file contains some of the most expensive routines in gmod so this would be a place where it matters.

jackson-c:
Sure, it's a matter of nano-seconds for one execution. But you gotta multiply it out:

160ns wasted per "ply.variable" call
"ply.variable" calls are made on about ~30 times during CalcMainSequence hook (at most 53 times)
CalcMainSequence is called per player per tick, so for a 64 player 33 tick server:
160ns * 30 * 33 * 64 = 10ms

So 10ms is being completely thrown out every second with this unnecessary indexing.

> I would recommend using playerMetatable.functionName for storage since the entity metatable might have a different implementation than the player.

The player metatable does not implement a lot of functions that are shared between all entities - rather, it inherits them from the entity metatable. When __index is called on a player, it searches first through the player metatable, then the entity metatable, then the player's unique entity table.

Kefta:
I know, that's why I'm saying to localise the player metatable functions instead of the entity's when you're using it just for a player. That way, in-case the player metatable has a different implementation it will use that. This skips all of the __index calls except for the one time localisation.

Jake3231:
What's the status of this request? If it can really make a big difference, I think that we should resolve the conflicts it has, and try to get this merged.

cannon:
It does make a massive difference. Literally half the impact.

No idea why this almost invisible optimization is getting support but my massive one, which includes all its changes anyway, isn't.
#1399

Kefta:
It would just be better to optimise the entity __index. This won't be merged because the improvements come at a cost of extremely messy code. It will also not work with metatable overrides since you are caching the functions at the top.

GitSparTV:
PR #1399 was merged but it didn't optimize all things mentioned in this PR.
Also this PR fixed variable names, which is a good suggestion.

PR #1174, opened 2016-05-11
Remove "INSTANT KILL" Message for Spectators

markusmarkusz:
At the moment, if somebody is spectating a owner of a knife he could see the "INSTANT KILL" Message. And i think it shouldn't be.
With this it'll be removed. This would check if the owner of the knife is alive.

(Sorry, for my bad english. I hope you can understand what i want.)

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_knife.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_knife.lua
@@ -278,7 +278,7 @@ if CLIENT then
       local tr = self.Owner:GetEyeTrace(MASK_SHOT)
 
       if tr.HitNonWorld and IsValid(tr.Entity) and tr.Entity:IsPlayer()
-         and tr.Entity:Health() < (self.Primary.Damage + 10) then
+         and tr.Entity:Health() < (self.Primary.Damage + 10)  and self.Owner:Alive() then
 
          local x = ScrW() / 2.0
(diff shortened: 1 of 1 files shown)


svdm:
I'm not sure I agree. If someone is spectating the owner of the knife in first-person, he should see the message. Otherwise, a spectator could see someone insta-kill with a knife when the message was not shown, they could be confused about how the knife works.

In general if I spectate someone in first-person, I expect to see the same HUD that they see.

markusmarkusz:
I think it is useless to see it.
If someone will be killed with a knife I think everyone should know that it was a instant kill.
If a you kill someone with a knife it'll be automaticly a instant kill.

(Hope you can underatand this ^^.)

meharryp:
The knife isnt always and instant kill though. It does 50 damage.

markusmarkusz:
But it is an instant kill.
If the target hp are less then 50 it's either an instant kill. If the knife would kill the target it's an instant kill.
The message will be shown if the knife would kill the target. This means it is everytime a instant kill when the knife kills someone.

markusmarkusz:
Am I the only one here who thinks that this is useless.

BlackCetha:
I dont see any reason to change what a spectator sees compared to what the player they are spectating sees.
Hell, thats the point of spectating, right?

markusmarkusz:
Right, thats the point of spectating.
If no one is agreeing with me, I think the pull request could be closed.
I think that this is useless, but everyone has his own opinion :).

PR #1182, opened 2016-05-22
TTT: Replace most UniqueID usage with SteamID

zerfgog:
Replaces lots of UniqueID usage with SteamID, preserving all necessary functionality.


willox:
I can see quite a few of these changes not being compatible with certain addons

Kefta:
Fix 'em

zerfgog:
I'll find out what changes break what and reply here.

zerfgog:
- DNA Sample field `killer_uid` -> `killer_sid` change breaks nothing (on GitHub)
- SCORE.Events `uid` -> `sid` field change breaks nothing (on GitHub) (surprisingly)
- `ttt_transfer_credits` concommand data change breaks nothing (on GitHub)
- `KARMA.RememberedPlayers` index change breaks nothing (on GitHub)
- `GiveEquipmentWeapon` timer name change breaks nothing (on GitHub)

but the big one:
- ragdoll (CORPSE) `uqid` -> `sid` field change breaks:
  - TTT [portable life monitor](https://github.com/Fortune117/Old-TTT-Addons/tree/master/PLM) addon
  - [ULX TTT Corpses](https://github.com/NekuMitsurugi/ULX-TTT-Corpses) addon
  - [TTT Remote Detonator](https://github.com/Exho1/RandomLuaStuff/blob/master/lua/weapons/weapon_ttt_rdetonate.lua) addon
  - @Kefta's [Entity Crash Catcher](https://github.com/Kefta/Entity-Crash-Catcher) addon, which is no longer supported and now moot, but several servers probably still use

Now so far these breaks are fairly minor as I doubt any servers use those addons (with the exception of the last one). However there are some extremely major addons that this change breaks:
- [TTT ULX Commands](https://github.com/bender180/Trouble-in-Terrorist-Town-ULX-Commands/blob/master/lua/ulx/modules/sh/ttt_admin.lua)
- [TTTDamagelogs](https://github.com/Tommy228/TTTDamagelogs/blob/master/lua/damagelogs/ulx/sv_autoslay.lua)

However, these addons only rely on this field in one or two places each. Simple changes can be made to fix this behavior. But there is one more addon that will definitely cause grief:

- TTT Defibrillator.

This addon is found on many servers and all of them must use this feature to operate successfully. The two versions that are on GitHub ([willox'](https://github.com/willox/archive/blob/master/gmod-defibrillator-master/lua/weapons/weapon_ttt_defib.lua) and [another one](https://github.com/Nomad923/newprimettt/blob/master/terrortown/entities/weapons/weapon_ttt_defib/shared.lua)) are both broken by this, for example.
This will probably cause problems, as there are also versions on the workshop.

Kefta:
Revert it, submit pull requests to all of the addons changing that one thing, warn people of deprecation. We can tell people how to fix it if they ask on Facepunch. I can help contact the authors on Steam if you wish for the workshop versions.

zerfgog:
Reverted.

Bo98:
How about, for backwards compatibility, just simply defining both a field for UniqueID and Steam ID and using the Steam ID one throughout TTT?

```
rag.uqid = ply:UniqueID()
rag.sid = ply:SteamID()
```

That should get the best of both worlds.

zerfgog:
@Bo98 Good idea. Done.

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
```diff
@@ -380,7 +380,9 @@ function CORPSE.Create(ply, attacker, dmginfo)
 
    -- flag this ragdoll as being a player's
    rag.player_ragdoll = true
-   rag.uqid = ply:UniqueID()
+   rag.sid = ply:SteamID()
+
+   rag.uqid == ply:UniqueID() -- backwards compatibility; use rag.sid instead
```
svdm: You've got an `=` too many there

zerfgog: How embarrassing. Fixed.

markusmarkusz:
This killed every usage with bots and every equipment menu reskin and every custom score event. :(

robotboy655:
Should've used SteamID64 tbh

markusmarkusz:
This would also break things. SteamID64 returns no value clientside.

markusmarkusz:
What can be used to identify players and bots on client and server?
Player:AccountID() doesn't work with bots.
Player:SteamID() has two problems: On server every bot has “BOT“ as SteamID. On client it returns no “NULL“ for bots.
Player:SteamID64() has one problem: On client it returns no value for bots.
Player:UniqueID() can cause problems.
I assume that Player:UserID() wouldn't be a good solusion.

Kefta:
There isn't any default solution for unique IDs that handle bots included, unfortunately. It would be fine if the SteamID function had a separate numerical universe for bots or something that would solve conflicts, but I doubt it'd ever be changed.

PR #1185, opened 2016-05-30
Fix rare cl_hudpick error

gspetrou:
Sometimes, for example if the player spawns on a weapon while they are still loading into the game, the game will try to show the hudpickup panel before LocalPlayer() returns a a valid player, causing an error.


svdm:
One wonders why the HUD is being rendered at all at that point in time...

Kefta:
This should be a bug report instead

PR #1186, opened 2016-05-31
Added player class Death hook

UselessGhost:
A curious omission from the player class library. Would make it easier (and cleaner) to write player classes without any extra hook nonsense.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/player.lua
+++ b/garrysmod/gamemodes/base/gamemode/player.lua
@@ -175,6 +175,8 @@ function GM:PlayerDeath( ply, inflictor, attacker )
 
 	end
 
+	player_manager.RunClass( ply, "Death", inflictor, attacker )
+
 	if ( attacker == ply ) then
 	
 		net.Start( "PlayerKilledSelf" )
--- a/garrysmod/gamemodes/base/gamemode/player_class/player_default.lua
+++ b/garrysmod/gamemodes/base/gamemode/player_class/player_default.lua
@@ -71,6 +71,9 @@ function PLAYER:SetModel()
 
 end
 
+function PLAYER:Death( inflictor, attacker )
+end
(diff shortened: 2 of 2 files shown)


thegrb93:
Shouldn't whatever gamemode that needs the hook implement it?

UselessGhost:
The player class library has loads of unused hooks in it.

robotboy655:
Thing is, if GM.PlayerDeath is overridden in custom gamemodes, the new PLAYER:Death will not be called because literally no custom gamemode is calling that player hook.

UselessGhost:
If a gamemode overrides GM:PlayerSpawn they need to do the same shit.

robotboy655:
It was in the base gamemode since the beginning of the player_manager library though, other gamemodes copied it if needed be.

Kefta:
Better late than never. New developers can now utilise this if they want,

UselessGhost:
So? Other gamemodes can copy this if need be too. It's an addition, so it's not like existing gamemodes will suddenly break if this is included.

aStonedPenguin:
No one even uses player classes.

Acecool:
They can have their use, but it is essentially a limited hook system which only calls 1 function rather than the entire list... ie in certain instances someone new to coding can use them to simplify a loadout for player classes, etc.. but those with more experience could just set up a list and not use a huge stack of if / else statements... etc...

I thought about integrating them into my framework, and I may do it ( if a games/ game enables it ) but It may not be worth the effort because a lot of the functionality can be built using other means, and much easier it seems..


So I've basically just been working on extending functionality of existing features and trying to make it as intuitive as possible rather than create new systems to work with or to replace old systems when they still work, and are named, well..

I personally thing it looks better to set things up that way ( team ) rather than create a bulky player class files which are pretty limited...

PR #1187, opened 2016-05-31
math.Approach Optimization

thegrb93:
Slight optimization to this function which is called very frequently. I recorded 25% speed increase. If the math functions were made local then it would be a 40% speed increase.

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -211,11 +211,11 @@ function math.Approach( cur, target, inc )
 
 	if (cur < target) then
 		
-		return math.Clamp( cur + inc, cur, target )
+		return math.min( cur + inc, target )
 
 	elseif (cur > target) then
 
-		return math.Clamp( cur - inc, target, cur )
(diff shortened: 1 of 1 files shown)


meepdarknessmeep:
this isn't even an optimization - it completely changes how it works

thegrb93:
No it doesn't. Explain what makes you think that.

thegrb93:
You and robotboy need to practice your math skills.

Kefta:
It does. For example, math.Approach( 100, 101, -50 ). Before, that would return 100, but now it returns 50. I know it's a bit illogical to use math.Approach that way in the first place, but it was a behaviour that your change affects.

Bo98:
https://github.com/garrynewman/garrysmod/blob/c4deb9d9ebf89eac19044b508b5775ff8ffb4be4/garrysmod/lua/includes/extensions/math.lua#L210

Kefta:
Oh, damn. @thegrb93 is right

robotboy655:
It's not about math skills, its about the one line GitHub chose to omit.

PR #1189, opened 2016-06-05
Add Vector:AsNormalized()

wyozi:


Kefta:
Does this really require another function?

wyozi:
It's quite trivial but I feel like it's warranted as normalizing a vector is fairly common thing to do.

Also afaik all engine methods that return a vector have to duplicate it to prevent stupid Lua code doing stupid things. Easier normalization in-place could reduce strain on gc, although I'm not sure if luajit optimizes this (especially in local variables where this method would matter).

Here's a random case I found in garrysmod code where the call could be chained: https://github.com/garrynewman/garrysmod/blob/451b4ff5d1aea7b9b06a8024ef706c248a79647e/garrysmod/lua/includes/extensions/client/panel/animation.lua#L315 I keep bumping into similar case quite often.

Kefta:
I don't see any issue with just normalising it before, but I guess it's up to personal preference. Either way, I don't think this should be in the main game.

robotboy655:
Vector:GetNormalized exists for 3 years.

Kefta:
@robotboy655 That creates a new copy of the vector. Different than this case.

zerfgog:
Nothing in this code base has method chaining really; it wouldn't make sense to only bring it to one class, even less sense to bring it to a single method.

Acecool:
I prefer chaining, some of the original functions simply modify itself and return nothing while others return a copy which is why the functions I added allowed chaining after modifying itself.. But, some of the original functions could be changed rather than adding a new one ( where the original behavior is preserved to allow for backwards compatibility, but adds a new arg so you can modify itself and chain... 

Or where functions modify their data but don't return anything could be modified to return self to allow for chaining, etc...

PR #1191, opened 2016-06-05
Optimize MouthMoveAnimation

JetBoom:
Why did we need to lookup and set the flex weight every frame for every player even if they're not speaking? Perf will still be lost but only if people are /actually/ speaking now.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/animations.lua
+++ b/garrysmod/gamemodes/base/gamemode/animations.lua
@@ -264,8 +264,8 @@ end
 --
 -- Moves the mouth when talking on voicecom
 --
-function GM:MouthMoveAnimation( ply )
-
+local function DoMouthMove( ply, weight )
+	
 	local FlexNum = ply:GetFlexNum() - 1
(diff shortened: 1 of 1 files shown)


mcNuggets1:
This should be merged

thegrb93:
This too!

Kefta:
@thegrb93 check out Matt's

thegrb93:
Yes whichever is best. Here's the other https://github.com/garrynewman/garrysmod/pull/1230

Kefta:
They conflict with each other

thegrb93:
I know, but it can be resolved.

robotboy655:
This was basically covered in https://github.com/Facepunch/garrysmod/commit/143b18a0e5afcc1a11dbd210829b9d5f045c54d6

If you still think you can do better, be my guest and redo this PR with the new changes.

thegrb93:
No it wasn't. 143b18a's performance is much worse.

Kefta:
This is a part of #1454 if that will ever be merged.

thegrb93:
That would be good

PR #1192, opened 2016-06-05
Fix stack overflow while loading

JetBoom:
Should fix https://github.com/Facepunch/garrysmod-issues/issues/2047

Either way, it's a horrible idea to index yourself inside __index. Why the error happens on line 25 instead of 23, I don't know. GetTable seems to be declared in engine or something because I can't find it.

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -22,7 +22,7 @@ function meta:__index( key )
 	--
 	local tab = self:GetTable()
 	if ( tab ) then
-		local val = tab[ key ]
+		local val = rawget( tab, key )
 		if ( val != nil ) then return val end
 	end
 


willox:
Does this fix an issue you can reproduce?

zerfgog:
I'm fairly sure that it's indexing the entity's GetTable(), not the entity's self. Very different things.

JetBoom:
Then why do people get stack overflows on fresh installs on that line?

willox:
> Then why do people get stack overflows on fresh installs on that line?

Line 25? They don't, as you have said.

> Either way, it's a horrible idea to index yourself inside __index.

This code isn't re-indexing the entity itself. It does make sense to use rawget rather than to invoke any metamethods that may exist on the table, but I don't intend on changing the current behaviour.

Regarding the engine error, I believe I've found the cause for that in the engine. So I'll just patch that. Are the stack overflows always followed by the engine error?

JetBoom:
tab somehow becomes == self so same thing.

If you found the actual cause then use that.

willox:
I found one potential cause for the engine error. The only reason I can think that this could cause the stack overflow is a little bad luck. Can you confirm that the stack overflow only occurs sometimes and not every time the engine error occurs?

JetBoom:
I can't get the error to come up on demand but someone posted a crash dump at the end of the issue thread. Personally I've only gotten lua panic to show up once or twice.

willox:
Ok. I'm going to close this (I'm pretty sure it won't affect the error you wanted to fix anyway.)

PR #1193, opened 2016-06-06
Fix the default value to accept invalid strings

dvdvideo1234:
The default value will ignore invalid number to sting inputs


dvdvideo1234:
According to https://wiki.garrysmod.com/page/Player/GetInfoNum

PR #1194, opened 2016-06-09
Update Local Viewmodel

MrUnknownGamer:
TTT doesn't really work with custom hands anyway. Plus, people complain about them not being able to use their custom crowbar models.


Kefta:
No

zerfgog:
This is terrible

svdm:
Does not make much sense to me.

PR #1197, opened 2016-07-01
DPanPanel

ottworks:


garrysmod/lua/vgui/dpanpanel.lua
```diff
@@ -0,0 +1,263 @@
...
+function PANEL:OnScroll( x, y )
+	
+	self:GetCanvas():SetPos( x, y )
+	
+end
+
+--[[---------------------------------------------------------
+Name: ScrollToChild
+-----------------------------------------------------------]]
+function PANEL:ScrollToChild( panel )
```
wyozi: Maybe call this `PanToChild`?

Mysterypancake1:
What do the 'do' things on line 110, 134 and 182 actually do?

Kefta:
Creates a new local environment since the variables he makes in those aren't used outside of the do blocks.

Kefta:
@0tt Not home atm, but I think you can use ipairs or numerical loops for the GetChildren since the keys are probably numbers

robotboy655:
Ok, so:
1) Please implement http://wiki.garrysmod.com/page/PANEL/GenerateExample
2) Live drag does not limit the drag of the panels
3) The following code produces infinite PerformLayouts calls ( it is called every frame )
```
local f = vgui.Create("DFrame")
f:SetSize( 500, 500 )
f:Center()
f:MakePopup()

local a = vgui.Create("DPanPanel", f)
a:Dock( FILL )
local b = a:Add("DButton")
local b = a:Add("DButton") b:SetPos( 100, 100 )
```
Use ***vgui_visualizelayout 1*** to see PerformLayouts live.

4) The following code completely breaks the panel functionality:
```
// f = dframe
local a = vgui.Create("DPanPanel", f)
a:Dock( FILL )
local b = a:Add("DButton") b:Dock(TOP)
```
5) The following code breaks the "snap to maximum canvas size" thing from 2), as well as makes the only draggable area the 5px of the margin around the second button:
```
//f = dframe
local a = vgui.Create("DPanPanel", f)
a:Dock( FILL )
local b = a:Add("DButton") b:Dock(TOP)
local b = a:Add("DButton") b:Dock(TOP) b:DockMargin(5,5,5,5)
```

6) The following code causes everything 5) and 3) causes + forces the canvas to the bottom of DPanPanel

0tt:
Why are you trying to dock to an infinite plane?

robotboy655:
It is not infinite. While it doesn't make much sense to use Dock() in DPanPanel, it shouldn't completely break the panel and cause unexpected behavior if done so.

Mysterypancake1:
Wouldn't this need to be added to the list [here](https://github.com/garrynewman/garrysmod/blob/451b4ff5d1aea7b9b06a8024ef706c248a79647e/garrysmod/lua/includes/vgui_base.lua) as well?

0tt:
Everything is better now. How do I get a panel's DockMargin?

Bo98:
panel:GetDockMargin()

robotboy655 (approved):

PR #1198, opened 2016-07-03
Update net.lua

jackson-c:
Stop arbitrarily converting all strings to lowercase; it's a big performance drain.

Time taken to index a table with a 20-character string directly: 0.0005 ms
Time taken to index a table with a 20-character string converted to lowercase on-the-fly: 0.15 ms

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -7,7 +7,7 @@ net.Receivers = {}
 --
 function net.Receive( name, func )
 		
-	net.Receivers[ name:lower() ] = func
+	net.Receivers[ name ] = func
 
 end
 
@@ -21,7 +21,7 @@ function net.Incoming( len, client )
 	
 	if ( !strName ) then return end
 	
-	local func = net.Receivers[ strName:lower() ]
+	local func = net.Receivers[ strName ]
 	if ( !func ) then return end
 
 	--


meepdarknessmeep:
Percent of time wasted in any one server converting strings to lowercase: <0.000000000000000001%

Kefta:
But it's unneeded. Regardless of how big of a performance save it is, doing something "just because" is a bad programming precedent.

jackson-c:
I think it's significant, actually. If your server was getting 1000 net messages a second, this pointless action would eat 150ms of runtime a second.

meepdarknessmeep:
what kind of fucking server gets 1000 net messages serverside a second

robotboy655:
I do agree with @meepdarknessmeep on this one, the "performance gain" would be pretty much irrelevant, and there's always a possibility some shitty addon registers "MyNetStuff" receiver and sends "mynetstuff" net messages, which would break if this is to be merged.

jackson-c:
If someone used a net message every tick on every client? 33 tick * 32 players is like 1000 a second.

also, language

Kefta:
I see the compatibility concern in this case, but the behaviour should be eventually deprecated, honestly.

@jackson-c Is unfortunately about right in this case; I have to do the same for my CS:S physics system since a lot of the methods called down the line from RunCommand don't have hook equivalents.

meepdarknessmeep:
couldn't we theoretically only lowercase on either the sending or receiving end instead of both? @robotboy655

meepdarknessmeep:
I mean the performance improvement will still do absolutely nothing, but we could probably only do it on the receiving end or something.

jackson-c:
You could make it so util.AddNetworkString automatically converts it to lowercase (C internal), and net.Recieve still automatically converts it to lowercase, but not net.Incoming. That way there would be no performance loss during actual runtime, and we would maintain "shitty addon compatibility".

meepdarknessmeep:
Doing stuff like that could cause a lot of problems too - some addons I know use networkstrings for more than just net messages.

jackson-c:
Wait never mind, that wouldn't work anyway, you'd still have to convert to lowercase on the sending side.

robotboy655:
But we are only doing it on the receiving end?

PR #1201, opened 2016-07-13
TTT: Fixed edge case where ply:Alive() returns false results

markusmarkusz:
Sometimes Player:Alive() returns a false result.
For example if you join for the first time.
This would fix this.
I don't think that there are addons which would be broken because of this change.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
@@ -7,6 +7,7 @@ local math = math
 
 function plymeta:IsTerror() return self:Team() == TEAM_TERROR end
 function plymeta:IsSpec() return self:Team() == TEAM_SPEC end
+function plymeta:Alive() return self:IsTerror() end
 
 AccessorFunc(plymeta, "role", "Role", FORCE_NUMBER)
 


bmwalters:
Have you tested this with addons such as SpecDM which heavily use this function?

markusmarkusz:
Yes.
This change would fix some bugs from the SpecDM.
This is the only place where Player:Alive() could make a problem after changing this:
https://github.com/Tommy228/TTT_Spectator_Deathmatch/blob/master/lua/cl_spectator_deathmatch.lua#L51

robotboy655:
I think a better way around this would be to Player.KillSilent the player on first spawn rather than overriding Player.Alive.

markusmarkusz:
When would be the best moment to use Player.KillSilent?
Maybe here?
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/init.lua#L306-#L316
Or in the PlayerInitialSpawn Hook?

markusmarkusz:
Player:KillSilent() isn't the best idea.
If you get forced to spectator and wait 2 round then Player:Alive() also returns true.

robotboy655:
The problem is not that it returns a false result, its the opposite it returns the correct result, the problem is that the player is actually alive (I am assuming ply:Spawn() is being called on spectating players).

This PR is utter nonsense.

PR #1202, opened 2016-07-14
HSL support

Nak2:
HSL offers an easy and fast solution to adjustment/lightness of a color.

300px-hsl_and_hsv

Added functions:
- HSLToColor(H,S,L)             -> Color struct
- ColorToHSL(Color struct)  -> H,S,L
- COLOR:ToHSL()             -> H,S,L

Additionally, I could add these functions as well:

```
function COLOR:lighten(amount)
    local h,s,l = ColorToHSL(self)
    return HSLToColor(h,s,l+amount/100)
end

function COLOR:darken(amount)
    return self:lighten(-amount)
end
```

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -55,6 +55,64 @@ function COLOR:__eq( c )
 end
 
 --[[---------------------------------------------------------
+	Converts a color to HSL
+-----------------------------------------------------------]]
+local clamp = math.Clamp
+local abs = math.abs
+local min,max = math.min,math.max
+function ColorToHSL(R,G,B)
+	if type(R)=="table" then
+		R,G,B = clamp(R.r,0,255)/255,clamp(R.g,0,255)/255,clamp(R.b,0,255)/255
+	else
+		R,G,B = R/255,G/255,B/255
+	end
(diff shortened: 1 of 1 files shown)


Kefta:
I would recommend only supporting ColorToHSL( color struct ) since that's what ColorToHSV does

Nak2:
Changed it to only support the color structure.

robotboy655:
Well your code doesn't even work - math extensions are loaded way after `color.lua` is, so math.Clamp is not available and therefore errors, so I will be adding this in C++. (soon)

Also your code produces non whole RGB numbers.

robotboy655:
Added the functions to Dev.

Nak2:
> Well your code doesn't even work - math extensions are loaded way after `color.lua` is, so math.Clamp is not available and therefore errors, so I will be adding this in C++. (soon)

I made this pull request **3 years** ago. I can only remember it was working before I put it on git.
Might have been some failed optimization to localize math.min and max. Can't remember.

> Also your code produces non whole RGB numbers.

If I've had any feedback, I could had rounded the numbers up. I tried to make sure that you can convert it back and forth without any problems. Again I haven't heard anything for **3 years**.

> Added the functions to Dev.

Nice. Could recommend considering a function to light or darken a color, as I found it useful in my projects.

PR #1206, opened 2016-07-16
Base (math-Lib): Changed name from math.Dist function

markusmarkusz:
According to wiki, the function math.Dist is deprecated.
So I changed everything from this function to math.Distance.
And math.Dist is now the alias from math.Distance.

The source of the information is the Wiki. http://wiki.garrysmod.com/page/math/Dist


robotboy655:
This is a very pointless chance but I will allow it

PR #1207, opened 2016-07-16
UniqueID2

uRandomAlex:
UniqueID2. Cached, with no collisions possible.

### DIFF
--- a/garrysmod/lua/includes/extensions/player_auth.lua
+++ b/garrysmod/lua/includes/extensions/player_auth.lua
@@ -56,7 +56,7 @@ end
 
 --[[---------------------------------------------------------
     Name: UniqueID2
-    Desc: Real unique ID.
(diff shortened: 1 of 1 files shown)


Kefta:
How is this the "real" UniqueID? Why is another needed?

uRandomAlex:
@Kefta It has no collisions and it's short, so it can be converted to number. All UID1 usage should be replaced with this.

Kefta:
Sorry, I was thinking of AccountID. Use that instead

bigdogmat:
You say we can convert this to a number if we want, but what if the ID part of the steam id is zero? Then it becomes a number only representing the account number. Same could be said for the universe, though that should always be 0 afaik.

If anything you should've implemented this like AccountID because then it's at least useful. Though AccountID is already being added next update.

uRandomAlex:
Sorry, I've not seen AccountID.

PR #1208, opened 2016-07-17
TTT: Fix Spaces

markusmarkusz:
Only the spaces between the names of the variables and their values have been adjusted.
And in some cases, duplicate or unneeded variables were removed.
And I changed the position in which one or other variable.
Have done then another. If this is considered useless or so, the PR can be easily geclosed.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
@@ -9,14 +9,15 @@
 --DISABLED
 --AddCSLuaFile()
 
-SWEP.HoldType = "normal"
+SWEP.HoldType               = "normal"
 
 if CLIENT then
-   SWEP.PrintName			= "Beacon"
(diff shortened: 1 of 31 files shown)


robotboy655:
berfore you think of it, Don't do it for the rest of the code.

markusmarkusz:
> Before you think of it, Don't do it for the rest of the code.
Maybe my English is not good enough and that's why I do not understand what you want, or I understand the sentence, but do not know what to tell me.
Can you give me might somehow explain what should I do now xD?

robotboy655:
I do not know how you can misunderstand that, but what I am saying is:

Do not create a Pull Request to clean the spaces for non TTT files in this Repository.

markusmarkusz:
I haven't planned to do this.
I only cleaned up this TTT weapons because they look a little bit strange for me.
And I think these weapons are the only files from TTT that needs to be cleaned up.

svdm:
I don't like "let's change some formatting" pull requests in general, but in this case the original files were quite messy, so I'll make an exception here.

PR #1209, opened 2016-07-19
TTT: Prevent Props be targeted by testers

markusmarkusz:
Example:
If a player (Innocent) the chute used on the Map "ttt_waterworld" for testing and a ragdoll carries with it comes a time the sound that he is Innocent and the other times he's Traitor, because of the ragdoll.
Therefore, would be a general inquiry as to whether the target is a player, in my opinion, from emergencies.

(Google Translator safes my live. xD)

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_logic_role.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_logic_role.lua
@@ -22,9 +22,8 @@ end
 
 
 function ENT:AcceptInput(name, activator)
-   if name == "TestActivator" then
-
-      if IsValid(activator) and activator:IsPlayer() and (self.Role == ROLE_ANY or activator:IsRole(self.Role)) then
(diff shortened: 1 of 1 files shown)


bmwalters:
> If a player (Innocent) uses the chute used on the map "ttt_waterworld" as a tester, and they carry a ragdoll with them, occasionally he is Innocent and other times he is a Traitor, because of the ragdoll.
> Therefore, in my opinion, a good check would be to see if the target is a player.

This feels like a feature to me (bringing a ragdoll into a tester).

markusmarkusz:
A ragdoll was only as an example. You can use every prop / entity.
And I think that shouldn't be so.

BlackCetha:
I agree with @bmwalters. This function has its gameplay purposes and shouldn't be changed.

svdm:
I believe mappers can prevent this if they want to by using a `filter_activator_class` to make sure only player entities are sent to `ttt_logic_role`. In the documentation for the entity in `ttt.fgd`, it's explicitly said that any entity that is not a player will always fire the OnFail output.

PR #1213, opened 2016-07-21
Fixed incorrect comment name in hook library

Kefta:


robotboy655:
Please no more comment changing PRs.

bmwalters:
Why not?

Kefta:
It's wrong?

Acecool:
I think it is a great idea that someone is updating incorrect comments. It'll help solve questions before they're even asked and merging shouldn't break anything either.

robotboy655:
@Acecool I think we can safely establish that 99% of the people don't bother trying to help themselves, not even with google, let alone comments inside code. Case and point: Any forum ever, including Facepunch. You could argue about the 1%, but I will argue that if they are smart enough to look in the base code comments, they are smart enough to figure it out without function name in the comments being wrong.

Acecool:
You may be right, but it doesn't hurt to correct the comments and seeing as it wouldn't break anything else merged.. why not allow Kefta to be the one to do it?

It may not be a huge deal to everyone but there are those that do read the comments and look at code before going to wiki, etc... I personally look through a lot of base code when I'm writing something new or changing behavior of something old ( while being backwards compatible; such as vgui elements ).

It doesn't hurt, it can help and the perfectionists will love you for it. It seems like a win win.

PR #1216, opened 2016-07-24
Include utf8 lib and coroutine extension on menu state init

bmwalters:


uRandomAlex:
Why? You can do this by yourself if you are making modification to main menu, it'll just increase RAM and CPU usage for other players.

robotboy655:
I don't see any reason for this either.

Kefta:
@urandomalex regardless of this was accepted or not, this doesn't affect RAM or CPU. That's like saying each Lua function you create takes up disk space

uRandomAlex:
@Kefta each Lua function takes up RAM. It's just some bytes, however. Because of laggy HTML menu we need to load as less Lua as we can.

Kefta:
It's absolutely negligible

PR #1218, opened 2016-07-27
TTT: Fix a small issue and remove some unused things

markusmarkusz:
Changes made:
- removed SWEP.LimitedStock because it's no longer buyable.
- removed SWEP:RemovePlayerSample because it's no longer used.
- removed unused if-query.
- removed unused sound.
- added LANG.TryTranslation to safely fix a issue in dna tooltip. Otherwise you can see for example "DNA sample found on pistol_name"

What about making names of the some variables general?
For example rename SWEP.ItemSamples to SWEP.Samples?
These changes aren't really needed so I haven't add them to this PR.
Should these changes be made?

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_wtester.lua
@@ -39,7 +39,6 @@ SWEP.Secondary.Delay       = 2
 SWEP.Kind                  = WEAPON_ROLE
 SWEP.CanBuy                = nil -- no longer a buyable thing
 SWEP.WeaponID              = AMMO_WTESTER
---SWEP.LimitedStock = true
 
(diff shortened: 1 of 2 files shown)


willox:
> For example rename SWEP.ItemSamples to SWEP.Samples?

You're probably best off trying to never rename anything that could be accessed by third-party scripts. You'd be surprised how often people rely on them.

markusmarkusz:
@willox 
Okay. I knew that this could happen.
This was the reason why I asked this.

markusmarkusz:
Changes made:
- removed unused LANG.GetTranslation
- removed unused LANG.GetParamTranslation
- added local variable for LANG.TryTranslation (Is this useful or not?)

svdm:
Indeed, I've often wanted to rename a bunch of stuff to be more clean and consistent (especially the `GAMEMODE.*` stuff in init.lua) but it will inevitably break a bunch of people's stuff, so it's not worth it.

PR #1222, opened 2016-08-02
DPanel to DFrame

MrUnknownGamer:
Very tiny fix,  but it may help in the future when someone is trying to make something for the scoreboard or modify it.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
@@ -426,7 +426,7 @@ function CLSCORE:ClearPanel()
       -- we need this hack as opposed to just calling Remove because gmod does
       -- not offer a means of killing the tooltip, and doesn't clean it up
       -- properly on Remove
-      gui.SetMousePos( ScrW()/2, ScrH()/2 )
+      input.SetCursorPos( ScrW()/2, ScrH()/2 )
       local pnl = self.Panel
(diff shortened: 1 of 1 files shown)


bmwalters:
But... the passed parameter in a lot of those situations IS a DPanel...

Kefta:
It's just a variable name

garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -426,7 +426,7 @@ function CLSCORE:ClearPanel()
       -- we need this hack as opposed to just calling Remove because gmod does
       -- not offer a means of killing the tooltip, and doesn't clean it up
       -- properly on Remove
-      input.SetCursorPos( ScrW()/2, ScrH()/2 )
+      gui.SetMousePos( ScrW()/2, ScrH()/2 )
```
Bo98: Why are you reverting other people's changes?

markusmarkusz:
```LUA
gui.SetMousePos
´´´
is deprecated.
See #1205

SpencerSharkey:
Wouldn't call this a "fix" 
DFrame = DPanel

bmwalters:
@SpencerSharkey No, there is a huge difference between the two panel types. The problem with this PR is that A) it might be unnecessary, B) it changes some places where they were ACTUALLY DPanel to "dframe" variable name.

PR #1223, opened 2016-08-03
TTT: Remove old fretta rests

markusmarkusz:
In GM 13 fretta is no longer available.
And there is also no replacement.
So this PR would remove all rests of fretta because they are no longer needed.

If this get aproved:
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua#L63-L69
Is this also a part of fretta vote?


bmwalters:
This may break the popular TTT Mapvote addon

robotboy655:
You should not be removing gamemode functionality, by removing hooks. ( TTTDelayRoundStartForVote )

markusmarkusz:
@robotboy655 @bmwalters 
I'll re-add the hook.
Until this I close this PR.

markusmarkusz:
Now it shouldn't break something.

svdm:
[This mapvote version](https://github.com/tyrantelf/gmod-mapvote/blob/master/lua/autorun/server/sv_autovote.lua) which has a bunch of workshop subscribers actually has TTT-specific stuff, but it doesn't actually seem to use hooks (whelp).

markusmarkusz:
I assume that no addon use this hook.
Normally addon creators use TTTEndRound or CheckForMapSwitch.
I think it's because TTTDelayRoundStartForVote is called in PrepareRound. So this is called after CheckForMapSwitch and TTTEndRound. And if there are no rounds left no new round would be started.
TTTDelayRoundStartForVote is maybe the dumbest function in TTT. It has no use.
And could it be that [this](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/init.lua#L730) should be:
```Lua 
timer.Simple(15, game.LoadNextMap())
```
game.LoadNextMap is not a function. game.LoadNextMap() is a function.
What about the things I also said? ("ttt_always_use_mapcycle"-ConVar and [cl_voice.lua L63-L69](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua#L63-L69))
They are also not used in anywhere.

bigdogmat:
> game.LoadNextMap is not a function. game.LoadNextMap() is a function.

`game.LoadNextMap` is a function, `game.LoadNextMap()` is a function call. Functions are first class values in Lua, and such they're passed just like normal variables.

markusmarkusz:
Ah.
```Lua
timer.Simple(15, game.LoadNextMap)
```
This is working. Maybe I misunderstood it.

PR #1228, opened 2016-08-12
TTT: Remove unused var and fix a small issue

markusmarkusz:
Just removed a useless var and fixed a small issue.

If you die a corpse will be created. (obvious)
If you reconnect, after your death, then the steam avatar wouldn't be shown in the corpse because after reconnecting Player:Alive() is returning a false result.


markusmarkusz:
Just added:
```lua
self:DrawShadow(false)
```
Because you can see the shadows from the world model (models/weapons/w_crowbar.mdl).

PR #1229, opened 2016-08-15
Update math extension

dvdvideo1234:
Added: math.Sign

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -1,3 +1,11 @@
+--[[---------------------------------------------------------
+	Name: Sign( num )
+	Desc: Takes the sign of a number, 0 when not relevant
+------------------------------------------------------------]]
+function math.Sign(num)
+  local num = tonumber(num) or 0
+  return (num ~= 0) and (num / math.abs(num)) or 0
+end
 
 --[[---------------------------------------------------------
 	Name: Distance( low, high )
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -3,7 +3,6 @@
 	Desc: Takes the sign of a number, 0 when not relevant
 ------------------------------------------------------------]]
 function math.Sign(num)
-  local num = tonumber(num) or 0
   return (num ~= 0) and (num / math.abs(num)) or 0
 end
 
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -3,7 +3,7 @@
 	Desc: Takes the sign of a number, 0 when not relevant
 ------------------------------------------------------------]]
 function math.Sign(num)
-  return (num ~= 0) and (num / math.abs(num)) or 0
+  return (num > 0 and 1) or (num < 0 and -1) or 0
 end
 
 --[[---------------------------------------------------------


garrysmod/lua/includes/extensions/math.lua
```diff
@@ -1,3 +1,11 @@
+--[[---------------------------------------------------------
+	Name: Sign( num )
+	Desc: Takes the sign of a number, 0 when not relevant
+------------------------------------------------------------]]
+function math.Sign(num)
+  local num = tonumber(num) or 0
```
AMD-NICK: I think it's unnecessary

thegrb93:
It has its uses. I think if then else is better than all this arithmetic though.
```
if a>0 then
    return 1
elseif a<0 then
    return -1
else
    return 0
end
```

robotboy655:
I think this whole PR is unnecessary.

dvdvideo1234:
Why.. There is no Sign function. I have to make it as a library every time

bigdogmat:
> I think it will crash when u get the sign of a string as you cannot compare string to number.

It'll just error, either way I think the point is barely any of the other math functions use tonumber on arguments. If someone is using a string input and is trying to use it as a number, they should purposely convert it before hand.

thegrb93:
Math functions don't need type safety imo.

AMD-NICK:
> It has its uses. I think if then else is better than all this arithmetic though.

this should be better
```lua
function math.Sign(num)
   eturn (num > 0 and 1) or (num < 0 and -1) or 0
end
```

dvdvideo1234:
What happens when you call it like this:

local a = math.Sin("50")
print("The result is "..a)

bigdogmat:
> If someone is using a string input and is trying to use it as a number, they should purposely convert it before hand.

dvdvideo1234:
@bigdogmat Ok then, removed the tonumber conversion

thegrb93:
local a = math.Sign(tonumber("50"))?

thegrb93:
My version is the fastest one.
```
require 'xbench'
--jit.off()

function math.Sign(num)
	return (num ~= 0) and (num / math.abs(num)) or 0
end

function math.Sign2(num)
	return (num > 0 and 1) or (num < 0 and -1) or 0
end

function math.Sign3(a)
	if a>0 then
		return 1
	elseif a<0 then
		return -1
	else
		return 0
	end
end

xbench.Compare({
    Old = function()
        math.Sign( 0.15 )
        math.Sign( -0.15 )
    end,
    New = function()
        math.Sign2( 0.15 )
        math.Sign2( -0.15 )
    end,
}, 1000000)

xbench.Compare({
    Old = function()
        math.Sign( 0.15 )
        math.Sign( -0.15 )
    end,
    New = function()
        math.Sign3( 0.15 )
        math.Sign3( -0.15 )
    end,
}, 1000000)
```
OLD: 0.11174105958327
NEW: 0.07191243159474
OLD: 0.10693967295873
NEW: 0.067574400211555

dvdvideo1234:
Nice, thanks for testing this for me ;)

AMD-NICK:
@thegrb93 fraud!11! :D
image

anyway my method looks better

```lua
require("xbench")

function math.Sign1(num)
	return (num ~= 0) and (num / math.abs(num)) or 0
end

function math.Sign2(num)
	return (num > 0 and 1) or (num < 0 and -1) or 0
end

function math.Sign3(a)
	if a > 0 then
		return 1
	elseif a < 0 then
		return -1
	else
		return 0
	end
end

xbench.Compare({
	Sign1 = function()
		math.Sign1( 0.15 )
		math.Sign1( 1 )
		math.Sign1( -0.15 )
	end,
	Sign2 = function()
		math.Sign2( 0.15 )
		math.Sign2( 0 )
		math.Sign2( -0.15 )
	end,
	Sign3 = function()
		math.Sign3( 0.15 )
		math.Sign3( 0 )
		math.Sign3( -0.15 )
	end,
}, 100000000)

-- SIGN2: 0.029420669774481
-- SIGN1: 0.034826290601643
-- SIGN3: 0.029995217636326
```

dvdvideo1234:
Bit it will look longer like one-liner

'''
function math.Sign3(a)
    if a > 0 then return 1 elseif a < 0 then return -1 else return 0 end
end
'''

Kefta:
My benchmarks were inconclusive because the difference ended up being purely based on processor tick time and no trial could prove which is the "fastest." I would say just go with Sign2 because it's the cleanest and doesn't call an outside method.

dvdvideo1234:
I bench my programs with my own code, that compares then for answer consistency :D

https://github.com/dvdvideo1234/ZeroBraineProjects/blob/master/dvdlualib/common.lua#L490-L553

dvdvideo1234:
Guys, aren't you gonna marge this already ? This modification is so simple but pretty convenient.

Kefta:
No PRs are being merged for whatever reason:

ExtReMLapin:
this is useless

dvdvideo1234:
What is so useless about it ... Not that I cannot do this myself

Acecool:
Yeah, don't add garbage code ( like the current situation with a lot of the networking functions that aren't type-bound and use string ) which promotes incorrect coding and typing practices ( and the reason the corrections had to be reverted; albeit I wouldn't have reverted the changes because too many old game-modes and addons were incorrectly coded if I was ever in such a situation ).

There are plenty of ways to do many different operations.. if the speed is negligible pick the cleanest one or do testing across more platforms and then choose.

Why not make your case / argument more compelling by providing examples where such a function has been useful to you.

dvdvideo1234:
Abs( ) has many useful features ... but ok....

PR #1232, opened 2016-08-17
TTT: Fix for #1195

markusmarkusz:
#1195 killed some brush entities from TTT.
This would fix it.

This issue was reported here:
https://facepunch.com/showthread.php?t=1531053


robotboy655:
Why is it using the direct variable over TriggerOutput?

svdm:
The TTT ents were written long before TriggerOutput's `data` argument existed (8fef618deea). They can/should be changed to just call self:TriggerOutput with the value that's currently being set in `op.param`.

markusmarkusz:
@svdm

I'll change it later.

PR #1234, opened 2016-08-26
TTT: Some Changes

markusmarkusz:
Because I have made several different changes in this Pull Request, I list them as concise as possible.

- GM.Customized got removed because it is from an older version.
- The mute variables refer now to the team variables.
- Added a ShouldScore() query at the point there deaths got added. (This helps some addons like the Power Round Addon.)
- Disabled the Muzzle Flashes for weapons, that don't need them.
- Made `PrintTraitors(ply)`, `PrintGroups(ply)` and `PrintReport(ply)` to local functions. I thought about if they have to be global.
- Added a query so detectives are shown as detectives not as innocents.

(I squash the commits if this got aproved.)

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/scoring.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/scoring.lua
@@ -181,7 +181,7 @@ function SCORE:ApplyEventLogScores(wintype)
    for k, e in pairs(self.Events) do
       if e.id == EVENT_KILL then
          local victim = player.GetBySteamID(e.vic.sid)
-         if IsValid(victim) then
+         if IsValid(victim) and victim:ShouldScore() then
             victim:AddDeaths(1)
          end
       end


Kefta:
Turning off CSMuzzleFlashes doesn't stop muzzle flashes; it's a variable that won't be used with non-guns. Best to just cut it out

garrysmod/gamemodes/terrortown/gamemode/shared.lua
```diff
@@ -218,3 +210,9 @@ DefaultEquipment = {
       "weapon_ttt_glock"
    }
 };
+
+-- Mute types
+MUTE_NONE   = 0
+MUTE_TERROR = TEAM_TERROR
+MUTE_ALL    = 2
+MUTE_SPEC   = TEAM_SPEC
```
bmwalters: I feel like this is error prone. It might be less error prone to have `ALL` equal the bitwise OR of TEAM_TERROR and TEAM_SPEC but numbers are probably better and make more sense simply by the fact that enums shouldn't depend on each other.

bmwalters: Actually I just read the code and wow, it actually checks the team against the MUTE enum.
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua#L189
That's kind of silly, why even have this enum? Just use bit flags and check if the Team is present in there.

markusmarkusz:
Better?

bmwalters:
That's not what I meant; I meant to remove the team enums entirely, and use bitwise operators to combine which teams are muted. `0 = no teams muted, `TEAM_TERROR | TEAM_SPEC` = all muted, etc

svdm:
The vast majority of these changes are pointless messing around, tweaking tiny things for the sake of changing something. As I've said in previous PRs, I'm not interested in merging that type of thing.

The only legitimate fix I'll accept is the ShouldScore one.

markusmarkusz:
@svdm
Okay.
I'll remove the other things from this PR.

nubpro:
May I know what exactly can we benefit on this?
It's a little vague.

Kefta:
It's just code cleanuo

nubpro:
I mean "Added a ShouldScore() query at the point there deaths got added. (This helps some addons like the Power Round Addon.)"

markusmarkusz:
@nubpro 
As an example:
With the Power Rounds Addon ([https://scriptfodder.com/scripts/view/401/power-rounds-murdertttphjbmbdrgw](https://scriptfodder.com/scripts/view/401/power-rounds-murdertttphjbmbdrgw)) you can create special rounds, like Deathmatch Rounds.
But if you just temporary override the ShouldScore() function, then only the points (frags) got blocked. But not the deaths.

PR #1236, opened 2016-08-31
Add table.ForceEmpty

Velkon:


garrysmod/lua/includes/extensions/table.lua
```diff
@@ -716,3 +716,13 @@ function table.GetKeys( tab )
 	return keys
 
 end
+
```
LennyPenny: I love this line of code, its like art

bmwalters:
really guys

Kefta:
You should empty _G if no arg is specified. Easy shortcut

PR #1237, opened 2016-08-31
Added table.ForceCount for even more table forcing.

tenrust:
Force some consistency on the table library.


danielga:
I was alerted to my ID being used in this commit. I do not condone the use of my credentials in such a manner. This commit is not GPG signed.

Kefta:
I've always wanted to count without actually knowing what the total is. Thanks!

bmwalters:
Make Garry's Mod `table` library great again? 🇺🇸

PR #1238, opened 2016-09-01
Added fallback for ammo item pickup name if no language string is ava…

Kefta:
…ilable

Instead of ammo showing up as "#556_ammo" if the language string is not available, it will show up as "556 ammo". I also removed useless nil checks for a table that's clearly defined at the top and removed inconsistant trailing whitespace.

Also, Weapon.GetPrintName is a metatable method.


Kefta:
I did change the loops to use ipairs, but table.insert adds new indicies to the last index regardless of previous holes, so pairs is staying.

robotboy655:
I don't think this is such a good idea for same reason we don't remove engine errors from console - errors like not supplying translations should be fixed, not avoided.

Kefta:
Fair enough, just thought it would be a good replacement since most ammo type names are descriptive enough

PR #1239, opened 2016-09-04
Added player.GetByName

markusmarkusz:
This function could be helpful. There are many addons that need a function like this.
But because there is no function by gmod they have create it. And there are also such functions for the SteamID64, SteamID32, UniqueID, ...
So why not for names? :)


robotboy655:
We have been over this.

https://github.com/garrynewman/garrysmod/pull/846

markusmarkusz:
Oh. Haven't seen this.

PR #1240, opened 2016-09-06
TTT Trap Hook

MrUnknownGamer:
Allow us to hook on to when a trap is called. I have had this work with TTT Damagelogs by Tommy228

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
@@ -74,7 +74,6 @@ end
 function ENT:TraitorUse(ply)
    if not (IsValid(ply) and ply:IsActiveTraitor()) then return false end
    if not self:IsUsable() then return false end
-   local desc = self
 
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/terrortown/entities/entities/ttt_traitor_button/init.lua
```diff
@@ -74,6 +74,7 @@ end
 function ENT:TraitorUse(ply)
    if not (IsValid(ply) and ply:IsActiveTraitor()) then return false end
    if not self:IsUsable() then return false end
+   local desc = self
```
meharryp: If it's only used once don't bother putting it in a variable.

bmwalters:
Dunno about the argument order; in a hook called `TTTTrapActivated` I'd expect the first argument to be, y'know, the trap that was activated

svdm:
I'd swap the arguments yeah.

Perhaps the hook should just be called `TTTTraitorButtonActivated`? A lot of these buttons are simply linked to doors, which are not exactly traps.

PR #1243, opened 2016-09-25
Update rope.lua

kulcris:
Add owner to ropes when they are spawned

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/rope.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/rope.lua
@@ -21,9 +21,9 @@ function TOOL:LeftClick( trace )
 
 	-- If there's no physics object then we can't constraint it!
 	if ( SERVER && !util.IsValidPhysicsObject( trace.Entity, trace.PhysicsBone ) ) then return false end
-	
+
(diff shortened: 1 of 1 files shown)


Kefta:
Should just cache the GetOwner call instead of calling it five times.

kulcris:
did that throughout entire file,

robotboy655:
Is there any reason to do it? And to do it only on the rope tool?

kulcris:
rope tool is one of the most abused tools in the game, setting the owner will allow someone to create a way to see who created the giant male appendage on the wall without going through miles of logs. 

Should probably be done with the other constraint tools as well though

robotboy655:
You can already do this with cleanup.GetList()[UniqueIDOfaPlayer]["ropeconstraints"]

kulcris:
you cant find who owns specific ropes with that though.

robotboy655:
Sure you can. Here are 5 lines of code that do this exact same thing:
```
for id, ply in pairs( player.GetAll() ) do
	for id, ent in pairs( cleanup.GetList()[ ply:UniqueID() ]["ropeconstraints"] ) do
		ent:SetOwner( ply )
	end
end
```
But it works for all constraints.

kulcris:
ah, awesome. thanks for that didnt realize that is how the table was setup.

PR #1245, opened 2016-09-25
Consistency bodygroups -> skin property

FPtje:
Effects have a field `ent.AttachedEntity`. When you have an effect that allows its skin to be changed through the properties menu, the actual `:SetSkin` method of the property is called with the `prop_effect`. When you have an effect that allows its bodygroup to be changed from the properties menu, the actualy `:SetBodyGroup` method of the property is called **not** with `ent` , but with `ent.AttachedEntity`. This is inconsistent and causes FPP to bug out. 

This PR fixes that inconsistency. I tested it and setting body groups still works. As for addon breakage, unless addons specifically hardcoded that shit's fucked for bodygroups, this shouldn't break anything.

### DIFF
--- a/garrysmod/lua/autorun/properties/bodygroups.lua
+++ b/garrysmod/lua/autorun/properties/bodygroups.lua
@@ -33,12 +33,11 @@ properties.Add( "bodygroups", {
 
 	MenuOpen = function( self, option, ent, tr )
 
-		if ( IsValid( ent.AttachedEntity ) ) then ent = ent.AttachedEntity end
-
+		local target = IsValid( ent.AttachedEntity ) and ent.AttachedEntity or ent
 		--
 		-- Get a list of bodygroups
 		--
-		local options = ent:GetBodyGroups()
+		local options = target:GetBodyGroups()
 
 		--
 		-- Add a submenu to our automatically created menu option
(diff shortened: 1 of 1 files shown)


robotboy655:
What's up with the unnecessary changes on the top?

FPtje:
The top change:

```lua
local target = IsValid( ent.AttachedEntity ) and ent.AttachedEntity or ent
```

is necessary. Leaving that change out will leave `target` undefined. What do you mean?

Kefta:
He's saying why is it necessary to create a new local instead of just changing ent? You're now altering the self:Filter( ent, player ) call.

Kefta:
Ignore my previous comment

FPtje:
I'll reply anyway. It's relevant:

1. That's consistent with what skin does
2. Filter expects to find an effect, not a prop_dynamic
3. `ent` is needed in the `:SetBodyGroup` call

So yeah, both `target` and `ent` are needed.

robotboy655:
Changing it from ```ent``` to ```target``` is completely pointless?

FPtje:
No, it isn't pointless. `ent` in the case of an effect is a prop_effect,
target is its attached prop_dynamic. Both those entities need to be
referred to.

PR #1246, opened 2016-09-25
Consistency bodygroups -> skin property

FPtje:
Effects have a field `ent.AttachedEntity`. When you have an effect that allows its skin to be changed through the properties menu, the actual `:SetSkin` method of the property is called with the `prop_effect`. When you have an effect that allows its bodygroup to be changed from the properties menu, the actualy `:SetBodyGroup` method of the property is called **not** with `ent` , but with `ent.AttachedEntity`. This is inconsistent and causes FPP to bug out. 

This PR fixes that inconsistency. I tested it and setting body groups still works. As for addon breakage, unless addons specifically hardcoded that shit's fucked for bodygroups, this shouldn't break anything.


FPtje:
God damnit

PR #1247, opened 2016-09-26
Fixed material tool reload function not working

NO-LOAFING:
(it was accidentally named "TOOL:RightClick()" instead of "TOOL:Reload()", causing it to override the existing right click function)


Kefta:
Now you have a new feature

robotboy655:
Clearly @aStonedPenguin has no idea what he is talking about, this is an oversight on my part, I forgot to rename the function when copying it a few weeks ago in this commit https://github.com/garrynewman/garrysmod/commit/1acd2a1057274906bc706b59137c33caed24793e while expanding feature set of default Sandbox tools.

aStonedPenguin:
i could have sworn that has always been a thing

PR #1250, opened 2016-09-29
Add hook to set scoreboard row color per player.

JLVerified:
Removes the hard-coding and opens it up for easier customization.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
@@ -69,7 +69,13 @@ local namecolor = {
    default = COLOR_WHITE,
    admin = Color(220, 180, 0, 255),
    dev = Color(100, 240, 105, 255)
-};
+}
+
+local rolecolor = {
+   default = Color(0, 0, 0, 0),
+   traitor = Color(255, 0, 0, 30),
+   detective = Color(0, 0, 255, 30)
+}
 
 function GM:TTTScoreboardColorForPlayer(ply)
    if not IsValid(ply) then return namecolor.default end
@@ -83,13 +89,15 @@ function GM:TTTScoreboardColorForPlayer(ply)
 end
 
 function GM:TTTScoreboardRowColorForPlayer(ply)
+   if not IsValid(ply) then return rolecolor.default end
+
    if ply:IsTraitor() then
-      return Color(255, 0, 0, 30)
+      return rolecolor.traitor
    elseif ply:IsDetective() then
-      return Color(0, 0, 255, 30)
+      return rolecolor.detective
    end
 
-   return Color(0, 0, 0, 0)
+   return rolecolor.default
 end
 
 local function ColorForPlayer(ply)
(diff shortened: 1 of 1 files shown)


Kefta:
The type check isn't really necessary. If someone returns something invalid in the hook, they'll notice the error immidiately.

JLVerified:
I was trying to maintain consistency, seeing as the function directly above has the same behavior. Should it be removed?

Kefta:
You don't need the local function at all; you can just integrate it into the main function

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -105,13 +123,10 @@ function PANEL:Paint()
...
-   if ply:IsTraitor() then
-      surface.SetDrawColor(255, 0, 0, 30)
-      surface.DrawRect(0, 0, self:GetWide(), SB_ROW_HEIGHT)
-   elseif ply:IsDetective() then
-      surface.SetDrawColor(0, 0, 255, 30)
-      surface.DrawRect(0, 0, self:GetWide(), SB_ROW_HEIGHT)
-   end
+   local c = RowColorForPlayer(ply)
+
+   surface.SetDrawColor(c.r, c.g, c.b, c.a)
```
markusmarkusz: You could use `clr(c)` here.
But this change isn't necessary.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -105,13 +123,10 @@ function PANEL:Paint()
...
+   surface.DrawRect(0, 0, self:GetWide(), SB_ROW_HEIGHT)
```
fruitwasp: PANEL.Paint has `width` and `height` parameters. Let's use them.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -96,6 +106,14 @@ local function ColorForPlayer(ply)
...
 
+local function RowColorForPlayer(ply)
+   if IsValid(ply) then
+      return hook.Call("TTTScoreboardRowColorForPlayer", GAMEMODE, ply)
+   end
+
+   return Color(0, 0, 0, 0)
+end
+
 function PANEL:Paint()
```
fruitwasp: `width` and `height` parameters to use on [line 129](https://github.com/garrynewman/garrysmod/pull/1250#discussion-diff-81411277R129)

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -82,6 +82,16 @@ function GM:TTTScoreboardColorForPlayer(ply)
...
 
+function GM:TTTScoreboardRowColorForPlayer(ply)
+   if ply:IsTraitor() then
+      return Color(255, 0, 0, 30)
+   elseif ply:IsDetective() then
+      return Color(0, 0, 255, 30)
+   end
+
+   return Color(0, 0, 0, 0)
+end
```
svdm: This is creating new Color objects for every single row that's rendered. It would be more efficient to create the colors once and just return them in this function. Same as is being done with `namecolor` right above here.

PR #1251, opened 2016-10-02
TTT: draw weapon Help before crosshair

markusmarkusz:
If you disable the crosshair the help from a hud wouldn't get drawn.
This shouldn't be so because you disable the crosshair, not the help.

By the way why is the crosshair drawn in SWEP:DrawHUD()?
Wouldn't it be better if it is drawn in [SWEP:DoDrawCrosshair()](http://wiki.garrysmod.com/page/WEAPON/DoDrawCrosshair)? Then developers can hide the crosshair for special equipments without overwriting the complete function.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -133,8 +133,11 @@ if CLIENT then
    local crosshair_size = CreateConVar("ttt_crosshair_size", "1.0", FCVAR_ARCHIVE)
    local disable_crosshair = CreateConVar("ttt_disable_crosshair", "0", FCVAR_ARCHIVE)
 
-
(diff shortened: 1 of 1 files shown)


markusmarkusz:
@svdm 
What about the thing that I mentioned?
> By the way why is the crosshair drawn in SWEP:DrawHUD()?
Wouldn't it be better if it is drawn in [SWEP:DoDrawCrosshair()](http://wiki.garrysmod.com/page/WEAPON/DoDrawCrosshair)? Then developers can hide the crosshair for special equipments without overwriting the complete function.

svdm:
I don't remember why DoDrawCrosshair wasn't used, perhaps simply because the CS:S SWEPs I based things on didn't use it.

I'm not sure if it's practical to change it now. I think it would lead to custom TTT SWEPs getting double crosshairs if they've overridden DrawHUD to do their own crosshair.

PR #1253, opened 2016-10-06
Base: Make SWEP.PrintName shared

markusmarkusz:
I also cleaned the two edited files a bit up (removed some tab spaces).
If you don't like this "cleanups" then I'll remove them :P.


markusmarkusz:
@robotboy655 
Better?

Kefta:
@robotboy655 It's good now?

PR #1259, opened 2016-10-17
Remove restrictions on sv_mincmdrate and sv_minupdaterate

Jiwer:
See my thread here for details.

https://facepunch.com/showthread.php?t=1537473


uRandomAlex:
wtf this has to do with cmdrate and updaterate

PR #1260, opened 2016-10-18
Moved const table outside of function

Kefta:


Weol:
Why tho?

Kefta:
There's no need to redeclare it every time?

Weol:
Nevermind, was under the impression that the table was global

PR #1262, opened 2016-10-20
TTT: Add support for SWEP:HUDShouldDraw()

markusmarkusz:
With this you can easily hide HUD elements.
It's just the same like the it's done in the base gamemode.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_hud.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_hud.lua
@@ -371,6 +371,15 @@ function GM:HUDShouldDraw(name)
       if name == v then return false end
    end
 
+   if IsValid(LocalPlayer()) then
+      local wep = LocalPlayer():GetActiveWeapon()
+
+      if IsValid(wep) and wep.HUDShouldDraw != nil then
+         local val = wep:HUDShouldDraw(name)
+         if val != nil then return val end
+      end
+   end
+
    return true
 end
 
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_hud.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_hud.lua
@@ -371,15 +371,6 @@ function GM:HUDShouldDraw(name)
       if name == v then return false end
    end
 
-   if IsValid(LocalPlayer()) then
-      local wep = LocalPlayer():GetActiveWeapon()
-
-      if IsValid(wep) and wep.HUDShouldDraw != nil then
-         local val = wep:HUDShouldDraw(name)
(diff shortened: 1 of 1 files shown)


Bo98:
Another possibility is just to call the base gamemode's HUDShouldDraw method (in place of the `return true` line). That way, TTT won't need to be updated if there are any changes to the base gamemode.

markusmarkusz:
@Bo98, yes this would/could be possible.
But I've got some errors that occured when I tested it.
I'm not sure if I made something wrong but I only changed:
```lua
return true
```
to
```lua
self.BaseClass:HUDShouldDraw(name)
```
Basically this should work but It hasn't worked for me.

Bo98:
Make sure you are still returning a value.

i.e.

```lua
return self.BaseClass:HUDShouldDraw(name)
```

markusmarkusz:
@Bo98, I've done that. But I forgot to mention this `return` in that example. Sorry.

Bo98:
No worries.

I forgot to say, calling self.BaseClass like that isn't quite correct since you are passing the base gamemode's table into the function instead of TTT's. The more correct way would be:

```
return self.BaseClass.HUDShouldDraw(self, name)
```

Though this could possibly break derived gamemodes because self.BaseClass could get overridden. The real correct way is:

```
-- Beginning of file or something
DEFINE_BASECLASS("gamemode_base")

-- ...

    return BaseClass.HUDShouldDraw(self, name)
```

markusmarkusz:
@Bo98 
I'll test that later and change it if it works fine for me.

Bo98:
I have no idea if it actually fixes the errors you're getting since you didn't say what those errors were, but the way I said is the more correct way of calling baseclass methods so it should be done that way anyway.

markusmarkusz:
There aren't any error messages.
`self.BaseClass.HUDShouldDraw(self, name)` just returned nil so no HUD was drawn. But I'll recheck this.

markusmarkusz:
Now I know what my problem was.
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/base/gamemode/cl_init.lua#L49-L67
In the base gamemode there is no check if there is really something that the function returns.

My SWEP:HUDShouldDraw() function was this:
```lua
function SWEP:HUDShouldDraw(name)
    if name == "TTTTargetID" then
        return false
    end
end
```
It does only return false for the case that name is "TTTTargetID". In any other case it returns nil.
So HUDShouldDraw works with nil.
I've just added a nil check to the function at the beginning.
`if val != nil then return val end`
But the base gamemode doesn't have any nil check. This was the reason why no HUD was drawn anymore when I tested this.

PR #1263, opened 2016-10-21
Base: Add nil check for HUDShouldDraw Hook

markusmarkusz:
Added a nil check to make it more consistent with the HUDShouldDraw-Hook.

At the moment:
The HUDShouldDraw-Hook doesnt't need to return something.
SWEP:HUDShouldDraw() must return something or all HUD parts are invisible.


robotboy655:
```
function GAMEMODE:HUDShouldDraw( d ) print( d ) end

function SWEP:HUDShouldDraw( d ) print( d ) end
```

Both produce exact same result, so it is consistent. SWEP:HUDShouldDraw is meant to work in the exact same way as GM:HUDShouldDraw as far as I can tell, and so it does.

PR #1267, opened 2016-10-28
TTT: Remove fingerprint functionality from beacons

Faerachnidendro:
~snipped previous description.

Actual issue stems from planted beacon entity receiving reference to beacon weapon's fingerprints table, instead of a new/cached table of it's own - entity attempts to merge the weapon's fingerprints table, with itself, instead of merging it's own fingerprints with that of the weapon(?)

gamemodes/terrortown/entites/weapons/weapon_ttt_beacon.lua @ lines 99 & 151

Latest update: Can't concoct a viable work around for beacons to utilize fingerprints, recommend removing fingerprint functionality from beacons.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -55,7 +55,6 @@ function ENT:UseOverride(activator)
 
       if IsValid(wep) then
          wep.fingerprints = wep.fingerprints or {}
-         -- Fix c-side error which hangs server on beacon pickup, giving only an ambiguous blue "[ERROR]" message in console.
          table.Add(wep.fingerprints, #wep.fingerprints + 1, prints)
       end
    end
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -55,8 +55,7 @@ function ENT:UseOverride(activator)
 
       if IsValid(wep) then
          wep.fingerprints = wep.fingerprints or {}
-         -- Fix error which hangs server on beacon pickup
-         table.Add(wep.fingerprints, #wep.fingerprints + 1, prints)
+         table.Add(wep.fingerprints, prints)
       end
    end
 end
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -50,13 +50,7 @@ function ENT:UseOverride(activator)
          wep = activator:Give("weapon_ttt_beacon")
       end
 
-      local prints = self.fingerprints or {}
       self:Remove()
-
-      if IsValid(wep) then
-         wep.fingerprints = wep.fingerprints or {}
-         table.Add(wep.fingerprints, prints)
-      end
    end
 end
 
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -55,6 +55,7 @@ function ENT:UseOverride(activator)
 
       if IsValid(wep) then
          wep.fingerprints = wep.fingerprints or {}
+         -- Fix error which hangs server on beacon pickup
          table.Add(wep.fingerprints, #wep.fingerprints + 1, prints)
       end
    end
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_beacon.lua
@@ -55,7 +55,8 @@ function ENT:UseOverride(activator)
 
       if IsValid(wep) then
          wep.fingerprints = wep.fingerprints or {}
-         table.Add(wep.fingerprints, prints)
+         -- Fix c-side error which hangs server on beacon pickup, giving only an ambiguous blue "[ERROR]" message in console.
+         table.Add(wep.fingerprints, #wep.fingerprints + 1, prints)
       end
    end
 end
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua
@@ -96,8 +96,6 @@ function SWEP:BeaconDrop()
          beacon:SetOwner(ply)
          beacon:Spawn()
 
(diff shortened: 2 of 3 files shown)


bigdogmat:
Uh, `table.Add` doesn't have a third argument ([source](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/includes/extensions/table.lua#L104-L115))

Faerachnidendro:
Then why is my code working?
Not to be a dick but I can't argue with results.

bigdogmat:
I mean, you can't argue with source either. Are you sure that part of the code is running, i.e putting a print before the function is ran and after will result and both prints being ran? If they both run fine then I can only assume something is overwriting `table.Add`, try putting this in console `PrintTable(debug.getinfo(table.Add))`

Faerachnidendro:
See edits to above post :P
Bad troubleshooting on my part, will either edit this pull or replace it with an actual fix.

bigdogmat:
Ah, ok

Faerachnidendro:
AH!
I see where I've gone wrong...

table.Add is misusing table.insert, in the way(or similar) that i thought table.Add was being misused.

I'm new to this git hub business - am I able to entirely change this pull or should I close it and make a new one?

willox:
`table.Add( { 1, 2, 3 }, { 4, 5, 6 } )` is meant to return `{ 1, 2, 3, 4, 5, 6 }`. Your changes would break that. I'd suggest you just try to fix any issues TTT has in TTT specifically rather than changing any behaviour in the libraries every other addon uses.

Faerachnidendro:
Yeah I just caught on to that xD
was reading over it now and thinking of dirrent work arounds, as I'm pretty sure the issue isn't with the fingerprints table being added, but how table.Add handles the table - unless of course I DID miss something else in the ttt_beacon entity?

Either way, my bad for the not so useful attempt at a fix here, feel free to close and delete this one until I figure it out :P

bigdogmat:
The only thing that would cause `table.Add` to hang the server would be if it was called with a huge table as it's source. _Nothings_ wrong with the way the function itself handles tables.

Faerachnidendro:
wait, so, if table.Add does intend for the dest table to remain sequential, shouldn't table.insert be providing #dest + 1 as arg 2? with v from the source loop as arg3?

bigdogmat:
If `table.insert` is not given a third argument then it'll use the second argument as the value, and push onto the end of the table

Faerachnidendro:
I understand that this is the expected behavior, but must admit, it has always worked falwlessly when providing the desired index, and errored in some fasion otherwise - personal experience, no modification of either table.Add or table.insert

bigdogmat:
As I forgot in my earlier comment, `table.Add` does type checking on both arguments, if source is not a table then it'll return dest.

Faerachnidendro:
Yeah I saw that, I'm not saying you're wrong so to speak - even the wiki confirms the expected behavior of table.insert, but in practice(personally), table.insert is unreliable when not given the desired index.

bigdogmat:
`table.insert` when used without a index isn't unreliable at all, if it is then you're using it incorrectly. By default it tries to push it into the end of the table, if the table has holes in it then it isn't sequential and as such `table.insert` has no idea where to put the value. As such it's undefined behavior

Faerachnidendro:
If the table wasn't sequential, that would be the error message, no? (genuinely asking)

bigdogmat:
No, just the behavior of where it places the value would be undefined

Faerachnidendro:
Heh, fair enough.
Just recall seeing errors about non sequential tables before( probably from iterator functions ).

Alright so then I guess it goes back to something else being wrong with the ttt_beacon entity.... got me fuckin stumped now though.

Faerachnidendro:
I'm gonna close this for now until I can figure out what's going on - but, that said, the blue "[ERROR]" message and server hang come back if I revert my changes to the ttt_beacon entity; the call to table.Add on line 58 of gamemodes/terrortown/entities/entities/ttt_beacon.lua, is throwing the error.

bigdogmat:
Then that means the issues manifests from adding the entities fingerprints table to the weapons fingerprints tables. So there's your thing to go off of

Faerachnidendro:
Well yeah, but nothing is out of the ordinary there, unless it has something to do with the source table(ent's fingerprints) being empty?

Faerachnidendro:
Oh for the love of... figured it out... it's sending the exact same table as both arg1 and arg2...

Faerachnidendro:
gamemodes/terrortown/entities/weapons/weapon_ttt_beacon.lua @ line 99

Planted beacon entity recieves reference to weapon's fingerprints table, instead of new/cached fingerprints table of it's own - genius.

bigdogmat:
I thought about this, though that would result in an out of memory crash, or at least it should.

Faerachnidendro:
Well to be fair, you also thought that the only way table.Add would hang the server, is if the supplied tables were overpopulated - indeed, perhaps this explains the hang perfectly?

willox:
What error did you get in your server's console, anyway?

Faerachnidendro:
literally "[ERROR]" in in 0,255,255 teal, nothing more, nothing less(except the hang)

bigdogmat:
That's weird, it should crashed the entire server with an out of memory error, instead it prevents it and prints out a blue `[ERROR]` in console

willox:
I guess if Lua runs out of memory in a protected context it's just going to error and then run the GC afterwards and carry on chugging. I'd expect whatever code is stitching the error message together to actually handle not being able to allocate a buffer for the message though.

Faerachnidendro:
I should be able to knock up a fix to the way the beacon works, to avoid this issue resurfacing in this context - as for the "[ERROR]" thing, maybe an if dest == source then return dest end condition being added to table.Add ?

Faerachnidendro:
At this point, I'm not seeing any reason for beacons to even record fingerprints - let alone an efficient and consistent way to cache the fingerprints of each placed beacon, poping them in and out as required on beacon spawn/pickup.
Mostly due to the way weapon_ttt_beacon removes and re-adds itself to the user's inventory, when exhausting all ammo or when picking up a placed beacon, having had no ammo left - using the weapon's SWEP:Clip1() value as a means for determining which fingerprints to give a beacon, results in inconsistencies which occasionally lead to inappropriate erasing of previously cached fingerprints.

The best course of action to rectify the error being thrown by the ttt_beacon entity, in my opinion, would be to remove all fingerprint related code for beacons, which shouldn't change much of anything about how they're used(if anyone even does use them).

svdm:
There may be a fix involving copying the fingerprints to the beacon instead of setting a reference, and removing duplicate prints after merging them back to the weapon on pickup, but whether any of the fingerprint behavior still makes sense after that is another thing.

Faerachnidendro:
That's what I tried at first, but was too tired to pursue it.
Might have another bash at it again soon.

PR #1268, opened 2016-10-28
TTT: Added hooks to allow easier modification of corpse id/search behavior.

Faerachnidendro:
I would like to propose these hook additions which would allow modders (if not just myself, others also) to easily create weapons and entities and other conditions which interrupt the ability for players to identify/search a corpse - without making changes to the corpse.lua file.
I am happy to privately share examples where these are sorely needed, for at least my own purposes - but let your imagination run wild a bit yourself ;)

**_"TTTCanIdentifyCorpse"_ (ply, corpse, corpse_was_traitor)**
Called to determine if a player can identify the corpse they are trying to inspect; return true to allow, false to disallow.

**_"TTTCanSearchCorpse"_ (ply, corpse, corpse_was_traitor)**
Called to determine if a player can view the search interface for the corpse they are trying to inspect; return true to allow, false to disallow.

**_"TTTOnCorpseCreated"_ (corpse)**
Called after a player has been killed and their corpse has been created and spawned with all essential variables set.

Ideally, this would be something for traitors to benefit from only( and in fair ways pertaining to the use of equipment items ), though it could be abused in ways that really ruin the gameplay - that said, if someone were determined enough to do that, they would almost certainly have done so already with or without such hooks.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
@@ -36,6 +36,12 @@ end
 -- If detective mode, announce when someone's body is found
 local bodyfound = CreateConVar("ttt_announce_body_found", "1")
 
+-- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse searching.
+function GM:TTTInterruptCorpseIdentify(identifier,corpse,corpse_is_traitor)
+   -- return true to interrupt corpse identification
+   return false
+end
+
 local function IdentifyBody(ply, rag)
    if not ply:IsTerror() then return end
 
@@ -48,6 +54,11 @@ local function IdentifyBody(ply, rag)
    local finder = ply:Nick()
    local nick = CORPSE.GetPlayerNick(rag, "")
    local traitor = (rag.was_role == ROLE_TRAITOR)
+   
+   -- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse identification.
+   if hook.Run("TTTInterruptCorpseIdentify", ply, rag, traitor) then
+      return
+   end
 
    -- Announce body
    if bodyfound:GetBool() and not CORPSE.GetFound(rag, false) then
@@ -169,6 +180,12 @@ local function bitsRequired(num)
    return bits
 end
 
+-- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse searching.
+function GM:TTTInterruptCorpseSearch(identifier,corpse,corpse_is_traitor)
+   -- return true to interrupt corpse search
+   return false
+end
+
 -- Send a usermessage to client containing search results
 function CORPSE.ShowSearch(ply, rag, covert, long_range)
    if not IsValid(ply) or not IsValid(rag) then return end
@@ -189,6 +206,11 @@ function CORPSE.ShowSearch(ply, rag, covert, long_range)
    local words = rag.last_words or ""
    local hshot = rag.was_headshot or false
    local dtime = rag.time or 0
+   
+   -- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse searching.
+   if hook.Run("TTTInterruptCorpseSearch", ply, rag, traitor) then
+      return
+   end
 
    local owner = player.GetBySteamID(rag.sid)
    owner = IsValid(owner) and owner:EntIndex() or -1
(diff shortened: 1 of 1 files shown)


Faerachnidendro:
Curiosity, any particular reason for the thumbs down there?
I'm open to criticism/questions/opinions here :)

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
```diff
@@ -36,6 +36,12 @@ end
 -- If detective mode, announce when someone's body is found
 local bodyfound = CreateConVar("ttt_announce_body_found", "1")
 
+-- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse searching.
+function GM:TTTInterruptCorpseIdentify(identifier,corpse,corpse_is_traitor)
```
svdm: Spaces after commas, same for the other hook below.

svdm: Also, I would suggest `TTTCanIdentifyCorpse` and `TTTCanSearchCorpse` as hook names, with the default implementation returning true. This fits in a bit better with standard hooks like PlayerCanPickupWeapon.

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
```diff
@@ -48,6 +54,11 @@ local function IdentifyBody(ply, rag)
    local finder = ply:Nick()
    local nick = CORPSE.GetPlayerNick(rag, "")
    local traitor = (rag.was_role == ROLE_TRAITOR)
+   
+   -- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse identification.
+   if hook.Run("TTTInterruptCorpseIdentify", ply, rag, traitor) then
+      return
+   end
```
svdm: Move above the block of `local`s

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
```diff
@@ -189,6 +206,11 @@ function CORPSE.ShowSearch(ply, rag, covert, long_range)
    local words = rag.last_words or ""
    local hshot = rag.was_headshot or false
    local dtime = rag.time or 0
+   
+   -- opportunity for addons to introduce ways for traitors(ideally) to interrupt corpse searching.
+   if hook.Run("TTTInterruptCorpseSearch", ply, rag, traitor) then
+      return
+   end
```
svdm: Also move above the locals, but after the `IsOnFire` check, unless you move that to the hook's default implementation.

robotboy655:
@svdm, if you could look into the PRs on page 2, I'd appreciate that.

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
Faerachnidendro: woops, hasty fingers, sorry

Faerachnidendro: I opted for default implementation as false(or nil), in case modders wanted to easily break out of all hook checks without interrupting the id/search, but have no arguments against your suggestion there.

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
Faerachnidendro: I was going to leave this as is, so that modders are less inclined to override this default behavior of preventing id/search for burning corpses - are you sure you'd prefer this to be in the GM hook func?

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
fruitwasp: the `traitor` variable is one of the locals, which is used in the hook call, so it only the other locals should be moved below the hook call.

garrysmod/gamemodes/terrortown/gamemode/corpse.lua
Faerachnidendro: Done, hard to decide if IsOnFire check should be within hook's default implementation, thoughts?

Faerachnidendro:
2 errors in merged commit, have been fixed in my branch, should (I/we) revert this?

svdm:
I believe it's easiest/cleanest if you create a new PR for those fixes.

Faerachnidendro:
Righto, will do :)

PR #1270, opened 2016-11-01
TTT: Fix mistakes in corpse.lua relating to previous commit.

Faerachnidendro:
This fixes several mistakes/omissions regarding the last commit I made for this file, which introduced 3 new hooks to allow modders to more-easily modify the default corpse id/search behavior.

**_"TTTCanIdentifyCorpse"_ (ply, corpse, corpse_was_traitor)**
Called to determine if a player can identify the corpse they are trying to inspect; return true to allow, false to disallow.

**_"TTTCanSearchCorpse"_ (ply, corpse, covert, long_range, corpse_was_traitor)**
Called to determine if a player can view the search interface for the corpse they are trying to inspect; return true to allow, false to disallow.

**_"TTTOnCorpseCreated"_ (corpse, corpse_player)**
Called after a player has been killed and their corpse has been created and spawned with all essential variables set.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
@@ -192,7 +192,7 @@ function CORPSE.ShowSearch(ply, rag, covert, long_range)
       return
    end
    
-   if not hook.Run("TTTCanSearchCorpse", ply, rag, (rag.was_role == ROLE_TRAITOR)) then
+   if not hook.Run("TTTCanSearchCorpse", ply, rag, covert, long_range, (rag.was_role == ROLE_TRAITOR)) then
       return
    end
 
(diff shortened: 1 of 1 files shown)


Faerachnidendro:
o_O
Why's this telling me there's a conflict? What've I done now? :sob:

svdm:
The PR includes commits that add all the stuff that the previous PR also added, which leads to a conflict git can't resolve automatically.

If I read them right, the last three commits of this PR contain your fixes, correct? If that is the case, I can simply apply the changes myself.

fruitwasp:
@svdm last 4 commits

Faerachnidendro:
Correct, the last 4, sorry for the mess around.

svdm:
Changes applied in  d7aa60b

PR #1271, opened 2016-11-06
TTT: Fixed very old C4 Bug

markusmarkusz:
This bug exists since 9 January 2014.
It causes that the following part of the C4 doesn't work:
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua#L423-L440

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
@@ -457,9 +457,8 @@ end
 ---- Communication
 
 local function C4ConfigHook()
-   local idx = net.ReadUInt(16)
+   local bomb = net.ReadEntity(16)
 
-   local bomb = ents.GetByIndex(idx)
    if IsValid(bomb) then
       if not bomb:GetArmed() then
          ShowC4Config(bomb)
@@ -471,12 +470,9 @@ end
 net.Receive("TTT_C4Config", C4ConfigHook)
 
 local function C4DisarmResultHook()
-   local idx = net.ReadUInt(16)
-   local result = {}
-
+   local bomb = net.ReadEntity()
    local correct = net.ReadBit() == 1
 
-   local bomb = ents.GetByIndex(idx)
    if IsValid(bomb) then
       if correct and disarm_success then
          disarm_success()
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
@@ -457,7 +457,7 @@ end
 ---- Communication
 
 local function C4ConfigHook()
-   local bomb = net.ReadEntity(16)
+   local bomb = net.ReadEntity()
 
    if IsValid(bomb) then
       if not bomb:GetArmed() then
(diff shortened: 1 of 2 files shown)


svdm:
Nice find. I don't think the writing code is the problem here though. Looking at the client code for that net message, it seem that the problem is simply that the corresponding `ReadUInt` will read 16 bits instead of 15: https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua#L474

This micro-optimisation was introduced in the PR converting TTT to the `net` library, and it's a lesson on why such bit-counting optimisations are not worth it when they're used on messages that are rarely sent in the first place.

I think my preferred fix would be replacing this UInt nonsense with `WriteEntity` and `ReadEntity`. That's going to do the same thing without the risk of getting the number of bits wrong when writing/reading.

markusmarkusz:
I'll do that on saturday when I'm back at home.

Am 07.11.2016 3:09 nachm. schrieb "Stefan van der Meer" <
notifications@github.com>:

markusmarkusz:
@svdm 
>I think my preferred fix would be replacing this UInt nonsense with WriteEntity and ReadEntity. That's going to do the same thing without the risk of getting the number of bits wrong when writing/reading.

Done.

When I tested the changes I saw this message when I disarmed the bomb:
```
Create Stream Failed error 41
Failed to load sound "ttt\wirecut.mp3", file probably missing from disk/repository
```
Is this problem related to the lua code of the bomb? (I'm sure it isn't but asking is the best thing.)
Btw. the sound files exists.
```
] lua_run_menu print(file.Exists("gamemodes/terrortown/content/sound/ttt/wirecut.mp3", "GAME"))
true
```

garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
```diff
@@ -457,9 +457,8 @@ end
 ---- Communication
 
 local function C4ConfigHook()
-   local idx = net.ReadUInt(16)
+   local bomb = net.ReadEntity(16)
```
Bo98: Don't need that 16 anymore.

svdm:
No clue why the sound would not be found. It used to work, of course, but given that this net msg bug has existed for a while something must have changed in the meantime. Possibly in gmod's mounting of a gamemode's `content/sound` directory.

willox:
Sound errors suck. Error 41 is BASS_ERROR_FILEFORM which means the file format isn't supported. Your other MP3 files work, so I guess it's something specifically wrong with wirecut.mp3.

robotboy655:
Should just use .wav for sound effects tbh

markusmarkusz:
I'll convert them and test it.
If it works then I'll upload them later in a PR.

Bo98:
Converting MP3 to WAV is useless unless it fixes the error. You've already lost quality if you're converting from the MP3, so no benefit will be gained.

Re-encoding the MP3 might be enough to fix this though.

markusmarkusz:
Then can you do that? Or tell me how to do that?

svdm:
I've converted it to WAV and though I don't have time to test it in-game right now, it at least makes certain audio players happier than the MP3 version. Not sure why because it's a pretty standard MP3 (128kbps CBR mono), maybe the fact that it's less than a second long is somehow an issue.

Either way I'll test the WAV soon and commit it if it works out.

Bo98:
So where there other audio players complaining about the MP3 file?

PR #1272, opened 2016-11-06
TTT: Added some useful things

markusmarkusz:
A small overview of the changes and some informations about them:
- I changed the ConCommand that requires `sv_cheats 1`. Now the Flag `FCVAR_CHEAT` is used because this is a bit better because it tells the client that the command requires `sv_cheats 1`.
- Added a hook to prevent using a traitor button. (Could be useful for some addons.)
Hook Name: `TTTCanUseTraitorButton`
Arguments: `ttt_traitor_button-Entity; Player that used the Button` (`self, ply`)
Return: Return false to prevent using it. You can also return a message that is shown up if you aren't allowed to use the button. (Example: `return false, "Not allowed."`)
- Added a second argument to `TTTEndRound` (Serverside only) that gives the information if the round end was triggered by the map.
(Could be useful for some addons.)
- Added a hook that is triggered when a user toggles his disguiser.
Hook Name: `TTTCanToggleDisguiser`
Arguments: `player that toggles the disguiser; boolean that tells to which state the disguiser would be changed.`
Return: Return false to prevent toggling the disguiser.
(This is very useful for addons like the [PowerRound-Addon](https://scriptfodder.com/scripts/view/401/) or the [TTTDamagelog-Addon](https://github.com/Tommy228/TTTDamagelogs/blob/master/lua/damagelogs/damagelog_events/misc.lua#L22-L29).)

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
@@ -126,18 +126,16 @@ if SERVER then
    end
 
    local function TestHat(ply, cmd, args)
-      if cvars.Bool("sv_cheats", 0) then
-         local hat = ents.Create("ttt_hat_deerstalker")
+      local hat = ents.Create("ttt_hat_deerstalker")
 
-         hat:SetPos(ply:GetPos() + Vector(0,0,70))
-         hat:SetAngles(ply:GetAngles())
+      hat:SetPos(ply:GetPos() + Vector(0,0,70))
+      hat:SetAngles(ply:GetAngles())
 
-         hat:SetParent(ply)
+      hat:SetParent(ply)
 
-         ply.hat = hat
+      ply.hat = hat
 
-         hat:Spawn()
-      end
+      hat:Spawn()
    end
-   concommand.Add("ttt_debug_testhat", TestHat)
+   concommand.Add("ttt_debug_testhat", TestHat, nil, nil, FCVAR_CHEAT)
 end
--- a/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
@@ -308,4 +308,4 @@ function PANEL:DoRightClick()
    menu:Open()
 end
 
-vgui.Register( "TTTScorePlayerRow", PANEL, "Button" )
+vgui.Register( "TTTScorePlayerRow", PANEL, "DButton" )
--- a/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
@@ -425,11 +425,17 @@ local function OrderEquipment(ply, cmd, args)
 end
 concommand.Add("ttt_order_equipment", OrderEquipment)
 
+function GM:TTTCanToggleDisguiser(ply, state)
+   -- Can be used to prevent players from using this button.
+   return true
+end
+
 local function SetDisguise(ply, cmd, args)
    if not IsValid(ply) or not ply:IsActiveTraitor() then return end
 
    if ply:HasEquipmentItem(EQUIP_DISGUISE) then
       local state = #args == 1 and tobool(args[1])
+      if not hook.Run("TTTCanToggleDisguiser", ply, state) then return end
 
       ply:SetNWBool("disguised", state)
       LANG.Msg(ply, state and "disg_turned_on" or "disg_turned_off")
@@ -438,11 +444,11 @@ end
 concommand.Add("ttt_set_disguise", SetDisguise)
 
 local function CheatCredits(ply)
-   if cvars.Bool("sv_cheats", false) and IsValid(ply) then
+   if IsValid(ply) then
       ply:AddCredits(10)
    end
 end
-concommand.Add("ttt_cheat_credits", CheatCredits)
+concommand.Add("ttt_cheat_credits", CheatCredits, nil, nil, FCVAR_CHEAT)
 
 local function TransferCredits(ply, cmd, args)
    if (not IsValid(ply)) or (not ply:IsActiveSpecial()) then return end
--- a/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
@@ -425,9 +425,9 @@ local function OrderEquipment(ply, cmd, args)
 end
 concommand.Add("ttt_order_equipment", OrderEquipment)
 
-function GM:TTTCanToggleDisguiser(ply, state)
+function GM:TTTToggleDisguiser(ply, state)
    -- Can be used to prevent players from using this button.
-   return true
+   -- return false to prevent it.
 end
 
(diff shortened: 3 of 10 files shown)


markusmarkusz:
Shit. I see that I've made a lot of things wrong with the hooks. :c
I'll correct them on saturday. Ignore this until I've made the changes. Thanks.

mcNuggets1:
Add C4 hooks :D

markusmarkusz:
I could do this, but not yet. I'm not at home this week.

Am 07.11.2016 12:59 nachm. schrieb "mcNuggets1" <notifications@github.com>:

markusmarkusz:
First I changed `TTTCanToggleDisguiser` to `TTTToggleDisguiser`. And from now on it can be used to just check who switched the disguiser to which state. It only blocks disguising if you return false.
I also removed a the MapWin argument again. This way would never work.
These are possible other ways:
- GAMEMODE.MapWin is set to true if the end was triggered by the map.
- a new custom GAMEMODE-Var is added and set to true if the end was triggered by the map.
- GAMEMODE.MapWin isn't set to WIN_NONE after TTTCheckForWin knows if the map triggers round end.
- We leave it as it is.
Chose the way you like.

markusmarkusz:
@mcNuggets1 
> Add C4 hooks :D

For which events? (Disarm, Arm, Destroy, Pickup, Explode?)
Also what about [TTTDamagelogs](https://github.com/Tommy228/TTTDamagelogs)?
Should the hooks have the same names like the one from TTTDamagelogs? The functions are overwritten so it shouldn't be the biggest problem. [Mainly because Tommy's Damagelog breaks some C4 functions.](https://github.com/Tommy228/TTTDamagelogs/blob/master/lua/damagelogs/sv_stupidoverrides.lua).

markusmarkusz:
I hope the way I calculate the hp is correct.

garrysmod/gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
```diff
@@ -126,18 +126,16 @@ if SERVER then
...
 
-         ply.hat = hat
+      ply.hat = hat
 
-         hat:Spawn()
-      end
+      hat:Spawn()
    end
-   concommand.Add("ttt_debug_testhat", TestHat)
+   concommand.Add("ttt_debug_testhat", TestHat, nil, nil, FCVAR_CHEAT)
```
bmwalters: Should this function even exist?

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -308,4 +308,4 @@ function PANEL:DoRightClick()
    menu:Open()
 end
 
-vgui.Register( "TTTScorePlayerRow", PANEL, "Button" )
+vgui.Register( "TTTScorePlayerRow", PANEL, "DButton" )
```
bmwalters: What's the point of this change?

markusmarkusz: Actually, this change doesn't really matter.
[Button](http://wiki.garrysmod.com/page/Category:Button) is deprecated. Changing this does not make a big difference. But I had the time and making a new PR for something like this doesn't really make sense.

garrysmod/gamemodes/terrortown/entities/entities/ttt_hat_deerstalker.lua
markusmarkusz: This may be useful for model creators.
They can test if the hat fits to the models head.

mcNuggets1:
TTT Damagelogs is an outdated addons with a lot of bugs.

markusmarkusz:
@mcNuggets1 
> TTT Damagelogs is an outdated addon with a lot of bugs, even without C4 changes

Yes, TTT Damagelogs is outdated. The same thing is with the SpecDM.
But I don't want to break a addon that is on almost every TTT server.
It would be nice if the addon is “officially“ continuted. Perhaps by someone else. But until that I really want to avoid breaking one of these two addons.

mcNuggets1:
Progress is the future.

By the way I don't think adding a hook breaks TTTDamagelogs, it just makes the solution of TTTDamagelogs integrating C4 hooks useless.

markusmarkusz:
> @mcNuggets1 
> > Add C4 hooks
>
> For which events? (Disarm, Arm, Destroy, Pickup, Explode?)
Also what about [TTTDamagelogs](https://github.com/Tommy228/TTTDamagelogs)?
Should the hooks have the same names like the one from TTTDamagelogs? The functions are overwritten so it shouldn't be the biggest problem. [Mainly because Tommy's Damagelog breaks some C4 functions.](https://github.com/Tommy228/TTTDamagelogs/blob/master/lua/damagelogs/sv_stupidoverrides.lua).

I asked how the hooks should be named.
I know that this shouldn't break anything.
But what if:
I add the hook “TTTPickupC4“.
I create an addon that uses this hook.
A server use this addon and the damagelog.
=> My addon is broken.
If I call the hook “TTTC4Pickup“ it wouldn't be broken.

My question was if I should name these hooks like the one from the Damagelog or if should give them other names.

mcNuggets1:
Ask svdm.

It is not my gamemode.

However if the hooks would do the same it would not break anything either.

markusmarkusz:
I hate GitHub.
I'll fix this merge conflicts as soon as possible.

MinIsMin:
@mcNuggets1
> TTT Damagelogs is an outdated addon with a lot of bugs, even without C4 changes

Can you list these bugs?
These C4 hooks are coming in handy.
Currently I'm rewriting the Damagelogs in consultation with Tommy.

markusmarkusz:
@MinIsMin
https://facepunch.com/showthread.php?t=1500900
Btw. I have a DamageLog that is 10 MB big.


I'll add these hooks when I know how to fix these annoying merge conflicts.

markusmarkusz:
@svdm
Would you accept this PR?
Any suggestions?

Should I create a seperate PR for the C4 Hooks? I really don't know how to fix these merge conflicts.

svdm:
Sure, make a separate PR. I actually much prefer a PR to only make changes related to a single feature/bug, that makes it faster to review and merge.

markusmarkusz:
@svdm
Okay. Thanks for the hint. I will keep this in mind in the future.

garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
```diff
@@ -425,11 +425,17 @@ local function OrderEquipment(ply, cmd, args)
...
+   -- Can be used to prevent players from using this button.
+   -- return false to prevent it.
+end
+
 local function SetDisguise(ply, cmd, args)
    if not IsValid(ply) or not ply:IsActiveTraitor() then return end
 
    if ply:HasEquipmentItem(EQUIP_DISGUISE) then
       local state = #args == 1 and tobool(args[1])
+      if not hook.Run("TTTToggleDisguiser", ply, state) == false then return end
```
markusmarkusz: Can it be that I have the disguised? I'll test this again later.
(Because I'm checking if the hook returns anything other then false. So if it returns nil it wouldn't work, would it?)

willox: returning nil will have the same effect as returning false there, which i think you don't want

PR #1273, opened 2016-11-12
Add table.ShallowCopy and use it instead of table.Copy in SortedPairs* and RandomPairs

bmwalters:
Previously, said functions were unnecessarily creating deep copies of the input table, which is slower than a simple shallow copy. I have also made it a function in the table library so others can use it without writing extra code.

This fixes Facepunch/garrysmod-issues#2183

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -514,7 +514,7 @@ end
 -----------------------------------------------------------]]
 function SortedPairs( pTable, Desc )
 
-	pTable = table.Copy( pTable )
+	pTable = table.ShallowCopy( pTable )
 
 	local SortedIndex = {}
 	for k, v in pairs( pTable ) do
@@ -539,7 +539,7 @@ end
 -----------------------------------------------------------]]
 function SortedPairsByValue( pTable, Desc )
 
-	pTable = table.Copy( pTable )
+	pTable = table.ShallowCopy( pTable )
 
 	local SortedIndex = {}
 	for k, v in pairs( pTable ) do
@@ -568,7 +568,7 @@ end
 -----------------------------------------------------------]]
 function SortedPairsByMemberValue( pTable, pValueName, Desc )
 
-	pTable = table.Copy( pTable )
+	pTable = table.ShallowCopy( pTable )
 	Desc = Desc or false
 
 	local pSortedTable = table.ClearKeys( pTable, true )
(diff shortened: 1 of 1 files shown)


bmwalters:
Also can someone verify my line endings? I committed Windows line endings because it seemed like that's what the rest of the file was using, but I'm not sure if that's correct.

markusmarkusz:
The line endings are correct.
All lua files use them (`CR LF`).

bmwalters:
You know what? I didn't even see #1167. That one's better, aside from the coding standards issues.

Kefta:
Many requests just aren't merged even if there aren't problems

bmwalters:
@Kefta true.
So... I decided to rewrite my own SortedPairs* and RandomPairs from scratch to see what I ended up with. Here's what I made:
```lua
local function GenericSortedPairs(tab, sortfunc)
	local keys = table.GetKeys(tab)

	table.sort(keys, sortfunc)

	local i = 0

	return function()
		i = i + 1
		if i <= #keys then
			local key = keys[i]
			return key, tab[key]
		end
	end
end

function SortedPairs(tab, descending)
	return GenericSortedPairs(tab, function(a, b)
		if descending then
			return a > b
		else
			return a < b
		end
	end)
end

function SortedPairsByValue(tab, descending)
	return GenericSortedPairs(tab, function(a, b)
		if descending then
			return tab[a] > tab[b]
		else
			return tab[a] < tab[b]
		end
	end)
end

function SortedPairsByMemberValue(tab, valuename, descending)
	return GenericSortedPairs(tab, function(a, b)
		if descending then
			return tab[a][valuename] > tab[b][valuename]
		else
			return tab[a][valuename] < tab[b][valuename]
		end
	end)
end

function RandomPairs(tab)
	return GenericSortedPairs(tab, function()
		return math.random() > 0.5
	end)
end
```

It's quite elegant and reads well (compared to both how it is currently and with #1167) and should perform better than current (only copying keys, no more __sortedindex garbage) and about the same as #1167. Additionally, the GenericSortedPairs helper function I wrote could actually be quite useful for people; they wouldn't be confined to the existing *Pairs functions; they could simply write their own sorting function if they wish.

Thoughts?

Faerachnidendro:
This was starting to bug me just the other week, +1 for this pr to be merged.

robotboy655:
I have looked into this PR, and just like #1167 this one has a problem, due to how SortedPairsByMemberValue works, it alters the input table, which is unacceptable.

PR #1274, opened 2016-11-12
Streamline and better options

nubpro:
Streamlined damage logs and added a convar option to allow or disallow detective to damage their own health station. By default, they can't.

Thanks!

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_health_station.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_health_station.lua
@@ -136,31 +136,6 @@ function ENT:Use(ply)
    end
 end
 
--- traditional equipment destruction effects
-function ENT:OnTakeDamage(dmginfo)
-   if dmginfo:GetAttacker() == self:GetPlacer() then return end
-
-   self:TakePhysicsDamage(dmginfo)
-
-   self:SetHealth(self:Health() - dmginfo:GetDamage())
-
-   local att = dmginfo:GetAttacker()
-   if IsPlayer(att) then
-      DamageLog(Format("%s damaged health station for %d dmg",
-                       att:Nick(), dmginfo:GetDamage()))
-   end
-
-   if self:Health() < 0 then
-      self:Remove()
-
-      util.EquipmentDestroyed(self:GetPos())
-
-      if IsValid(self:GetPlacer()) then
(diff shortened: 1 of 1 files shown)


bmwalters:
In the future, when you create a pull request that does multiple things, changes should be in separate commits for clarity. For example, I would create one commit that moves the `OnTakeDamage` function into the `if SERVER` block, one commit that changes the DamageLog format, and one that adds the ConVar, instead of putting it in one commit and putting this information in the description.

As far as the actual change, the name ""ttt_detective_dmg_selfhs" is a bit cumbersome, perhaps "ttt_damage_own_healthstation" is better?

garrysmod/gamemodes/terrortown/entities/entities/ttt_health_station.lua
```diff
@@ -171,5 +146,31 @@ if SERVER then
...
+      if dmginfo:GetAttacker() == self:GetPlacer() and not ttt_detective_dmg_selfhs:GetBool() then return end
+   
+      self:TakePhysicsDamage(dmginfo)
+
+      self:SetHealth(self:Health() - dmginfo:GetDamage())
+
+      local att = dmginfo:GetAttacker()
+      local placer = self:GetPlacer()
+      if IsPlayer(att) then
+         DamageLog(Format("DMG: \t %s [%s] damaged health station [%s] for %d dmg", att:Nick(), att:GetRoleString(), placer:Nick(), dmginfo:GetDamage()))
```
svdm: You'll have to check here if `IsPlayer(placer)`, because that player may have disconnected after placing the health station. If I recall correctly, other damagelog messages tend to print something like "disconnected" as the nickname if that happens.

garrysmod/gamemodes/terrortown/entities/entities/ttt_health_station.lua
```diff
@@ -171,5 +146,31 @@ if SERVER then
          nextcharge = CurTime() + self.RechargeFreq
       end
    end
-end
 
+   local ttt_detective_dmg_selfhs = CreateConVar("ttt_detective_dmg_selfhs", "0") -- 0 as detective cannot damage their own health station
```
svdm: The suggested `ttt_damage_own_healthstation` would be more clear

nubpro:
Thank you so much for the feedback, I've modified the scripts accordingly.

PR #1275, opened 2016-11-13
Fix CallOnRemove on Players

Python1320:
Fix http://wiki.garrysmod.com/page/Entity/CallOnRemove on players.
The player entity in EntityRemoved is not the player entity in PlayerDisconnected, it does not even have the old player's entity table.

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -139,7 +139,8 @@ local function DoDieFunction( ent )
 
 end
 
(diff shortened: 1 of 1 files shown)


mcNuggets1:
Should CallOnRemove even work on players and isnt there a better way of fixing that?

Kefta:
I would say EntityRemoved not being called for players is the actual issue.

Python1320:
Better way is to fix in engine, but until then this is the way to fix it. It probably should work, since Players are entities. Current situation is definitely wrong though.

robotboy655:
This should be fixed in engine since last update.

PR #1277, opened 2016-11-14
Fix Anti-AFK Exploit

end360:
Fixes an exploit to bypass the AFK switcher. The exploit is to use the alias console command to overwrite the "ttt_spectator_mode" console command. This fixes it because toggle treats ttt_spectator_mode as a ConVar, which it is, whereas just calling ttt_spectator_mode 1 treats it as a concommand, and the alias exploit overwrites the concommand (I don't actually know if this is how this works, I don't know the engine of the game, I'm just making a conjecture of why this works.) If needed I can provide the exploit but I'm sure you can find it out if you're on Github.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
@@ -161,9 +161,9 @@ end
 concommand.Add("ttt_force_detective", force_detective)
 
 
-local function force_spectate(ply, cmd, arg)
+local function force_spectate(ply, arg)
    if IsValid(ply) then
-      if #arg == 1 and tonumber(arg[1]) == 0 then
+      if !arg then
          ply:SetForceSpec(false)
       else
          if not ply:IsSpec() then
@@ -179,6 +179,7 @@ local function force_spectate(ply, cmd, arg)
       end
    end
 end
-concommand.Add("ttt_spectate", force_spectate)
-
+net.Receive("ttt_spectate", function(l,pl)
+   force_spectate(pl,net.readBool())
+end)
 
--- a/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
@@ -179,6 +179,7 @@ local function force_spectate(ply, arg)
       end
    end
 end
+concommand.Add("ttt_spectate", force_spectate)
 net.Receive("ttt_spectate", function(l,pl)
    force_spectate(pl,net.ReadBool())
 end)
--- a/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
@@ -181,6 +181,6 @@ local function force_spectate(ply, cmd, arg)
 end
 concommand.Add("ttt_spectate", force_spectate)
 net.Receive("ttt_spectate", function(l, pl)
-   force_spectate(pl, null, net.ReadInt(1))
+   force_spectate(pl, null, net.ReadInt(2))
 end)
 
--- a/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
@@ -1,4 +1,4 @@
-
+util.AddNetworkString("ttt_spectate")
 function GetTraitors()
    local trs = {}
    for k,v in ipairs(player.GetAll()) do
@@ -180,7 +180,7 @@ local function force_spectate(ply, arg)
    end
 end
 concommand.Add("ttt_spectate", force_spectate)
-net.Receive("ttt_spectate", function(l,pl)
-   force_spectate(pl,net.ReadBool())
+net.Receive("ttt_spectate", function(l, pl)
+   force_spectate(pl, net.ReadBool())
 end)
 
--- a/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
@@ -1,4 +1,3 @@
-util.AddNetworkString("ttt_spectate")
 function GetTraitors()
    local trs = {}
    for k,v in ipairs(player.GetAll()) do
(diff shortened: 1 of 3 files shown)


bmwalters:
So what you're saying is, running `alias ttt_spectator_mode echo` prevents the default AFK checker from running. Your proposed fix is to use the `toggle` concommand which correctly treats the named convar as a convar, regardless of alias. Wouldn't a cleaner fix be:
```lua
GetConVar("ttt_spectator_mode"):SetInt(1)
```

markusmarkusz:
Is this really useful?
Can you set an alias for concommands?
If so then this wouldn't really help because changing ttt_spectator_mode only executes ttt_spectate. So if it is possible to blocl ttt_spectate this wouldn't help.

I think it would be better to make a function like Player:ForceSpectator() that forces user in spectator mode whether ttt_spectate or ttt_spectator_mode are working.

markusmarkusz:
`toggle` doesn't work because it's blocked.
```
] ttt_spectate 
*DEAD* markusmarkusz: hi 
] ttt_spectator_mode 0
*DEAD* markusmarkusz: hi 
] lua_run_cl GetConVar("ttt_spectator_mode"):SetInt(1)
*DEAD* markusmarkusz: hi 
] lua_run_cl RunConsoleCommand("toggle","ttt_spectator_mode", 1)
RunConsoleCommand: Command is blocked! (toggle)[markusmarkusz|2|STEAM_0:1:83494461] Lua Error:
RunConsoleCommand: Command is blocked! (toggle)
```
Neither the way from @bmwalters nor the one from @end360 would solve the problem.

robotboy655:
Just use net messages

end360:
Yeah I guess we'll just have to do what @robotboy655 suggested.

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -179,6 +179,7 @@ local function force_spectate(ply, cmd, arg)
       end
    end
 end
-concommand.Add("ttt_spectate", force_spectate)
-
+net.Receive("ttt_spectate", function(l,pl)
```
fruitwasp: register the network string using [](http://wiki.garrysmod.com/page/util/AddNetworkString)

fruitwasp: space between function arguments

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -179,6 +179,7 @@ local function force_spectate(ply, cmd, arg)
       end
    end
 end
-concommand.Add("ttt_spectate", force_spectate)
-
+net.Receive("ttt_spectate", function(l,pl)
+   force_spectate(pl,net.ReadBool())
```
fruitwasp: space

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -179,6 +179,7 @@ local function force_spectate(ply, cmd, arg)
       end
    end
 end
-concommand.Add("ttt_spectate", force_spectate)
```
markusmarkusz: Don't delete the concommand.
This would break things.
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/cl_help.lua#L162-#L168
Also there could be addons which are using this addon.

end360: Whoops, meant to put it back in because I thought of that while doing some of the other stuff.

Acecool:
Using concommands is an old way to network from client to server and if you allow user input without checking it then exploits are born...

Replace them with net messages, but you can check for afk status using server-only ( including changing the team, etc... ) methods such as checking the move hooks for movement ( should work in spectator mode ) and entity angle for camera movement... If you're going to work on this system, instead of killing a player who is afk and leaving a body ( which can lead to people getting shot ), remove the body too... Add something useful in addition to removing the exploit..

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -161,9 +161,9 @@ end
 concommand.Add("ttt_force_detective", force_detective)
 
 
-local function force_spectate(ply, cmd, arg)
+local function force_spectate(ply, arg)
```
markusmarkusz: The concommand can't work this way.
The callback has the cmd as the second argument. So the concommand for backwards compatibility wouldn't work.

markusmarkusz:
@Acecool 
This PR is used to fix a harmless exploit. If you want to change/add other things to the force spectator system then a new PR would be better. That's the way I think.

willox:
Don't mix WriteBool and ReadInt :(

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -163,7 +163,7 @@ concommand.Add("ttt_force_detective", force_detective)
 
 local function force_spectate(ply, cmd, arg)
    if IsValid(ply) then
-      if #arg == 1 and tonumber(arg[1]) == 0 then
+      if arg == 0 then
```
bigdogmat: This is going to be a table if it's called by the concommand

Kefta:
null is not a valid keyword in Lua

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -1,4 +1,4 @@
-
+util.AddNetworkString("ttt_spectate")
```
markusmarkusz: I would personally prefer to move this to init.lua.
https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/init.lua#L125-#L156

robotboy655:
How many goddamn commits do you need to change 20 lines of code? How about you first finish your code and THEN propose it to be added into the game?

end360:
I need at least 13 commits

robotboy655:
Then go and finish your 13 commits, combined them all into a single one, only THEN come back and try again.

PR #1279, opened 2016-11-19
Fix Anti-AFK Exploit

end360:
Fixes an exploit to bypass `ttt_spectator_mode` changes using the `alias` command, preventing the server from forcing you to be a spectator for being AFK. Uses net commands. Has been refined and tested (in a 4 player P2P "singleplayer" server).

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
@@ -53,7 +53,7 @@ function GM:Initialize()
    self.BaseClass:Initialize()
 
    net.Start("TTT_Spectate")
-     net.WriteBool(GetConVar("ttt_spectator_mode")):GetBool()
+     net.WriteBool(GetConVar("ttt_spectator_mode"):GetBool())
    net.SendToServer()
 end
 
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
@@ -53,7 +53,7 @@ function GM:Initialize()
    self.BaseClass:Initialize()
 
    net.Start("TTT_Spectate")
-     net.WriteInt(math.Clamp(GetConVar("ttt_spectator_mode"):GetInt(),0,1),2)
+     net.WriteBool(GetConVar("ttt_spectator_mode"):GetBool()
    net.SendToServer()
 end
 
@@ -369,9 +369,9 @@ function CheckIdle()
          RunConsoleCommand("say", "(AUTOMATED MESSAGE) I have been moved to the Spectator team because I was idle/AFK.")
 
          timer.Simple(0.3, function()
-                              RunConsoleCommand("ttt_spectator_mode", 1) -- If they want to bypass it now they still can but this way they still get fspecced
+                              RunConsoleCommand("ttt_spectator_mode", 1)
                                net.Start("TTT_Spectate")
-                                 net.WriteInt(1,2)
+                                 net.WriteBool(true)
                                net.SendToServer()
                               RunConsoleCommand("ttt_cl_idlepopup")
                            end)
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
@@ -53,7 +53,7 @@ function GM:Initialize()
    self.BaseClass:Initialize()
 
    net.Start("TTT_Spectate")
-     net.WriteBool(GetConVar("ttt_spectator_mode"):GetBool()
+     net.WriteBool(GetConVar("ttt_spectator_mode")):GetBool()
    net.SendToServer()
 end
 
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
@@ -52,7 +52,9 @@ function GM:Initialize()
 
    self.BaseClass:Initialize()
 
-   RunConsoleCommand("ttt_spectate", GetConVar("ttt_spectator_mode"):GetInt())
+   net.Start("TTT_Spectate")
+     net.WriteInt(math.Clamp(GetConVar("ttt_spectator_mode"):GetInt(),0,1),2)
+   net.SendToServer()
 end
 
 function GM:InitPostEntity()
@@ -367,7 +369,10 @@ function CheckIdle()
          RunConsoleCommand("say", "(AUTOMATED MESSAGE) I have been moved to the Spectator team because I was idle/AFK.")
 
          timer.Simple(0.3, function()
-                              RunConsoleCommand("ttt_spectator_mode", 1)
+                              RunConsoleCommand("ttt_spectator_mode", 1) -- If they want to bypass it now they still can but this way they still get fspecced
+                               net.Start("TTT_Spectate")
+                                 net.WriteInt(1,2)
+                               net.SendToServer()
                               RunConsoleCommand("ttt_cl_idlepopup")
                            end)
       elseif CurTime() > (idle.t + (idle_limit / 2)) then
--- a/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
@@ -1,4 +1,3 @@
-
 function GetTraitors()
    local trs = {}
    for k,v in ipairs(player.GetAll()) do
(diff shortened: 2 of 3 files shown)


Kefta:
Just use Read/Write bool

garrysmod/gamemodes/terrortown/gamemode/traitor_state.lua
```diff
@@ -180,5 +179,7 @@ local function force_spectate(ply, cmd, arg)
    end
 end
 concommand.Add("ttt_spectate", force_spectate)
-
+net.Receive("TTT_Spectate", function(l, pl)
+   force_spectate(pl, nil, {net.ReadInt(2)})
```
Faerachnidendro: If you decide to go with net.WriteBit for the net send, then you will need to change this line too.

**Change:**
force_spectate(pl, nil, {net.ReadInt(2)})

**To:**
force_spectate(pl, nil, {net.ReadBit()})

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
```diff
@@ -52,7 +52,9 @@ function GM:Initialize()
 
    self.BaseClass:Initialize()
 
-   RunConsoleCommand("ttt_spectate", GetConVar("ttt_spectator_mode"):GetInt())
+   net.Start("TTT_Spectate")
+     net.WriteInt(math.Clamp(GetConVar("ttt_spectator_mode"):GetInt(),0,1),2)
```
Kefta: @Faerachnidendro Why would you want to use WriteBit for a bool?

bmwalters:
@Faerachnidendro recently net.Read/WriteBool were added to the library. These functions are effectively the same as net.Read/WriteBit but have better semantics when you're dealing with boolean values.

Faerachnidendro:
Oh I know they are, but, looking at what's excpeted by force_spectate, using net.Write/ReadBit seems more appropriate and perhaps marginally faster(negligible if at all though).

Kefta:
WriteBit isn't semantically "more correct" than a bool. Sure it's infintesmally slower due to another function call, but that's really negligible for where it's used.

Faerachnidendro:
@Kefta I didn't mean "more appropriate" in the sense that one does the job better than the other(as said, they're essentialy the same, poor choice of words there on my part), even Write/ReadBool could be used for the same result:
force_spectate(pl, nil, {net.ReadBool() and 1 or 0})

My intent was the same result without the need for math.Clamp to reign in unexpected values, my preference in this case would be for net.Write/ReadBit, so that the boolean-to-number conversion is done clientside instead of serverside, but, yes, Write/ReadBool absolutely provides the same functionality.

willox:
I'd consider WriteBit and ReadBit deprecated tbh. The confusion between number/book that they introduce is just dangerous.

Faerachnidendro:
Can't disagree there, especially in relation to cvars.
Would you be considering removing them, if they're already to be considered deprecated? (Just so I can go about fixing my code ahead of time)

willox:
i'm not going to remove anything

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
svdm: WriteBool is appropriate here

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
```diff
@@ -367,7 +369,10 @@ function CheckIdle()
          RunConsoleCommand("say", "(AUTOMATED MESSAGE) I have been moved to the Spectator team because I was idle/AFK.")
 
          timer.Simple(0.3, function()
-                              RunConsoleCommand("ttt_spectator_mode", 1)
+                              RunConsoleCommand("ttt_spectator_mode", 1) -- If they want to bypass it now they still can but this way they still get fspecced
```
svdm: The comment on this line only makes sense in context of this PR, so you should either explain more on what the concommand + net message are trying to do, or just remove it entirely.

end360:
The reason I'm not using WriteBool is because of the concommand compatibility. It sends a table for arg `local function force_spectate(ply, cmd, arg)` so instead of using booleans I just used a 2 bit integer. It's 2 bits because the wiki says to add one because "Garry's Mod is using the first bit for purposes other than storing your integer's value." The wiki also says net.WriteBool is an alias of net.WriteBit so I don't see why you guys want me to change it.

bigdogmat:
> It's 2 bits because the wiki says to add one because "Garry's Mod is using the first bit for purposes other than storing your integer's value." The wiki also says net.WriteBool is an alias of net.WriteBit so I don't see why you guys want me to change it.

Just a note here, the reason it says this for `net.WriteInt` is because 1 bit is taken up by the sign bit, use `net.WriteUInt`. Although you could just add another argument that you could check ahead of the table, thus allowing you to use `net.WriteBool`.

willox:
That wiki description definitely needs rewriting

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
```diff
@@ -53,7 +53,7 @@ function GM:Initialize()
    self.BaseClass:Initialize()
 
    net.Start("TTT_Spectate")
-     net.WriteInt(math.Clamp(GetConVar("ttt_spectator_mode"):GetInt(),0,1),2)
+     net.WriteBool(GetConVar("ttt_spectator_mode"):GetBool()
```
Jvs34: You dropped a )

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
```diff
@@ -52,7 +52,9 @@ function GM:Initialize()
 
    self.BaseClass:Initialize()
 
-   RunConsoleCommand("ttt_spectate", GetConVar("ttt_spectator_mode"):GetInt())
+   net.Start("TTT_Spectate")
+     net.WriteBool(GetConVar("ttt_spectator_mode")):GetBool()
```
bigdogmat: And you put it in the the wrong place..

end360: oops

svdm:
So this looks done. Have you tested it with all the changes you made since the original PR?

end360:
Not yet, my PC died so I had to wait on a power supply to come in. I'll test it now.

edgarasf123:
What's the point of adding networked variables if they're called once player has initiated?
I believe you can achieve same effect with PlayerInitialSpawn  hook.

MrUnknownGamer:
Tested this and it doesn't work. Shows the text in chat saying you won't spawn but it'll still spawn you.

I recommend we open this again

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
```diff
@@ -368,6 +370,9 @@ function CheckIdle()
 
          timer.Simple(0.3, function()
                               RunConsoleCommand("ttt_spectator_mode", 1)
+                               net.Start("TTT_Spectate")
+                                 net.WriteBool(true)
+                               net.SendToServer()
```
Bo98: This stuff should be in the cvar change callback.

markusmarkusz:
@MrUnknownGamer 
If you create a server with "Create new game", then it is like this.
This isn't related to his changes.

garrysmod/gamemodes/terrortown/gamemode/cl_init.lua
markusmarkusz: No. Because then this PR wouldn't make sense.
You can overwrite ttt_spectator_mode. This is the reason why he created this PR.
So if you do this net message in the cvar change callback then this wouldn't get called if you overwrite the convar with alias.

Bo98: I edited my comment before you replied. Reread it now.

markusmarkusz: It should not be a problem if you are moved twice into spectator mode.
This shouldn't impact anything.

Bo98: It means the player is respawned twice. Minor but it is something I was making sure was known.

DirkyJerky:
Wait could you not just use `SetConVar` in client lua and not have to make a custom network packet at all?

I'm not familiar with the exploit I may be wrong.

markusmarkusz:
Because he can alias “ttt_spectate“.
Changing “ttt_spectator_mode“ wouldn't work in this case.

DirkyJerky:
Oh I see the work is *in* `ttt_spectate`. Thanks

edgarasf123:
Why would anyone exploit this anyways? Seems pointless exploit.

end360:
Just because an exploit seems pointless doesn't mean it shouldn't be fixed.

PR #1280, opened 2016-11-19
TTT: Force all players to innocents on preparing

markusmarkusz:
Advantage:
Traitor things can't be accessed by users that aren't traitor anymore. (Some traitor things (like rooms or secrets) are made with [ttt_logic_role](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/entities/ttt_logic_role.lua). This entity only checks the role and not if the round is active.)

Disadvantage:
There is a very small chance that some addons could use this. But I'm not sure at the moment.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_logic_role.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_logic_role.lua
@@ -23,13 +23,16 @@ end
 
 function ENT:AcceptInput(name, activator)
    if name == "TestActivator" then
+      if IsValid(activator) and activator:IsPlayer() then
+         local activator_role = (GetRoundState() == ROUND_PREP) and ROLE_INNOCENT or activator:GetRole()
 
-      if IsValid(activator) and activator:IsPlayer() and (self.Role == ROLE_ANY or activator:IsRole(self.Role)) then
-         Dev(2, activator, "passed logic_role test of", self:GetName())
-         self:TriggerOutput("OnPass", activator)
-      else
-         Dev(2, activator, "failed logic_role test of", self:GetName())
-         self:TriggerOutput("OnFail", activator)
+         if self.Role == ROLE_ANY or self.Role == activator_role then
+            Dev(2, activator, "passed logic_role test of", self:GetName())
+            self:TriggerOutput("OnPass", activator)
+         else
+            Dev(2, activator, "failed logic_role test of", self:GetName())
+            self:TriggerOutput("OnFail", activator)
+         end
       end
 
       return true
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
@@ -193,11 +193,11 @@ function GM:HUDDrawTargetID()
          _, color = util.HealthToString(ent:Health())
       end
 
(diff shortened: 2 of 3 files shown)


end360:
An example of rooms that use ttt_logic_role that can be used if you were traitor on the last round is the door to the vending machine on ttt_waterworld

svdm:
This change would break the role selection when the round begins. See https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/init.lua#L853
The currently assigned role is used in the selection process to lower the chances of being traitor multiple rounds in a row.

Perhaps an alternative fix would be to change ttt_logic_role to consider every player to be innocent if the round is preparing.

garrysmod/gamemodes/terrortown/entities/entities/ttt_logic_role.lua
```diff
@@ -24,7 +24,7 @@ end
 function ENT:AcceptInput(name, activator)
    if name == "TestActivator" then
 
-      if IsValid(activator) and activator:IsPlayer() and (self.Role == ROLE_ANY or activator:IsRole(self.Role)) then
+      if IsValid(activator) and activator:IsPlayer() and (self.Role == ROLE_ANY or activator:IsActiveRole(self.Role)) then
```
svdm: Just using IsActiveRole was my first thought as well, but it's not ideal. It will return false even if the `ttt_logic_role` is checking for ROLE_INNOCENT, if it's during the prep or post round. Similarly a check for traitors will be false if it's post round, so a traitor in a traitor-room would be stuck in there for the whole post-round phase.

Perhaps something like this

    local activator_role = (GetRoundState() == ROUND_PREP) and ROLE_INNOCENT or activator:GetRole()

Which you could then compare to `self.Role`. Of course you'd also need to do the IsPlayer check before that.

markusmarkusz:
It should work. (I tested it and it worked for me.)

mcNuggets1:
This change is needed.
Because TargetID shows detectives on preparing and a traitor mark in the scoreboard, even a red or blue crosshair.

svdm:
@Faerachnidendro We don't actually want the finished/post-round phase to be handled the same way. Players who were traitor during the round should still be able to operate e.g. traitor-only doors in the post-round phase. That's why I specifically suggested the check on prep. Ideally it would be a check for waiting or prep, but that makes it more involved for very little benefit (during waiting everyone should be spectator'd anyway).

markusmarkusz:
That traitors are displayed in the scoreboard as traitors is a wanted behavior, isn't it?
I can change that players which were detectives aren't shown as detectives in preparing. Should I?

markusmarkusz:
Now everyone should be happy. I squashed the commits together.

bmwalters:
You seemed to miss the commit message there. Can't say I've seen that 😂

markusmarkusz:
Google translator has eaten the message. Sorry. 🙄

Faerachnidendro:
@svdm Aah, makes sense I guess. Doesn't really change much to use a traitor button during post-round anyway, the round being over and all.

bmwalters:
@markusmarkusz 
```bash
git checkout patch-3
git commit --amend
```

mcNuggets1:
Will this be merged :L?

PR #1281, opened 2016-11-27
Fix No Dropdown on Scoreboard

MrUnknownGamer:
If the round isn't active, the mute button will disappear and nothing will drop down. I.E. Suspect, Kill, etc.

Stops you from dropping down until the round begins to fix this bug. Why would you need to mark anyone as "suspect" or kill anyway in preround?

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
@@ -281,6 +281,7 @@ function PANEL:PerformLayout()
 end
 
 function PANEL:DoClick(x, y)
+   if GetRoundState() ~= ROUND_ACTIVE then return end
    self:SetOpen(not self.open)
 end
 


svdm:
If you have a "dropdown" open when the round ends, I think this change makes it so you can no longer close it. Also, I think a better way to fix the problem is to check `self.info` in DoClick, and simply always set open to false if it's nil. That should handle all cases where you might try to open a row that has no info to show (= empty dropdown) regardless of round state.

MrUnknownGamer:
> If you have a "dropdown" open when the round ends, I think this change makes it so you can no longer close it. Also, I think a better way to fix the problem is to check self.info in DoClick, and simply always set open to false if it's nil. That should handle all cases where you might try to open a row that has no info to show (= empty dropdown) regardless of round state.

Sorry I'm not really following, can you explain a bit more? <3

svdm:
OK, so when you click a scoreboard row it toggles `self.open` to show what you call the "dropdown" and the code calls `self.info`. The problem with the change in the PR is that if you have opened it while the round is active, and then the round ends, you can no longer close it.

You're correct that if the round isn't active yet nothing "drops down". Clicking will set `self.open` to true, but there will be nothing for `self.info` to display, so all that happens is the mute button goes away. (This all happens in PerformLayout.)

So ideally we would fix the part where clicking sets `self.open` to be true even when `self.info` does not contain anything to display. If we keep that from happening, the problem is fixed without relying on the round state.

I haven't tested it but I think something like this would work:
```
function PANEL:DoClick(x, y)
   if self.info then
      self:SetOpen(not self.open)
   else
      self:SetOpen(false)
   end
end
```

MrUnknownGamer:
Ahh ok, ima do that real quick on local server

siasur:
"If the round isn't active, the mute button will disappear [...]" I want to mute people in every state of the game... there are assholes everywhere...

PR #1282, opened 2016-11-30
Base: Add Identifier to RunString on lua_run Entity

markusmarkusz:
If a map causes an lua error message, it is indicated that it comes through this by adding the map name as identifier.

Can be very useful when searching the source of an error message that is caused by this.

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/lua_run.lua
+++ b/garrysmod/gamemodes/base/entities/entities/lua_run.lua
@@ -5,6 +5,9 @@ ENT.DisableDuplicator = true
 AccessorFunc( ENT, "m_bDefaultCode", "DefaultCode" )
 
 function ENT:Initialize()
+
+	self.RunStringID = game.GetMap() .. ":" .. self:MapCreationID()
+
 end
 
 function ENT:KeyValue( key, value )
@@ -38,7 +41,7 @@ function ENT:RunCode( activator, caller, code )
 
 	self:SetupGlobals( activator, caller )
 
-		RunString( code )
+		RunString( code, self.RunStringID )
 
 	self:KillGlobals()
 


garrysmod/gamemodes/base/entities/entities/lua_run.lua
```diff
@@ -38,7 +38,7 @@ function ENT:RunCode( activator, caller, code )
 
 	self:SetupGlobals( activator, caller )
 
-		RunString( code )
+		RunString( code, game.GetMap() )
```
JamesxX: Might be better to also include the hammer ID of the lua_run entity so as to further narrow down any errors.

```
RunString( code, game.GetMap()  .. ":" .. self:MapCreationID() )
```

markusmarkusz: I think this is acceptable. I'll add it later.

JamesxX: As concatenation isn't expensive, and there aren't many run_lua entities per map, in general, it probably won't impact performance in any meaningful way, so it doesn't matter when it's done really.

markusmarkusz:
Is that okay?

JamesxX:
Looking good

garrysmod/gamemodes/base/entities/entities/lua_run.lua
```diff
@@ -1,10 +1,14 @@
 
 ENT.Type = "point"
 ENT.DisableDuplicator = true
+ENT.MapCreationID = -1
 
 AccessorFunc( ENT, "m_bDefaultCode", "DefaultCode" )
 
 function ENT:Initialize()
+
+	self.MapCreationID = self:MapCreationID()
```
bigdogmat: This overwrites `MapCreationID`, name it something else as it'll break any addon that attempts to call it. Also, if you are going to put this here, do what I suggested and put all of it here, e.g.

```lua
self.RunStringID = game.GetMap() .. ":" .. self:MapCreationID()
```

markusmarkusz:
Oh fuck.
Completly forgot that this overwrites Entity:MapCreationID() on this Entity. :/

garrysmod/gamemodes/base/entities/entities/lua_run.lua
willox: Good shout, but it wouldn't actually override the method (because the entity functions have priority over any values set on the entity.)

So the commit actually errors as it is when trying to concat the values on line 45.

bigdogmat: Huh, never knew that. Good to know.

willox:
You're adding the map name to the string twice. I think you'd have noticed both the mistakes you've ran in to if you tried testing it.

markusmarkusz:
I'm feeling dumb right now. :C

Kefta:
Looks good now.

robotboy655:
Didn't know this existed when I made this commit: https://github.com/Facepunch/garrysmod/commit/1f0fa0e78dd0e80113727e393ff7bd01aba5887f#diff-059d98c26cce7848ec787feb83394481 which sort of fulfills this PR.

I feel like `EntIndex()` is better than `MapCreationID()` in this case because even though this entity is meant for use through hammer, `MapCreationID()` will be -1 for entities made through Lua and that is not helpful for what this change is meant to accomplish.

PR #1283, opened 2016-11-30
TTT: Add C4 Hooks

markusmarkusz:
Here are the C4 Hooks.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
@@ -466,7 +466,7 @@ if SERVER then
    end
 
    local function ReceiveC4Config(ply, cmd, args)
-      if (not IsValid(ply)) or (not ply:IsTerror()) or (not ply:Alive()) or #args != 2 then return end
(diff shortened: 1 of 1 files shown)


Kefta:
Good DeMorgan's law

mcNuggets1:
Perfeeect.

Acecool:
Actually quite useful but don't forget to create the base functions. If they don't exist the hook won't be called. I'd recommend sticking them in gamemode or GM...

markusmarkusz:
No. They don't have to exist as Gamemode Functions.

Kefta:
@Acecool You're thinking of gamemode.Call

Acecool:
I always had issues; if I didn't make the blank function the hook.Add wouldn't run. I use gamemode for GM extensions ( so my addons will work regardless of whether they're used in gamemodes/ or addons/ because addons/ load prior to GM / GAMEMODE but gamemode exists )...

May have been an old glitch.. I'll double-check it next time I run my dev server.

Kefta:
I mean, you can see it here: https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/includes/modules/hook.lua#L121

PR #1284, opened 2016-12-01
Corrected wrong var reference 

markusmarkusz:
`expected` isn't defined in this case.
I'm sure that `percent` is right here.


Bo98:
Looks like a copy-paste error.

Probably went unnoticed as that piece of code will rarely be run, if ever, because `vgui_workshop` will have been created in the other hooks.

PR #1287, opened 2016-12-03
TTT: Increasing Performance in preparing

markusmarkusz:
This change should improve performance. I noticed a visible improvement in tests.
It would be great if someone can confirm that these changes are beneficial. (To ensure that these improvements are not related to my PC or my server.)

Openly the question would be whether it is wise, `ents.TTT.GetSpawnableSWEPs()` to edit. I do not know how likely it is that someone is using this feature. I have modified that the table no longer has the weapon tables. The weapon class is enough.

In fact, you do not have to copy the table because it is not modified anyway. Should we omit the copy at `rndpairs`?

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_random_ammo.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_random_ammo.lua
@@ -8,7 +8,7 @@ function ENT:Initialize()
    local ammos = ents.TTT.GetSpawnableAmmo()
 
    if ammos then
-      local cls = table.Random(ammos)
+      local cls = ammos[math.random(1, #ammos)]
       local ent = ents.Create(cls)
       if IsValid(ent) then
          ent:SetPos(self:GetPos())
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_random_weapon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_random_weapon.lua
@@ -14,7 +14,7 @@ end
 function ENT:Initialize()
    local weps = ents.TTT.GetSpawnableSWEPs()
    if weps then
-      local w = table.Random(weps)
+      local w = weps[math.random(1, #weps)].ClassName
       local ent = ents.Create(w)
       if IsValid(ent) then
          local pos = self:GetPos()
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_random_weapon.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_random_weapon.lua
@@ -15,7 +15,7 @@ function ENT:Initialize()
    local weps = ents.TTT.GetSpawnableSWEPs()
    if weps then
       local w = table.Random(weps)
-      local ent = ents.Create(WEPS.GetClass(w))
+      local ent = ents.Create(w)
       if IsValid(ent) then
          local pos = self:GetPos()
          ent:SetPos(pos)
--- a/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
@@ -2,11 +2,16 @@
 
 ents.TTT = {}
 
+local use_weapon_spawn_scripts = CreateConVar("ttt_use_weapon_spawn_scripts", "1")
+local weapon_spawn_count = CreateConVar("ttt_weapon_spawn_count", "0")
+
 local table = table
 local math = math
 local pairs = pairs
+local ipairs = ipairs
 
 local function ReplaceSingle(ent, newname)
+
    -- Ammo that has been mapper-placed will not have a pos yet at this point for
    -- reasons that have to do with being really annoying. So don't touch those
    -- so we can replace them later. Grumble grumble.
@@ -60,6 +65,12 @@ local function ReplaceAmmoSingle(ent, cls)
    end
 end
 
+local function ReplaceAmmo()
+   for _, ent in ipairs(ents.FindByClass("item_*")) do
+      ReplaceAmmoSingle(ent)
+   end
+end
+
 local hl2_weapon_replace = {
    ["weapon_smg1"] = "weapon_zm_mac10",
    ["weapon_shotgun"] = "weapon_zm_shotgun",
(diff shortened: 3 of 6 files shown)


bigdogmat:
You're `TableRandom` function is doing the exact same as `return tbl[math.random(1, #tbl)]`, no need to loop over the table.

Also, because spots is always going to be sequential,  your `rndpairs` function could simply be written like 
```lua
local function rndpairs(tab)
  local keys  = table.GetKeys(tab)
  local count = #keys

  return function()
    if count == 0 then return end

    local index = table.remove(keys, math.random(1, count))
    count = count - 1

    return index, tab[index]
  end
end
```

Could probably be written better, but it's better than what you currently have

markusmarkusz:
@bigdogmat 
> You're TableRandom function is doing the exact same as return tbl[math.random(1, #tbl)], no need to loop over the table.

Huh. This has happened because I used the table.Random function when editing, then copied it and changed it and then probably overlooked. Will improve that later. Sorry.

I'll take a look at the rndpairs function also later.
Thabks for feedback.

svdm:
Could you clarify when you are seeing a performance increase? At a glance, much of what this PR changes is only ever called if a map is being used that does not have a rearm script (so a CS:S/HL2DM/TF2 map without a script). I'm pretty sure this is very rare these days, as nearly every map is either TTT-specific or has is used with a rearm script.

Have you measured for each of your changes whether they actually help, and if so, how much? I have doubts about some cases, for example in `RemoveWeaponEntities`, using `ents.FindByClass` will execute the search in C++, which should be pretty fast. Whether your new approach is faster or slower can depend on the number of entities in the map versus the number of weapon classes being removed.

markusmarkusz:
`ents.FindByClass` uses intern` ents.GetAll` if I am not mistaken. If you only test `ents.GetAll` once, this is faster. I will test more maps (with and without Rearm Script) and check if you notice improvements. It would be great if someone else then confirmed these improvements.

markusmarkusz:
Damn it. I've made a few mistakes again. I'll fix them later.

Kefta:
I expected the single GetAll to be faster, however, it seems that it is much slower than the current implementation -- tested with 100,000 iterations. Actually, an optimised version in where one loop is saved through merging (checking weapon_zm_improvised in weapon_*) would be the best. Using a numeric for loop would be ever-so-slightly more efficient, but it is really up to preference at that level.

markusmarkusz:
I'll simply leave it as it is. I've made too much nonsense. Do not make sense to continue producing any other crap.
I hope this does not misunderstand anyone, but when I look at what I have changed so far, I ask myself seriously what I have done in part.

markusmarkusz:
To explain why I closed this PR now: I have tested the code and noticed improvements. Now, after running these tests on another server, I notice much less of these improvements. Since I have coded so much nonsense in some places and have often broken the code due to carelessness, I decided to leave it as it is. In fact, there are parts in this PR that should fix minor errors and improve performance. Since I am not so sure, if these changes are worthwhile, I list the times, where I am sure that they would be worth. If someone confirms that these changes, I'll either re-open this PR, or create a new one. I would be glad if you would answer me and tell me if the changes listed below are useful:
- Using [this](https://github.com/markusmarkusz/garrysmod/blob/patch 8/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua#L264-L276) function instead of RandomPairs (propably useful for some other parts of this code. For example at the function “SpawnWillingPlayers“ in init.lua).
- Using [this](https://github.com/markusmarkusz/garrysmod/blob/patch-8/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua#L326) method instead of table.Random because table.Random loops through the table twice.)
- Moving some tables out of functions. For example [here](https://github.com/markusmarkusz/garrysmod/blob/patch-8/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua#L350-L354) or [here](https://github.com/markusmarkusz/garrysmod/blob/patch-8/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua#L363-L377).
- Calling [this](https://github.com/markusmarkusz/garrysmod/blob/patch-8/garrysmod/gamemodes/terrortown/gamemode/init.lua#L374-L380) on every cleanup. (Addons can execute game.CleanupMap(). So every weapons spawned by TTT would be removed.)
- Calling [this](https://github.com/markusmarkusz/garrysmod/blob/patch-8/garrysmod/gamemodes/terrortown/gamemode/init.lua#L460-L464) only on FirstRound is better than calling it everytime a new round begins.

willox:
You'll just be wasting time unless you actually discover real bottlenecks with a profiler or something

markusmarkusz:
@willox
I've used addons, such as FProfiler (from FPjte) for testing, for every change. These addons also showed improvements. Therefore, I was confused with the tests of Kefta and after I had tested on another server, I was really confused. Came out for some reason results with substantial differences.

Kefta:
Can you provide benchmarked tests or find issues with the code I provided?

markusmarkusz:
The code you have provided is actually faster than my implementation, according to the Bechmark script (which you used for this). Nevertheless, according to FProfiler my code was better than the previous one. Makes me little sense, since it is slower according to the benchmark. I can maybe make a few tests again tomorrow and post the results here.

markusmarkusz:
Right at the beginning I would like to mention that I will no matter what will not open this pull request again.
I have noticed that the code is broken in different places.
Nevertheless, I will make these tests. If the improvements are worth it, I would create a new pull request.


These tests were done with [bench](https://github.com/SuperiorServers/dash/blob/master/lua/dash/libraries/bench.lua):
Testing the difference between "rndpairs" and "RandomPairs": https://gist.github.com/markusmarkusz/d5b710a23c87a459a482877034efe347
Testing the difference between "table.Random" and "table[math.random(1, #table)]": https://gist.github.com/markusmarkusz/e5611a63e1f9ef12cf2ac0bc41128abc

The most expensive function calls (tested with FProfiler):
Without any change: 
image
With some other changes (like precaching convars, removing useless function calls, using ipairs in numeric tables instead of pairs and some other things): 
image

I can post here later what I've changed. Then maybe someone could confirm if it really is improvements.

markusmarkusz:
It took me a little longer to upload the code. I had something to do in the past few days and therefore no time to upload the code.

First of all, this code has of course merge conflicts. I've taken these changes directly from my test server. This means that the changes made on GitHub aren't included.

Second: Here is a test done with [bench](https://github.com/SuperiorServers/dash/blob/master/lua/dash/libraries/bench.lua) showing the difference between the caching of ConVars.
https://gist.github.com/markusmarkusz/77f7525f5583d229e09fa4bf3ded8e05

And last but not least, it would be great if someone test these changes. I've mainly used [FProfiler](https://github.com/FPtje/FProfiler) and [bench](https://github.com/SuperiorServers/dash/blob/master/lua/dash/libraries/bench.lua) for tests.
The results are visible in the previous comment.

markusmarkusz:
Is it worth to create a new PR for these changes?

svdm:
Judging from the profiling results the gains seem very small, most of the time is being spent in C++ in the form of game.CleanUp and spawning entities. As a result I really doubt stuff like messing with ipairs and caching convars is going to have a noticeable impact.

Kefta:
When you're doing that many ent loops, it actually has a small but noticeable effect in my experience.

PR #1288, opened 2016-12-09
Add player_manager.GetPlayerClasses() and PLAYER:ClassChanged()

LeQuackers:
Getting the table of registered player classes is a pain and PLAYER:Destroy() would be useful for removing things that aren't garbage collected, such as client-side props and projected textures.


neico:
Maybe squash the commits?

Also, makes me wonder why there's no table.Copy variant for shallow copies (but in your case table.Copy might be more desireable to actually prevent modification)

bmwalters:
@neico #1273

garrysmod/lua/includes/modules/player_manager.lua
```diff
@@ -301,6 +301,18 @@ AddValidHands( "dod_american", "models/weapons/c_arms_dod.mdl",		1,		"10000000"
 
 local Type = {}
 
+function GetTable()
+
+	local tbl = {}
+	
+	for k, v in pairs( Type ) do
```
GitSparTV, line 308: Couble be moved to table.Copy, but this is fine.

garrysmod/lua/includes/modules/player_manager.lua
```diff
@@ -361,8 +373,11 @@ local function LookupPlayerClass( ply )
 	--
 	-- Check the cache
 	--
-	local method = ply.m_CurrentPlayerClass;
-	if ( method && method.Player == ply && method.ClassID == id && method.Func ) then return method end
+	local method = ply.m_CurrentPlayerClass
+	if ( method && method.Player == ply ) then
+		if ( method.ClassID == id && method.Func ) then return method end -- current class is still good, behave normally
+		if ( method.Destroy ) then method:Destroy() end -- the class id changed, destroy the old class
```
GitSparTV, line 379: Do you use PLAYER:Destroy to clean up player data on class change?

GitSparTV (commented):
Bumping + commeting

garrysmod/lua/includes/modules/player_manager.lua
LeQuackers: Yes, some examples include removing clientside models, projected textures, and resetting bone manipulations.

garrysmod/lua/includes/modules/player_manager.lua
```diff
@@ -296,6 +296,12 @@ AddValidHands( "dod_american", "models/weapons/c_arms_dod.mdl", 1, "10000000" )
 
 local Type = {}
 
+function GetTable()
+
+	return table.Copy( Type )
```
robotboy655, line 301:
```
] lua_run PrintTable(player_manager.GetTable())
> PrintTable(player_manager.GetTable())...

[ERROR] lua/includes/modules/player_manager.lua:301: attempt to index global 'table' (a nil value)
  1. GetTable - lua/includes/modules/player_manager.lua:301
   2. unknown - lua_run:1
```

Also since `player_manager` library doesn't only handle player classes, the function name should point to the fact that it's the player classes table, not player models, etc., so something like ```player_manager.GetPlayerClasses()``` or something.

robotboy655 (changes requested):

garrysmod/lua/includes/modules/player_manager.lua
robotboy655: Also also is everyone happy with the callback being called `Destroy`?

GitSparTV: For the consistency I would propose "OnRemove" "OnPlayerClassChanged" "OnChange" or something similar

GitSparTV: However, we have Destroy methods in some classes in Gmod.
But this case this is more like DFrame "OnClose" rather than physics object destroying

robotboy655: Thing is nothing here is actually removed, and none of the other PLAYER: callbacks use `On` in their names. Something like `PLAYER:SwitchedAway` or `PLAYER:ChangedFrom` would better represent what is actually happening when the callback gets called.

GitSparTV: ClassChanged

robotboy655 (approved):

PR #1289, opened 2016-12-09
Add phys_pushscale to fist force

Kefta:
Also, removed the SERVER check from the physobj force application in the case that an entity has a shared physics object


Kefta:
Also, shouldn't it be 1/mass so heavier objects get pushed less?

PR #1290, opened 2016-12-09
Add CompoundLerp

LeQuackers:
Allows you to lerp variables smoothly every frame regardless of FPS/tick rate in Think/Render hooks, for example:

```Lua
-- after 1 second self.Var will be 95% from its initial position to 8
self.Var = CompoundLerp( 0.95, FrameTime(), self.Var, 8 )

-- after 3 seconds self.Var will be 65% from its initial position to 8
self.Var = CompoundLerp( 0.65, 0.33 * FrameTime(), self.Var, 8 )
```

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -219,6 +219,22 @@ function Lerp( delta, from, to )
 end
 
 --[[---------------------------------------------------------
+	Compound lerp
+-----------------------------------------------------------]]
+function CompoundLerp( percent, delta, from, to )
+	
+	if ( percent >= 1 ) then return to end
+	if ( percent <= 0 ) then return from end
+	
+	if ( delta <= 0 ) then return from end
+	
+	local t = 1 - ( 1 - percent ) ^ delta
+	
+	return from + t * ( from - to )
+	
+end
+
+--[[---------------------------------------------------------
 	Convert Var to Bool
 -----------------------------------------------------------]]
 function tobool( val )


LeQuackers:
CompoundLerp doesn't fit too well in my opinion, anyone have a better name?

GitSparTV:
isn't this math.Approach? 🤔

GitSparTV:
Nvm you just using Lerp wrong ahahaha

thegrb93:
Not wrong, but it's equivalent to `Lerp( 1 - ( 1 - t ) ^ delta, from, to )` so just use that instead.

GitSparTV:
Could be added as example on wiki. So it won't be forbidden

LeQuackers:
It's a math function. Saying someone is using Lerp wrong when it's common to do or gives the desired output is dumb.

The reason this pull request exists is because people use ```current = Lerp(FrameTime(), current, target)``` for smoothing, which is normally fine but sucks as the frame time approaches 1.

You can use ```Lerp( 1 - ( 1 - t ) ^ delta, from, to )``` if you want, but most people don't know how to fix the previously stated issue, hence this pull request.

GitSparTV:
You still use Lerp wrong, you are passing the output value to `from` parameter, this makes lerp non-linear

LeQuackers:
that's the point

Kefta:
Then it's non-linear interpolation. Don't use lerp for this.

LeQuackers:
That's why this pull request exists; to replace ```current = Lerp(FrameTime(), current, target)``` with an actual function.

GitSparTV:
It won't be called lerp

GitSparTV:
Also there is a similar function I mentioned earlier: math.Approach

robotboy655:
Just use a more appropriate math function + `math.Remap` from a website like this: https://easings.net/ for the desired effect.

This way you will not end up with a forever changing number that will never reach the `to` value, and you won't have any issues with with frame time, since you'd wouldn't be using `FrameTime()`.

Here's an example adapted from that website and the example from Wiki's Lerp page:
```

function easeOutCubic( x )
	return 1 - math.pow( 1 - x, 3 )
end

function easeOutCubicRemapped( t, from, to )
	local t = math.Clamp( t, 0, 1 )
	local y = easeOutCubic( t )
	return math.Remap( y, 0, 1, from, to )
end

local start, oldhp, newhp = 0, 0, 0
local animationTime = 0.5

hook.Add("HUDPaint", "EaseOutExample", function()
	local hp = LocalPlayer():Health()
	local barW = ScrW() - 200 

	if newhp ~= hp then
		oldhp = newhp
		start = SysTime()
		newhp = hp
	end

	local smoothHP = easeOutCubicRemapped( (SysTime() - start) / animationTime, oldhp, newhp)

	draw.RoundedBox( 4, 100, 200, barW, 100, color_black )
	draw.RoundedBox( 4, 100, 200, math.max( 0, smoothHP ) / LocalPlayer():GetMaxHealth() * barW, 100, color_white )
end)
```

PR #1291, opened 2016-12-10
Squared velocity lengths

Luabee:
Computing the square root of a number is expensive. By switching to LengthSqr and Length2DSqr, we skip this square-rooting altogether.

This only works for comparisons like these, since for any two numbers `x` and `y`, where `x < y`, we can assume that `x*x < y*y`.


Kefta:
You can just change the mul operations into plain numbers

PR #1292, opened 2016-12-10
Add util.Spring

LeQuackers:
Allows you to calculate spring movement regardless of FPS/tick rate without spazzing and add instantaneous velocity at different times. Also supports vectors, angles, and integers.

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -223,6 +223,132 @@ end
 
 
 --
+-- Spring
+--
+--
+local T = 
+{
+	
+	--
+	-- Allows you to add velocity to the spring
+	--
+	AddVelocity = function( self, addVel )
+	
+		local CT = CurTime()
+		
+		local t		= CT - self.Time
+		local d		= self.DecayRate
+		local c		= self.Cycle
+		local b		= self.Pos
+		local vel	= self.Vel
+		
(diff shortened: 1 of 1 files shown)


LeQuackers:
This still feels like it's missing something, any suggestions?

Kefta:
Maybe support dampening states?

LeQuackers:
By dampening states do you mean something similar to Lerping between rates of decay? If I update the position and velocity when the rate of decay changes it might work, gonna have to test it.

Kefta:
Yeah, the damping coefficient will cause the decay to affect the spring faster or slower

LeQuackers:
RIP last comment. Lerping rate of decay is effected by FPS/tick rate. I could still try to make Lerping velocity and rate of decay work if I made it into a calc 3 problem, but I'm lazy, so I'll probably add functions similar to AddVelocity/AddVelocityAtTime for the rate of decay in the meantime.

LeQuackers:
Added/Removed/Renamed stuff. Still need to figure out a good way to allow people to use different variables for time with this, like UnPredictedCurTime or a custom time system. Might let util.Spring have an optional function parameter that gets called instead of CurTime.

GitSparTV:
Does it have a purpose in GMod? Looks very auxiliary

Kefta:
What Spar said. This can be library implemented.

PR #1293, opened 2016-12-11
Add Coherent Noise Utility Functions

LeQuackers:
Useful in animations that require natural looking movement. The seeds for the noise functions should be offsets with respect to time, for example:

`local vec1 = util.NoiseVector( time )`
`local vec2 = util.NoiseVector( time + 88 )`

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -160,6 +160,81 @@ function util.NiceFloat( f )
 end
 
 
+--
+-- Coherent noise stuff
+--
+function util.Noise( n )
+
+	math.randomseed( math.Round( 100 * n ) )
+	
+	math.random()
+	math.random() -- scramble the LCG a bit
+	math.random()
+	
+	return math.random()
+	
+end
+
+function util.CoherentNoise( n )
+
+	local n1 = math.floor( n )
+	local n2 = n1 + 1
+	
+	local t = n - n1
+	local t2 = t * t
+	local t3 = t * t2
+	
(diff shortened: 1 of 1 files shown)


Kefta:
I don't understand this one

LeQuackers:
The util.Noise function allows you to gernerate the same result given a specific input. The util.CoherentNoise function uses this to create a cubic hermite spline with random positions and tangents using the floor and ceil of its input as the seeds and then interpolates between them using the fractional component of its input. For example:

`print( util.Noise( 6 ), util.Noise( 6 ) ) -- this will display the same number twice`

`local a = util.CoherentNoise( 5.3 ) -- uses 5 and 6 as seeds, t equals 0.3`
`local b = util.CoherentNoise( 5.4 ) -- uses 5 and 6 as seeds, t equals 0.4`
`local c = util.CoherentNoise( 5.9 ) -- uses 5 and 6 as seeds, t equals 0.9`

`local d = util.CoherentNoise( 6.5 ) -- uses 6 and 7 as seeds, t equals 0.5`
`local e = util.CoherentNoise( 7.6 ) -- uses 7 and 8 as seeds, t equals 0.6`

` -- this will produce coherent noise that is seamless because the starting seed is equal`
` -- to the last seed used and the ending seed is equal to the next seed that will be used`
`print( util.CoherentNoise( CurTime() ) )`

It would probably be easier to see the effects in movement or color.

Mysterypancake1:
I made a quick test and it looks nice
```
hook.Add( "CalcView", "Test", function( ply, pos, ang, fov )
	local n = 0.5 - util.CoherentNoise( CurTime() * 2 )
	return { angles = ang + Angle( n, n, n ) * ply:GetVelocity():Length() * 0.025 }
end )
```

PR #1294, opened 2016-12-11
Add CHSpline

LeQuackers:
Similar to the Lerp and BSpline functions, however it allows you to set the starting and ending tangents. Works with integers and vectors, might work with angles.


robotboy655 (approved):

robotboy655:
I will move this to `math` library, as it makes more sense there, next to the other spline function

PR #1295, opened 2016-12-13
Add Entity:CallOnTransmitStateChanged

LeQuackers:
I read issue [#2637](https://github.com/Facepunch/garrysmod-issues/issues/2637) from the bug report section, I think this would help a bit.

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -141,6 +141,57 @@ end
 
 hook.Add( "EntityRemoved", "DoDieFunction", DoDieFunction )
 
+if ( CLIENT ) then
+
+	--[[---------------------------------------------------------
+		Name: CallOnTransmitStateChanged
+		Desc: Call this function when this entity's transmit state changes.
+		Calls the function like Function( <entity>, <should transmit>, <optional args> )
+	-----------------------------------------------------------]]
+	function meta:CallOnTransmitStateChanged( name, func, ... )
+
+		local mytable = self:GetTable()
+		mytable.OnTransmitFunctions = mytable.OnTransmitFunctions or {}
+
+		mytable.OnTransmitFunctions[ name ] = { Name = name, Function = func, Args = { ... } }
+
+	end
+
+	--[[---------------------------------------------------------
+		Name: RemoveCallOnTransmitStateChanged
+		Desc: Removes the named hook
+	-----------------------------------------------------------]]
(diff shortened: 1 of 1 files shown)


mcNuggets1:
You are doing some great pull requests ;)

willox:
Example usage?

LeQuackers:
In kefta's case:

```Lua
ply:CallOnTransmitStateChanged( "clothes", function( self, shouldTransmit )
	if ( shouldTransmit ) then
		
		clothes:SetMoveParent( self )
		clothes:SetNoDraw( false )
		
	else
		
		clothes:SetNoDraw( true )
		
	end
end )
```

willox:
I'm not sure this is really a good solution to the issue. You're still going to want to remove the clientside entity when the player is removed, how would you handle that?

mcNuggets1:
atleast its a solution ;)

LeQuackers:
You have a good point and I'm still thinking of a solution. The purpose of the Entity:CallOnTransmitStateChanged function was to prevent client side models/hats/whatever from breaking if a player moved out of someone's PVS.

willox:
Entities aren't removed when they leave the PVS. That'll should happen if they are outside the PVS and the client receives a full update from the server.

Ideally if a player _is_ removed, the clientside entities should also be removed until that player is re-created.

aStonedPenguin:
Why would you just not use the hook in your own code?

LeQuackers:
@willox I was under the impression that if an entity was out of a player's PVS it would become dormant client side and call its remove function because of how [C_EntityFlame](https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/game/client/c_fire_smoke.cpp#L294-L323) works, but if I'm wrong then I'm wrong.

@aStonedPenguin You also have a good point, but if it's only a few client side models I'd rather use something like this because it's somewhat managed in my opinion.

LeQuackers:
Yep, I'm wrong read partition->Remove() as Remove() form [C_BaseEntity](https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/game/client/c_baseentity.cpp#L2098).

LeQuackers:
Regardless of the Entity:Remove/Entity:Initialize mistake, Entity:CallOnTransmitStateChanged would still be helpful when dealing with interactions between client side models and PVS stuff.

Kefta:
Yeah, I personally think this is a bit of a hacky solution -- the best would be to stop players from being removed and recreated instantly in lagspikes. I personally use a PVS module to handle fullupdates myself, which solves a majority of these issues.

PR #1297, opened 2016-12-16
TTT: Fix enter button for disguiser

markusmarkusz:
Calling `WEPS.DisguiseToggle(ply)` in server state doesn't make any sense.
It has to be client side.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_keys.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_keys.lua
@@ -113,3 +113,10 @@ function GM:KeyRelease(ply, key)
    end
 end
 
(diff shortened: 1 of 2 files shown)


Kefta:
It used to work -- what commit broke it?

markusmarkusz:
It seems like that this is a issue for a long time. (This function is the same function like 4 years ago.)

Kefta:
Yeah, I don't remember it working properly since GM12. Always thought it was my numpad

svdm:
Judging from the comment it used to be in the KeyRelease hook. I vaguely remember something changing about (numpad) key handling/bindings in a gmod update long ago, probably was moved out at that point and broken either then or later.

PR #1298, opened 2016-12-16
Add hook SANDBOX:ShouldHideTool to hide tools

JamesxX:
Allow addons to hide tools from the tool menu. It is already possible to disallow the use of a tool by a player by means of the SANDBOX:CanTool hook, however, tools still remain visible in the tool menu.

Adding this hook would allow server owners who wish to restrict tool use to also hide these tools.

This feature is requested from the following thread: https://facepunch.com/showthread.php?t=1545453
Feature was also requested here: https://github.com/Facepunch/garrysmod-requests/issues/612

Example usage:
```lua
hook.Add( "ShouldHideTool", "TESTING", function( tool )
	
	if ( tool.ItemName == "hoverball" ) then return true end

end)
```

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/toolpanel.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/toolpanel.lua
@@ -66,6 +66,8 @@ function PANEL:AddCategory( Name, Label, tItems )
 
 	for k, v in SortedPairs( tools ) do
 
+		if ( hook.Run( "ShouldHideTool", v.ItemName ) ) then continue end
+		
 		local item = Category:Add( v.Text )
 
 		item.DoClick = function( button )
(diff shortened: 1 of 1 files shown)


aStonedPenguin:
You should probably pass the table and not just the name.

SpencerSharkey:
yeah... ditto @aStonedPenguin. would be a bit more helpful w/ the tool itself.

JamesxX:
Done, Updated first post aswell

GitSparTV:
This is very useful especially for servers that wants to hide specific tools.
Spawnmenu is hard to customize so this hook will definitely help.

GitSparTV:
Hoewever, hiding tools from spawnmenu doesn't prevent them from using

JamesxX:
> This is very useful especially for servers that wants to hide specific tools.
> Spawnmenu is hard to customize so this hook will definitely help.

Thanks, though this was suggested so long ago I almost had to remind myself what exactly I proposed. The change seemingly won't be implemented, so it might be worth checking dates in the future!

Best wishes

FlorianLeChat:
The same thing could be done for spawnmenu tabs.. 😞

REALalphas:
Good idea, add same to spawnmenu

robotboy655:
This implementation does not have an effect live, which may be unwanted in case of things like hiding the tools/tabs for non admins only, which can happen mid game, not just on loading time.

PR #1299, opened 2016-12-17
Allow weapons and entities to set their own classname

Kefta:
"How could this possibly be useful?" Some weapons and entities are made to override ones created by the map, or default engine instances. For example, a CS:S weapon pack would do this to make weapons on CStrike maps usable, however, sometimes, class names will conflict with other addons that attempt to do the same thing. I made a convar in my weapon/entity pack to let the server owner decide whether to override default entities with SENTs or SWEPs, and this setting is taken into account when creating the weapon. I currently have to use a nasty __newindex hack to get the ClassName overwritten since weapons/ents.Register will force the name of the file upon the weapon, and this simple change will remove the need for that while not breaking any addons. The engine does the same exact thing with LINK_ENTITY_TO_CLASS since it manually sends in what it wants the entity class name to be, which is almost always different than the file name.


Kefta:
Squashed

Kefta:
This is more than 5 years old and my oldest PR, please consider merging this simple quality-of-life feature that breaks nothing.

PR #1300, opened 2016-12-21
Weapons and entities now copy their userdata and metatables

Kefta:
Deep copies are now made of weapons and entities in Get. Also, BaseClass keys are no longer set in any table except the base SWEP/ENT table. Lastly, the SWEPs base is now set in Register. Matrices are not copied due to https://github.com/Facepunch/garrysmod-requests/issues/880, but the code to do so is sitting in a comment.


garrysmod/lua/includes/modules/scripted_ents.lua
```diff
@@ -22,18 +22,35 @@ BaseClasses[ "filter" ] = "base_filter"
...
-
 	end
 
-	t[ "BaseClass" ] = base
+end
+
+local function DeepCopy( t )
+
+	local new = {}
+	setmetatable( new, debug.getmetatable( t ) )
```
bmwalters: Shouldn't this just be `getmetatable( t )`?

Kefta: This is what table.Copy uses: https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/includes/extensions/table.lua#L28

Not sure if they're equivalent

bmwalters: The difference is that debug.getmetatable ignores the `__metatable` of the object. Since it's a copy method, I'd say we want to use the `debug` version.

mcNuggets1:
This has conflicts, it seems.

Kefta:
No clue why the diff is so messy, but a bit annoying constantly resolving conflicts for PRs ignored for years.

neico:
I assume line-ending changes are the culprit for the diff mess.

Kefta:
Yeah I noticed that when ignoring them after I made the change, but I resolved the conflicts through the GitHub editor which I assumed would maintain line endings.

robotboy655 (changes requested):
These and merge conflicts need to be addressed before this can be merged. (It will be if the issues are addressed)

robotboy655:
Closing due to lack of activity.

robotboy655:
It's been enough time. You can open a new one if you ever get around to it.

Kefta:
Yeah I started rewriting this and got distracted, I'm gonna go with a different solution so a new PR would be better

PR #1302, opened 2016-12-23
Added trigger entity type

ForrestMarkX:
Added a trigger entity type that simulates the way CBaseTrigger works

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/base_trigger.lua
+++ b/garrysmod/gamemodes/base/entities/entities/base_trigger.lua
@@ -14,7 +14,7 @@ end
 	Name: IsTouchedBy
 -----------------------------------------------------------]]
 function ENT:IsTouchedBy(ent)
-	return self.Entities[ent:EntIndex()]
+	return self.Entities[ent]
 end
 
 --[[---------------------------------------------------------
@@ -23,7 +23,7 @@ end
 -----------------------------------------------------------]]
 function ENT:StartTouch(ent)
 	if ( !self:PassesTriggerFilters(ent) ) then return end
-	self.Entities[ent:EntIndex()] = ent
+	self.Entities[ent] = ent
 	
 	self:Input("OnStartTouch", self, ent)
 end
@@ -34,7 +34,7 @@ end
 -----------------------------------------------------------]]
 function ENT:EndTouch(ent)
 	if ( !self:IsTouchedBy(ent) ) then return end
-	self.Entities[ent:EntIndex()] = nil
+	self.Entities[ent] = nil
 	
 	self:Input("OnEndTouch", self, ent)
 
@@ -60,7 +60,7 @@ end
 -----------------------------------------------------------]]
 function ENT:Touch(ent)
 	if ( !self:PassesTriggerFilters(ent) ) then return end
-	if ( !self.Entities[ent:EntIndex()] ) then self.Entities[ent:EntIndex()] = ent end
+	if ( !self.Entities[ent] ) then self.Entities[ent] = ent end
 	
 	self:Input("OnTouch", self, ent)
 end
--- a/garrysmod/gamemodes/base/entities/entities/base_trigger.lua
+++ b/garrysmod/gamemodes/base/entities/entities/base_trigger.lua
@@ -1,0 +1,95 @@
+
+ENT.Base = "base_entity"
+ENT.Type = "trigger"
+
+--[[---------------------------------------------------------
+	Name: Initialize
+-----------------------------------------------------------]]
+function ENT:Initialize()
+	self:SetTrigger(true)
+	self.Entities = {}
+end
+
+--[[---------------------------------------------------------
+	Name: IsTouchedBy
+-----------------------------------------------------------]]
+function ENT:IsTouchedBy(ent)
+	return table.HasValue(self.Entities, ent)
+end
+
+--[[---------------------------------------------------------
+	Name: StartTouch
+	Desc: Called when a entity starts touching us
+-----------------------------------------------------------]]
+function ENT:StartTouch(ent)
+	if not self:PassesTriggerFilters(ent) then return end
+	table.insert(self.Entities, ent)
+	
+	self:Input("OnStartTouch", self, ent)
+end
+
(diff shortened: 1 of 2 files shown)


neico:
Wait there wasn't a basetrigger entity?
You got my vote then, even though the syntax looks a bit out of the norm... (ex. I'd avoid using all those table functions where there's faster alternatives, spaces, and C style operators)

I probably would've mentioned it in the title instead of being so ambiguous

ForrestMarkX:
Yeah I can easily convert the LUA style to C++ if requested

bmwalters:
Just as general Lua tips (not in support or against this PR):
Instead of:
```
local t = {}
table.insert(t, 5)
print(table.HasValue(t, 5)) -- true
print(table.HasValue(t, 3)) -- false
table.RemoveByValue(t, 5)
print(table.HasValue(t, 5)) -- false
```
Use
```
local t = {}
t[5] = true
print(t[5] == true) -- true
print(t[5] == true) -- false
t[5] = nil
print(t[5] == true) -- false
```

The latter form is much faster.

ForrestMarkX:
Interesting, thanks for the tip. I'll update the code

bmwalters:
You don't need to use `:EntIndex()` either. Anything can be a table key, even entities. It won't slow down your code.

neico:
Unless he uses a different for loop (pairs() ipairs())

ForrestMarkX:
Yeah, I figured using a number would be better but can't use EntIndex because of the stuff in EndTouch

bmwalters:
@neico No, he still couldn't use `ipairs` even if he used numbers. Entities touching this entity are not guaranteed to have sequential EntIndexes. `ipairs` only works when the keys are numeric **and sequential**.

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
```diff
@@ -0,0 +1,95 @@
...
+function ENT:Initialize()
+	self:SetTrigger(true)
+	self.Entities = {}
+end
+
+--[[---------------------------------------------------------
+	Name: IsTouchedBy
+-----------------------------------------------------------]]
+function ENT:IsTouchedBy(ent)
+	return self.Entities[ent]
```
neico: shouldn't that be `return IsValid( self.Entities[ent] )` or something like that?
Else it'll return the entire entity instead of a boolean.

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
```diff
@@ -0,0 +1,95 @@
...
+-----------------------------------------------------------]]
+function ENT:EndTouch(ent)
+	if ( !self:IsTouchedBy(ent) ) then return end
+	self.Entities[ent] = nil
+	
+	self:Input("OnEndTouch", self, ent)
+
+	local bFoundOtherTouchee = false
+	local iSize = #self.Entities
+	for i=iSize, 0, -1 do
```
bmwalters: These lines won't work anymore with these changes. Use:
```
for hOther, _ in pairs(self.Entities) do
```

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
```diff
@@ -0,0 +1,95 @@
...
+	end
+end
+
+--[[---------------------------------------------------------
+	Name: Touch
+	Desc: Called when a entity touches us
+-----------------------------------------------------------]]
+function ENT:Touch(ent)
+	if ( !self:PassesTriggerFilters(ent) ) then return end
+	if ( !self.Entities[ent] ) then self.Entities[ent] = ent end
```
neico: change into IsTouchedBy?

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
bmwalters: There's not really a problem with returning the entity, but if you cared I would prefer `return self.Entities[ent] != nil`

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
ForrestMarkX: They should as this is just getting the tables length then doing a for loop going down

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
bmwalters: He could, but doing this saves the overhead of that function call. It might make sense if we want more logic for `IsTouchedBy`, but for now it's probably worth it to duplicate this code.

ForrestMarkX: Yeah, it's basically the same

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
ForrestMarkX: Well the check would return false anyway if the entity isn't valid and true otherwise

garrysmod/gamemodes/base/entities/entities/base_trigger.lua
bmwalters: No, because if you look lower then you are using the line `local hOther = self.Entities[i]`. Here, you are indexing the Entities table with the key of `i`. Now, the keys are no longer numbers. `self.Entities[i]` will be nil.

bmwalters: Additionally, you cannot use the length operator to accurately find the size of a table that is not both numeric and sequential.

ForrestMarkX: Ohh yeah I see now.

robotboy655:
The "brush" entity type already is a CBaseTrigger, but without the outputs and PassesTriggerFilters check.

I thought about adding that stuff back in 2016, but again, I am afraid it will cause problems for existing code.
(https://wiki.garrysmod.com/page/ENTITY/PassesTriggerFilters)

This pull request wasn't tested and doesn't actually work, you can't add custom entity types like this.

PR #1303, opened 2016-12-23
Ported UTIL_ClipTraceToPlayers to util.ClipTraceToPlayers

ForrestMarkX:
Since there's no way to call it from LUA I decided to port over UTIL_ClipTraceToPlayers to LUA. I'd need someone to look this over though as some functions I do not know if they work the same as they did in the C++ version.

Check the comments I left

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -362,7 +362,7 @@ local function DistanceToTrace( pos, traceStart, traceEnd, along, pointOnTrace )
 		along = rangeAlong
 	end
 
-	local range
+	local range = 0
 	if rangeAlong < 0.0 then
 		range = -(pos - traceStart):Length()
 
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -394,7 +394,7 @@ function util.ClipTraceToPlayers( vecAbsStart, vecAbsEnd, mask, filter, tr )
 	local trace = { start = vecAbsStart, endpos = vecAbsEnd, filter = filter, mask = bit.bor(mask, CONTENTS_HITBOX) }
 	
 	for _, ply in pairs(player.GetAll()) do
-		if !IsValid(ply) or !ply:Alive() then continue end
+		if !IsValid(ply) || !ply:Alive() then continue end
 		if ply:IsDormant() then continue end
 
 		local range = DistanceToTrace( ply:WorldSpaceCenter(), vecAbsStart, vecAbsEnd )
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -347,3 +347,64 @@ function util.RemovePData( steamid, name )
 	sql.Query( "DELETE FROM playerpdata WHERE infoid = "..SQLStr(name) )
 	
 end
+
+--[[---------------------------------------------------------
+   Name: ClipTraceToPlayers
+   Desc: Traces across players and returns a modified trace
+-----------------------------------------------------------]]
+local function DistanceToTrace( pos, traceStart, traceEnd, along, pointOnTrace )
+	local to = pos - traceStart
+	local dir = traceEnd - traceStart
+	//Dunno if NormalizeInPlace translates to GetNormalized or Normalize
+	local length = dir:GetNormalized()
+
+	local rangeAlong = dir:Dot(to)
+	if along then
+		along = rangeAlong
+	end
+
+	local range
+	if rangeAlong < 0.0 then
+		range = -(pos - traceStart):Length()
+
+		if pointOnRay then
+			pointOnRay = traceStart
+		end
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/util.lua
```diff
@@ -347,3 +347,64 @@ function util.RemovePData( steamid, name )
...
 end
+
+--[[---------------------------------------------------------
+   Name: ClipTraceToPlayers
+   Desc: Traces across players and returns a modified trace
+-----------------------------------------------------------]]
+local function DistanceToTrace( pos, traceStart, traceEnd, along, pointOnTrace )
+	local to = pos - traceStart
+	local dir = traceEnd - traceStart
+	//Dunno if NormalizeInPlace translates to GetNormalized or Normalize
```
bmwalters: FYI, NormalizeInPlace means that it directly mutates the given vector (in place). Looking at the documentation for both GetNormalized and Normalize, GetNormalized returns a new vector while leaving the source unmutated. Normalize mutates the original vector. Thus, Normalize would be the correct translation for NormalizeInPlace. This would mean that there wouldn't be a return value, so you'd have to be aware of that.

bmwalters: The benefit of in-place mutation is that you can skip the overhead of creation of a new object.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -347,3 +347,64 @@ function util.RemovePData( steamid, name )
...
+	local trace = { start = vecAbsStart, endpos = vecAbsEnd, filter = filter, mask = bit.bor(mask, CONTENTS_HITBOX) }
+	
+	for _, ply in pairs(player.GetAll()) do
+		if !IsValid(ply) || !ply:Alive() then continue end
+		if ply:IsDormant() then continue end
+
+		local range = DistanceToTrace( ply:WorldSpaceCenter(), vecAbsStart, vecAbsEnd )
+		if (range < 0.0 || range > maxRange) then continue end
+
+		//I couldn't find the code for ClipRayToEntity to hopefully util.TraceEntity works the same - FMX
```
Kefta: TraceEntity does not work the same. See my request here: Facepunch/garrysmod-requests#664

Kefta:
If you're adding more Source functions to Garry's Mod, I did a bunch here if you want to take any code: https://github.com/Kefta/gslib/tree/master/lua/code_gs/lib

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -347,3 +347,63 @@ function util.RemovePData( steamid, name )
...
 end
+
+--[[---------------------------------------------------------
+   Name: ClipTraceToPlayers
+   Desc: Traces across players and returns a modified trace
+-----------------------------------------------------------]]
+local function DistanceToTrace( pos, traceStart, traceEnd, along, pointOnTrace )
+	local to = pos - traceStart
+	local dir = traceEnd - traceStart
+	local length = dir:Normalize()
```
bmwalters: > This would mean that there wouldn't be a return value, so you'd have to be aware of that.

Kefta:
Normalize does not return the radius in Lua, unfortunately. You'll have to calculate it manually

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -355,7 +355,9 @@ end
 local function DistanceToTrace( pos, traceStart, traceEnd, along, pointOnTrace )
 	local to = pos - traceStart
 	local dir = traceEnd - traceStart
-	local length = dir:Normalize()
+	
+	dir:Normalize()
+	local length = dir
```
Bo98: Are you not looking for this to be `dir:Length()` (and be _before_ the normalise line)?

ForrestMarkX: I don't believe so as the C++ code does "float length = dir.NormalizeInPlace();"

Kefta:
@ForrestMarkX You are correct, NormalizeInPlace in C++ returns the vector length, however, in Lua, Normalize returns nothing. Right now, you are normalizing the vector then setting length to a vector. What you need to do is:

local length = dir:Length()
dir:Normalize()

willox:
util.TraceLine already calls UTIL_ClipTraceToPlayers. Where else are you going to need to call it?

Kefta:
@willox I call it after a hull trace for shotgun pellets in a custom FireBullets function. But I did not know that it was called in TraceLine; how else have the trace functions been modified in Garry's Mod?

ForrestMarkX:
@willox I did not know this either

Kefta:
Did you test to check if they're the same? Alsor, please squash commits.

ForrestMarkX:
@Kefta They seem to be but I don't have a reliable way to test. Also I don't know how to squash commits without screwing up the entire branch

Kefta:
This is called internally now and thus isn't necessary.

GitSparTV:
@Kefta Close?

PR #1304, opened 2016-12-23
Added table.InheritNoBaseClass and table.Inherit now recursively inherits sub-tables

Kefta:
Additionally, table.Inherit now has an additional argument to choose what value it wants to use to store the baseclass


robotboy655:
Wouldn't it be more sensible to add a new function with the new functionality as `table.InheritDeep` or `table.InheritRecursive`, rather than changing an existing function and unnecessary risk breaking existing scripts and then adding the old function under a different name?

PR #1305, opened 2016-12-24
Added ammo accessors and convar support

Kefta:
Adds accessors for all ammo keys, as well as adding support for custom keys to be added. Also, DMG_SNIPER and the ammo flags enums were added since they were missing.

This solves https://github.com/Facepunch/garrysmod-requests/issues/857, https://github.com/Facepunch/garrysmod-requests/issues/646 / https://github.com/Facepunch/garrysmod-requests/issues/861

"This is messy and should be done in the engine." I agree, but this is still a viable solution as the maxcarry accessor was added over a year ago and that was it. Side benefit is that these being implemented in Lua is slightly faster, although it's pretty negligible.

### DIFF
--- a/garrysmod/lua/includes/extensions/game.lua
+++ b/garrysmod/lua/includes/extensions/game.lua
@@ -1,54 +1,266 @@
 
+-- Half-Life damage type
+DMG_SNIPER = bit.lshift( DMG_BUCKSHOT, 1 )
+
+-- Ammo flags
+AMMO_FORCE_DROP_IF_CARRIED = 0x1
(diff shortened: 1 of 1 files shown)


Kefta:
Just noticed an issue with the convar support; will fix soon.

Kefta:
Fixed convar support, and custom ammo keys with non-number types. I also added a key "displayname" to the ammo structure as a shortcut for having to add the ammo's display name to the language table. Lastly, DMG_MISSILEDEFENSE was added from HL2.

gspetrou:
Bump, this should really be looked into.

robotboy655:
Most of this was added into the game recently, any missing enums should be reported, so that they are all defined in 1 place.

PR #1306, opened 2016-12-24
Add Upvalue Utility Functions

LeQuackers:
The intended use of these functions are to allow developers to modify modules/extensions without rewriting the whole thing. In my case, I needed a way to get a list of registered player classes as well as add a PLAYER:Destroy hook to the player_manager module.
```Lua
-- as of this post player_manager.GetTable() doesn't exist
local Type = util.GetTableUpvalue( player_manager, "Type" )
function player_manager.GetTable() return Type end

PrintTable( player_manager.GetTable() )
```

```Lua
-- adding a PLAYER:Destroy hook is too lengthy, here's another basic example
LocalPlayer():UniqueIDTable("meh table")["foo"] = "bar"

local g_UniqueIDTable = util.GetTableUpvalue( FindMetaTable( "Player" ), "g_UniqueIDTable" )
function player.GetUniqueIDTables() return g_UniqueIDTable end

PrintTable( player.GetUniqueIDTables() )
```

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -160,6 +160,93 @@ function util.NiceFloat( f )
 end
 
 
+--
+-- util.GetTableUpvalues( table )
+-- returns all upvalues from a table of functions
+--
+function util.GetTableUpvalues( tbl )
+
+	local upvalues = {}
+
+	for k, v in pairs( tbl ) do
+	
+		if ( isfunction( v ) ) then
+		
+			local i = 1
+			while ( true ) do
+			
+				local key, value = debug.getupvalue( v, i )
+				
+				if ( key == nil ) then break end
+				upvalues[ key ] = value
+				i = i + 1
+				
+			end
+		
+		end
+		
+	end
+	
+	return upvalues
+	
+end
+
+--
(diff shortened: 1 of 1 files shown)


LeQuackers:
I noticed the emojis, any particular reason this would be a bad idea?

neico:
The entire system is designed to not allow modifications, especially the player_manager one.
The fact that you rely on the debug lib makes it not any better (it might not get removed but who knows...).

If you look at other modules then you can clearly see that some modules account for modifications.

No one's stopping you from writing a utility lib that everyone can use, but imo this has no place in vanilla gmod. (ex. try searching for "debug." in the vanilla lua files, there might be close to 0 results)

edgarasf123:
There is a really good reason why private variables should stay private, similarly why classes in C++ have private members.

The developers who create modules doesn't have to worry about breaking stuff by changing how a module works inside as long as public methods and the general functionality of the module stays the same.

Now, image people start accessing private variables of a modules and incorporating them inside their gamemodes/addons, and now no changes possible to a module without breaking the other addons/gamemodes.

This should stay away from vanilla, last thing we want is encourage developers to use this.

LeQuackers:
@edgarasf123 You got me with the conflicting addons example, RIP pull request.

PR #1307, opened 2016-12-24
Added weapons/scripted_ents.PreRegistered hook

Kefta:
This allows modification to the registration table before it's registered, or block it completely. This would be useful for addons that are overriding default Lua weapons or servers that want to edit weapons in workshop addons without decompiling.
100th PR :)

### DIFF
--- a/garrysmod/lua/includes/modules/scripted_ents.lua
+++ b/garrysmod/lua/includes/modules/scripted_ents.lua
@@ -52,6 +52,8 @@ end
 
 function Register( t, name )
 
+	if ( hook.Run( "PreSENTRegister", t, name ) == true ) then return end
+
 	local Base = t.Base
 	if ( !Base ) then Base = BaseClasses[ t.Type ] end
 
--- a/garrysmod/lua/includes/modules/scripted_ents.lua
+++ b/garrysmod/lua/includes/modules/scripted_ents.lua
@@ -52,7 +52,7 @@ end
 
 function Register( t, name )
 
-	if ( hook.Run( "PreRegisterSENT", t, name ) == true ) then return end
+	if ( hook.Run( "PreRegisterSENT", t, name ) == false ) then return end
 
 	local Base = t.Base
 	if ( !Base ) then Base = BaseClasses[ t.Type ] end
--- a/garrysmod/lua/includes/modules/scripted_ents.lua
+++ b/garrysmod/lua/includes/modules/scripted_ents.lua
@@ -52,7 +52,7 @@ end
 
 function Register( t, name )
 
-	if ( hook.Run( "PreSENTRegister", t, name ) == true ) then return end
+	if ( hook.Run( "PreRegisterSENT", t, name ) == true ) then return end
 
 	local Base = t.Base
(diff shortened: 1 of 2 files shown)


markusmarkusz:
The names sound bad. Maybe something like PreWeaponRegister and PreSENTRegister or something like that.
(Also, it might be good to call it with the GAMEMODE table. Could be useful for gamemodes.)

Kefta:
@markusmarkusz These hooks are gamemode independent, and thus should not use the GM table. Also, those hook names do sound better; second opinions?

markusmarkusz:
When you call the hook with the GAMEMODE table, gamemode creators can limit weapons to a special weapon base by simply using this hook.
This is much easier.

As an example if I want that my example gamemode should only allow weapons with the weapon base `weapon_example_base` I can simply do: (Since `weapons.PreRegister` can't work as function name in this case I use `PreWeapon` as name.)
```lua
function GM:PreWeapon(t, name)
    if t.Base ~= "weapon_example_base" then return true end
end
```

This is my opinion. No idea how others think about it, but I think it would be easier this way.

Jvs34:
Just use hook.Run instead of Call, you don't have to fiddle with any gamemode stuff and gamemode makers will be able to simply add a GM:Whatever on their own and everyone is happy.

Acecool:
Not a bad idea; I personally extract the gma file and delete the lua files but that means loading time is longer ( and editing these files and putting them in an addon hasn't worked for me in the past ).

I'm not a fan of the name PreSENTRegister ( ie UpperCamelCase that bleeds into the next word - SENT is an accepted term in the game but 3 words bleeding over is ugh )..

How about one of these:
PreRegisteredSENT
PreRegisteredScriptedEnt
PreRegisteredScriptedEntity

fruitwasp:
`PreRegisterScriptedEntity` & `PreRegisterWeapon`

bmwalters:
@fruitwasp currently there are two hooks which use "SENT" in the name and 0 which use "ScriptedEntity" (the two being SANDBOX:PlayerSpawnSENT and SANDBOX:PlayerSpawnedSENT)

fruitwasp:
@bmwalters might be better to use SENT yeh. I have edited my comment.

Acecool:
They have SENT at the end so 3 words don't bleed through with UpperCamelCase.. I would support it being PreRegisteredSENT ( A PreRegisteredSWEP may be nice too but if the SENT works for ALL ents such as nextbot, sweps, scripted ents, etc... then we don't need it ).

Kefta:
I'm all for PreRegisterSENT, but should it be PreRegisterSWEP or PreRegisterWeapon?

markusmarkusz:
I am of the opinion that PreRegisterSWEP is better. They are Scripted WEaPons.

Kefta:
Updated

Kefta:
Could this be looked into merging? Having an official way to modify entities pre-registration would be great for changing workshop weapons/entities without re-registration hacks.

Kefta:
Every 15 seconds, a new workshop addon is published using an weapons.Register or InitPostEntity hack to replace entities. When will these atrocities come to an end?

PR #1308, opened 2016-12-24
Fixed endroundtimer not stopping when forcing roundrestart

GamefreakDE:
I always wondered why the round was restarting twice when forcing roundrestart post round.


svdm:
Nice, though would've been prevented if timer.Stop printed a (dev) message when you try to stop a timer that does not exist.

PR #1316, opened 2017-01-04
Add an option to disable halo on physgun unfreeze

MrJegorN:
I need this because the material on my SENT is not rendering properly with these halos


Kefta:
You didn't add it as an argument to the function.

robotboy655:
Use http://wiki.garrysmod.com/page/halo/RenderedEntity

thegrb93:
I wish halos would just render the Sent's model instead of invoking its draw function. Doing so assumes the Sent isn't using stencils.

PR #1317, opened 2017-01-04
TTT: Fix Error?

markusmarkusz:
```
[ERROR] gamemodes/terrortown/gamemode/cl_voice.lua:656: attempt to compare number with nil
  1. Draw - gamemodes/terrortown/gamemode/cl_voice.lua:656
   2. unknown - gamemodes/terrortown/gamemode/cl_hud.lua:350
```
I'm not sure why `client.voice_battery` becomes nil.
But I am very sure that it was not an addon.
This solution may not be the best. Rather, it would be better to know why `client.voice_battery` was nil.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
@@ -653,7 +653,7 @@ end
 local speaker = surface.GetTextureID("voice/icntlk_sv")
 function VOICE.Draw(client)
    local b = client.voice_battery
-   if b >= battery_max then return end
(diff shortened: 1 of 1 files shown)


svdm:
I suppose this could happen in an edge case where `VOICE.Draw` is called before the client state has been initialised after the map loads (because that happens in the clientside InitPostEntity). Do you know if this error occurred right after a map loaded?

markusmarkusz:
I'll check tomorrow if I still have the server logs.

markusmarkusz:
Apparently an addon (possibly Pointshop 2) was responsible for it. At the server start there was an error message, which I had overlooked. After a server restart the error was away.

PR #1318, opened 2017-01-04
Fixed string.NiceTime

CornerPin:
Method for calculation length of the year "60 * 60 * 24 * 7 * 52" replaced with "60 * 60 * 24 * 365", because 7*52 is 364.


robotboy655:
It's pointless to combine 7*52 but not 60*60*24

Kefta:
@robotboy655 you misunderstood -- 7*52 = 364; the multiplier should be 365.

Acecool:
365.25 is accurate..

robotboy655:
Fair enough

PR #1319, opened 2017-01-04
Update ConVar retrieval

MrUnknownGamer:
Updated all GetConVar{type}("thing") to GetConVar("thing"):Get{type} since GetConVar{type} is deprecated

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_equip.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_equip.lua
@@ -325,8 +325,6 @@ local function TraitorMenuPopup()
    dfields.desc:SetContentAlignment(7)
    dfields.desc:MoveBelow(dfields.type, 1)
 
-   local iw, ih = dinfo:GetSize()
-
    local dhelp = vgui.Create("ColoredBox", dinfobg)
    dhelp:SetColor(Color(90, 90, 95))
    dhelp:SetSize(diw, dih - 205)
--- a/garrysmod/gamemodes/terrortown/gamemode/player.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player.lua
@@ -507,7 +507,7 @@ local function CheckCreditAward(victim, attacker)
 
    -- DETECTIVE AWARD
    if IsValid(attacker) and attacker:IsPlayer() and attacker:IsActiveDetective() and victim:IsTraitor() then
-      local amt = GetConVarNumber("ttt_det_credits_traitordead") or 1
+      local amt = GetConVar("ttt_det_credits_traitordead"):GetInt() or 1
       for _, ply in pairs(player.GetAll()) do
          if ply:IsActiveDetective() then
             ply:AddCredits(amt)
@@ -523,7 +523,7 @@ local function CheckCreditAward(victim, attacker)
       local inno_alive = 0
       local inno_dead = 0
       local inno_total = 0
-      
+
       for _, ply in pairs(player.GetAll()) do
          if not ply:GetTraitor() then
             if ply:IsTerror() then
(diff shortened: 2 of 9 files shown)


garrysmod/gamemodes/terrortown/gamemode/player.lua
```diff
@@ -546,9 +546,9 @@ local function CheckCreditAward(victim, attacker)
       end
 
       local pct = inno_dead / inno_total
-      if pct >= GetConVarNumber("ttt_credits_award_pct") then
+      if pct >= GetConVar("ttt_credits_award_pct"):GetInt() then
```
willox: This'll screw up even with the default setting of `ttt_credits_award_pct 0.35` because GetInt returns a whole number.

garrysmod/gamemodes/terrortown/gamemode/shared.lua
```diff
@@ -143,7 +143,7 @@ local ttt_playercolors = {
 
 CreateConVar("ttt_playercolor_mode", "1")
 function GM:TTTPlayerColor(model)
-   local mode = GetConVarNumber("ttt_playercolor_mode") or 0
+   local mode = GetConVar("ttt_playercolor_mode"):GetInt() or 0
```
Bo98: Meh, I'd use the `ConVar` object returned by the `CreateConVar` literally two lines up but I guess this doesn't matter anymore now `GetConVar` caches things.

Kefta:
Why not just cache each ConVar in the file instead of calling GetConVar in the function?

garrysmod/gamemodes/terrortown/gamemode/shared.lua
markusmarkusz: Storing the ConVar Object returned by CreateConVar is still faster.

robotboy655:
I feel like this is a pointless change that has no benefits. GetConVarNumber & co already cache the convars anyway

Kefta:
@robotboy655 Still faster to cache the convar in the file. It skips 1-2 global function calls and a table lookup. Also, if you feel like that's the case, you should remove the deprecation from the wiki page.

garrysmod/gamemodes/terrortown/gamemode/cl_equip.lua
```diff
@@ -325,8 +325,6 @@ local function TraitorMenuPopup()
    dfields.desc:SetContentAlignment(7)
    dfields.desc:MoveBelow(dfields.type, 1)
 
-   local iw, ih = dinfo:GetSize()
```
Kefta: Wrong change?

robotboy655:
@Kefta There isn't any dramatic difference in performance between how it before and after this PR. It would be pretty much same. The point is to save the convar objects to a variable and use that variable in code that is ran often, not to do GetConVar( "x" ):GetX(), which is literally what GetConVarNumber( "x" ) does.

garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
```diff
@@ -196,7 +196,7 @@ if CLIENT then
    -- Perform the gesture using the GestureRunner system. If custom_runner is
    -- non-nil, it will be used instead of the default runner for the act.
    function plymeta:AnimPerformGesture(act, custom_runner)
-      if GetConVarNumber("ttt_show_gestures") == 0 then return end
+      if GetConVar("ttt_show_gestures"):GetInt() == 0 then return end
```
markusmarkusz: Im my opinion GetBool would be better here.

Kefta:
@robotboy655 Yeah, that's what I was saying.

markusmarkusz:
https://gist.github.com/markusmarkusz/77f7525f5583d229e09fa4bf3ded8e05
A benchmark that was done before GetConVar cached everything. But the more important here is GetConVarNumber and ConVar:GetFloat().
The difference isn't very big.

PR #1321, opened 2017-01-04
Allow for custom tooltip panels

PatrickRatzow:
This allows for Tooltip to be a custom panel if you do PANEL:SetTooltipPanel( panel ) on it.

If no panel are set it will just default to DToolTip.

Very neat if you switch between several different tooltips for different menus.

### DIFF
--- a/garrysmod/lua/includes/util/tooltips.lua
+++ b/garrysmod/lua/includes/util/tooltips.lua
@@ -55,7 +55,15 @@ function ChangeTooltip( panel )
 	
 	if ( !Text && !Panel ) then return end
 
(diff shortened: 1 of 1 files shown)


Kefta:
The code would probably look better as Tooltip = vgui.Create( Panel or "DToolTip" ); probably just a personal preference, though

PatrickRatzow:
I'd do that myself but I don't see Garry do that in his code and I want the coding style to be like his, consistency is key.

Kefta:
Fair enough

robotboy655:
http://wiki.garrysmod.com/page/Panel/SetTooltipPanel is used for a different purpose.

PR #1322, opened 2017-01-05
Crowbar and knife not testing if player is terror

GamefreakDE:
The Crowbar and the knife it not testing if the hitEnt is terror, so I wondered if it wouldnt be good to test it?
I dont know where to suggest this so I just made a pull request ^^

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_knife.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_knife.lua
@@ -81,7 +81,7 @@ function SWEP:PrimaryAttack()
       edata:SetNormal(tr.Normal)
       edata:SetEntity(hitEnt)
 
-      if hitEnt:IsPlayer() or hitEnt:GetClass() == "prop_ragdoll" then
+      if hitEnt:IsPlayer() and hitEnt:IsTerror() or hitEnt:GetClass() == "prop_ragdoll" then
          util.Effect("BloodImpact", edata)
       end
    else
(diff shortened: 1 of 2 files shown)


Kefta:
In what situation would the player not be a terrorist?

MinIsMin:
When being a ghost (in SpecDM).

markusmarkusz:
This solution isn't enough. There are some more ways to hit SpecDM users with knives or crowbars.

GamefreakDE:
Then tell me which ones, Ill add them

markusmarkusz:
Example: Thrown knife.
[Here](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_improvised.lua#L186-L194) you have to block damage if hitEnt is a player and not alive.

Also almost every weapon that use traces makes problems with the SpecDM.

GamefreakDE:
I dont need to block the damage, its enough to do that in EntityTakeDamage, I just need to make that It doesnt sticks, you are right

Bo98:
A much better solution would be to use trace filtering so that a player behind a SpecDM player can be hit.

GamefreakDE:
The Problem is:
Things the trace should not hit. Can also be a table of entities or a function with one argument:
Entity ent - The entity that the trace hit
Using a function here is super slow. Try to avoid it. <-- Thats not good

GamefreakDE:
And @markusmarkusz you forget that the Crowbar should still be able to damage entities besides players, I only care about the blood effect.

markusmarkusz:
If you attack a SpecDM User with a crowbar there will be an error message iirc.

GamefreakDE:
I fixed it already in my local test branch, I will make a pull request soon.

Bo98:
> Using a function here is super slow. Try to avoid it. <-- Thats not good

So you'd prefer the crowbar and knife to occasionally do no damage? Primary/SecondaryAttack isn't called every frame so it shouldn't be that much of an issue. DrawHUD might be.

I guess you could also create a table of players that are not terrorists to avoid a function.

PR #1323, opened 2017-01-05
TTT: Call PlayerLoadout as hook

markusmarkusz:
Useful for addons that want to block loadouts or something like this.
But maybe not necessary.


MrUnknownGamer:
Why not use hook.Run()? That way, you don't have to hook onto the global variable GAMEMODE when running the hook.

markusmarkusz:
There is no big difference.

PR #1325, opened 2017-01-14
Making a hook to prevent people from buying Equipment

GamefreakDE:
So, long story short, I wanted and hook to prevent people from buying stuff, If you think that things dont work then please let me know.

LG Gamefreak


GamefreakDE:
I fucked up

robotboy655:
What a suprise

PR #1326, opened 2017-01-15
Function to avoid overriding Equipment IDs

GamefreakDE:
Making Equipment in TTT was always really hard due to the high probability of someone overriding your number, this should fix it. Open to any suggestions.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/equip_items_shd.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/equip_items_shd.lua
@@ -24,6 +24,7 @@ EQUIP_ARMOR    = 1
 EQUIP_RADAR    = 2
 EQUIP_DISGUISE = 4
 
+EQUIP_COUNT = 4
 
 -- Icon doesn't have to be in this dir, but all default ones are in here
 local mat_dir = "vgui/ttt/"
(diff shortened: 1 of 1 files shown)


markusmarkusz:
Why not just do it like AddScoreGroup and GROUP_ *?
That would be much easier in my opinion.

GamefreakDE:
Do you mean like this?

function AddEquipmentID(name)  -- Utility function to register a new ID
  if _G["EQUIP_"..name] then error("ID of name '"..name.."' already exists!") return end
  EQUIP_COUNT = EQUIP_COUNT * 2
  _G["EQUIP_"..name] = EQUIP_COUNT
end

markusmarkusz:
Yes, I think this would be much easier and cleaner.

GamefreakDE:
The problem is, wouldnt this cause problems with people who just write EQUIP_MARTY = 1024 as an example? I dont know how we should fix this because most past addon creators wont change their addons.

markusmarkusz:
There would be compatibility problems. I would say we're doing it just like with the UniqueID to SteamID conversion. This has also killed some addons and destroyed all bot functions in TTT. However, this affected only a few addons.
I also don't know many addons who use this equipment table. There aren't many of them. It shouldn't do much damage. Especially not with EQUIP_MARTY, since you need to have 8 different items which use this equipment table.. And that would be really much.

GamefreakDE:
The Problem is that in my workshop are actually do many equipment items, that's the reason why I made the pull :P

svdm:
I don't like hacking the global table to generate the `EQUIP_*` global variable. It's going to do unexpected things if `name` contains spaces or special characters, etc. It's also not required for equipment to have their ID in a global with a name like EQUIP_* at all.

If someone wants to make an addon with 3 pieces of equipment and store their ID in a table, that's perfectly fine. It's up to the addon author to check if a player has the equipment and make it do something, and add it in the right UI places, so as long as the ID is *somewhere* it's fine.

Why not something like this, which is cleaner and more flexible:
```
EQUIP_MAX = 4

function GenerateNewEquipmentID()
   EQUIP_MAX = EQUIP_MAX * 2
   return EQUIP_MAX
end
```

Which you would use in an addon like:
```MY_EQUIP_THING = GenerateNewEquipmentID()``

I realise the `G[...]` method is the same as AddScoreGroup, but I don't like it there either and wouldn't mind replacing it with a cleaner solution (although I'll have to keep most of the current version for compatibility).

GamefreakDE:
What about my first commit, how do you think about that?

PR #1327, opened 2017-01-18
Prevent People from buying equipment

GamefreakDE:
Added new hook "TTTCanOrderEquipment" to prevent people from buying equipment.
Open for any suggestions.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
@@ -334,6 +334,11 @@ local function HasPendingOrder(ply)
    return timer.Exists("give_equipment" .. tostring(ply:SteamID()))
 end
 
+function GM:TTTCanOrderEquipment(ply, id, is_item)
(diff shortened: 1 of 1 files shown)


markusmarkusz:
Something like TTTCanOrderEquipment might be better. (Consistence with other hooks like TTTCanIdentifyCorpse, etc.)

windowsok:
Thanks for
  the information your article brings. I see the novelty of your writing, I
  will share it for everyone to read together. I look forward to reading many
  articles from you.
<a
  href="https://wingsio0.com">wingsio</a>

Kamshak:
Can this be fixed to use

``` if hook.Run("TTTCanOrderEquipment", ply, id, is_item) == false then return end ```

That way you could add specific logic and not return anything for items you don't want to handle (so that other addons also get a chance to change the result)

sannysc:
The behavior you described is already supported. If you don't want to deal with the item, then don't return anything, and the hook system will move on to the next listener. It is only when you return something other than ```nil``` does the hook stop.

PR #1329, opened 2017-01-20
Add PlayerCheckLimit hook

Python1320:
This could be a welcome addition to allow per-player limits without everyone having to resort to detouring CheckLimit.

Thoughts? Should we instead replace the convar result?

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/player_extension.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/player_extension.lua
@@ -14,10 +14,14 @@ function meta:CheckLimit( str )
 	local c = cvars.Number( "sbox_max" .. str, 0 )
 	local count = self:GetCount( str )
 	
-	local ret = hook.Run( 'PlayerCheckLimit', self, str, c, count )
-	if ( ret != nil ) then return ret end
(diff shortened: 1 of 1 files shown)


thegrb93:
This is a good addition. It might be useful to also pass the limit and the player's current count for convenience.

mcNuggets1:
This could be really useful.

Python1320:
Added the information

robotboy655:
I have a few problems with this PR:
1) The `c` and `count` arguments should be switched around to be in more expected order of `currentCount, maxCount`.
2) Returning false doesn't display "You've hit the limit" error message to the player

Here's my proposal:
```
function meta:CheckLimit( str )

	-- No limits in single player
	if ( game.SinglePlayer() ) then return true end

	local c = cvars.Number( "sbox_max" .. str, 0 )
	local count = self:GetCount( str )

	local ret = hook.Run( "PlayerCheckLimit", self, str, count, c )
	if ( ret != nil ) then
		if ( !ret && SERVER ) then self:LimitHit( str ) end
		return ret
	end

	if ( c < 0 ) then return true end

	if ( count > c - 1 ) then
		if ( SERVER ) then self:LimitHit( str ) end
		return false
	end

	return true

end
```

PR #1330, opened 2017-02-04
Fix YouTube videos not loading in DHTML

ZapsterStudios:
Should resolve the following error when trying to view YouTube videos:
``Uncaught TypeError: Property 'warn' of object #<error> is not a function``

### DIFF
--- a/garrysmod/lua/vgui/dhtml.lua
+++ b/garrysmod/lua/vgui/dhtml.lua
@@ -16,7 +16,7 @@ function PANEL:Init()
 	-- Implement a console.log/warn - because awesomium doesn't provide it for us anymore.
 	--
 	self:AddFunction( "console", "log", function( param ) self:ConsoleMessage( param ) end )
-	self:AddFunction( "console", "warn", function( param ) self:ConsoleMessage( param ) end )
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dhtml.lua
```diff
@@ -13,9 +13,10 @@ function PANEL:Init()
 	self.Callbacks = {}
 
 	--
-	-- Implement a console.log - because awesomium doesn't provide it for us anymore.
+	-- Implement a console.log/warn - because awesomium doesn't provide it for us anymore.
 	--
 	self:AddFunction( "console", "log", function( param ) self:ConsoleMessage( param ) end )
+	self:AddFunction( "console", "warn", function( param ) self:ConsoleMessage( param ) end )
```
bmwalters: Might want to print a prefix or something to distinguish between a normal message and a warning. Maybe you could give the message a different color.

ZapsterStudios: Good idea. Added!

SpencerSharkey: can we also add error while we're at it?

PR #1332, opened 2017-02-15
TTT: Enable scoreboard to be sorted by different criteria other than score

DirkyJerky:
Reasoning: 
- Sorting by score isn't very useful. 
- An unsorted scoreboard is very difficult to use efficiently on large servers
--- It can take several seconds to find a specific person on it, especially if some of them are color coded.
- Most people can search an alphabetically sorted list far quicker than one that is not.

Current status:
- Titles in the column header row are clickable to make the scoreboard sort by that criteria
- TTTScoreboardColumns hook _extended_ so a sorting function can be optionally added for sortable custom columns
- [Pseudo-markdown pseudo-documentation for said hook](https://github.com/garrynewman/garrysmod/pull/1332#issuecomment-282517423)

Pictures!: [Name](https://i.imgur.com/e9gA63T.jpg) | [Role - T](https://i.imgur.com/38aroS3.jpg) | [Role - D](https://i.imgur.com/gyK3kHW.jpg) | [Role - I](https://i.imgur.com/sInYCdA.jpg) | [Score](https://i.imgur.com/IjhQCxy.jpg) | [Karma](https://i.imgur.com/1w36ZVk.jpg) | [Deaths](https://i.imgur.com/jVdJN1A.jpg) | [Ping](https://i.imgur.com/NuCZCyI.jpg) | [Custom column test](https://i.imgur.com/J4TLatR.jpg)


garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -18,6 +18,83 @@ local function CompareScore(pa, pb)
...
+   local at = 0
+   local bt = 0
+   for k,v in pairs(t) do
+      if v == a:Nick() then
+         at = k
+      else
+         bt = k
+      end
+   end
+   return bt > at
```
bmwalters: Couldn't this all be replaced by:
```lua
local function CompareRowsByName(rowa, rowb)
    local plya = rowa:GetPlayer()
    if not IsValid(plya) then return false end

    local plyb = rowb:GetPlayer()
    if not IsValid(plyb) then return true end

    return plya:GetName() < plyb:GetName() -- might be >
end
```

garrysmod/gamemodes/terrortown/gamemode/init.lua
```diff
@@ -91,6 +94,8 @@ CreateConVar("ttt_det_credits_starting", "1")
 CreateConVar("ttt_det_credits_traitorkill", "0")
 CreateConVar("ttt_det_credits_traitordead", "1")
 
+-- Other
+CreateConVar("ttt_scoreboard_sorting", "1", FCVAR_NOTIFY + FCVAR_REPLICATED)
```
bmwalters: Instead of a ConVar, why not allow the clients to sort by clicking on the column headings or something?

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -18,6 +18,83 @@ local function CompareScore(pa, pb)
...
+
+   local a = pa:GetPlayer()
+   local b = pb:GetPlayer()
+
+   if not IsValid(a) then return false end
+   if not IsValid(b) then return true end
+
+
+
+
```
bmwalters: You seem to have quite a bit of whitespace here and above.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -18,6 +18,83 @@ local function CompareScore(pa, pb)
...
+   return a:Ping() > b:Ping()
+end
+
+local function CompareDeath(pa, pb)
+   if not IsValid(pa) then return false end
+   if not IsValid(pb) then return true end
```
bmwalters: The rows should always be valid at call time so these checks can be removed.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -18,6 +18,83 @@ local function CompareScore(pa, pb)
    return a:Frags() > b:Frags()
 end
 
+local function CompareAlpha(pa, pb)
+   if not IsValid(pa) then return false end
+   if not IsValid(pb) then return true end
+
+   local a = pa:GetPlayer()
+   local b = pb:GetPlayer()
```
bmwalters: Might want to use some more descriptive variable names for the rows and players. It took me longer to figure out the purpose of this code than if they had been properly labelled.

Kefta:
A few things: you should use bit.* operations for bit flags instead of normal arithmetic. In this case, CreateConVar actually supports sending in a table of flags that it will merge internally. Next, you should always test pull requests before submitting them; community feedback is great, but it works even better on working code. Also, why are you doing all of that extra sorting work in CompareAlpha?

DirkyJerky:
I just finished testing it; it works fine.

I was not the one that created the initial code; I will work on cleaning it up now

And sorting by clicking on column headers sounds like a fabulous idea.

garrysmod/gamemodes/terrortown/gamemode/init.lua
DirkyJerky: Good idea; but I believe I need to keep the ConVar as I could not find a shared object needed to host the variable. There may be one though; I'm not very familiar with the codebase.

bmwalters: It could be a property of the scoreboard object

bmwalters:
@DirkyJerky @Kefta 
https://github.com/bmwalters/garrysmod/commit/d8b7007288ce525f47a34ae20bf5a1bc5ffa5710

DirkyJerky:
Great job @bmwalters :D
We could probably make a few fake column labels alongside the other ones to set the name or role sorting options; 
Maybe make the text of the labels could change color a bit if it is selected too? 
I'll try it tomorrow.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
```diff
@@ -112,6 +112,23 @@ function PANEL:Init()
 
    self.ply_groups = {}
 
+   concommand.Add( "ttt_scoreboard_setsort", function( _, _, args )
```
bmwalters: i honestly don't think we need this

markusmarkusz: A convar to save the order would be much better, wouldn't it?

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -116,11 +104,40 @@ end
...
+         if plya:Frags() == plyb:Frags() then return plya:Deaths() < plyb:Deaths() end
+
+         return plya:Frags() > plyb:Frags()
+      elseif self.sort_mode == "role" then
+         if plya:GetRole() == plyb:GetRole() then return plya:GetName() < plyb:GetName() end
+
+         return plya:GetRole() > plyb:GetRole()
+      elseif self.sort_mode == "karma" then
+         return plya:GetBaseKarma() < plyb:GetBaseKarma()
+      else
```
markusmarkusz: Maybe add a hook, so you can also sort custom columnes too?

DirkyJerky:
Alright I added two more buttons to click to sort on; and made the active sorting option get highlighted.

DirkyJerky:
The role sorting option doesn't seem to be working...

DirkyJerky:
Every option is working smoothly it seems. <INSERT PICTURES>

DirkyJerky:
In my personal opinion; It would be a much better option to store the current sort option as a ConVar (with ARCHIVE flag set), because then it will be persistent across sessions.

bmwalters:
@DirkyJerky 
The extra column background created by the Name and Role headings and the headings themselves don't look very good IMO. I think a good way to sort by name may be clicking on the "Terrorists (10)" heading area of whatever the topmost score group is. We might be able to just forego sorting by role if there isn't a good way to fit it into the design as the row highlighting sort of makes it easy to view without grouping. I'd be interested in the input of others here.

About the ConVar, I don't think the current sort option is something that needs to persist. It's trivial to interact with the headings to change the sorting back when wanted.

Lastly, maybe the green from the "Terrorists (10)" heading background could be re-used for the text color.

Nice job getting some stuff made; it's good to have working prototypes of these features.

DirkyJerky:
@bmwalters 
I agree the name/role buttons need a better spot, but the score groups headers are a very meh spot, especially that they aren't even on the same row (and there are up to three of those headers able to be viewed simultaniously)

I also agree the ConVar isn't necessary... but strictly speaking having the scoreboard able to be sorted at all isn't necessary either.
Its about convienience: An end user who, for example, would like it to be sorted ascending via Karma, would have to fix it every time they started playing.
Putting it in a ConVar has no disadvantages for the end user and a few simple advantages for the them aswell.

DirkyJerky:
Oh I see yea it makes an extra shade column under the Role header; I'll see what I can do

markusmarkusz:
Adding these columns could cause problems with custom columns.
You can add the ConVars to the Help Screen (HELPSCRN) and make it like it is with the languages. Then users can easily decide if they want that the scoreboard is sorted by name.

DirkyJerky:
@markusmarkusz That was exactly what I was thinking!

DirkyJerky:
So; Do we even need the column headers to be clickable anymore? I feel without a Name/Role button to press it would be very misleading/frustrate-able for some users.

garrysmod/gamemodes/terrortown/gamemode/lang/german.lua
```diff
@@ -236,6 +236,11 @@ L.help_settings     = "Einstellungen"
 L.help_settings_tip = "Spielerseitige Einstellungen"
 
 -- Settings
+L.set_title_sb = "Anzeigereinstellungen"
```
markusmarkusz: Use “Scoreboard-Einstellungen“.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
```diff
@@ -134,14 +137,18 @@ function PANEL:Init()
...
+   self:AddColumn( GetTranslation("sb_ping"), nil, nil,         "ping" )
+   self:AddColumn( GetTranslation("sb_deaths"), nil, nil,       "deaths" )
+   self:AddColumn( GetTranslation("sb_score"), nil, nil,        "score" )
 
    if KARMA.IsEnabled() then
-      self:AddColumn( GetTranslation("sb_karma") )
+      self:AddColumn( GetTranslation("sb_karma"), nil, nil,     "karma" )
    end
 
+   -- Reusing a few existing translations
```
markusmarkusz: Remove it completly. There is no use for these columns anymore.

Bo98:
Having the other columns (score etc.) clickable would still be useful imo since numeric sorting is more common than just name sorting.

bmwalters:
@markusmarkusz I don't really think there's a need to squash commits during development of a branch here; it would actually be harmful to do so because we lose the history and the previous iterations.
svdm manually merges so he can squash while merging if he wishes.

GamefreakDE:
@bmwalters I think that sorting by name should still be default or at least available because I am the kind of person who wants to find names quick and names are the best for that.

DirkyJerky:
The main reason I started this was because I wanted to be able to sort
scoreboard by name; I surveyed a bunch of people on my regular server and
the major consensus was that sorting alphabetically would be the
main/default option they would use the majority of the time.

And <previous> made a very good point; The other options would be mostly
useful as a secondary alternative just for the statistics, to be used for
only a few moment by some.

What I will try to do next:
- Re-add the Name/Role clickable headers,
- But align them to the left, (next to the portrait) - Add a title label
showing that these buttons are there to change the sorting options

I'm thinking as a final product it would look like (aligned to the left):
'Sort by: | Name | Role'

DirkyJerky:
Looks like: https://i.imgur.com/8eOReuz.jpg

Name is currently highlighted with COLOR_LGRAY; I couldn't get the changes to the highlight color to show up for some damn reason. I know I changed the file...

markusmarkusz:
Am I the only one thinks that this becomes even more complicated every time? xD

Bo98:
The massive gap 'role' and 'karma' will probably confuse some into thinking that only 'name' and 'role' are sortable.

Though having it too close to the rest isn't much better when some are column headers and some not.

Maybe a gap between 'name' and 'role' will help. Or maybe people seeing 'karma' selected (default setting) will give people the picture. The selected header colouring definitely needs to be more obvious though - the contrast is minimal atm.

DirkyJerky:
Many commits does not necessarily mean adding complexity. A lot of them are just tweaking what already exists.

bmwalters:
This is a mess

Kefta:
You made ColorForPlayer a global function

bigdogmat:
I highly suggest testing changes before pushing them.

DirkyJerky:
@bigdogmat I do

Kefta:
@DirkyJerky https://github.com/garrynewman/garrysmod/pull/1332/commits/6ca09876d8d290e711e6dcc1a7c5b3a465621b12 Doesn't seem the case

DirkyJerky:
@Kefta True I don't test thoroughly before I commit; But I do before I push

DirkyJerky:
I'm not necessarily expecting every one commit will make it to the final product (if it does get merged). A lot of these added technically unnecessary components can be easily removed.

Kefta:
Pushing == commits in this context.

DirkyJerky:
I squashed the optional feature into one commit

DirkyJerky:
I extended (fully backward compatible; I tested) the options available to use in the "TTTScoreboardColumns" hook so that in addition to plugins being able to provide custom data columns, they can also provide custom sorting functions so that (when one is provided in the API) end-users can use the clickable column header to sort by the sort function provided by the plugin.

Would ideally be documented somewhere for addon developers, I just don't know where to put it. 'http://ttt.badking.net/guides/hooks' is already out of date-ish

DirkyJerky:
And tested the color sort too (I am quite happy with it :D)
https://i.imgur.com/kI8rJ0S.jpg https://i.imgur.com/sJ2UZzH.jpg

DirkyJerky:
And tested the color sort too (I am quite happy with it :D)
https://i.imgur.com/kI8rJ0S.jpg https://i.imgur.com/sJ2UZzH.jpg

garrysmod/gamemodes/terrortown/gamemode/lang/german.lua
markusmarkusz: Please upsate this.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
```diff
@@ -132,35 +197,129 @@ function PANEL:Init()
...
+   end
+
+   table.insert( table_to_add, lbl )
    return lbl
 end
 
+-- Apparently the following doesnt work
+-- PANEL.AddColumn = column_label_work(self.cols)
+function PANEL:AddColumn( label, func, width, sort_id, sort_func ) 
+   column_label_work( self, self.cols, label, func, width, sort_id, sort_func )
```
markusmarkusz: Would break things since it doesn't return the DLabel anymore.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -54,7 +54,7 @@ function PANEL:Init()
    self:SetCursor( "hand" )
 end
 
-function PANEL:AddColumn( label, func, width )
+function PANEL:AddColumn( label, func, width, _, _ )
```
markusmarkusz: Why?

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
Kefta: Also, remove the comment. It doesn't work since you're trying to set PANEL.AddColumn to a return value, not a function.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
DirkyJerky: To correspond with the two new parameters in the AddColumn function of sb_main

Unless the version of lua GMod uses is really screwed up the change won't make a lick of difference.

DirkyJerky:
TTTScoreboardColumns (pnl)
===

/Client/
Called when the scoreboard is initialized. Additional custom columns can be added by calling `pnl:AddColumn`

`pnl:AddColumn(heading, func, [width, [sort_id, sort_func]])`
  Required
    `pnl` is the panel passed to this hook as an argument.
    `heading` is your column heading text.
    `func` is a function you define that will be called to get up-to-date text for the column.
      `function(ply, label)`
        `ply` is the player on that row.
        `label` is the DLabel object (which can be used to set text colour).
        
  Optional
    `width` is the width of the column in pixels; Defaults to 50.

    If both of the following are provided, the top column label will become clickable in order to sort by the contents of the column
    `sort_id` is an identifier that will be used to reference the accompanying sort function
    `sort_func` is the comparative function used to sort the column.
      `function(plya, plyb) `
        `plya` 
        `plyb` are the players to be compared.
        Return 
          >0 ->`plya` should be placed below `plyb` when in ascending order,
          <0 ->`plyb` below `plya`
          0 -> Will sort by name

`pnl:AddFakeColumn(heading, _, [width, [sort_id, sort_func]])`
  Same as `AddColumn` except it will not create an actual data column, just the header
  Use when you want to provide sorting functionality but not have to display a data column

Examples:
Show user IDs on the scoreboard:
  `pnl:AddColumn("ID", function(ply) return ply:UserID() end)`
Same but sortable:
  `pnl:AddColumn("ID", function(ply) return ply:UserID() end, nil, "steam_id", function(plya, plyb) return plya:UserID() - plyb:UserID() end)`
Same but offers just the sorting functionality
  `pnl:AddColumn("ID", nil, nil, "steam_id", function(plya, plyb) return plya:UserID() - plyb:UserID() end)`

Kefta:
Did you use Google translate for the non-English translations? If so, remove them and wait for an actual speaker to PR the new meanings in.

DirkyJerky:
Oh yea I was meaning to look for some pre-existing translations and replace those. Thanks for reminding me

DirkyJerky:
From memory I recall that I made scoreboard settings menu in the F1 Settings thing just as an optional thing; 
I believe we can just do without it entirely (which includes all the translation placeholder additions as well).

Thoughts anyone?

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
```diff
@@ -85,6 +88,68 @@ function ScoreGroup(p)
...
+   --    unless maybe there are two colors that differ by one unit in hue
+   --    and are polar opposites in brightness, which would be extremely rare in reality.
+   local lightness = (cMax + cMin) / 2
+
+   -- Subtract so brighter hue ties come first (less = higher on board)
+   ret = ret - lightness * 0.00001
+
+   return ret
+end
+
```
svdm: Yeah nah, just kill the whole color sort thing.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
```diff
@@ -22,6 +22,9 @@ surface.CreateFont("treb_small", {font = "Trebuchet18",
                                   size = 14,
                                   weight = 700})
 
+CreateClientConVar("ttt_scoreboard_sorting", "name", true, false, "name | role | karma | score | deaths | ping")
+CreateClientConVar("ttt_scoreboard_ascending", "false", true, false, "true | false")
```
svdm: Would be more consistent as 0/1 like every boolean cvar.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua
```diff
@@ -132,35 +197,127 @@ function PANEL:Init()
...
+
+      self_:ApplySchemeSettings()
+   end
+end
+
+-- For headings only the label parameter is relevant, second param is included for
 -- parity with sb_row
-function PANEL:AddColumn( label, func, width )
-   local lbl = vgui.Create( "DLabel", self )
+local function column_label_work(self_, table_to_add, label, _, width, sort_identifier, sort_func )
```
svdm: The `_` is pointless here, it's a local function with no equivalent in sb_row, so there's no parity to keep.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -116,11 +101,37 @@ end
...
+
+      if sort_func != nil then
+         comp = sort_func(plya, plyb)
+      end
+
+      if comp != 0 then
+         return comp > 0
+      end
+
+      return plya:GetName() > plyb:GetName()
```
svdm: This will sort "Z" before "a". You will have to upper or lowercase both strings to get a somewhat sane ordering.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -116,11 +101,37 @@ end
...
+   end)
+
+   if GetConVar("ttt_scoreboard_ascending"):GetBool() then
+      self.rows_sorted = table.Reverse(self.rows_sorted)
```
svdm: This seems wasteful. Why not just invert the value of `comp` in the above function?

DirkyJerky:
Done.
Thanks for the review!

DirkyJerky:
For the new translation, I was going to scour the internet to see if I could possibly find some existing locale files for the single translation, but I quick checked Google Translate (clichè I know) and through their "community translate" program they seem to have a whole bunch of "verified" translations, one of which is the one I need. Example: https://translate.google.com/#en/pt/Sort%20By%3A

Could we use those?

Kefta:
Don't put in empty translations; let it fall back to English

Kefta:
No, I meant delete the translation from the other files. The code will automatically make it fall back to english.

markusmarkusz:
German Translation:
`Sortiere nach:`

Kefta:
Spanish is "Ordene por:"

Kefta:
Can you squash?

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
```diff
@@ -116,11 +101,41 @@ end
...
+      local plya = rowa:GetPlayer()
+      local plyb = rowb:GetPlayer()
+
+      if not IsValid(plya) then return false end
+      if not IsValid(plyb) then return true end
+
+      local comp = 0 -- Lua doesnt have an Ordering enumeration, I think?
+      
+      local sort_mode = GetConVar("ttt_scoreboard_sorting"):GetString()
+      local sort_func = sboard_panel.sort_table[sort_mode]
```
svdm: This use of `sboard_panel` breaks things. `UpdateSortCache` is called while initializing the scoreboard. At that point, `sboard_panel` is nil (because the panel is still being created).

As a result, if 2+ players exist when you first open the scoreboard, then this `sboard_panel.sort_table` access will of course break and throw your face full of errors.

Note that simply not sorting if `sboard_panel == nil` probably will not work, because it will just leave the sort cache in an unsorted state until something triggers an update.

robotboy655:
@Kefta, PRs can now be automatically Squashed when merging

Kefta:
Ah, didn't know

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_team.lua
DirkyJerky: Ugh I am bad I should have investigated this more earlier.

I had uncorrectly assumed that because existing code in `sb_row.lua` and `sb_team.lua` used that variable without any nil checks, that it was something out of the scope of this PR. My bad.

DirkyJerky:
I believe the best but messy option is as follows:

- Replace `sboard_panel` with `self:GetParent():GetParent():GetParent` (nesting dolls are fun)

Goal is the top `TTTScoreboard` panel, which defines its `self.sort_table` in its `INIT()` [[]](https://github.com/DirkyJerky/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua#L141)

Child #1 is `TTTPlayerFrame` [[]](https://github.com/DirkyJerky/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua#L116)

Child #2 is just a `Panel` used as a canvas, returned by `TTTPlayerFrame:-:GetCanvas()` [[]](https://github.com/DirkyJerky/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua#L440)
[[]](https://github.com/DirkyJerky/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua#L446)

(Target) Child #3 is one of the `TTTScoreGroup`s, each of which has the previous `Panel` as its parent [[]](https://github.com/DirkyJerky/garrysmod/blob/master/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_main.lua#L120)

svdm:
Honestly it's cleaner to just move the `sort_table` stuff into a global `sboard_sort`. It's basically constant anyway. Then you can use that `sboard_sort` wherever you need it in the scoreboard panels without the nesting mess.

DirkyJerky:
Will test at home this afternoon.

MinIsMin:
The "Sort by" string is clipping into "Name" in some languages (tested with German and Spanish) and doesn't look well.
German Scoreboard Image

Better use a variable width and not 50 (maybe :SizeToContents(), GetPos + space)

PR #1333, opened 2017-02-19
 Add serverside functions for printing colored text to chat

bmwalters:
Fixes https://github.com/Facepunch/garrysmod-requests/issues/122

Note that this function overwrites the existing `ply:ChatPrint` function with this new behavior. Since the previous `ply:ChatPrint` function takes a single string for its parameter, the new function is backwards-compatible.

For the implementation, I offloaded as much of the work as possible to `chat.AddText` so any future enhancements to that function should work for people using these functions.

### DIFF
--- a/garrysmod/lua/includes/init.lua
+++ b/garrysmod/lua/includes/init.lua
@@ -123,3 +123,4 @@ if ( CLIENT ) then
 
 end
 
+include ( "util/chat_print.lua" )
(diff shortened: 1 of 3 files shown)


aStonedPenguin:
Do this yourself in your own code. Imo there should be nothing more then an api to construct your own utils for the exact reason you made this commit, the ideal functionality of them is subjective.

Kefta:
If this is added as a default feature, it should be done in the engine.

GamefreakDE:
In my Opinion I really would love such a feature, but probably not how it is done right now. I use Color Messages everytime I make an addon so this would make working so much easier, I like the idea, but not in this state probably.

Kefta:
People should not be sending constant strings/color objects over the network. Have the data available on the client and send a net message from the server when you need it printed.

thegrb93:
So basically still "Implement your own"

PR #1334, opened 2017-02-26
Updated weapon_base

markusmarkusz:
This pull request is not really important.
I have only removed a semicolon, removed two useless functions (SWEP:OnRemove() is already defined in shared.lua) and moved a few clientside only variables to cl_init.lua.

You can accept the pull request, make suggestions, or reject it. In this case I really don't care about it.


robotboy655:
While those variables are indeed only used clientside by default, I don't think we should move them to cl_init.lua. Some addons may find those useful on the server, and in this case those variables will be filled to their default values if a custom SWEP doesn't set them. Most weapons are defined shared only anyway.

PR #1335, opened 2017-02-26
Improve server browser

stickz:
- Weight all servers bellow 30ms the same because there is very little difference in ping.

- Stop servers with only a few people from filling up the ranking.

- Stop the list from filling up with all high slot servers because having more players weights them higher.

- Sink servers that are almost full to help seed other servers. And to ensure successful connectivity.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -235,7 +235,7 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	data.recommended = data.ping;
 	
 	if ( data.recommended <= 30 ) data.recommended = 30 // Weight all servers bellow 30ms the same	
-	if ( data.players < 5) data.recommended += ( 5 - data.players ) * 5 // Decrease the number of empty-like servers
+	if ( data.players < 7) data.recommended += ( 7 - data.players ) * 5 // Decrease the number of empty-like servers
 	
 	if ( data.players == 0 ) data.recommended += 100; // Server is empty
 	
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -245,6 +245,10 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	if ( data.players >= 16 ) data.recommended -= 40;
 	if ( data.players >= 32 ) data.recommended -= 20;
 	if ( data.players >= 64 ) data.recommended -= 10;
+	
+	// Increase the impact of the server's ping on high slot servers, so they don't dominate the list
+	if ( data.maxplayers >= 50 ) data.recommended += 10
+	if ( data.maxplayers >= 100 ) data.recommended += 15
 
 	data.listen = data.desc.indexOf('[L]') >= 0;
 	if ( data.listen ) data.desc = data.desc.substr( 4 );
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -233,6 +233,11 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	data.hasmap = DoWeHaveMap( data.map );
 
 	data.recommended = data.ping;
+	
+	if ( data.recommended <= 30 ) data.recommended = 30 // Weight all servers bellow 30ms the same	
+	if ( data.players < 5) data.recommended += 20 // Decrease the number of empty-like servers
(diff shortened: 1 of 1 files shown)


stickz:
@robotboy655 Do you have any thoughts on these improvements to the server browser?

garrysmod/html/js/menu/control.Servers.js
```diff
@@ -233,14 +233,26 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
...
 	data.recommended = data.ping;
+	
+	if ( data.recommended <= 30 ) data.recommended = 30 // Weight all servers bellow 30ms the same	
+	if ( data.players < 5) data.recommended += 20 // Decrease the number of empty-like servers
+	
 	if ( data.players == 0 ) data.recommended += 100; // Server is empty
-	if ( data.players == data.maxplayers ) data.recommended += 75; // Server is full
+	
+	if ( data.players == data.maxplayers ) data.recommended += 37; // Server is full
+	if ( data.players == data.maxplayers * 0.95 ) data.recommended += 37; // Server is almost full
```
Bo98: Should this be `>=` rather than `==`?

aStonedPenguin:
Sink the ranking of those dirty 1%er popular servers!!! Redistribute the players back to the 99%!


...This is dumb.

garrysmod/html/js/menu/control.Servers.js
stickz: Thanks, i'll fix that right away!

stickz:
@aStonedPenguin Greatly boosting them to the point where they overpower other servers is "dumb" to. If you look at the logic in this file, they're still boosted by double what I propose offsetting.

It seems like this algorithm was intended to make servers that have people on them appear higher on the list. Not to boost up servers with higher slot counts, to the point where they dominate Garry's Mod.

KingofBeast:
#UpWithBubbles #DownWithAir

Kefta:
#DownWithMonopolies

meharryp:
The server browser is shit but I don't think making the most popular servers smaller, medium population servers is the right way to do it.

stickz:
@meharryp Medium population servers are most impacted in a positive way by these changes. More of them appear on top and it's not all large servers that are appearing in the first few choices anymore.

MrUnknownGamer:
Even aStonedPenguin says this is stupid. Why would you do this.

uRandomAlex:
Faking ping or any other information in serverbrowser is stupid. You should find another way to change server's position.

stickz:
I think people need to see this visually. This is what the server browser looks like with all these improvements. A health balance of different server sizes. And only the most relevant choices on top.

Kefta:
dontbelievehislies

NotSoSuper:
@Kefta 
image

aStonedPenguin:
[IMAGE ALT TEXT HERE](https://youtu.be/R_FQU4KzN7A)

stickz:
If anyone wants to try these improvements open [this link](https://raw.githubusercontent.com/stickz/garrysmod/patch-2/garrysmod/html/js/menu/control.Servers.js).

Right click, Save As and Save Here. Then simply launch Garry's Mod.
`C:\Program Files (x86)\Steam\steamapps\common\GarrysMod\garrysmod\html\js\menu`

Servius:
Looks fine. Does this impact filters negatively in any way? As in, if I were to tap the "players" button would it still organize the servers by player count?

SpencerSharkey:
Do you know how old this is?

Why cares what order the servers are listed in. When I look at a server list I typically expect to see it sorted by player-count and region. 

I don't need the game to try and decide what server is best based on some non-quantifiable metrics. Just list servers, and let me decide where to go. You'll find that ordering the server list isn't going to make your somewhat-popular-but-not-the-most-popular server any better.

stickz:
@robotboy655 This should be closed.  #1358 is designed to replace this pull request. An exponential equation is a more narrowly focused and accurcate approach with virtually no negatives or unpopularlity.

@Servius @SpencerSharkey See pull request #1358. It improves the quantifiable metrics of ping and player count when sorting by ranking. Sorting by player count or ping is still optional and not impacted.

The reason why ranking exists is to balance player count and ping. It's naive to say you don't want mathematics to help put the best server options on the top, using these two factors.  The problem with the current system is it's modeled with linear and if statements heavily damage ranking accuracy.

Servius:
@SpencerSharkey This wouldn't effect filters meaning it could still display the best ping and highest player count. It just removes a lot of the unwanted servers from the list like p2p serves and small 1 pop serves from the list.

aStonedPenguin:
I maintain my stance from your last PR.

[IMAGE ALT TEXT HERE](https://youtu.be/R_FQU4KzN7A)

PR #1337, opened 2017-03-06
Rework part of server list

meepdarknessmeep:
- Servers with ping <= 75 don't get penalized
- [LANG] tags in host names are now parsed into a flag icon and get more rank
- Server-is-full check is now properly performed with bots
- [L]isten server tags are now properly removed
- Gamemodes generate only one random, stopping new servers in gamemodes with 0 people from spazzing around

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -447,6 +447,9 @@ end
 function SWEP:SetIronsights(b)
    self:SetIronsightsPredicted(b)
    self:SetIronsightsTime(CurTime())
+   if CLIENT then
+      self:CalcViewModel()
+   end
 end
 function SWEP:GetIronsights()
    return self:GetIronsightsPredicted()
@@ -462,8 +465,6 @@ function SWEP:SetIronsightsPredicted() end
 -- Set up ironsights dt bool. Weapons using their own DT vars will have to make
 -- sure they call this.
 function SWEP:SetupDataTables()
-   -- Put it in the last slot, least likely to interfere with derived weapon's
-   -- own stuff.
    self:NetworkVar("Bool", 3, "IronsightsPredicted")
    self:NetworkVar("Float", 3, "IronsightsTime")
 end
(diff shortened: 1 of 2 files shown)


uRandomAlex:
gr8 for pull #1337

mcNuggets1:
I don't know what the flag change looks like?

HandsomeMatt:
@mcNuggets1 it puts a flag before the server name if the server name contains [language] in the server name.

e.g: My TTT Server [es] would display `resource/localization/es-ES.png` before it.

If the language matches your local language, the server is ranked up more.

meepdarknessmeep:
it actually only looks at the first language.length + 3 letters to check if it is "[es] " - but other than that matt is right

robotboy655:
It looks like this:
!(http://i.imgur.com/nvlZTqR.png)[]

But it misses a lot of the tags

meepdarknessmeep:
It uses the languages available to gmod only - if you add more language flags in the resource folder it will automatically add more tags.
otherwise servers with [GER] can just switch directly to [DE], or other things such as that.

Another problem I had with it (that we could fix with a resource file change) is that the spanish flag is from spain, not spain/mexico. We could probably have some of those flags blended together or something.

Kefta:
Problem with that there are a lot of South American countries that also speak Spanish -- we ultimately can't fit everyone. We could separate Spanish and Castilian and have a Mexican and Spanish flag, respectively.

meepdarknessmeep:
The problem is not having one for each country, it's for having the people know what language it is. I would suggest the English flag be changed to American/UK and the spanish one Mexico/Spain

mcNuggets1:
Well, I don't like the flags.

meepdarknessmeep:
Thanks, can you now give criticism about what part you don't like?

mcNuggets1:
I just don't like the look and I am already used to searching for tags.

meepdarknessmeep:
Did you even try the changes? You can still search for language tags...

What about the look makes you upset?

mcNuggets1:
I just don't like how it looks like.

Making a new column for flags which could be sorted would be really cool, tho.

Kronos9247:
It would be very usefull to have

meepdarknessmeep:
um, whoops
data is gone

meepdarknessmeep:
not like it was going to be pushed anyways

Kefta:
Darn

PR #1338, opened 2017-03-07
TTT: Fixed some cases where Player:Alive() isn't equal to Player:IsTraitor()

markusmarkusz:
- Some small micro optimizations.
- Removed useless GAMEMODE:PlayerSpawnAsSpectator(ply)
- Use Player:KillSilent() to prevent Player:Alive() from returning a different result than that of Player:IsTraitor().
- Moved Player:GetForceSpec() shared by networking force_spec.
- Fixed a problem at ScoreGroup() that is caused because Player:Alive() now has the same result like Player:IsTraitor().

I hope you like my great English. And yes, the text may be full of mistakes.

What about breaking other Addons?
It could be that an addon is broken by these changes. But there are also a few addons which would work because these changes.
I doubt that Addons will override Get/SetForceSpec. I don't know if there are people that are stupid enough to use Player.force_spec instead of Player:Get/SetForceSpec.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -521,8 +521,9 @@ function IncRoundEnd(incr)
 end
 
 function TellTraitorsAboutTraitors()
+   local plys = player.GetAll()
    local traitornicks = {}
-   for k,v in pairs(player.GetAll()) do
+   for k,v in ipairs(plys) do
       if v:IsTraitor() then
          table.insert(traitornicks, v:Nick())
       end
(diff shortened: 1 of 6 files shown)


svdm:
I'm not convinced this is a good idea. The fact that spectators who are not dead are different from dead terrorists is used in a couple of checks in TTT itself, like:

https://github.com/garrynewman/garrysmod/blob/784cd57576d85712fa13a7cea3a9523b4df966b0/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua#L82
https://github.com/garrynewman/garrysmod/blob/784cd57576d85712fa13a7cea3a9523b4df966b0/garrysmod/gamemodes/terrortown/gamemode/player_ext.lua#L186

The bugs caused by this PR would be pretty subtle. I'm sure it's possible to implement those differently, but SWEPS/addons might be using the same distinction and would get similar subtle bugs.

markusmarkusz:
You're right. These two functions could be fixed if necessary, this would be fixed with a small "workaround".
It's probably better if we don't make these changes but look for possible problems in TTT itself that are related to this. If there are such, they could be fixed if necessary.
Perhaps I make an new PR for the other changes (networking force_spec and the small optimizations). But only if that is worth and if I find something useful or any bug.

PR #1341, opened 2017-03-14
Add TTTSelectRoles Hook

GamefreakDE:
Useful for addons that want to do custom calculations or stuff before the roles are selected.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -864,6 +864,8 @@ function SelectRoles()
    local det_count = GetDetectiveCount(choice_count)
 
    if choice_count == 0 then return end
+   
+   hook.Call("TTTSelectRoles", GAMEMODE, choices, prev_roles)
 
    -- first select traitors
    local ts = 0
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -865,7 +865,10 @@ function SelectRoles()
 
    if choice_count == 0 then return end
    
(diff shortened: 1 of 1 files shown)


svdm:
I don't want addons to influence traitor selection, so you should either remove the `choices` and `prev_roles` tables, or make copies and pass those to the hook.

GamefreakDE:
Is there a bigger reason for this or do you just dislike addons modifying base game mode parts? I wanted to add a new Role and didn't want to override functions too much. I am still happy with the copies, just wondering why you don't like it I'm the first place.

markusmarkusz:
He doesn't
> like that addons get the raw tables

because he doesn't
> want addons to influence traitor selection

GamefreakDE:
I know that its not good to ask you, svdm, here,and maybe you wont even read this, but I realized that my choice to add the hook at this place was a wrong decision, could I make another Pull request to move to hook under line 933? And would I be allowed to take the normal tables then because it doesnt influence traitor selecting anywhere? I made a dumb decision and I dont think that this hook as any use there.

svdm:
What is the goal of the hook exactly? If it was called at the end of SelectRoles, you could pretty much just use TTTBeginRound instead.

GamefreakDE:
I wanted to add new roles to TTT, but now I realize how dumb that was because it's such a hard thing to do and it would require to change nearly the whole gamemode. Tbh I don't think this hook has any use so it can be deleted imo. Sorry, but I didn't though about my idea and just created this pull request without thinking about the result.

xxx0365:
About not wanting addons to influence traitor selection, they already can, so why not have a cleaner way of doing it?

monster010:
@xxx0365 Because a cleaner was is better than a way that is dirty. Good Example: The netcode from source is not so clean, that you notice that when you shoot, you don't always is the player, even if blood splashes.

PR #1342, opened 2017-03-17
Fixed RoundedBox(Ex) using incorrect corner textures.

ARitz-Cracker:
People who want rounded boxes shouldn't get a pixel-y mess when the border side is between 16 and 32 ☹️


neico:
From the code it sounds more like it needs a 32x32 texture variant and an extra check instead of changing a hard-coded number.

While we're at it maybe limit the border size to something like 64 so that there won't be any more textures needed in the future (or provide them up front), and warn about it.

PS: I always found using textures as a "overlay" to realize rounded borders a bit dodgy but whatever...

Kefta:
Yeah, a 32 version should just be made.

ARitz-Cracker:
@neico drawing textures takes less time than drawing polys.
@Kefta You're probably right. This is just making sure that what's available preforms as expected.

robotboy655:
added 32 and 512 versions

PR #1344, opened 2017-03-27
Made hook.Add/Remove not silently fail on arg type mismatch

Kefta:
In the rare case a typo was made for the hook args, it is now noticeable.

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -1,8 +1,10 @@
-local gmod                        = gmod
-local pairs                        = pairs
-local isfunction        = isfunction
-local isstring                = isstring
-local IsValid                = IsValid
+local gmod = gmod
+local pairs = pairs
+local isfunction = isfunction
+local isstring = isstring
+local IsValid = IsValid
+local type = type
+local error = error
 
 module( "hook" )
 
@@ -12,8 +14,9 @@ local Hooks = {}
     Name: GetTable
     Desc: Returns a table of all hooks.
 -----------------------------------------------------------]]
-function GetTable() return Hooks end
-
+function GetTable()
+	return Hooks
+end
 
 --[[---------------------------------------------------------
     Name: Add
@@ -22,11 +25,11 @@ function GetTable() return Hooks end
 -----------------------------------------------------------]]
 function Add( event_name, name, func )
 
-	if ( !isfunction( func ) ) then return end
-	if ( !isstring( event_name ) ) then return end
+	if ( !isstring( event_name ) ) then error( "bad argument #1 to 'Add' (string expected, got " .. type( event_name ) .. ")" ) end
+	if ( !isfunction( func ) ) then error( "bad argument #3 to 'Add' (function expected, got " .. type( func ) .. ")" ) end
 
-	if (Hooks[ event_name ] == nil) then
-			Hooks[ event_name ] = {}
+	if ( Hooks[ event_name ] == nil ) then
+		Hooks[ event_name ] = {}
 	end
 
 	Hooks[ event_name ][ name ] = func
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/modules/hook.lua
```diff
@@ -41,7 +44,7 @@ end
 -----------------------------------------------------------]]
 function Remove( event_name, name )
 
-	if ( !isstring( event_name ) ) then return end
+	if ( !isstring( event_name ) ) then error( "bad argument #1 to 'Add' (string expected, got " .. type( event_name ) .. ")" ) end
```
neico: 'Add' -> 'Remove'

neico:
What did those first lines do to you to deserve this much bashing? (alas, why the space changes?)

Kefta:
Each of them was a space; I don't understand why considering they weren't aligned, and everything else in GMod uses tab alignment, thus I just removed them since I added new locals anyway.

garrysmod/lua/includes/modules/hook.lua
Kefta: Thanks!

Kefta:
Should the second argument be type-checked? The wiki says the identifier can be a string, table, or entity/panel, and that numbers/bools aren't allowed, but everything is allowed right now.

robotboy655:
Please add newlines, and perhaps it still should be error() because this doesn't look terribly helpful:
image

Or at least put the `return`s back in so it functions like it used to.

thegrb93:
Why doesn't ErrorNoHalt provide the stacktrace?

Kefta:
I'm thinking of PRing an ErrorNoHaltStack function.

robotboy655:
I am afraid adding stack trace to `ErrorNoHalt` will cause problems for existing code, especially since `ErrorNoHalt` doesn't add its own new line.

Kefta:
It would be a separate function, literally called ``ErrorNoHaltStack``. Here's my first stab at it:
```
function ErrorNoHaltStack(sMessage, uLevel --[[= 1]])
	if (not isstring(sMessage)) then
		error(string.format("bad argument #1 to 'ErrorNoHaltStack' (string expected, got %s)", type(sMessage)), 2)
	end
	
	if (uLevel == nil) then
		uLevel = 2
	elseif (isnumber(uLevel)) then
		uLevel = uLevel + 1
	else
		error(string.format("bad argument #2 to 'ErrorNoHaltStack' (number expected, got %s)", type(uLevel)), 2)
	end
	
	local tInfo = debug.getinfo(uLevel, "nS")
	
	if (tInfo == nil) then
		ErrorNoHalt(string.format("[ERROR] %s\n", sMessage))
		
		return
	end
	
	local sSource = tInfo.source
	
	if (sSource == "[C]") then
		sMessage = string.format("[ERROR] %s\n", sMessage)
	else
		sMessage = string.format("[ERROR] %s:%i: %s\n", sSource, tInfo.lastlinedefined, sMessage)
	end
	
	local iStackLevel = 1
	
	::StackLoop::
	
	sMessage = string.format("%s%u. %s - %s:%i\n", sMessage, string.rep(" ", iStackLevel + 1), iStackLevel, tInfo.name, tInfo.short_src, tInfo.lastlinedefined)
	uLevel = uLevel + 1
	tInfo = debug.getinfo(uLevel, "nS")
	
	if (tInfo ~= nil) then
		iStackLevel = iStackLevel + 1
		
		goto StackLoop
	end
	
	ErrorNoHalt(sMessage)
end
```

Might be better to add in C but this has the benefit of actually altering the stack trace along with the error level argument unlike ``error``.

thegrb93:
you can `repeat until` instead of goto might be preferred but lgtm

garrysmod/lua/includes/modules/hook.lua
```diff
@@ -22,11 +25,11 @@ function GetTable() return Hooks end
 -----------------------------------------------------------]]
 function Add( event_name, name, func )
 
-	if ( !isfunction( func ) ) then return end
-	if ( !isstring( event_name ) ) then return end
+	if ( !isstring( event_name ) ) then ErrorNoHalt( "bad argument #1 to 'Add' (string expected, got " .. type( event_name ) .. ")" ) end
+	if ( !isfunction( func ) ) then ErrorNoHalt( "bad argument #3 to 'Add' (function expected, got " .. type( func ) .. ")" ) end
```
Bo98: If you are using `ErrorNoHalt` then you probably want the `return` still there to prevent further errors.

Kefta:
This PR now depends on https://github.com/Facepunch/garrysmod/pull/1587.

PR #1347, opened 2017-04-03
Change gmod_tool's holdtype to revolver

Kefta:
The viewmodel has two hands!


Kefta:
Just thought I'd throw in that the medkit does, as well, but there is no two-handed substitute for a semi-passive holdtype.

PR #1349, opened 2017-04-08
TTT: Corpses get skin&bodygroups

aceradryd:
TTT: Corpses get the skin and bodygroups from the dead Player.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/corpse.lua
@@ -387,8 +387,8 @@ function CORPSE.Create(ply, attacker, dmginfo)
    rag:SetPos(ply:GetPos())
    rag:SetModel(ply:GetModel())
    rag:SetSkin(ply:GetSkin())
-   for id in pairs(ply:GetBodyGroups()) do
-      rag:SetBodygroup(id,ply:GetBodygroup(id))	
+   for key,value in pairs(ply:GetBodyGroups()) do
+      rag:SetBodygroup(value["id"],ply:GetBodygroup(value["id"]))	
    end
    rag:SetAngles(ply:GetAngles())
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/terrortown/gamemode/corpse.lua
```diff
@@ -386,6 +386,11 @@ function CORPSE.Create(ply, attacker, dmginfo)
 
    rag:SetPos(ply:GetPos())
    rag:SetModel(ply:GetModel())
+   rag:SetSkin(ply:GetSkin())
+   bodyGroupTable = ply:GetBodyGroups()
+   for id in pairs(bodyGroupTable) do
```
svdm: You're creating a `bodyGroupTable` global var here. Just make it `for id in pairs(ply:GetBodyGroups()) do`

aceradryd:
I have changed the code.

Yes, you can set your skin/bodygroup for example with the "Enhanced PlayerModel Selector" and then the corpse should have the skin/bodygroup also for example for the "Advanced Disguiser".

Kefta:
Shouldn't you be using the "id" key in the value table instead of using the key? Don't know if bodygroups work sequentially or not.

aceradryd:
I dont know, what do you mean, Kefta?

Kefta:
GetBodyGroups returns a table with the bodygroup id key in it. Currently, you are just using the table index. This should work so long as the bodygroups are sequential  (1, 2, 3 ,..., k) but I'm not sure if that's guaranteed or not.

aceradryd:
Okay, i have changed the code for this case. Now it use the key from the table.

BlackCetha:
You should use the dot-notation for accessing string-indexed tables (`table.index`) and add a space after the comma to keep the style consistent with the codebase.

aceradryd:
Changed

PR #1350, opened 2017-04-10
Added draw.DrawTextWordWrap()

EstevanTH:
Hi!

`draw.DrawTextWordWrap()` allows to draw a text with word wrap (if too many words on a line, draw on next line). Spaces and tabs are displayed as spaces. New lines properly displayed. Alignment is fine too.

I know that you do not like my coding style and the performance of my code. I know that whatever I do, it already exists anyway. Please stay sweet, I did my best to add a nice feature.

Momo :heart:

### DIFF
--- a/garrysmod/lua/includes/modules/draw.lua
+++ b/garrysmod/lua/includes/modules/draw.lua
@@ -228,8 +228,18 @@ do
 							insert( textLines, line )
 						else
 							local spaceExploded = {}
+							local wasEmpty = true
 							for word in gmatch( line, "[^%s]*" ) do
-								insert( spaceExploded, word )
+								if len( word )==0 then
+									if wasEmpty then
+										insert( spaceExploded, word )
+									else
+										wasEmpty = true -- avoid output of an extra space after every word
+									end
+								else
+									wasEmpty = false
+									insert( spaceExploded, word )
+								end
 							end
 							local startConcat = 1
 							while startConcat<=#spaceExploded do -- while there are words to process
--- a/garrysmod/lua/includes/modules/draw.lua
+++ b/garrysmod/lua/includes/modules/draw.lua
@@ -261,6 +261,7 @@ do
 		local lineW,lineH
 		local lineX,lineY
 		lineY = y
+		local fullHeight = 0
 		SetTextColor( color.r,color.g,color.b,color.a )
 		for _,line in ipairs( textLines ) do
 			lineW,lineH = GetTextSize( line )
@@ -268,7 +269,9 @@ do
 			SetTextPos( lineX,lineY )
 			DrawText( line )
 			lineY = lineY+lineH
+			fullHeight = fullHeight+lineH
 		end
+		return fullHeight
 	end
 end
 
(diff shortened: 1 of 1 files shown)


EstevanTH:
Here is my test code:

```lua
local text = "\n"..[[
Library versions should be specified for shared objects if the function interfaces are expected to change (C++ public/protected class definitions), more or fewer functions are included in the library, the function prototype changes (return data type (int, const int, ...) or argument list changes) or data type changes (object definitions: class data members, inheritance, virtual functions, ...).

One can look at the symbol version used:	nm csub1.o


Symbol table '.dynsym' contains 144 entries:
	Num:    Value          Size Type    Bind   Vis      Ndx Name
		0: 0000000000000000     0 NOTYPE  LOCAL  DEFAULT  UND 
		1: 0000000000003b30     0 SECTION LOCAL  DEFAULT   10 
		2: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND getenv@GLIBC_2.2.5 (4)
		3: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND free@GLIBC_2.2.5 (4)
		4: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND ferror@GLIBC_2.2.5 (4)
		5: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fread@GLIBC_2.2.5 (4)
		6: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND fclose@GLIBC_2.2.5 (4)
		7: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND __stack_chk_fail@GLIBC_2.4 (5)
		8: 0000000000000000     0 FUNC    GLOBAL DEFAULT  UND memset@GLIBC_2.2.5 (4)
		9: 0000000000000000     0 NOTYPE  WEAK   DEFAULT  UND __gmon_start__
...
...
]]
hook.Add( "HUDPaint", "draw_DrawTextWordWrap_test", function()
	local x,y = 64,128
	local w = ScrW()-128
	draw.RoundedBoxEx( 0, x,y, w, ScrH(), Color( 0,0,0,192 ) )
	draw.DrawTextWordWrap( text, "DebugFixed", x,y, w, Color( 255,0,0 ), TEXT_ALIGN_CENTER )
end )
```

EstevanTH:
I think that there is something missing: returning the height of the rendered text.

I have got to add that! :smile:

EstevanTH:
Oops, I have the same problem that I had on new lines: `string.gmatch()` adds an additional space after the first space following a word...
That's another thing to fix!

Damianu:
Soo its like markup.parse() and markup:Draw() ?

EstevanTH:
Just like [`draw.DrawText()`](http://wiki.garrysmod.com/page/draw/DrawText), but with word-wrap and no tab support (it's too complex to put both into the same function).

robotboy655:
This is basically MarkupObject and I'd rather avoid having multiple functions/features that do the same/similar thing.

PR #1352, opened 2017-04-19
TTT: Remove version

markusmarkusz:
There is no reason that TTT has his own version system.
Nobody really cares about it.
Nobody updates the version.
Nobody reads it.
Trust me.

Either we remove the version completely, or we use the version of GMod.
(If you want the first option: Should I remove [these](https://github.com/markusmarkusz/garrysmod/blob/patch-2/garrysmod/gamemodes/terrortown/gamemode/lang/english.lua#L1042) versions too?)

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -161,7 +161,6 @@ util.AddNetworkString("TTT_Spectate")
 ---- Round mechanics
 function GM:Initialize()
    MsgN("Trouble In Terrorist Town gamemode initializing...")
-   ShowVersion()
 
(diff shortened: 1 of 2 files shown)


svdm:
nah

markusmarkusz:
Why should TTT have its own version numbering?
[I totally agree with](https://github.com/garrynewman/garrysmod/pull/1159#issuecomment-212012440) @Fuzzik:
> On a serious note, I think at this point it would be better to version TTT based off of Gmod's version, since that's when clients are receiving the patches anyway. This system never really made sense to me since TTT was included in Gmod. Just a thought.

The version is only updated every few years, since it is not relevant.

svdm:
I have started updating it again.

PR #1354, opened 2017-04-23
Small Resource Location Fixes

Jake3231:
A few tiny fixes that resolves some issues in the template html files, allowing some resources to be properly referenced.

### DIFF
--- a/garrysmod/html/template/addon_list.html
+++ b/garrysmod/html/template/addon_list.html
@@ -75,8 +75,8 @@
 					<controls>
 
 						<left>
-							<control ng-hide="file.local || file.rated" ng-click="Rate( file, true )"><img src='img/thumb-up.png' /></control>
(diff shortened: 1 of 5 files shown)


willox:
doesn't this just break the menu? the paths are already correct

Jake3231:
`doesn't this just break the menu? the paths are already correct`

In my testing, this doesn't break anything. It just resolves some html errors that the user will never see.

willox:
in my testing it stops the images from loading. i guess it'd fix the images not loading if you try to directly load the templates as html documents, but you're not meant to do that

Jake3231:
Okay.  On macOS this doesn't cause any issues.  It does also resolve any issues that occur when loading these files in a browser.

Jake3231:
It seems as this is no longer necessary.

PR #1355, opened 2017-04-23
Update Fonts In the Menu for macOS

Jake3231:
This ensures that Helvetica Neue is used on some of the menu buttons and input boxes.  On macOS, fonts were some serif font, but this makes sure everything is the proper font.


Bo98:
They used to work fine but broke after some OS update (Yosemite maybe?). It's an Awesomium problem that is probably addressed in CEF, if/when that comes for macOS.

Jake3231:
Okay. You're right.  Would this affect any users of 10.9 or earlier?  If this is included, it'll ensure that the latest versions of macOS are showing the proper behavior.  It'd be nice to see CEF replace Awesomium, since it seems to be causing a lot of issues.

Jake3231:
Just as an update, I have a theory as to why this issue may be happening in the first place.  OS X 10.11 El Capitan changed the system font to Apple's proprietary San Francisco font.  If Gmod attempts to use the system font on 10.11 or later, it may not know how to handle San Francisco.  The reason I assume this is that macOS used to use Helvetica Neue as the system font, so that may've caused things to just work out okay until the switch to San Francisco.

robotboy655:
Correct me if I am wrong, but this font is already set for the `BODY` element in the same file, which would affect all the buttons, etc., making this PR do nothing.

PR #1356, opened 2017-04-30
Added SetInterval and GetInterval to DNumberWang

kpjVideo:
Allows for an easier way to add and subtract numbers higher/lower than 1 in DNumberWang.


bigdogmat:
The point of `AccessorFunc` is that it adds Setter/Getter functions.

PR #1357, opened 2017-04-30
added SetTitleColor and GetTitleColor  to DFrame

bull29:
it's a part of dlabel, so why not?  

it's not quite easy for beginners to figure otherwise

similar to #1336


bull29:
It really doesn't look like this is ever getting merged.

Kefta:
Reopen this

bull29:
💯

robotboy655:
Use Derma skins to style derma components. We do not need every `DLabel` method on the `DFrame`'s label duplicated on `DFrame` itself.

PR #1358, opened 2017-05-01
Use exponential equation to model ping on server browser

stickz:
This pull request improves the server browser, by changing how ping is modeled. It prioritizes lower latency servers in a different way, which puts more good options higher up on the server browser.

**Exponential equation:** `0.05 * ping ^ 1.615`
**Sample values:** `30ms = 12.4, 60ms = 37.2, 90ms = 71.6, 120ms = 114, above 130ms = actual ping`

**How does it work?**
It puts moderate pings at a stronger position on the list if the server has lots of people playing. If the server ping is above 130ms, it reverts to the default, so clients in exotic places aren't negatively impacted.

**But, I don't like playing at higher pings**
You don't have to play with a higher ping. This just gives the user the choice to, if they're looking for more servers; instead of filling the server browser up with choices which only have a few people playing.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -232,7 +232,7 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 
 	data.hasmap = DoWeHaveMap( data.map );
 
-	data.recommended = math.min(0.006 * Math.pow(data.ping, 2.05), data.ping); // Model ping using an exponential equation
+	data.recommended = Math.min(0.006 * Math.pow(data.ping, 2.05), data.ping); // Model ping using an exponential equation
 
 	if ( data.players == 0 ) data.recommended += 100; // Server is empty
 	if ( data.players == data.maxplayers ) data.recommended += 75; // Server is full
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -232,7 +232,7 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 
 	data.hasmap = DoWeHaveMap( data.map );
 
-	data.recommended = Math.min(0.006 * Math.pow(data.ping, 2.05), data.ping); // Model ping using an exponential equation
+	data.recommended = Math.min(0.05 * Math.pow(data.ping, 1.615), data.ping); // Model ping using an exponential equation
 
 	if ( data.players == 0 ) data.recommended += 100; // Server is empty
 	if ( data.players == data.maxplayers ) data.recommended += 75; // Server is full
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -232,7 +232,8 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 
 	data.hasmap = DoWeHaveMap( data.map );
 
-	data.recommended = data.ping;
+	data.recommended = 0.006 * Math.pow(data.ping, 2.05) // Model ping using an exponential equation
+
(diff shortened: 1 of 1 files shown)


stickz:
DarkRP screenshots.

**Before** (8 servers with less than 11 players on top)


**After** (4 servers with less than 11 players on top)

stickz:
Post reserved for screenshots tomorrow.

BlackCetha:
I think this would be a good change for the points mentioned.
One thing though, why did you choose 1.615 as the exponent?

stickz:
I choose 1.615 as the exponent because I wanted to change how pings up to 130ms are modeled. And also because of the 0.05 base, which nicely distinguishes between the lowest pings on the spectrum.

I tried a much lower base (0.003) and higher exponent (2.05), but it didn't procedure as good results.

Servius:
@stickz In the future it may be nice to tag an image in the description for pull requests. 

Couldn't a set of "quick filters" be setup to decide between exponents?

aStonedPenguin:
I maintain my stance from your last PR.

[IMAGE ALT TEXT HERE](https://youtu.be/R_FQU4KzN7A)

Jvs34:
"You can't handle my strongest redirect servers, traveler"

uRandomAlex:
When will you provide us with screenshots?

thegrb93:
So this puts lower latency servers up slightly higher on the list. I don't see any difference, just more math the already slow browser has to process.

Servius:
@thegrb93 It cuts out the one slot solo player servers and it puts more focus on the medium pop to high pop servers.

thegrb93:
What's this have to do with population? All it does is nonlinearize the ping score.

stickz:
@thegrb93 When the ping score is de-linearized it's also lowered, so the player count score has a larger impact on the ranking; thus putting servers with more players closer to the top. However, the ping score is lowered in a way where it still perfers to put the lower ping options on the top, like people want.

Computers can procress the type of math used in this pull request very fast. It will not add more than a few hundrend miliseconds (0.3 seconds in the worst case scenario) to the server browser loading times.

thegrb93:
I know how the math works. What I want to know is why not improve the way it uses player count to score rather than overcomplicating ping score. There's barely any difference with this.

Servius:
Why not just have it so you can enable or disable the use of this? If it's disabled by default I don't think anyone will complain.

Kefta:
Not everything needs a setting, and this is one of those cases

Servius:
Well if it's something that would benefit some people and _negatively_ impact some other people then why not have a little check box to turn this on and off ?

Kefta:
How does it negatively impact people? And there is no reason to have a setting for something as simple and obscure to the average user as this

Servius:
It wouldn't actually _negativley_  impact anyone but people are still sitting there complaining about this. So the compromise would be to make it optional.

Kefta:
No one is complaining about the impact of the change, but rather, the lack there of.

stickz:
This change acts as a catalyst to help make future improvements to the server browser better. There are zero negatives and nobody would want something 100% beneficial in all circumstances to be optional.

As the screenshot in the second posts illustrates, it pushes more options with higher player counts up to the top of the server browser. Even if the difference is small, there's no reason why anyone wouldn't want it.

BlackCetha:
I agree with @stickz.
A few milliseconds of difference in latency shouldn't impact the sorting as much as it does right now when there is a more relevant factor, playercount, to consider aswell.

From the perspective of the player, what's more of a difference: 5ms more latency or 5 less players?
With the current implementation, its entirely possible that a few half-empty servers with good ping (or redirect servers) fill up the first 10 entries or so. This change will reduce the number of times that this is happening.

I think most people only ever join the first few servers on the list, so those should be chosen based on what makes a actual difference for the players.

robotboy655:
As stated before this doesn't bring much meaningful changes and there has been other improvements to the "Rating" of a server since this PR.

PR #1359, opened 2017-05-02
Added a hook to disable the drawing of a targetid

tobiasx:
To allow for more customization.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
@@ -151,6 +151,9 @@ function GM:HUDDrawTargetID()
    local ent = trace.Entity
    if (not IsValid(ent)) or ent.NoTarget then return end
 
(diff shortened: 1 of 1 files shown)


Bo98:
Any reason why you can't use `HUDShouldDraw`, checking "TTTTargetID" is the parameter?

garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
```diff
@@ -151,6 +151,9 @@ function GM:HUDDrawTargetID()
    local ent = trace.Entity
    if (not IsValid(ent)) or ent.NoTarget then return end
 
+   local c = hook.Call("TTTDontDrawTargetID", GAMEMODE, trace)
```
markusmarkusz: Please use a better name. Like TTTDrawTargetID or something like this.

markusmarkusz:
This would be useless since DrawPropSpecLabels is called before the hook.

PR #1265, opened 2016-10-28
Add convar that allows players to disable viewing custom sv_loadingurls

bmwalters:
Fixes Facepunch/garrysmod-requests#653
The convar name and help text are whatever; they can be changed if someone has a better suggestion.

PR #1361, opened 2017-05-10
Let clients decide if they want to see custom loading screens

LuaTenshi:
Let clients decide if they should see custom loading screens for servers or if they should just see the standard loading screen.


bigdogmat:
#1265

LuaTenshi:
I could have sworn that thread was closed a long time ago. Glad to see that it's still going oh well. Let's remind everyone with yet another pull to make the same change I guess...

PR #1362, opened 2017-05-11
Removed a critically important unmeant "local" operator

Tripperful:
Considering on the comment, the **outer** "trace" variable should be overwritten if we perform the new trace, since we're checking it further. **Otherwise this whole block of code is useless.**

### DIFF
--- a/garrysmod/lua/includes/modules/properties.lua
+++ b/garrysmod/lua/includes/modules/properties.lua
@@ -101,9 +101,9 @@ function GetHovered( eyepos, eyevec )
 		filter = filter
 	} )
 
-	// Hit COLLISION_GROUP_DEBRIS and stuff
+	-- Hit COLLISION_GROUP_DEBRIS and stuff
 	if ( !trace.Hit || !IsValid( trace.Entity ) ) then
-		local trace = util.TraceLine( {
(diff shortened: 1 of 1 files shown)


Tripperful:
Another solution would be to just remove [**this whole block of code**](https://github.com/garrynewman/garrysmod/blob/master/garrysmod/lua/includes/modules/properties.lua#L104-L113), it would leave the behavior as it is but optimize the function a bit, avoiding an unnecessary table creation and trace every frame.

Kefta:
I think it should be left in, otherwise, entities that don't fall under MASK_SOLID won't be hit. You could argue to do the MASK_ALL trace from the beginning, but I believe the behaviour Garry was shooting for was to give solid entities priority over non-solids.

Kefta:
Also, you can prevent another table from being created by setting the output key of the second trace to "trace", and maybe save the table being sent into TraceLine to be reused for the next trace. Then, at max, only two tables are used.

robotboy655:
This isn't from Garry, this is from my latest commit in that file when I fixed the trace not hitting some props on HL2:DM maps. I will merge this when our backend changes are finished, right now its a PITA to get anything onto the internal repo.

PR #1365, opened 2017-05-14
Add global function Print

NeatNit:
Print behaves like PrintTable for tables, and like print for all other cases. Very useful for debugging, especially with lua_run and lua_run_cl, where I constantly find myself switching back and forth between print and PrintTable.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -90,6 +90,20 @@ function PrintTable( t, indent, done )
 end
 
 --[[---------------------------------------------------------
+	Either print or PrintTable, depending on the arguments
+-----------------------------------------------------------]]
+function Print(...)
+	if select("#", ...) == 1	-- only 1 element
+		and type(...) == "table"	-- which is a table
+		and (getmetatable(...) == nil or not debug.getmetatable(...).__tostring)	-- that has no custom tostring function
+	then
+		PrintTable(...)
+	else
+		print(...)
+	end
+end
+
+--[[---------------------------------------------------------
 	Returns a random vector
 -----------------------------------------------------------]]
 function VectorRand()


NeatNit:
Relevant facepunch post: https://facepunch.com/showthread.php?t=1440586&p=52226862&viewfull=1#post52226862

Kefta:
Why not loop through each element and print tabular elements as tables and non-tabular elements normally?

NeatNit:
I felt that this is a more useful and intuitive thing to do. I suppose a function that does what you said can be done as well, but I would call it `PrintAll`. `Print` implies that it shares a strong resemblance to `print` - which it does.

Kefta:
I would just put the varargs in a table and loop through them, but that's just me.

NeatNit:
That would easily trip over nils with undefined behavior that would *probably* work fine in all of your tests and would *probably* work fine in all cases thanks to a few optimizations that happen to do the same thing you intended, but it'd still be undefined behavior.

Kefta:
You should print nils with it. If someone does Print ({},nil,5), then nil should be printed

Kefta:
Oh, you mean with table counting. Yeah, you're right.

SwadicalRag:
```lua
function Print(...) return PrintTable {...} end
```

:v

NeatNit:
>```Lua
>function Print(...) return PrintTable {...} end
>```
>:v

I know this is a joke, but for the sake of anyone who doesn't get it: loads of problems with this.

1. it ignores __tostring
2. for non-table inputs, it prints them as "1\t=\t<value>\n2\t=\t<value>" etc, which takes up more space than `print(...)` and is, in my opinion, ugly.
3. It skips nils
4. It doesn't differentiate between nil and no value (applicable only to the last argument)

dvdvideo1234:
Can you also add PrintExtract which van extract your table as a Lua source recursively:

```lua
function PrintExtract(tT,sS)
  local vS, vT, vK, sS = type(sS), type(tT), "", tostring(sS or "Data")
  if(vT ~= "table") then
    print("{"..vT.."}["..tostring(sS or "Data").."] = <"..tostring(tT)..">",nil)
    return nil 
  end
  if(next(tT) == nil) then print(sS.." = {}"); return nil end
  print(sS.." = {}")
  for k,v in pairs(tT) do
    if(type(k) == "string") then vK = sS.."[\""..k.."\"]"
    else vK = sS.."["..tostring(k).."]" end
    if(type(v) ~= "table") then
      if(type(v) == "string") then print(vK.." = \""..v.."\"",nil)
      else print(vK.." = "..tostring(v),nil) end
    else PrintExtract(v,vK) end
  end
end
```

AMD-NICK:
PrintTable({any_values})

PR #1366, opened 2017-05-17
Fix: Detective Role-Pick

Asorie:
Fix to Detective Role-Pick. Sometimes it was possible to get the detective role, even if the players karma was lower than "ttt_detective_karma_min" - value.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -911,23 +911,26 @@ function SelectRoles()
       local pick = math.random(1, #choices)
       local pply = choices[pick]
 
-      -- we are less likely to be a detective unless we were innocent last round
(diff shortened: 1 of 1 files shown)


Kefta:
You are calling GetBaseKarma on pply before checking if they're valid. Also, TTT uses spaces not tabs

Asorie:
@Kefta You're right, i changed it now.

markusmarkusz:
TTT uses three spaces.

Asorie:
@markusmarkusz Is this really so important? What about the code itself?

svdm:
This is working as intended, see [docs on site](http://ttt.badking.net/config-and-commands/convars#TOC-Traitor-and-Detective-counts):

> ttt_detective_karma_min (def. 600):
> If a player's Karma falls below this point, his chances of being selected as detective are reduced. 

So your chances are reduced, but not zero.

PR #1367, opened 2017-05-18
Add constructor arguments to PANEL:Init

sannysc:
Currently there is no way to initialize a panel with some data, even though there are panels which rely on external data as a dependency. This leaves us having to create our own function just to initialize the panel, and we have to call that function after the panel was already supposed to be initialized through ```PANEL:Init```. Isn't that the whole purpose of ```PANEL:Init```? To initialize the panel and prepare it for consumption? Take an inventory system, as an example.

You have an item panel, which shows an item in your inventory. This item panel has a label and a spawn icon. Obviously this item panel relies on an actual item to show anything useful. An item panel without an item doesn't make sense. So you have an item's data that you perhaps got from a database. You pass that data to the panel, which gives the panel the necessary information to display that item.

```lua
function PANEL:Init(itemData)
	self.Label = self:Add("DLabel")
	self.SpawnIcon = self:Add("SpawnIcon")

	self.Label:SetText(itemData.Name)
	self.Label:SetColor(itemData.Color)

	self.SpawnIcon:SetModel(itemData.Model)
end

vgui.Register("MyItemPanel", PANEL, "Panel")
```

Then when you want to create the panel.

```lua
local panel = vgui.Create("MyItemPanel", nil, nil, item)
```

The two functions I've edited already take varargs, they just don't pass them to ```PANEL:Init```.

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel/scriptedpanels.lua
+++ b/garrysmod/lua/includes/extensions/client/panel/scriptedpanels.lua
@@ -23,7 +23,7 @@ function vgui.Create( classname, parent, name, ... )
 
 		local metatable = PanelFactory[ classname ]
 
-		local panel = vgui.Create( metatable.Base, parent, name or classname )
+		local panel = vgui.Create( metatable.Base, parent, name or classname, ... )
 		if ( !panel ) then
 			Error( "Tried to create panel with invalid base '" .. metatable.Base .. "'\n" );
 		end
@@ -34,7 +34,7 @@ function vgui.Create( classname, parent, name, ... )
 
 		-- Call the Init function if we have it
 		if ( panel.Init ) then
-			panel:Init(...)
+			panel:Init( ... )
 		end
 
 		panel:Prepare()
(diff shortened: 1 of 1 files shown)


bull29:
you should incorporate this into Panel.Add as well

willox:
I'm not sure this is ever going to work nicely with the automatic constructor chaining we've got. It doesn't really make sense to pass the params to the Init method of every sub-class, and I don't see a way that we could specify how the pass different params to the sub-classes.

sannysc:
The other option is not calling the base constructor automatically, which, now that I think about it, makes less sense than just doing this.
```lua
function Base:Init(someParameter)
	...
end

function Child:Init(anUnrelatedParameter, andAnotherOne)
	self.BaseClass.Init(self, "someParameter for the parent")
	...
end
```
If we remove calling the base constructor automatically, then there would be no weird samey-parameter-chaining shenanigans. Children can choose how to handle calling the parent, and can take different parameters of their own. Would you be willing to accept this if the second commit is reverted @willox?

sannysc:
Nevermind. That wouldn't work either because then it would be calling the base's Init twice. There is too much that would need to be changed to make this happen.

PR #1368, opened 2017-05-18
Added player.GetByAccountID( number AccountID )

Tripperful:
In some cases this function is more optimal to use since you compare numbers instead of strings when iterating through players.


Kefta:
https://github.com/garrynewman/garrysmod/pull/1309

PR #1371, opened 2017-06-01
Update of sql.TableExists function

SleepyMode:
The current sql.TableExists function will return true if query variable r will return true, and false if it won't.
If not found, the query will either return nil or false from my past experience, which both resort to false in case they are put in an if, so you can simply return r instead of returning either true or false.

### DIFF
--- a/garrysmod/lua/includes/util/sql.lua
+++ b/garrysmod/lua/includes/util/sql.lua
@@ -34,9 +34,7 @@ function sql.TableExists( name )
 
 	local r = sql.Query( "select name FROM sqlite_master WHERE name="..SQLStr( name ).." AND type='table'" );
 	
-	if ( r ) then return true end
-	
-	return false
+	return r
 
 end
 


BlackCetha:
This is bad because it changes output values of a function that is already in use - this shouldn't be done anyways.
Additionally in this case, it doesn't have any advantages over the old code and may break code that relies on this function outputting boolean values instead of `true/false/nil`.
Suppose the following code:
```lua
if sql.TableExists( "test" ) == false then
  sql.Query( "CREATE TABLE test (d INTEGER)" )
end
```

This code works fine without your change but breaks with it.
You may say "But nobody does that!", and you may be right.
But how can you/we be sure?

bigdogmat:
`sql.Query` will only return false if it errors. If the query returned no data, then it returns nil. So there's no problem with the way it is currently.

stickz:
The purpose of this function is to return a boolean value (true or false) if a table in a database exists. This commit breaks that by returning the table name or nil instead. Lua works definitely from other languages. The following code checks if the query result is not nil and returns true if so.
`If (r) then return true end`

If you really want to simplify this code, a ternary operator could be used.
```
function sql.TableExists( name )
	local r = sql.Query( "..." );
	return r and true or false
end
```

PR #1373, opened 2017-06-04
Simplify SQL wrappers with ternary operators

stickz:
This pull request makes the SQL wrappers faster. Commenting is added so readability isn't sacrificed.


Kefta:
It actually doesn't make it faster. Ternary operators don't exist in Lua -- instead, it evaluates the value after the and so it actually adds an extra check at the gain of shorter code.

AMD-NICK:
Beauty code is a good code!

PR #1374, opened 2017-06-07
Fix anticheat thing

Velkon:
Idk why it exists but now it's super unbypassable ****

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
@@ -117,10 +117,10 @@ if CLIENT then
 
       return false
    end
-   local ply = LocalPlayer()
+   local ply = LocalPlayer
    local gmod_GetWeapons = plymeta.GetWeapons
    function plymeta:GetWeapons()
-      if self != ply then
+      if self != ply() then
          return {}
       else
          return gmod_GetWeapons(self)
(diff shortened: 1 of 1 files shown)


Tenrys:
Isn't LocalPlayer() NULL until some time after init?

Kefta:
Until InitPostEntity, yeah.

Tenrys:
So this won't work, correct?

bull29:
it won't.

Velkon:
I wrote this on my phone but I'll fix it when I get home unless someone wants 2 do it

thegrb93:
Just add `local LocalPlayer = LocalPlayer` next to `local gmod_GetWeapons = plymeta.GetWeapons`. I still don't see the point though.

Velkon:
Web editor is shit on phone

Kefta:
This should probably be fixed internally, too.

Velkon:
There are legitimate uses for getweapons, unless you mean LocalPlayer being nil

Kefta:
What's the use of GetWeapons working for other players clientside? And LocalPlayer is never nil, it's null which means it can still be evaluated as invalid.

Velkon:
You can still get a weapon list without getweapons but I guess it would only save on networking if you ever need to use it client-side

Kefta:
I'm surprised they're networked at all to the local client. I would think that since the weapon is dormant when its holstered, the server wouldn't network anything about it to non-owning clients.

PR #1375, opened 2017-06-19
Added achievement progress on right-click remove

CornerPin:


garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/remover.lua
```diff
@@ -79,6 +79,8 @@ function TOOL:RightClick( trace )
 
 	end
 
+	self:GetOwner():SendLua( string.format( "for i=1,%i do achievements.Remover()end", count ) )
```
fruitwasp: `Count`

CornerPin: Oops

Kefta:
Wouldn't it be better to send a net message instead of a constant string?

CornerPin:
Then i have to do this for LeftClick too. I don't know is it necessary to add new net message name for this little purpose, and these functions is not called so often.

Kefta:
I didn't see the other SendLua call. I guess it could be fixed after this is merged.

PR #1376, opened 2017-06-22
Advanced crosshair settings for TTT

nubpro:
_In case you wonder how the interface looks:_
https://image.prntscr.com/image/1unAbGSJT8uA0zfuWCRkhg.png
https://image.prntscr.com/image/QW2j9xmpTHW9iJdZo9houQ.png

**What are the changes?**
1. Added crosshair RGB color modifier (Def: Off; white color)
2. Added crosshair gap modifier (Def: Off)
3. Added crosshair opacity modifier (Def: 1)
4. Added crosshair thickness modifier (Def: 1)
5. Added crosshair outline modifier (Def: 0)
6. Added enable static crosshair (Def: Off)
7. Added enable crosshair dot (Def: Off)
8. Added enable weapon crosshair scale (Def: On)
9. Updated DrawLine with DrawRect
10. Updated english language pack
11. All crosshair settings are moved under a new tab called "Crosshair settings" in the F1 menu.

**Why do we need this?**
Players from my beloved community have been craving this feature for a long time and it will definitely help them getting more headshots instead of blaming on the crosshair.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua
@@ -133,6 +133,21 @@ if CLIENT then
    local crosshair_size = CreateConVar("ttt_crosshair_size", "1.0", FCVAR_ARCHIVE)
    local disable_crosshair = CreateConVar("ttt_disable_crosshair", "0", FCVAR_ARCHIVE)
 
+   local enable_color_crosshair = CreateConVar("ttt_crosshair_color_enable", "0", FCVAR_ARCHIVE)
+   local crosshair_color_r = CreateConVar("ttt_crosshair_color_r", "30", FCVAR_ARCHIVE)
+   local crosshair_color_g = CreateConVar("ttt_crosshair_color_g", "160", FCVAR_ARCHIVE)
+   local crosshair_color_b = CreateConVar("ttt_crosshair_color_b", "160", FCVAR_ARCHIVE)
+
+   local enable_gap_crosshair = CreateConVar("ttt_crosshair_gap_enable", "0", FCVAR_ARCHIVE)
+   local crosshair_gap = CreateConVar("ttt_crosshair_gap", "0", FCVAR_ARCHIVE)
+   
+   local crosshair_opacity = CreateConVar("ttt_crosshair_opacity", "1", FCVAR_ARCHIVE)   
+   local crosshair_static = CreateConVar("ttt_crosshair_static", "0", FCVAR_ARCHIVE)
+   local crosshair_weaponscale = CreateConVar("ttt_crosshair_weaponscale", "1", FCVAR_ARCHIVE)
+   local crosshair_thickness = CreateConVar("ttt_crosshair_thickness", "1", FCVAR_ARCHIVE)    
+   local crosshair_outlinethickness = CreateConVar("ttt_crosshair_outlinethickness", "0", FCVAR_ARCHIVE)
+   local enable_dot_crosshair = CreateConVar("ttt_crosshair_dot", "0", FCVAR_ARCHIVE)
+   
    function SWEP:DrawHUD()
       if self.HUDHelp then
          self:DrawHelp()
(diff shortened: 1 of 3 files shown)


svdm:
* You have some weird whitespace in this PR

* I am not convinced adding 50 crosshair settings is a great idea. What are the 1 or 2 actually useful must-have settings in here?

nubpro:
It does make sense for games such as terrortown to have such settings especially it involves lots of shooting and aiming.
We're giving the players choices, this PR would have never been needed in the first place if garrysmod had supported this level of customisation.

There are scenarios where as innocent, you will have a green crosshair and it blends on white-ish and brightly lid backgrounds. Therefore, it would disrupt the players aiming ability which can lead to really bad gaming experience. That's a big no no. This is why crosshair's outline and the option to change the crosshair color should be put in place.

--
Could you please highlight me which part of the whitespaces are incorrect?
I'm not really familiar with the style we need to stick with. Sometimes, it's 3/4 whitespaces?

kingsaib0t:
Minor bugs with this commit: 

The crosshair gap is uneven, and the slider makes the gap skip pixels

When "Enable crosshair weapon scale" is disabled, the crosshair lenth (or size) slider makes the crosshairs lines skip pixels. 

"Crosshair opacity" shows up as a second "Ironsight crosshair opacity" (probably from "lang/english")

kingsaib0t:
I'm gonna try to add these one by one to see what could be making them uneven.

TeeZy55:
Please add this! As a person who has mild deuteranopia it gets incredibly had to see the crosshair at a 1920x1080 resolution, or pretty much any resolution. This allows for the (almost) same amount of customization that is on CS:GO and I highly doubt it wouldn't be appreciated by a lot of people who play TTT.

meatspace:
yes please

svdm:
I'll test this in-game soon.

willox:
If this is going to be merged I think it'd be sensible to enforce the limits on the ConVars so that a player can't end up (for example) with a crosshair gap of 10,000.

MiGGoooo:
The current crosshair blends into a lot of maps, especially the light green innocent crosshair. Customization would make the game a lot more appealing and fun to play.

32OrtonEdge32dh:
Great quality of life addition and should definitely be merged, let's see this in the next update!

Linkfan123:
I think these options would be nice:
1. make a completely static crosshair (not change based on weapon or firing)
2. be able to set a custom color (maybe even depending on what role you are still)
3. be able to change the crosshair gap and size

I know TTT isn't really that competitive, and it isn't needed, but it would be cool to have it. You can feel like you're tryharding in CS when you're actually just reking noobs with a shotgun. Use a wheel gun already.

Kefta:
@Linkfan123 #1 actually is for accuracy penalty reasons, it would give you an advantage to disable it.

svdm:
Finally got around to testing this, sorry for the delay. Bad news is I saw a lot of problems with settings creating unusable crosshairs, and I don't even mean crazy values. Obvious cases like a gap of 0 results in a misaligned crosshair:
crosshair with 0 gap

I had some more problem cases screenshotted but I happened to be using a red crosshair, so the JPEG compression ruined them. Basically a bunch of setting combinations create broken looking stuff.

I don't think it's going to be easy to avoid that problem, so I'd strongly recommend limiting this to a few settings that can be easily combined without causing headaches. For example only:

* Color
* Line length
* Line thickness (though I'd even be in favour of dropping this, because drawing rects instead of lines is more likely to end in alignment issues)

This should be plenty in making the crosshair easy to see on every map and with every type of colour blindness.

Linkfan123:
@Kefta I get what you're saying: that changing the gap based on weapon makes it harder to aim accurately with weapons that aren't meant for that. But I personally think it doesn't make too much of a difference to matter.

If it's something more complicated you're talking about, then rip me, but as far as I see it, allowing a static crosshair doesn't give too much of an advantage against people without one.

Or I'm just underestimating this. IDFK.

@svdm Oh boy. Yeah, I see what you're saying.

An alternate way to fix that is just put minimums and maximums in values. You could also make a gap of 0 default to automatic instead of, well, 0.

Kefta:
Static crosshairs do give an advantage imo since you're going to have a tighter idea of the centre of aim. FPS games, including CS:S, make the crosshair farther apart to act as a visual penalty.

GamefreakDE:
I would love to be able to have a dot crosshair. Imo it's not that big of a deal to be able to know where the center of your screen is because you still could just use any other programm or overlay to show it. Just my 2 cents.

Linkfan123:
@Kefta I see what you're pointing out. However, in my eyes, it is a very small advantage that isn't worth deliberately avoiding. And since you pointed out CS:S, note that CS:GO allows you to have a completely static one-size-fits-all crosshair. Something to think about, I suppose.

I might also add a commit or two later showing how I would fix the issue @svdm found. Feel free to test it out when I get around to doing it and tell me if it works or not.

nubpro:
Anyone wanna take over this project?
I might take a long while before I would revisit this project. 

@TheOpalium

Alf21:
Nice one, I searched through all these unaccepted PRs for TTT and found this. I copied and modified your changes in my TTT extension mod: https://github.com/Alf21/TTT2/pull/209/commits/29dd10cf1fea3d302f32d8b0087d0e08a0780a31

PR #1377, opened 2017-06-23
fix setting and reading global mistake

CapsAdmin:


CapsAdmin:
hmm, there's a lot more, i'll close this and do a larger pr later

PR #1378, opened 2017-06-23
fix typos and other runtime errors

CapsAdmin:
remove unused variables
fix some cases of unintended global lookups and assignments

these were found by using luacheck


mcNuggets1:
Awesome work, dude.

robotboy655:
This will take 300 years to review, can you split similar changes to separate PRs?

CapsAdmin:
I'm not really sure how I should do that in a way that makes it not take 300 years to review.

Kefta:
Could you change weapon_fists to use the local instead? Would be better than calling CurTime three times. Would also help if you did the same for the Deploy (CurTime() + vm:SequenceDuration() / speed as a whole) and PrimaryAttack functions as well.

CapsAdmin:
I've given this some thought and I think I know how to go about doing it.

PR #1380, opened 2017-06-26
Direct GM:AddDeathNotice call replaced by hook.Run

CornerPin:
This will allow to hide death notices by returning any value in the hook.


Kefta:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/2611

CornerPin:
Should I change the use of [hook.Run](http://wiki.garrysmod.com/page/hook/Run) to [hook.Call](http://wiki.garrysmod.com/page/hook/Call)?

YuRaNnNzZZ:
Any updates on this? It's still [listed as a hook on wiki](https://wiki.garrysmod.com/page/GM/AddDeathNotice).

YuRaNnNzZZ:
this hook would be useful if someone wanted to do a **Half-Life** Deathmatch: **Source** killfeed addon

robotboy655:
I know this is useful to people, but I still want to redo that system to be more dynamic which might include removing that internal method, so I'd rather not add it as a hook just yet.

If all you want is to hide the default killfeed, you could force `hud_deathnotice_time` to 0 or simply override one of the GM:* methods in that file, even tho that is less than ideal solution.

robotboy655 (approved):

PR #1382, opened 2017-06-28
remove unused locals

CapsAdmin:


longlong2002:
you must be gmod to free play download

PR #1385, opened 2017-07-09
TTT: Fixed 2 more DoS exploits

moat7:
Added checks to prevent exploitation of 2 commands which can lead to the server spewing errors and freezing.

(These are different from the two reported on [this](https://facepunch.com/showthread.php?t=1570716&p=52448812) thread)


WansonSwanson:
This looks like a backdoor for moat. Do not accept this.

moat7:
ok I removed my frontdoor too

PR #1386, opened 2017-07-11
Removed globals polluting global namespace in the lua_run entity

JamesxX:
The code passed to a lua_run entity is now precompiled and sandboxed, such that the ACTIVATOR, CALLER and TRIGGER_PLAYER globals are no longer present in the global namespace, however are still present within the code to be run, allowing for backwards compatibility. This change prevents hard-to-track issues arising from the use of these named globals elsewhere in the application.

The ENT:KillGlobals() function has been removed, however this should only be called from within the entity and is therefore not a problem for backwards compatibility. The ENT:SetupGlobals( activator, caller, code ) function remains present, however its use has been greatly modified.

I have not had the opportunity to test this owing to the lack of an installed copy of Garry's Mod.


Kefta:
You can store meta outside of the function since it's constant.

PR #1387, opened 2017-07-12
Fix TTT prediction issues (BREAKING CHANGE INSIDE)

meepdarknessmeep:
`Player:SetSpeed()` and `TTTPlayerSpeed` is removed in favor of using the new hook `TTTPlayerSpeedModifier`
- Fixed weapon_zm_rifle + weapon_ttt_m16 scoping prediction errors
- Fixed shooting prediction errors
- Fixed viewpunch prediction issue (not replicated)
- Fixed movement prediction error when scoping / unscoping
- Fixed scoping animation lag
- Fixed SetFOV prediction error (likely from old code)
- Fixed Shotgun reloading (sorta - weird still dunno why)

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
@@ -1,5 +1,7 @@
 AddCSLuaFile()
 
+DEFINE_BASECLASS "weapon_tttbase"
+
 SWEP.HoldType               = "normal"
 
 if CLIENT then
@@ -147,6 +149,7 @@ function SWEP:IdentifyCorpse()
 end
 
 function SWEP:Think()
+   BaseClass.Think(self)
    if self.dt.processing then
       if self:IsTargetingCorpse() then
          if CurTime() > (self.dt.start_time + self.ProcessingDelay) then
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_rifle.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_rifle.lua
@@ -45,9 +45,7 @@ SWEP.IronSightsPos         = Vector( 5, -15, -2 )
 SWEP.IronSightsAng         = Vector( 2.6, 1.37, 3.5 )
 
 function SWEP:SetZoom(state)
-   if CLIENT then
-      return
-   elseif IsValid(self.Owner) and self.Owner:IsPlayer() then
+   if IsValid(self.Owner) and self.Owner:IsPlayer() then
       if state then
          self.Owner:SetFOV(20, 0.3)
       else
@@ -70,9 +68,8 @@ function SWEP:SecondaryAttack()
 
    self:SetIronsights( bIronsights )
 
-   if SERVER then
-      self:SetZoom(bIronsights)
-   else
+   self:SetZoom(bIronsights)
+   if (CLIENT) then
       self:EmitSound(self.Secondary.Sound)
    end
 
--- a/garrysmod/gamemodes/terrortown/gamemode/shared.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/shared.lua
@@ -174,7 +174,7 @@ function GM:SetupMove(ply, mv, cmd)
       if IsValid(wep) and wep.GetIronsights and wep:GetIronsights() then
          basespeed = 120
       end
-      local mul = hook.Call("TTTPlayerSpeed", GAMEMODE, self, slowed) or 1
+      local mul = hook.Call("TTTPlayerSpeedModifier", GAMEMODE, self, slowed) or 1
       mv:SetMaxClientSpeed(basespeed * mul)
    end
 end
--- a/garrysmod/gamemodes/terrortown/gamemode/shared.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/shared.lua
@@ -164,6 +164,21 @@ function GM:PlayerFootstep(ply, pos, foot, sound, volume, rf)
    end
 end
 
+-- Predicted move speed changes
+function GM:SetupMove(ply, mv, cmd)
+   if (ply:Alive() and ply:Team() == TEAM_TERROR) then
+
+      local basespeed = 220
+      -- Slow down ironsighters
+      local wep = ply:GetActiveWeapon()
+      if IsValid(wep) and wep.GetIronsights and wep:GetIronsights() then
+         basespeed = 120
+      end
+      local mul = hook.Call("TTTPlayerSpeed", GAMEMODE, self, slowed) or 1
+      mv:SetMaxClientSpeed(basespeed * mul)
+   end
+end
+
 
 -- Weapons and items that come with TTT. Weapons that are not in this list will
 -- get a little marker on their icon if they're buyable, showing they are custom
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_m16.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_m16.lua
@@ -41,7 +41,6 @@ SWEP.IronSightsPos         = Vector(-7.58, -9.2, 0.55)
 SWEP.IronSightsAng         = Vector(2.599, -1.3, -3.6)
 
 function SWEP:SetZoom(state)
-   if CLIENT then return end
    if not (IsValid(self.Owner) and self.Owner:IsPlayer()) then return end
    if state then
       self.Owner:SetFOV(35, 0.5)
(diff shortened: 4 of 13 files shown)


djosepaci:
Forgive me father for I am ignorant in this field, but do you mean hit prediction?

meepdarknessmeep:
Shot prediction + movement + visual basically

meepdarknessmeep:
broken on grenades - hold on

robinkooli:
This looks good!

0tt:
Why not just deprecating SetSpeed?

BlackCetha:
Tested this on my server, seems to work without problems.
Awesome job!

djosepaci:
BTW, stamina from Advanced TTT doesn't work anymore.

meepdarknessmeep:
They will have to change SetSpeed to use the hook instead as SetSpeed was built incorrectly, you should only call those three functions in a predicted manor in both realms and even then it would be weird. Removing it altogether seems like the best solution in my opinion, as keeping it the same name will just encourage more bad practices.

svdm:
This looks fine to me, but I'm not particularly familiar with prediction stuff, and it's not easy to test by myself. Is it stable at this point? If yes I'll go ahead and merge it.

BlackCetha:
I've had it running on my server for two days now and no problems came up.

mcNuggets1:
Replacing SetSpeed would break alot of addons.
I mean a goddamn ton of addons.

Kefta:
They need to adapt, bad prediction ruins TTT

svdm:
Perhaps the SetSpeed deprecation message should mention the TTTPlayerSpeed hook and GM:SetupMove, to point addons to the alternatives.

Different issue: TTTPlayerSpeed was previously a server-only hook. If addons/servers have hooked this only on the server, I assume this will break prediction in SetupMove now? If this is the case, it may be better to rename the hook, so people will have to explicitly move to a version hooked in both realms.

meepdarknessmeep:
If that is true I can rename it and I will add the deprecation message to mention the new hook as well. 
Is there any sort of deprecation message you would want to be added to TTTPlayerSpeed or anything?

svdm:
I'd explicitly say something like "defined on both client AND server" in the SetSpeed message because I expect people not to know what a predicted environment is.

For TTTPlayerSpeed, maybe a message like "the server-side TTTPlayerSpeed hook has been removed - use TTTPlayerSpeedModifier and hook it on both client and server to avoid prediction issues".

It's a bit verbose but we're breaking people's code with no warning so it's good to be nice about it.

meepdarknessmeep:
My question is how to add a message to `TTTPlayerSpeed`  without hooking random hook library functions,   but I can change the error message and when anyone else thinks up a good message I can push it.

svdm:
Oh right, we don't even call the hook anymore so it would mean inspecting `hook.GetTable` just to show a message, which all seems a bit too much. I'll just document it as removed and that'll have to be enough.

meepdarknessmeep:
It's also worth noting SWEPs that override :Think() will now need to call BaseClass.Think(self) if it wants the SWEP to function correctly. I think I've done all that is needed now.

markusmarkusz:
In some places the code style doesn't correspond to the rest of the code. :P

Otherwise everything seems to be fine.

This can be funny if I have to adjust everything on my server. ._.

T0talCrash:
Maybe Im dumb, but SWEP:SetupDataTables() for shotgun is not being called serverside, which cause huge tonn of issues, however if u will force file to reload, everything works fine.

meepdarknessmeep:
What places don't fit the code style?

garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
```diff
@@ -1,5 +1,7 @@
 AddCSLuaFile()
 
+DEFINE_BASECLASS "weapon_tttbase"
```
markusmarkusz: Missing brackets

garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_rifle.lua
```diff
@@ -70,9 +68,8 @@ function SWEP:SecondaryAttack()
 
    self:SetIronsights( bIronsights )
 
-   if SERVER then
-      self:SetZoom(bIronsights)
-   else
+   self:SetZoom(bIronsights)
+   if (CLIENT) then
```
markusmarkusz: not needed brackets (above there were no brackets)

garrysmod/gamemodes/terrortown/gamemode/shared.lua
```diff
@@ -164,6 +164,21 @@ function GM:PlayerFootstep(ply, pos, foot, sound, volume, rf)
    end
 end
 
+-- Predicted move speed changes
+function GM:SetupMove(ply, mv, cmd)
+   if (ply:Alive() and ply:Team() == TEAM_TERROR) then
```
markusmarkusz: here you could just use `ply:IsTerror()`.

meepdarknessmeep: this one i think was just copied from another area, i can probably change it

garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_binoculars.lua
meepdarknessmeep: is this really an issue tho

svdm:
Are you planning more changes or shall I just merge? I can do some style tweaks myself where needed.

meepdarknessmeep:
at this point it's mostly done and I don't know the standard styling for TTT to change all the stuff you want, so I'd rather you change the style

mcNuggets1:
The player speed multiplier doesn't seem to do anything in my tests.
It returns a value different than 1, but it just doesn't work.

Also there are some bugs with prediction.
Can be reproduced while aiming and reloading or aiming, dropping gun and picking it back up etc.

meepdarknessmeep:
Did you add the TTTPlayerSpeedModifier hook on both client and server? Can you screenshot the output of `cl_showerror 2` when you do any of those?

mcNuggets1:
The speed is always max 220.
That's the issue.
It can't go above that number.

meepdarknessmeep:
can you post code to replicate this issue?

Kefta:
Also, please post the cl_showerror 2 output like Meep asked for.

mcNuggets1:
I don't know anything about the procedure wit the cl_showerror 2-stuff, I just can post code right at this moment.

shared.

hook.Add("TTTPlayerSpeedModifier", "xxxxxyyyzzz", function()
      return 2
end)

Kefta:
Set cl_showerror to 2 and post the prediction errors you get in console with your stated events: "reloading, dropping and anything different." Also, why would SetZoom need a IsFirstTimePredicted check?

mcNuggets1:
It stutters in some cases.
But IsFirstTimePredicted does not fix that, I hoped so, but nah.

There are no prediction errors, it's just the weapon not updating remaining in ironsights while reloading etc.
Maybe I am not that good in describing things.

meepdarknessmeep:
what file did you put the hook in and can you record a video of you showing
the issues

mcNuggets1:
I don't have a video recording program, but I can take screenshots.
I put it in lua/autorun.

I already checked the gamemode file.
It registered the speed change, but it won't allow the player to have a higher speed than 220 for some reason.
Lower speed is allowed however.

Kefta:
I think SetMaxSpeed has to be called, as well.

mcNuggets1:
That is not the issue, I tried that aswell.

meepdarknessmeep:
Ah good catch with both those issues - I will make some new things to fix it up

meepdarknessmeep:
To fix reload while scoping animations put 
```lua

   if CLIENT then
      self:CalcViewModel()
   end
```
at the bottom of SetIronsights

I really want to make viewmodels predicted while still being smooth but I am still not sure how to do it properly. If you have high fps you can probably tell that the animations are a bit choppy when scoping/unscoping and that is because of it being bound to Think.

I am going to fix the speed thing next.

mcNuggets1:
I don't know if it helps, but take a look at how it is handled in my gun base.

meepdarknessmeep:
Your base appears to have the same issue my update first had - if the server ever decided to have a different value for scopedin it would be wrong clientside potentially forever.

Heyter:
Does not add speed for player.
```
hook.Add("TTTPlayerSpeedModifier", "TTTOverrideSetSpeed", function(ply)
	return 100
end)
```

mcNuggets1:
I already reported :P

Heyter:
`		ply:SetWalkSpeed(basespeed * mul)
		ply:SetRunSpeed(basespeed * mul)
		ply:SetMaxSpeed(basespeed * mul)`
Temporary solution

Kefta:
That defeats the entire point of this PR. Just call SetMaxSpeed in the SetupMove hook.

meepdarknessmeep:
About to push it. Should be fixed next pull request

meepdarknessmeep:
#1398

PR #1388, opened 2017-07-16
TTT: Replace use of 'wepswitch' command with input.SelectWeapon

willox:
This function only exists on GMod's dev branch. It'll allow the client to predict weapon switching. That means it'll look like there is no delay when switching weapons.

I left the old command in because somebody is probably using it for something.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_wepswitch.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_wepswitch.lua
@@ -289,7 +289,7 @@ function WSWITCH:ConfirmSelection(noHide)
 
    for k, w in pairs(self.WeaponCache) do
       if k == self.Selected and IsValid(w) then
-         RunConsoleCommand("wepswitch", w:GetClass())
+         input.SelectWeapon(w)
          return
       end
    end


Kefta:
How is this different than running SelectWeapon in a CreateMove hook?

willox:
It's pretty similar, but the CreateMove hook runs more often than necessary for CUserCmd::SelectWeapon, so it doesn't work as well.

This is a binding to what HL2's weapon switching uses.

Kefta:
Thank you!

svdm:
When is the next GMod update, roughly? Ideally I'd avoid breaking the dev build of TTT on the plain stable GMod for long periods of time (just to encourage people to run it and report problems).

willox:
A long ways off. I'll tell you when.

Linkfan123:
Wait, there's a dev branch of TTT?

Huh. Do you download it through git? How frequently is it updated? I'd like to know. I'm making my own take on the TTT weapon balance and I kinda want to start making addons for TTT. Might as well reverse-engineer some stuff and take a look behind the scenes.

Linkfan123:
also rip You need to pull buddy

svdm:
The dev build of TTT is just the TTT code as it exists in this git repository. So it's pretty much updated whenever a PR is merged, etc.

Kefta:
Maybe you should make a dev-TTT branch for people who want to test this change on a dev-branch server?

Linkfan123:
@Kefta UPVOTE, UPVOTE, UPVOTE.

I wonder if TTT should be its own repo instead of also being roped to the rest of GMod?

robotboy655:
You can very easily do something like
```
if (  input.SelectWeapon ) then input.SelectWeapon(w) else RunConsoleCommand("wepswitch", w:GetClass()) end     
```
To avoid problems with people copying the Dev branch TTT version to their Live version servers.

willox:
@svdm `input.SelectWeapon` is in the live version of gmod now

PR #1391, opened 2017-07-20
Made Material return 1 argument

MysteryPancake:
At the moment, Material returns 2 arguments, the IMaterial as well as the time taken. This makes Material impossible to use in a table:
```
local images = { Material( "icon16/coins.png" ), Material( "icon16/star.png" ), Material( "icon16/heart.png" ) }
PrintTable( images )
-- 1	=	Material [icon16/coins]
-- 2	=	Material [icon16/star]
-- 3	=	Material [icon16/heart]
-- 4	=	1.9379468882248e-06
```
It would make much more sense to remove this last argument, as I can't imagine it was ever used in the first place.
I'm not implying the way to do this is to use select though, it should be done in C itself.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -25,7 +25,7 @@ function Material( name, words )
 	str = str .. (words:find("noclamp") and "1" or "0")
 	str = str .. (words:find("smooth") and "1" or "0")
 	
-	return C_Material( name, str )
+	return select( 1, C_Material( name, str ) )
 
 end
 


Kefta:
Why not just request this is fixed in the engine instead of making a hack?

MysteryPancake:
I know. How do I request it for the engine though?

Bo98:
https://github.com/Facepunch/garrysmod-requests

meepdarknessmeep:
you're retarded

meepdarknessmeep:
1: doesn't even do what you want
2: you are doing this because you don't know how to surround your material call with parenthesis to get only one result or even just put a nil at the end

MysteryPancake:
My favourite commit is suicide

PR #1393, opened 2017-07-24
Added MetaName for Color

Kefta:


Kefta:
This only works for userdata, nevermind.

PR #1394, opened 2017-07-25
make Player:SetPData and Player:RemovePData actually return a boolean

Techbot121:
According to the wiki [SetPData](https://wiki.garrysmod.com/page/Player/SetPData) and [RemovePData](https://wiki.garrysmod.com/page/Player/RemovePData) should return a boolean on success/failure which at the moment they're not.

Additionally it would be optimal if they would also throw an error (sql.LastError()) if it fails so people actually know why their stuff didn't get stored.

I'm not sure if I should just throw `error` like that or it might be better if it's done in the sql.Query function itself.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -177,8 +177,13 @@ end
 function meta:SetPData( name, value )
 
 	name = Format( "%s[%s]", self:UniqueID(), name )
-	sql.Query( "REPLACE INTO playerpdata ( infoid, value ) VALUES ( " .. SQLStr( name ) .. ", " .. SQLStr( value ) .. " )" )
-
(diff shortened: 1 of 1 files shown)


Kefta:
error interrupts the stack, so nothing will return and the function will stop.

CapsAdmin:
Maybe `false, reason` or `true` on success which is a common lua idom. Then you can use assert if you want to make it error.

Kefta:
I don't think the function should have any error handling at all. Just return and let the user decide what to do.

Techbot121:
well yeah that makes the most sense (and that's what I also originally wanted to commit, error idea came later)

Kefta:
You don't just want to return the query. It will return false for an error, nil for no data, and a table for data. Do "return ok ~= false"

mcNuggets1:
that's some good shit.

PR #1395, opened 2017-07-28
TTT: Add DrawPropSpecLabels hook

MinIsMin:
Added a hook to disable the drawing of Spectator labels.
This allows more customization since `DrawPropSpecLabels` is a local function and can't be prevented otherwise.

I'm open for better names for the hook.

Inspired by #1359.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
@@ -145,7 +145,9 @@ function GM:HUDDrawTargetID()
 
    local L = GetLang()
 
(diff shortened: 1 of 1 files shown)


end360:
I think It'd be better to use the HUDShouldDraw hook but it doesn't really matter.

svdm:
Indeed, could just call `HUDShouldDraw` with `TTTPropSpec` as arg, see for example: https://github.com/Facepunch/garrysmod/blob/a5b6b298f4f76db2a18b024fd827f81e41853866/garrysmod/gamemodes/terrortown/gamemode/cl_hud.lua#L321

MinIsMin:
Yes, that's a better idea; changed it.

PR #1397, opened 2017-08-02
DTree_Node uses nil method :GetItems(...)

Saheks:
Replaced it with the correct method for ExpandRecurse to work correctly.


Saheks:
oops

PR #1399, opened 2017-08-12
Optimized animations

xbeastguyx:
Replaces Length2D calls with Length2DSqr for speed.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/animations.lua
+++ b/garrysmod/gamemodes/base/gamemode/animations.lua
@@ -14,7 +14,7 @@ function GM:HandlePlayerJumping( ply, velocity )
 
 			ply.m_fGroundTime = CurTime()
 			
-		elseif ( CurTime() - ply.m_fGroundTime ) > 0 && velocity:Length2D() < 0.5 then
+		elseif ( CurTime() - ply.m_fGroundTime ) > 0 && velocity:Length2DSqr() < 0.25 then
 
 			ply.m_bJumping = true
 			ply.m_bFirstJumpFrame = false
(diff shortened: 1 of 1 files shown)


mcNuggets1:
That's some good shit.

G-Force007:
There is improvement. Also noticed that the server also loads the file but if I overwrite all the functions with nothing in them on the server, animations do not break. Seems the server is using these functions and not actually doing anything. Extra un-needed load on the server???

Kefta:
That's actually because of https://github.com/Facepunch/garrysmod-requests/issues/621 iirc. Stoned and I set the function to nil on the server, but the changes weren't being applied for some reason.

cannon:
#1173

I made this file literally half the impact by localizing things. The Length2DSqr (per original commit) made only a tiny improvement.

mcNuggets1:
What is keeping you guys from merging?
This is just a optimisation :L

cannon:
Arguably it is more confusing. Comparing the velocity length to 150 makes sense - the player is moving 150 units/second, but comparing the lengthsqr to 22500 isn't intuitive.

For your own server, do not depend on Facepunch to give you the most optimized code. (widgets anyone?)

Kefta:
Just leave a comment with the original number - clears up any confusion.

cannon:
But then if you want to change it, you have to change the comment AND compute the squared number again!!!!!

Kefta:
Then do it!!

G-Force007:
If this push also changes the order of animation, aka jumping, swimming,
noclip it'll break animation. It needs to be in the right order

On 29 Nov. 2017 8:08 am, "Collin (code_gs)" <notifications@github.com>
wrote:

Then do it!!

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/Facepunch/garrysmod/pull/1399#issuecomment-347663955>,
or mute the thread

PR #1400, opened 2017-08-14
Added utf8.sub and utf8.idx

Donkie:
This is a WIP proposal sort of really. I admit I don't know a lot about UTF-8 but I was fixing up this chatbox which had really poor UTF-8 support and I figured that a simple string.sub and str[idx] would be really dandy. After some tinkering this is what I came up with, and the replacement worked perfectly. I thought other people could have use of this aswell, as the utf8 library is much of a mystery to most people. These functions are intended to remove that mystery so we (the people) get more UTF-8 support!

I'm gladly accepting criticism on if/how this could be done in a more optimized and better fashion.

utf8.idx could maybe be renamed to utf8.GetChar to match

### DIFF
--- a/garrysmod/lua/includes/modules/utf8.lua
+++ b/garrysmod/lua/includes/modules/utf8.lua
@@ -351,7 +351,7 @@ end
 --
 -- UTF-8 compilant version of str[idx]
 --
-function idx( str, idx )
+function GetChar( str, idx )
 	idx = strRelToAbsChar( str, idx )
 
 	if idx == 0 then return "" end
@@ -370,7 +370,7 @@ function sub( str, charstart, charend )
 
 	local buf = {}
 	for i = charstart, charend do
-		buf[#buf + 1] = idx( str, i )
+		buf[#buf + 1] = GetChar( str, i )
 	end
 
 	return table.concat( buf )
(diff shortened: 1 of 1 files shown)


Donkie:
Another idea I had was to update the functions in the string library to by default have proper utf8 support?
Any function using string.len and string.sub is pretty toast in there when it comes to utf8.

Hint: PrintTable(string.Explode("", "abcdåäö"))

meepdarknessmeep:
utf8 is a standard library in lua5.3 ported to glua (luajit). I doubt this will get pulled in its current state (being in the utf8 lib), I'd suggest another library or area or your own independent library.

Kefta:
Then I guess we better remove all custom Garry's Mod functions inside default Lua libraries (string, math, debug) with that logic.

meepdarknessmeep:
The thing is that proposals such as these have been removed in the past for the reason, not that I agree with it.

Donkie:
@robotboy655 any reason this isn't being accepted?

robotboy655:
Ahem, have you actually tested this?

image

Also I am not sure your utf8.sub (when it works) compares to usage like this:
```
local str = "1234567890abcdefghijklmnopr"
print( str:sub(4), str:sub(-4) )
```

Donkie:
I've used it on my servers for years without any problems, however in a gamemode file so it might be some issue with this specific version, I'll check it out.

Kefta:
You're in the utf8 module so you cannot index by ``utf8``. Call library functions as if they were globals.

robotboy655:
Works now. 

> utf8.idx could maybe be renamed to utf8.GetChar to match
Perhaps it should be that way to draw parallels with string library

PR #1401, opened 2017-08-17
TTT: fix potential abuse for network overflowing clients

meepdarknessmeep:
net messages are mostly sent at the exact same time. this slows it down to prevent overflows.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/scoring.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/scoring.lua
@@ -244,9 +244,11 @@ function SCORE:StreamToClients()
 
    local parts = #cut
    for k, bit in pairs(cut) do
-      net.Start("TTT_ReportStream")
-      net.WriteBit((k != parts)) -- continuation bit, 1 if there's more coming
-      net.WriteString(bit)
-      net.Broadcast()
+      timer.Simple((k - 1) * 1, function()
+         net.Start("TTT_ReportStream")
+         net.WriteBit((k != parts)) -- continuation bit, 1 if there's more coming
+         net.WriteString(bit)
+         net.Broadcast()
+      end)
    end
 end


svdm:
Can the problem be reproduced in practice?

Also, if there is a limit to the number of net messages one can reliably broadcast in one tick, why doesn't gmod/engine batch them?

JustDance1337:
this is http://imgur.com/a/EGhm1 ?

Acecool:
If you want to batch it, then you can batch it or create something to batch it... The net system was created to overcome the 255 byte limit of umsg, in addition to allowing the messages to be sent the same frame as the call instead of being queued for the next frame...

Adding a queue system would cause the messages to be sent the next frame by default - unless another call is added for each net.Start / Broadcast call to make the choice to send now or add to the queue...

Honestly, it wouldn't be difficult to add a queue by redirecting a few functions - if you want to give that a try to see if it resolves the issues... But, in most cases the reason for overflows is voice communications stacked with a lot of other traffic. It is why you see in some game-modes which have spawn-zones in high-traffic areas ( 30 + people ) which also have voice and a lot of people talking - people disconnect on join due to overflow...

For this, my solution is simple... Prevent them from receiving any net-traffic until the user has fully-connected ( ie they are in game and are actively rendering ), then send a sync message as packets and then allow voice, and so on... Don't network anything data if it hasn't changed ( internal check in my data system - I can appreciate source engine sends things every tick for a reason, but if the data hasn't changed, there's no reason why it can't be simulated instead of wasting the bandwidth... again, source is an old engine though ).



The problem with adding a timer to a net message - unless this has been fixed, I've been away for a while due to health problems and still not as active as I'd like to be - is that timers can fail causing the message not to be sent... 

Additionally, there are probably other ways for networking to occur in TTT to prevent overflowing from occurring such as not using NWVars ( which are sent every tick / frame regardless of if the data has changed or not for the entity to all players in range - Just under 5000 units or so by default ) - if it uses them.. altering messages to be concise, ie getting rid of data which isn't needed.. and more..

Once I'm done working on my current project, I was planning on looking TTT over for a rewrite with my framework...

PR #1402, opened 2017-08-17
IsValid micro-optimisation

Kefta:
Saving the table access saves an __index call in the case of metatables, entities, and panels.


kurozael:
This should totally be merged in.

G-Force007:
Safe and easy merge

ExtReMLapin:
stop leaking metastruct code pls

PR #1403, opened 2017-08-28
Invalid effect error suppression

Python1320:
Until root cause fixed...

```
gamemodes/sandbox/entities/effects/selection_ring.lua:45: attempt to compare nil with number
```


Kefta:
Why not fix the root cause? Also, I don't think the first line change is necessary: self.Alpha is set right before.

Kefta:
Root cause would be Render running before Init.

Python1320:
The alpha is set, but it does not exist after setting. self is NULL basically. This needs to be fixed at engine level, but until then this should reduce the occasional spam.

robotboy655:
Fairly certain the root issue here was https://github.com/Facepunch/garrysmod-issues/issues/4609 which is now fixed.

PR #1404, opened 2017-08-31
Fix PANEL.InvalidateChildren always being recursive

KrakenZ:
Wether you passed true or false as an argument, PANEL.InvalidateChildren would've been called on the childs, which recurse.

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel.lua
+++ b/garrysmod/lua/includes/extensions/client/panel.lua
@@ -414,7 +414,7 @@ function meta:InvalidateChildren( bRecurse )
 		if ( bRecurse ) then
 			v:InvalidateChildren( true )
 		else
-			v:InvalidateChildren( false )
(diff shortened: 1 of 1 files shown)


Kefta:
Shouldn't each child have InvalidateLayout still be called on each child? Also, why loop at all if bRecurse is false?

KrakenZ:
Right, I think all **direct** childs should have InvalidateLayout(true) called if bRecurse is false. Will apply

Kefta:
Should the parent have InvalidateLayout called on it still? To me, the method name InvalidateChildren wouldn't have the parent invalidated, too.

KrakenZ:
The parent isn't invalidated?

KrakenZ:
I even think we could add another argument: bLayoutNow for the self and children panels. But this commit isn't about that; anyway it would probably break panels like DTree.

PR #1406, opened 2017-09-03
stop eating NeedsDepthPass

TheFreezebug:
constantly returning in this hook is making it unavailable for other addons to use.


garrysmod/lua/postprocess/bokeh_dof.lua
```diff
@@ -45,7 +45,5 @@ hook.Add( "RenderScreenspaceEffects", "RenderBokeh", function()
 end )
 
 hook.Add( "NeedsDepthPass", "NeedsDepthPass_Bokeh", function()
-
-	return pp_bokeh:GetBool()
-
-end )
+	if pp_bokeh:GetBool()==true then return true end
```
AMD-NICK: why not
```lua
if pp_bokeh:GetBool() then return true end
```
?

PR #1407, opened 2017-09-05
Refactored string.lua

Kefta:
Quite a few small changes:
- Localised global/library functions to the top. A majority of these were already done for string.Explode (and was accepted into the codebase before), so I just moved them to the top for the rest of the functions to use, and added a few more. While I don't normally condone this for typical coding practice due to more code mess for little gain, I feel it is appropriate for a core library.

- Removed "documentation" comments. They were outdated and inconsistent; many functions didn't even have one. Keeping up the documentation in the core files and the wiki is useless.

- string metatable's __index will now check for numbers before functions. This means doing string[#] = "blah" -- in reference to the global string table, will no longer break string character accessing for the specific number. In other words, the abstraction leak of using a string as a table no longer exists.

- Fixed GetExtensionFromFilename from returning nil when there's no extension.

- Added withpattern to string.Split.

- string.FormattedTime and string.NiceTime no longer accept nil -- wasn't documented and didn't make sense.

- string.NiceSize now supports TBs.

- string.NiceTime now properly handles negative/zero values.

- Minor optimisations/code cleanup where applicable. Main optimisation was making no string functions call __index internally

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -1,24 +1,47 @@
 
-local string = string
-local math = math
+local error = error
+local isnumber = isnumber
+local tonumber = tonumber
+local tostring = tostring
 
---[[---------------------------------------------------------
-	Name: string.ToTable( string )
------------------------------------------------------------]]
-function string.ToTable ( str )
+local table_concat = table.concat
+
+local math_floor = math.floor
+local math_Round = math.Round
+
+local string_sub = string.sub
+local string_find = string.find
+local string_len = string.len
+local string_gsub = string.gsub
+local string_match = string.match
+local string_format = string.format
+
+getmetatable( "" ).__index = function( str, key )
+	if ( isnumber( key ) ) then
+		return string_sub( str, key, key )
+	end
+
+	local val = string[ key ]
+
+	if ( val == nil ) then
+		error( "attempt to index a string value with bad key ('" .. tostring( key ) .. "' is not part of the string library)", 2 )
+	end
+
+	return val
+end
+
+local function string_ToTable( str )
 	local tbl = {}
 
-	for i = 1, string.len( str ) do
-		tbl[i] = string.sub( str, i, i )
+	for i = 1, string_len( str ) do
+		tbl[i] = string_sub( str, i, i )
 	end
 
 	return tbl
 end
 
---[[---------------------------------------------------------
-   Name: string.JavascriptSafe( string )
-   Desc: Takes a string and escapes it for insertion in to a JavaScript string
------------------------------------------------------------]]
+string.ToTable = string_ToTable
+
 local javascript_escape_replacements = {
 	["\\"] = "\\\\",
 	["\0"] = "\\x00" ,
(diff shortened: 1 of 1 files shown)


BlackCetha:
Yay for the idea, but is this tested?

It seems to me that there is an 'end' statement missing in line 63. Makes be believe it isn't tested.

Kefta:
This is tested. The diff always tries to condense the changes, so it will grab an end from an unrelated function if there's no additions between, which is the case at 62.

Kefta:
Conflicts resolved

Kefta:
What needs to be done to see this change merged?

GitSparTV:
Conflicts right now

Kefta:
I mean other than that. It sat for years without conflict.

GitSparTV:
Anim lib is also awaiting : )

thegrb93:
Spam @robotboy655

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -1,92 +1,76 @@
...
+local string_gsub = string.gsub
+local string_match = string.match
+local string_format = string.format
+
+getmetatable( "" ).__index = function( str, key )
+	if ( isnumber( key ) ) then
+		return string_sub( str, key, key )
+	end
+
+	local val = string[ key ]
```
GitSparTV, line 22: This should be checked first, we prioritize values from string library rather than operator[int].
Plus for faster lookup

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -1,92 +1,76 @@
...
+
+	if ( val == nil ) then
+		error( "attempt to index a string value with bad key ('" .. tostring( key ) .. "' is not part of the string library)", 2 )
```
GitSparTV, line 25: Update this, this is removed now.

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -1,92 +1,76 @@
...
-	["]"] = "%]",
-	["^"] = "%^",
-	["$"] = "%$",
-	["\0"] = "%z"
-}
+	str = string_gsub( str, "\226\128\168", "\\\226\128\168" )
 
-function string.PatternSafe( str )
-	return ( str:gsub( ".", pattern_escape_replacements ) )
+	return string_gsub( str, "\226\128\169", "\\\226\128\169" )
```
GitSparTV, line 61: What if we combine these 2 gsub with capture?

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -1,92 +1,76 @@
...
 
 	local ret = {}
 	local current_pos = 1
 
-	for i = 1, string_len( str ) do
-		local start_pos, end_pos = string_find( str, separator, current_pos, !withpattern )
-		if ( !start_pos ) then break end
+	for i = 1, #str do
+		local start_pos, end_pos = string_find( str, sep, current_pos, withpattern )
+		if ( start_pos == nil ) then break end
```
GitSparTV, line 73: Any particular reason to change it to `== nil`? I'm not against it, just asking.

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -1,92 +1,76 @@
...
+	withpattern = !withpattern
 
 	local ret = {}
 	local current_pos = 1
 
-	for i = 1, string_len( str ) do
-		local start_pos, end_pos = string_find( str, separator, current_pos, !withpattern )
-		if ( !start_pos ) then break end
+	for i = 1, #str do
+		local start_pos, end_pos = string_find( str, sep, current_pos, withpattern )
```
GitSparTV, line 72: We can also localize `string_find` inside, but that's not a big deal.

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -96,255 +80,215 @@ function string.Explode(separator, str, withpattern)
...
---[[---------------------------------------------------------
-	Name: GetFileFromFilename( path )
-	Desc: Returns file with extension from path
-	Usage: string.GetFileFromFilename("garrysmod/lua/modules/string.lua")
------------------------------------------------------------]]
 function string.GetFileFromFilename( path )
-	if ( !path:find( "\\" ) && !path:find( "/" ) ) then return path end 
-	return path:match( "[\\/]([^/\\]+)$" ) or ""
+	if ( !string_find( path, "\\", 1, true ) && !string_find( path, "/", 1, true ) ) then return path end
+	return string_match( path, "[\\/]([^/\\]+)$" ) or ""
```
GitSparTV, line 109: What if `return string_match( path, "[\\/]([^/\\]+)$" ) or path`?
And remove line 108

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -96,255 +80,215 @@ function string.Explode(separator, str, withpattern)
...
-		return t .. pluralizeString( " second", t )
+		local t = math_floor( seconds )
+		return t .. ( t == 1 and " second" or " seconds" )
 	end
 
-	if ( seconds < 60 * 60 ) then
-		local t = math.floor( seconds / 60 )
-		return t .. pluralizeString( " minute", t )
+	-- 60 * 60
+	if ( seconds < 3600 ) then
```
GitSparTV, line 145: Side note: it's totally fine to write literal expressions. They are `consteval`

GitSparTV (changes requested):
Found some things I would consider to look at.
If you don't mind, code, I can PR in your PR to fix and fetch with upstream your PR to merge this PR to gmod. Or maybe you can set me as a collaborator in your fork.

garrysmod/lua/includes/extensions/string.lua
GitSparTV: It's intentional in your case and I agree with that

garrysmod/lua/includes/extensions/string.lua
Kefta: Didn't know that at the time I wrote this.

garrysmod/lua/includes/extensions/string.lua
Kefta: Will LuaJIT not optimise out the const local upval/treat it as a local?

garrysmod/lua/includes/extensions/string.lua
Kefta: Introducing type consistency into the codebase. Even though string.find will never return false, it's now more clear what the if-statement is actually checking.

garrysmod/lua/includes/extensions/string.lua
Kefta: I think the ability to alter string[1] is a too big of a break in abstraction to ignore. Faster access yes but if a dev really cared about saving cycles, they'd localise the funcs anyway.

garrysmod/lua/includes/extensions/string.lua
thegrb93: you mean like `string_gsub( str, "\226\128\16([89])", "\\\226\128\16%1" )`

GitSparTV: Yeah

garrysmod/lua/includes/extensions/string.lua
GitSparTV: Hm, apparently over-localizing introduces additional equality check, probably because upvalues has const attribute internally, while locals don't.

Over-localization is a trick I use for non-JIT code

PR #1410, opened 2017-09-10
Add table.FindMemberByKey( Table, key, val )

Tripperful:
Use case:

```Lua
local guys = {

	{
		age = 10,
		name = "John"
	},
	{
		age = 20,
		name = "Peter"
	},
	{
		age = 30,
		name = "Mike"
	},
}

print( table.FindMemberByKey( guys, "name", "Peter" ) )
```
Result:
```2       table: 0x2364fd78```

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -356,14 +356,14 @@ function table.ForceInsert( t, v )
 end
 
 --[[---------------------------------------------------------
-	Name: table.FindMemberByColumn( Table, columnName, columnValue )
-	Desc: Returns the first found entry which has value columnValue associated with key columnName
+	Name: table.FindMemberByKey( Table, key, val )
+	Desc: Returns the first found entry which has value "val" associated with key "key"
 -----------------------------------------------------------]]
-function table.FindMemberByColumn( Table, columnName, columnValue )
+function table.FindMemberByKey( Table, key, val )
 
 	for k, v in pairs( Table ) do
 		
-		if istable( v ) and v[columnName] == columnValue then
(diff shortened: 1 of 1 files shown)


dvdvideo1234:
Maybe nitpicking on mnemonics, but "GetMemberByColumn" might be better than "GetByMemberValue", because in Lua everything is a "value". Translates to "Retrieve the desired /member/ where its colum /name/ is /Peter/"

robotboy655:
I kinda find this function to be bad, BECAUSE what if there's 2 entries with the same "Name" or "Age" field? Which one should be returned then?

sannysc:
The first one it finds should be returned. This is already common behavior in the ```table``` library.

http://wiki.garrysmod.com/page/table/KeyFromValue
http://wiki.garrysmod.com/page/table/RemoveByValue

Tripperful:
Changes some things. Thanks for the feedback.

> **robotboy655**: ... if there's 2 entries with the same "Name" or "Age" field? Which one should be returned then?

Maybe it would then make sense to better rename it to table.FindMembersByColumn and make it return a table of all found keys? It would act like a "database query", returning a "res" table, which can be nil or a table containing one or more elements

sannysc:
Now this is even more confusing than before...There are no "columns" in tables. There are keys and there are values.

Tripperful:
Changed some things once again

Tripperful:
Hah, it's still rather confusing, the first thing I think of when I see such function name is "Oh, it probably just does ```return Table[key]```, what's the point..?". 

If someone has a good name for it - go ahead. Otherwise, if it will stay confusing, I will just close the PR. The community doesn't seem to like this proposal anyways.

PR #1176, opened 2016-05-15
Renamed string.StartWith to string.StartsWith

SomeSortOfDuck:
Added string.StartWith as an alias.

PR #1411, opened 2017-09-14
string.StartWith should be string.StartsWith

aStonedPenguin:
It's inconsistent with string.EndsWith and other languages that implement equivalent functions

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -298,7 +298,7 @@ function meta:__index( key )
 	end
 end
 
-function string.StartWith( String, Start )
+function string.StartsWith( String, Start )
 
    return string.sub( String, 1, string.len (Start ) ) == Start
 


neico:
Provide a backwards compatible alias, you most likely will break quite a few scripts with that change

PS: why aren't we using CI here yet? it's not like there aren't any linters for lua, even if a lot of stuff comes from the native side which can be polyfilled (and most likely automatically extracted).

Acecool:
And a lot of other functions should be: ( needle, haystack ) but are reversed and inconsistent. There are many issues to not having a coding standard which resolves these situations.

I agree, the function should be changed - unfortunately an alias would be required so I submit the following suggestion: Create an addon for the official version of Garry's Mod and call it Garry's Mod Deprecated Functions and Hotfixes - or something along those lines. Put the deprecated calls in, and allow them to exist for a year before removing them from the official folder. A year would give anyone more than enough time to update their code, and it would give server owners long enough to copy code over if an addon uses it and the author is no longer updating it if the code is obfuscated or otherwise closed...

sannysc:
Already denied.
#1176

PR #1412, opened 2017-09-18
TTT: Allow popups to be toggled with their hotkey

TheOpalium:
**QoL change**
Allows popups like the Traitor shop or the round end pop to be closed with their hotkey (_Default: C_) instead of using the mouse.
It's a small thing but it's a huge relief, really.

Can be changed in the settings menu, if somebody doesn't like (_but why the hell not lol_)

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
@@ -524,6 +524,12 @@ function CLSCORE:Reopen()
    end
 end
 
+function CLSCORE:Toggle()
+   if self.Panel and self.Panel:IsValid() then
+      self.Panel:SetVisible(not self.Panel:IsVisible())
+   end
+end
+
 local buff = ""
 local function ReceiveReportStream(len)
    local cont = net.ReadBit() == 1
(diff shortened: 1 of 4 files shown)


TheOpalium:
Also, apologies if the CVar is not located in the ideal place. It made sense to me to place it there, but you can move it as you see right.

garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -524,6 +524,12 @@ function CLSCORE:Reopen()
    end
 end
 
+function CLSCORE:Toggle()
+   if self.Panel and self.Panel:IsValid() then
```
sannysc: You should remove both ```if``` statements and use ```IsValid(self.Panel)``` instead.

garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -524,6 +524,12 @@ function CLSCORE:Reopen()
    end
 end
 
+function CLSCORE:Toggle()
+   if self.Panel and self.Panel:IsValid() then
+      self.Panel:SetVisible(not self.Panel:IsVisible())
```
sannysc: There's a function to do this. ```self.Panel:ToggleVisible()```

svdm:
To be honest I'd just remove the convar entirely, why disable it if there's no downside?

TheOpalium:
Honestly? I don't think it's needed either, but I was adding that in case anyone would prefer it as it is for some crazy reason. People don't always like these changes. 🤷‍♂️

PR #1415, opened 2017-09-26
Fix panel childrens remove when panel is removed

flydvorkin:
When panel is removed other child objects of this panel are not removed.

For example try:
menubar.Control:Remove()
and menu elements are stay and if click at its cause lua errors


robotboy655:
```menubar.Control:Remove()``` doesn't cause any Lua errors for me, the engine itself removes all child elements of a removed panel already.

PR #1416, opened 2017-10-03
Replace all self.Owner with self:GetOwner()

roboderpy:
https://github.com/Facepunch/garrysmod-issues/issues/3310


Kefta:
Might want to separate out TTT changes

PR #1417, opened 2017-10-03
Replace self.Owner with self:GetOwner() in base

roboderpy:


robotboy655:
Feels like a pointless change..

I know that it was said somewhere that you should use `self:GetOwner()` over `self.Owner`, but `self.Owner` is never going away and looks cleaner.

Kefta:
It's more efficient. self:GetOwner() calls __index once while self.Owner calls it an additional time instead of looking directly in the Entity metatable.

roboderpy:
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/entity.lua#L31

PR #1418, opened 2017-10-03
Replace self.Owner with self:GetOwner() in TTT

roboderpy:


svdm:
I was going to ask what the point of replacing it is, but then I noticed this is the first change I see in your PR:

    self:GetOwner() = data:GetEntity()

And I'm not going to waste any more time on this.

roboderpy:
updated

garrysmod/gamemodes/terrortown/entities/effects/teleport_beamdown.lua
```diff
@@ -8,7 +8,7 @@ local mid = 32
 local final_height = top
 
 function EFFECT:Init(data)
-   self.Owner = data:GetEntity()
+   self:GetOwner() = data:GetEntity()
```
ZehMatt: ??

Kefta:
@svdm self.Owner is deprecated. It only works because of an __index hack, and is slower.

PR #1419, opened 2017-10-10
Update the vector Lua extention to include various handy methods.

dvdvideo1234:


Kefta:
I would remove the nil check line, it will just lead to people not knowing when they sent in a bad value. Also, you can skip the Vector.Set line by using the Vector's copy-constructor. Ex. ``local v = Vector(self)``

dvdvideo1234:
If modifiers returned `*this` the constructor will become:

function meta:GetRotated(ang)
	return Vector(self):Rotate(ang)
end

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,164 @@ local meta = FindMetaTable( "Vector" )
...
+	ns, nv = self:Length(), vec:Length()
+	self:Mul(nv)
+	self:Add(vec:GetMul(ns))
+	return self
+end
+
+--[[---------------------------------------------------------
+Returns a copy of bisector of two vectors
+-----------------------------------------------------------]]
+function meta:Bisect(vec)
```
thegrb93: Typo

Kefta: Where?

thegrb93: Look 10 lines up

dvdvideo1234: Ten lines up is
```
Convets the vector to bisector of two vectors
function meta:Bisect(vec)
```

fruitwasp: `GetBisect` instead.

fruitwasp: GetBisect

dvdvideo1234: It seems correct o me: https://www.mathsisfun.com/geometry/bisect.html

thegrb93: It's still not fixed, dummy

robotboy655:
I feel like most of the stuff in here is useless and there are several problems with other stuff:
1) Do we really need stuff like `ToTable`, `ToAngle` (this one specially confusing because the name might make someone think its transforming a normalized/direction vector into an angle, which it does not), `GetSub/Mul/Add/Div`? `GetZero`? `Middle`?
2) You still have 2 Vector.Bisect methods
3) Your first vector.Bisect sets global variables.
4) Shouldn't ```GetNegate``` be ```GetNegated```? ```Normalize``` and `GetNormlized`
5) `GetMiddle` never uses its argument, and probably should be `GetCenter` or something
6) `GetResize` also never uses its argument, and should be `GetResized`

dvdvideo1234:
1. The idea is to change the data structure, but if you insist, I will remove them ;)
2. Thanks, fixed.
3. Thanks, fixed also
4. Fixed ;)
5. The `Middle` must use its argument for obtaining a [midpoint](https://en.wikipedia.org/wiki/Midpoint).
6. Thanks, also fixed ;)

Kefta:
SetEx should be added in the engine, not Lua. I requested it here https://github.com/Facepunch/garrysmod-requests/issues/515

Kefta:
I think these should be implemented in the engine for speed.

Kefta:
You should utilise vector:Unpack where you index x, y, and z to be more efficient. Also, using number indexing is slightly faster than string indexing (vec[1] vs vec.x).

dvdvideo1234:
> You should utilise vector:Unpack

Last time I tested at home Unpack was not pushed to production. I will try today as well.

> Also, using number indexing is slightly faster than string indexing (vec[1] vs vec.x)

Yes, this is true. [The official wiki says we must use X, Y, Z](https://wiki.facepunch.com/gmod/Vector). It is always better to stick to the documentation. ;)

Kefta:
That is an oversight of the wiki, I will change it. You should still use numerical indexing.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,176 @@ local meta = FindMetaTable( "Vector" )
...
+	local v = Vector(self)
+	return v:Bisect(vec)
+end
 
-	return Color( self.x * 255, self.y * 255, self.z * 255 )
+--[[---------------------------------------------------------
+Convets the vector to nagated version of itself
+-----------------------------------------------------------]]
+function meta:Negate()
+	self.x = -self.x
```
Kefta: Unpack here.

Kefta (changes requested):

Kefta:
Updated the wiki for the Vector and Angle pages on all possible ways to index them.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,176 @@ local meta = FindMetaTable( "Vector" )
 -- Nothing in here, still leaving this file here just in case
 
 --[[---------------------------------------------------------
-	Converts Vector To Color - alpha precision lost, must reset
+Converts Vector to Color - alpha precision lost, must reset
 -----------------------------------------------------------]]
-function meta:ToColor( )
+function meta:ToColor()
+	return Color(self.x * 255, self.y * 255, self.z * 255)
```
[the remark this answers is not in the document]
dvdvideo1234: `Color(x * 255, y * 255, z * 255)`

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,176 @@ local meta = FindMetaTable( "Vector" )
...
 -----------------------------------------------------------]]
-function meta:ToColor( )
+function meta:ToColor()
+	return Color(self.x * 255, self.y * 255, self.z * 255)
+end
+
+--[[---------------------------------------------------------
+Converts Vector to Table - removes all metrods from the copy
+-----------------------------------------------------------]]
+function meta:ToTable(arr)
```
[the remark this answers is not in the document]
dvdvideo1234: Done.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,176 @@ local meta = FindMetaTable( "Vector" )
...
+	if(arr) then return {self.x, self.y, self.z} end
+	return {x = self.x, y = self.y, z = self.z}
+end
+
+--[[---------------------------------------------------------
+Converts Vector to Angle - data is copied
+-----------------------------------------------------------]]
+function meta:ToAngle()
+	return Angle(self.x, self.y, self.z)
```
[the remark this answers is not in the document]
dvdvideo1234: Done.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done. Added single component trigger. Id only one is needed to be negated. When no arguments, all are negated.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
-function meta:ToColor( )
+function meta:ToColor()
+	local x, y, z = self:Unpack()
+	return Color(x * 255, y * 255, z * 255)
+end
+
+--[[---------------------------------------------------------
+Converts Vector to Table - removes all metrods from the copy
+-----------------------------------------------------------]]
+function meta:ToTable(arr)
```
Kefta: This argument is no longer needed

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
+	local x, y, z = self:Unpack()
+	return {x = x, y = y, z = z}
+end
+
+--[[---------------------------------------------------------
+Converts Vector to Array - removes all metrods from the copy
+-----------------------------------------------------------]]
+function meta:ToArray()
+	local x, y, z = self:Unpack()
+	return {x, y, z}
```
Kefta: ``return {self:Unpack()}``

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
+end
+
+--[[---------------------------------------------------------
+Converts Vector to Angle - data is copied
+-----------------------------------------------------------]]
+function meta:ToAngle()
+	local x, y, z = self:Unpack()
+	return Angle(x, y, z)
```
Kefta: ``return Angle(self:Unpack())``

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
+	local v = Vector(self)
+	v:Add(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Adds a unpacked x, y and z to a vector
+-----------------------------------------------------------]]
+function meta:AddEx(x, y, z)
+	self[1] = self[1] + x
```
Kefta: I think ``*Ex`` should be changed to ``*Unpacked`` for naming consistency. Also use ``Unpack`` and ``SetUnpacked`` here.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
+	self[2] = self[2] + y
+	self[3] = self[3] + z
+	return self
```
Kefta: I wouldn't return here for consistency with other vector modifying methods.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetBisected(...)
+	local v = Vector(self)
+	v:Bisect(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Convets the vector to nagated version of itself
+-----------------------------------------------------------]]
+function meta:Negate(bx, by, bz)
```
Kefta: I think these arguments are pointless. Negate should negate the entire vector and not just parts - that can be done inline in code if necessary instead of inducing a bunch of type checks.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,247 @@ local meta = FindMetaTable( "Vector" )
...
+end
+
+--[[---------------------------------------------------------
+Applies new basis to a vector by the three axises of an angle
+-----------------------------------------------------------]]
+function meta:Basis(ang)
+	local x = self:Dot(ang:Forward())
+	local y = self:Dot(ang:Right())
+	local z = self:Dot(ang:Up())
+	self[1], self[2], self[3] = x, y, z
```
Kefta: ``SetUnpacked`` here.

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Ok !

dvdvideo1234: Done with Add/Sub*

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: `Bisect` is a modifier also. As I remember `a: Add(b)` on vectors does not return anything. Sorry. This due to my personal coding style for returning `*this` in C++.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Ok. Thanks !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done !

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,262 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetNegated(...)
+	local v = Vector(self)
+	v:Negate(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Convets the vector to nagated version of itself
+-----------------------------------------------------------]]
+function meta:Nudge(num, vec)
```
Kefta: What are the use cases for this function? Where does the term "nudge" from?

I think it would just be better to write this inline as the multiplied vector will probably be reused in most cases.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,262 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetBasis(...)
+	local v = Vector(self)
+	v:Basis(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Projects the vector in to an axis defined by the second vector
+-----------------------------------------------------------]]
+function meta:Project(vec)
```
Kefta: It would be better to copy the original vector instead of using GetNormalized than setting it after finding the dot product.
```Lua
local v = Vector(self)
self:Normalize()
self:Mul(v:Dot(self))
```

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,262 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetProject(...)
+	local v = Vector(self)
+	v:Project(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Retrieves the angle between two vectors from 0 to math.pi
+-----------------------------------------------------------]]
+function GetAngle(vec)
```
Kefta: Missing ``meta:`` in front. This should probably renamed to AngleBetween to prevent confusion with vec:Angle(). Also, this could be more efficiently implemented as

```Lua
return math.acos(self:Dot(vec) / math.sqrt(self:LengthSqr() * vec:LengthSqr()))
```

using the formula ``cos(a) = dot(x, y)/(|x|*|y|)``. While you're at it, you should also add an additional optional argument to specify a normal vector representing the plane the vectors lie in. This can be written as

```Lua
return math.atan2(self:Determinant(vec, normal:GetNormalized()), self:Dot(vec))
```

using the formula ``tan(a) = determinant(x, y, n) / dot(x, y)``. Most library functions in GMod don't use type checking so you can get away with

```Lua
function meta:AngleBetween(vec, n)
	if (n == nil) then
		return ...
	end

	return ...
end
```

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,262 @@ local meta = FindMetaTable( "Vector" )
...
+	local y = self:GetSub(x)
+	local nx = x:Dot(vec)
+	local mx = nx/math.abs(nx)
+	return math.atan2(y:Length(), x:Length()*nx)
+end
+
+--[[---------------------------------------------------------
+Calculates 3x3 determinant by using other two vectors as rows
+-----------------------------------------------------------]]
+function meta:GetDeterminant(vec2, vec3)
```
Kefta: I would just make the name ``Determinant`` to be consistent with ``Vector:Length``.

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,262 @@ local meta = FindMetaTable( "Vector" )
 -- Nothing in here, still leaving this file here just in case
 
 --[[---------------------------------------------------------
-	Converts Vector To Color - alpha precision lost, must reset
+Converts Vector to Color - alpha precision lost, must reset
 -----------------------------------------------------------]]
-function meta:ToColor( )
+function meta:ToColor(num)
+	local x, y, z = self:Unpack()
+	return Color(x * 255, y * 255, z * 255, (num or 0) * 255)
```
Kefta: This should be ``num == nil and 1 or num`` (the ``== nil`` to prevent people accidentally sending in ``false`` as the argument and never noticing the difference, but more importantly having 1 as the default so the alpha defaults to 255).

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,237 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetBisected(...)
+	local v = Vector(self)
+	v:Bisect(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Convets the vector to nagated version of itself
+-----------------------------------------------------------]]
+function meta:Negate()
```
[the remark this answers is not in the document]
dvdvideo1234: Ok.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: A short term for `origin = origin + coefficient * direction`

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,262 @@ local meta = FindMetaTable( "Vector" )
...
+	local vn = vec:GetNormalized()
+	local nm = self:Dot(vn)
+	self:Set(vn)
+	self:Mul(nm)
+end
+
+--[[---------------------------------------------------------
+Projects the vector in to an axis defined by the second vector
+-----------------------------------------------------------]]
+function meta:GetProject(...)
```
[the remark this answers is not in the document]
dvdvideo1234: This must modify `self` this is different than `GetProject` which makes a copy. ;)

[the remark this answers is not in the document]
dvdvideo1234: This is done for better support where `A:getSomething` utilizes `A:Someting` and returns a copy ;)

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: No, because it need Normalized of the argument as it defines the base X coordinate axis. You have to store the argumant in `self` rather.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Ok. Fixed.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Yes, thank you. Fixed ;)

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: That sounds neat. Thank you !

dvdvideo1234: Never noticed the second parameter but it sounds great ;)

garrysmod/lua/includes/extensions/vector.lua
Kefta: Yeah you're correct, mixed those up.

garrysmod/lua/includes/extensions/vector.lua
Kefta: Ah, so this is the same as [VectorMA](https://github.com/VSES/SourceEngine2007/blob/43a5c90a5ada1e69ca044595383be67f40b33c61/src_main/public/mathlib/mathlib.h#L269). Like Negate, this should be binded from the engine then instead of Lua.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done !

```
local a = Vector(3,3)
local b = Vector(3,0)
local c = Vector(0,0,6)
local d = a:GetProject(b)
print(d, a:AngleBetween(b,c))
```

dvdvideo1234: VEC{3,0,0}	-0.78539816339745

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Ah, never new it has to be a dedicated engine function. I use it mery much in my addons, So I thought it must be needed here ;) Thank you !

dvdvideo1234: I will Remove Nudge. This is the exact thing I am talking about.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,237 @@ local meta = FindMetaTable( "Vector" )
...
+-----------------------------------------------------------]]
+function meta:GetNegated(...)
+	local v = Vector(self)
+	v:Negate(...)
+	return v
+end
 
-	return Color( self.x * 255, self.y * 255, self.z * 255 )
+--[[---------------------------------------------------------
+Convets the vector to nagated version of itself
```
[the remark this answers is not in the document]
dvdvideo1234: Yes, thanks ! Fixed.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,237 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetAdd(...)
+	local v = Vector(self)
+	v:Add(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Adds a unpacked x, y and z to a vector
+-----------------------------------------------------------]]
+function meta:AddUnpacked(x, y, z)
```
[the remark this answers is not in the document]
dvdvideo1234: Thank you. Fixed :)

garrysmod/lua/includes/extensions/vector.lua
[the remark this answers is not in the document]
dvdvideo1234: Written that way you only need to change the code in `meta:Project(...)` to adjust the behavior.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
 -- Nothing in here, still leaving this file here just in case
 
 --[[---------------------------------------------------------
-	Converts Vector To Color - alpha precision lost, must reset
+Converts Vector to Color - alpha precision lost, must reset
 -----------------------------------------------------------]]
-function meta:ToColor( )
+function meta:ToColor(num)
+	local a = (num == nil and 1 or num)
```
Kefta: This doesn't need to be a local - it can be placed inline in the Color call.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+	v:Add(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Adds a unpacked x, y and z to a vector
+-----------------------------------------------------------]]
+function meta:AddUnpacked(x, y, z)
+	local vx, vy, vz = self:Unpack()
+	self:AddUnpacked(vx + x, vy + y, vz + z)
```
Kefta: I think you meant to do SetUnpacked here, this is infinite recursion right now.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+	v:Sub(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Subtracts a unpacked x, y and z from a vector
+-----------------------------------------------------------]]
+function meta:SubUnpacked(x, y, z)
+	local vx, vy, vz = self:Unpack()
+	self:SubUnpacked(vx - x, vy - y, vz - z)
```
Kefta: Same here.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetNegated(...)
+	local v = Vector(self)
+	v:Negate(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Returns a copy of the nudged vector
+-----------------------------------------------------------]]
+function meta:GetNudge(...)
```
Kefta: This needs to be removed if Nudge is.

garrysmod/lua/includes/extensions/vector.lua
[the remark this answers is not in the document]
Kefta: It may be less duplicitous to have ``Get*`` call down to the non-returning alternatives, but it's slower. I guess it's not that big of a deal anyway.

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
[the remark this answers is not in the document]
Kefta: It may be less duplicitous to have Get* call down to the non-returning alternatives, but it's slower. I guess it's not that big of a deal anyway.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+function meta:GetBisected(...)
+	local v = Vector(self)
+	v:Bisect(...)
+	return v
+end
+
+--[[---------------------------------------------------------
+Convets the vector to negated version of itself
+-----------------------------------------------------------]]
+function meta:Negate()
```
Kefta: This should be removed for an engine implementation - I would go ahead and also remove GetNegated until the bindings are made so this can be used on people's servers in the meantime.

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Done !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Yes ! Thanks !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Same. Yes. Done !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Removed for you to do engine implementation.

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Ok. Done.

Kefta:
Looking good. Last thing to clean up is the style (spaces around parentheses and and brackets, C-styled logical operators) to match the codebase, and to change the name of the PR as this is no longer one method.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,230 @@ local meta = FindMetaTable( "Vector" )
...
+	local vs = Vector(self)
+	self:Set(vec)
+	self:Normalize()
+	self:Mul(vs:Dot(self))
+end
+
+--[[---------------------------------------------------------
+Projects the vector in to an axis defined by the second vector
+-----------------------------------------------------------]]
+function meta:GetProject(...)
```
Kefta: Shouldn't this be GetProjected to match GetBisected and GetResized?

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Yes, it should. Fixed.

Kefta: It's not

dvdvideo1234: Ah.... Noticed the code bas is `Add` -> `GetAdd`

dvdvideo1234: Reverted. Also Added ray intersection fraction calculator

garrysmod/lua/includes/extensions/vector.lua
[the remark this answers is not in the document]
dvdvideo1234: Yes. The performance difference is negligible and you only need to change one method in the future instead of both. If `Project` works, then `GetProject` works too ;)

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: On the other hand should I rename `GetAdd` `GetSub` to `GetAdded` `GetSubbed` .... :confused:

Kefta: I think all Get methods except for Midpoint and Add, Sub Mul, Div should end with ``ed`` since we currently have ``Unpack`` and ``SetUnpacked``. So ``GetProjected``, ``GetBisected``, ``GetResized``, and ``GetRotated``.

dvdvideo1234: Ah. Makes sense. Thank you! Updated !

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+The first origin /vO1/ is actually the current object ( self )
+Takes:
+	vD1 --> Direction of the first ray
+	vO2 --> Position origin of the second ray
+	vD2 --> Direction of the second ray
+Returns:
+	f1 --> Intersection fraction of the first ray
+	f2 --> Intersection fraction of the second ray
+-----------------------------------------------------------]]
+function meta:IntersectFraction( vD1, vO2, vD2 )
```
Kefta: You have a bug with:
```Lua
local dn = ( dx:Length() ) ^ 2
if( dn <= 0 ) then return nil end
```
since x^2 will always be >= 0.

I don't recognise this formula - what is the "intersection fraction"? Since this takes two independent vectors and directions, I think this would be better suited for the ``math`` or ``util`` libraries.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+	local x = self:Dot( ang:Forward() )
+	local y = self:Dot( ang:Right() )
+	local z = self:Dot( ang:Up() )
+	self:SetUnpacked( x, y, z )
+end
+
+--[[---------------------------------------------------------
+Applies new basis to a vector by the three axises of an angle
+-----------------------------------------------------------]]
+function meta:GetBasised( ... )
```
Kefta: "Basised" isn't a word, should still be ``GetBasis``.

Kefta (changes requested):

garrysmod/lua/includes/extensions/vector.lua
Kefta: Whoops, forgot to mention to leave Basis alone, as well.

dvdvideo1234: Ok, Reverted! Thanks.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -3,10 +3,257 @@ local meta = FindMetaTable( "Vector" )
...
+	local v = Vector( self )
+	v:Project( ... )
+	return v
+end
+
+--[[---------------------------------------------------------
+Retrieves the angle between two vectors from 0 to math.pi
+-----------------------------------------------------------]]
+function meta:AngleBetween( vec, nrm )
+	if( nrm == nil ) then
```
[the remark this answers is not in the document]
dvdvideo1234: Done !

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Should I remove it then ?

garrysmod/lua/includes/extensions/vector.lua
dvdvideo1234: Fixed. Thanks !

garrysmod/lua/includes/extensions/vector.lua
Kefta: Can you link to a mathematical/code source from where you got this formula from?

dvdvideo1234: Em, I know it from the university ... Maths part 2 analytical geometry. But I am sure that you can find it in the web also ;)

dvdvideo1234: It can be done with `== 0`, though I like <= better :D. Anywayn` is actually `|dx|^2` the length of the cross procuct of the directions. Maybe it is not needed in the vector library .Removed ;)

Kefta:
I think this is finally in a good place. There's some additional optimisations that can be done to save on Vector objects in ``Get*`` alternatives instead of just calling the original functions with a copy, but those can be done post-merge. Thanks for staying patient with all the corrections.

Kefta (approved):

dvdvideo1234:
Thank you too for adding all these ;)

dvdvideo1234:
Thoughts @thegrb93 , @fruitwasp is this any good ?

thegrb93:
The ToAngle needs to be removed as it will be confused with Vector:Angle

Kefta:
I think ``<operation>Unpacked`` should be added in the engine - it would be more efficient than creating a vector object simply to perform an operation on all three digits, and even more efficient than setting each component manually. They could even be made into optional args to be able to do two component operations in one function call instead of 4 (2 __newindex, 2 __index). I'll add them to my request on https://github.com/Facepunch/garrysmod-requests/issues/1442.

Now looking at it, ``ToAngle`` should be removed not for the confusion with ``Angle``, but rather the uselessness of the operation: I can think of no cases where you would go from a Vector to Angle unless it were to convert an impulse force, but that would require performing operations on the component anyway.

dvdvideo1234:
The APIs with `Unpacked` will work very nice for me, as I will remove like 100+ lines of code from my addon just for using these. Sometimes a copy of the vector of the executed operation is needed,  and for this `Get<Operation>Unpacked` is used.

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -259,3 +259,23 @@ function meta:AngleBetween( vec, nrm )
...
+--[[---------------------------------------------------------
+Recalculates the orthogonality of an up vector according to the
+other vector given as forward direction direction. The call
+changes the self vector assigned as an up direction
+ * self > The up direction being orthogonalized
+ * vF   > The forward direction for the orthogonalization process
+ * bN   > When set to true, normalizes the three unit vectors
+ * vR   > The function returns the right vector on success
+-----------------------------------------------------------]]
+function meta:Orthogonal(vF, bN)
```
Kefta: I don't think this should be part of the lib - it's arbitrary for the up vector to be the self arg. Maybe it should be a global function or thrown into util.

dvdvideo1234: Should I remove it then?

Kefta: I mean, it's your PR, it's ultimately up to Rubat. I just think it's vague implemented as a vector class method personally.

dvdvideo1234: You are probably right ;)

dvdvideo1234:
@Kefta @thegrb93,

Is it better for data conversion to just [run this](https://github.com/Facepunch/garrysmod/blob/0b20ee02bd0558f1841317fdb303fbfb19b83f63/garrysmod/lua/includes/extensions/util.lua#L106). It relies on the object factory to create an instance from any type. I see the conversion from a string is not needed. But I wonder, what is the difference between:

1. `Vector(1,2,3)`  This is the regular call
2. `Vector({1,2,3})` Table-array with 1,2,3
3. `Vector({x=1,y=2,z=3})` Table-hash with X,Y,Z
4. `Vector("1 2 3")` This according to the wiki parses the string to a vector. Notice the space.
5. `Vector("1,2,3")` What will happen when I give it another delimiter. Notice the comma. :P
6. `Vector(true,true,true)` This I think should produce `Vector(1,1,1)` dunno..

I cannot think of any suitable usage converting:
1. `vector` from `function`,
2. `vector` from `userdata`
3. `vector` from a `thread`.

Kefta:
2 doesn't work but can be achieved with ``Vector(unpack({1, 2, 3})``. 3, 5, and 6 don't work. You shouldn't adjust your functions at all for that behaviour - let the Vector constructor handle all type coercions. This PR is fine in its current state.

dvdvideo1234:
@Kefta Thanks!

Requested in https://github.com/Facepunch/garrysmod/pull/1619

dvdvideo1234:
@Kefta 
I see that [there is a  new wiki](https://wiki.facepunch.com/gmod/Angle:SetUnpacked). does this mean that the PR #1419 is merged already?

robotboy655:
> @ Kefta
> I see that [there is a new wiki](https://wiki.facepunch.com/gmod/Angle:SetUnpacked). does this mean that the PR #1419 is merged already?

No, I have added some methods that were requested [here](https://github.com/Facepunch/garrysmod-requests/issues/515)

CornerPin:
What is the point of having `GetAdd`, `GetSub`, `GetMul` and `GetDiv` when we already have `__add`, `__sub`, `__mul` and `__div`? The special metamethods do the same things and must be more efficient.

dvdvideo1234:
@CornerPin 

These are here for language richness. I can admit, they do the same stuff like the other ones, but I've never measured it.
The richest one language is, the better as you can so the same thing by utilizing different lines of code.
They are here for language richness mostly, for the users who prefer `local c = a:GetAdd(b)` more than `local c = a + b`.

dvdvideo1234:
@robotboy655 

Should I remove the `VECTOR:Get%` part of the PR as @CornerPin sais we can use `+, - * /` ?

I can leave only the methods that return nothing ?

garrysmod/lua/includes/extensions/vector.lua
```diff
@@ -2,11 +2,267 @@
 local meta = FindMetaTable( "Vector" )
 
 --[[---------------------------------------------------------
-	Converts Vector To Color - alpha precision lost, must reset
+Converts Vector to Color - alpha precision lost, must reset
 -----------------------------------------------------------]]
-function meta:ToColor()
+function meta:ToColor( num )
```
Grocel, line 7: This conflicts with an pre-existing meta:ToColor() in your PR.

dvdvideo1234:
Thanks @Grocel 

Updated: https://github.com/Facepunch/garrysmod/pull/1419/commits/9ceb4563a3ed545d9033bbf87eb512cb185db850

Grocel (approved):

Grocel:
You should give your PR a full revisit, though. Maybe close this one re-redo it from most up to date base GMod files from scratch. Some of the requested functions are already part of the game.

PR #1423, opened 2017-10-24
Fix DFrame buttons height

Tripperful:
If you have ever noticed, the buttons in DFrame (Close, Maximize and Minimize) are taller than they appear.
They are painted 24 pixels tall and seem to be aligned with the dark gray part of the header (which is 24 pixels tall), but they are actually not. These buttons are 31 pixels tall, that's why if you hover or click the frame just below those buttons they get triggered.

This PR makes these buttons 24 pixels tall to be aligned with the draggable header of the frame and adjusts the default skin to work correctly with this change.

### DIFF
--- a/garrysmod/lua/skins/default.lua
+++ b/garrysmod/lua/skins/default.lua
@@ -123,24 +123,24 @@ SKIN.tex.Tab_Bar					= GWEN.CreateTextureBorder( 128, 352, 127, 31, 4, 4, 4, 4 )
 
 SKIN.tex.Window = {}
 
-SKIN.tex.Window.Normal			= GWEN.CreateTextureBorder( 0, 0, 127, 127, 8, 32, 8, 8 )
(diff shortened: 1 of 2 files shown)


Tripperful:
In reality DFrame needs some more serious refactoring, somebody should make use of other buttons (Minimize and Maximize), I have also successfully added Set/GetHeaderHeight with proper elements scaling/margin adjustment (literally just made a use of a variable instead of hardcoded 24/31 constants everywhere).

Maybe it would also be a good idea to make the header a child panel of the whole frame, make it a DragBase and dock it and the buttons inside it, add ShowMinimizeButton and ShowMaximizeButton along with ShowCloseButton which is for now responsible for all three. 

Minimization and maximization would be easy enough to implement in Windows-like manner. 
Minimization would make windows thin in height and width (like 128x24 or whatever) and stack such minimized windows at the bottom of the screen/parent. 
Maximization would remember the current window size (to restore it later) and make the window fill the whole screen/parent. 

But these are pretty serious changes, not sure if they should be done.

PR #1424, opened 2017-11-02
[TTT] better cheating prevention for weapon dormancy

meepdarknessmeep:
will stop all clients besides owner to know the weapon even exists until it's pulled out and in pvs of others


Kefta:
Is there a reason why this isn't done in the engine already?

meepdarknessmeep:
because of unnecessary limitations that would not be needed in gamemodes that aren't ttt?

meepdarknessmeep:
think i might rewrite this a bit better based on feedback

PR #1427, opened 2017-11-08
Added volume to GameDetails

nurdism:
I, like many others, don't like having loading music when joining a server. I've added a simple way for web developers to get the preferred volume. I've added it to the end of GameDetails, so it won't break any existing pages out there. Just a nice thing to have for the peeps that don't like loading music.

### DIFF
--- a/garrysmod/lua/menu/loading.lua
+++ b/garrysmod/lua/menu/loading.lua
@@ -6,8 +6,6 @@
 --
 --=============================================================================--
 
-CreateClientConVar( "cl_loading_volume", "1.0", true, false )
-
 g_ServerName	= ""
 g_MapName		= ""
 g_ServerURL		= ""
@@ -253,10 +251,7 @@ function GameDetails( servername, serverurl, mapname, maxplayers, steamid, gamem
 	g_SteamID		= steamid
 	g_GameMode		= gamemode
 
-	local volume = 1.0
-	if ( ConVarExists( "cl_loading_volume" ) ) then
-		volume = GetConVar( "cl_loading_volume" ):GetFloat()
-	end
+	local volume = GetConVar( "volume" ):GetFloat()
 
 	MsgN( servername )
 	MsgN( serverurl )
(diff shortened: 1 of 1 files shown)


robotboy655:
I see a few problems with this:
1) There is no in-game slider or ANYTHING to tell the user that this convar exists and there is no way for average developer to know this either
2) I have high doubts anyone will go and change their loading screens to support this, there's basically no reason for people to do this because if they put music on their loading screen, they don;t want it turned off.

nerdism:
1. There's a lot of game options that don't have a slider or toggle, and that's what the wiki is for. 
2. I beg to differ, there are tons of communities that sate in the loading screen itself, if you don't like the music, tab out and mute awesomium. Having the option to get the preferred volume would be great. Even if the web developer decides to ignore it. 
3. This is a very wanted feature. Just google `gmod loading mute` and there's pages and pages of people asking the same thing.

ExtReMLapin:
why is sound on loading screen even supported

Lunaversitay:
I love the idea but like robogirl said there's just no way people are gonna update their loading screens, not even half of the people will know about it due to the fact they usually use github code (such as marcuz's loading screen) for theirs.

also lapin it's supported cause loading screens are directed to websites, so you can do whatever you want as long as awesomium doesn't outdate it by 2000 years :v

ExtReMLapin:
Yeah but we could kill some awesomium features for the loading screen

nerdism:
Yeah, but would it hurt to have the feature, even if its rarely used, hell it would take like < 5 lines of code to update any loading screen.

kanalumaddela:
Is it possible to use the volume music volume set in the audio tab in the options and pass that value? Also to address robot's 2nd point, I don't really think there would be major changes to the source of *most* loading screens. If anything they can just add the `volume` argument to `GameDetails()` and assign it to their existing variable which is global, right?

nerdism:
@kanalumaddela bring up a good point on **music volume** part. Does that slider affect anything in game? I know it's convar is `volume`.

robotboy655:
Consider this merged.
https://github.com/Facepunch/garrysmod/commit/2637ea029a051b88e3526b999c99acdd01c5240f

Had to do it manually due to conflicts, didn't realize you changed it to the built in volume convar which I am fine with.

PR #1428, opened 2017-11-15
Add min, max args to Vector/AngleRand

Kefta:
```
function VectorRand( min --[[= -1]], max --[[= 1]] )
function AngleRand( min --[[-90 pitch, -180 yaw/roll]], max --[[= 90 pitch, 180 yaw/roll]] )
```

Maintains previous behaviour, but the arguments now give forced mins/maxs into math.Rand. This also directly matches the behaviour of RandomVector and RandomAngle in the mathlib.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -90,13 +90,13 @@ function PrintTable( t, indent, done )
 end
 
 function VectorRand( min --[[= -1]], max --[[= 1]] )
-	min = min or -1
-	max = max or 1
+	min = min || -1
+	max = max || 1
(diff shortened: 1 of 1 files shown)


robotboy655:
Kinda weird that one function does this:
```
min = min or -1
max = max or 1
```
while the other one does this:
```
math.Rand( min or -90, max or 90 ), math.Rand( min or -180, max or 180 ), math.Rand( min or -180, max or 180 )
```
Should be either one or the other for both, shouldn't it?

Kefta:
It's because of maintaining backwards compatibility with the function's old defaults.

PR #1432, opened 2017-12-12
Add rounding to Angle and Vector

Tripperful:
- Add Angle:GetRound( decimals )
- Add Angle:GetFloor()
- Add Angle:GetCeil()
- Add Vector:GetRound( decimals )
- Add Vector:GetFloor()
- Add Vector:GetCeil()
- Add Vector:GetRotated( rotation ), a brother of Vector:GetNormalized()

### DIFF
--- a/garrysmod/lua/includes/extensions/angle.lua
+++ b/garrysmod/lua/includes/extensions/angle.lua
@@ -20,23 +20,23 @@ end
 --[[---------------------------------------------------------
 	Returns a rounded angle
 -----------------------------------------------------------]]
-function meta:Round( decimals )
+function meta:GetRound( decimals )
 
-	return Angle( math.Round( self[1], decimals ), math.Round( self[2], decimals ), math.Round( self[3], decimals ) )
+	return Angle( math.Round( self.p, decimals ), math.Round( self.y, decimals ), math.Round( self.r, decimals ) )
 end
 
 --[[---------------------------------------------------------
 	Returns a rounded down angle
 -----------------------------------------------------------]]
-function meta:Floor()
+function meta:GetFloor()
 
-	return Angle( math.floor( self[1] ), math.floor( self[2] ), math.floor( self[3] ) )
+	return Angle( math.floor( self.p ), math.floor( self.y ), math.floor( self.r ) )
 
 --[[---------------------------------------------------------
 	Returns a rounded up angle
 -----------------------------------------------------------]]
-function meta:Ceil()
+function meta:GetCeil()
 
-	return Angle( math.ceil( self[1] ), math.ceil( self[2] ), math.ceil( self[3] ) )
+	return Angle( math.ceil( self.p ), math.ceil( self.y ), math.ceil( self.r ) )
 
 end
--- a/garrysmod/lua/includes/extensions/angle.lua
+++ b/garrysmod/lua/includes/extensions/angle.lua
@@ -16,3 +16,27 @@ function meta:SnapTo( component, degrees )
 	return self
 
 end
+
+--[[---------------------------------------------------------
+	Returns a rounded angle
+-----------------------------------------------------------]]
+function meta:Round( decimals )
+
+	return Angle( math.Round( self[1], decimals ), math.Round( self[2], decimals ), math.Round( self[3], decimals ) )
+end
+
+--[[---------------------------------------------------------
+	Returns a rounded down angle
+-----------------------------------------------------------]]
+function meta:Floor()
+
+	return Angle( math.floor( self[1] ), math.floor( self[2] ), math.floor( self[3] ) )
+
+--[[---------------------------------------------------------
+	Returns a rounded up angle
+-----------------------------------------------------------]]
+function meta:Ceil()
+
+	return Angle( math.ceil( self[1] ), math.ceil( self[2] ), math.ceil( self[3] ) )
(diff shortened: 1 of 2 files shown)


sannysc:
These function names sound like they modify the existing object, but instead, they create a new one. And another thing, you should be using the named fields on the objects instead of ```1``` ```2``` ```3```.

```x``` ```y``` ```z``` for vectors
```p``` ```y```  ```r``` for angles.

**Comment by Kefta** [comment]
I don't see a point of these being in the core library either. Also, GetRotated has already been done in https://github.com/Facepunch/garrysmod/pull/1419

**Comment by Tripperful** [comment]
@sannysc
I was thinking if I should make them modify the original or return a new object.
I agree that such naming is inconsistent with vector/angle normalization/rotation functions naming, though there are not so many functions to talk about consistency. I work with vectors a lot and almost in all cases it's more handy to return a new object. First of all - this way you can do chains like 
```Lua
----------------------------------------------------------------
local gridPos = ent:GetPos():Round()
-- vs --
local gridPos = Vector( ent:GetPos() )
gridPos:Round()
----------------------------------------------------------------
----------------------------------------------------------------
----------------------------------------------------------------
render.DrawSprite( ent:GetPos():Floor(), ... )
-- vs --
local spritePos = ent:GetPos()
spritePos:Floor()
render.DrawSprite( spritePos, ... )
----------------------------------------------------------------
```
etc.

Of course you can do "return self" in functions which modify the original object, but that leads you to a habit to use such chains with these functions, which in many cases can lead to unintended side affects (modification of objects you were not trying to). In case of GetPos it wouldn't have side affects since GetPos returns a new object which you can modify without being afraid to move the original entity, but take the following example into consideration
```Lua
local points = { Vector( 1.1, 2.2, 3.3 ), ... }
for i, point in ipairs( points ) do DrawPoint( point:Ceil(), ... )
```
In this case, ceiling would change contents of the original points table, but it's not intentional. To avoid this, you'd need to explicitly create a new vector, copying the original one, and only then rounding it or whatever. It clutters the code and in my opinion is just unnecessary.

In my opinion, changing the original value and returning it should be a solution in an environment where you have getters returning a copy of object's field for you to work with, but while in GLua we usually work with tables, where using metatables (for creating getters and stuff) is an overkill for most simple tasks, such approach is not applicable.

GetNormalized appeared once not "just because", but because it is handy to use. I want these new functions to be also handy, I don't think creating 6 more functions (like GetRounded, etc.) is necessary for any reason other than following an arguable convention (yeah, maybe micro-optimization in some cases, but it's vice versa for other cases)

About vectors indexing
I don't see a reason why I should avoid using numerical indexes for vectors while Garry's Mod deliberately allows it. For angles it's even more fun, you can use 1,2,3/p,y,r/x,y,z. It's handy in some cases, for example
```Lua
for i = 1, 3 do
	if pos[i] > max[i] then max[i] = pos[i] end
	if pos[i] < min[i] then min[i] = pos[i] end
end
```
I have chosen numerical indexes here because they should be a tiny bit faster than string literals (I hope?)

@Kefta
My bad, I didn't make sure GetRotated wasn't already proposed. In any case, it's not merged, so I don't think that's much of a deal

Sorry for the long post, but I wanted to explain myself clearly (I doubt I succeeded though lol). If you have arguments against - I can change my mind

sannysc:
>I was thinking if I should make them modify the original or return a new object.
I agree that such naming is inconsistent with vector/angle normalization/rotation functions naming, though there are not so many functions to talk about consistency. I work with vectors a lot and almost in all cases it's more handy to return a new object.

The behavior doesn't need to be changed. The name is what doesn't make any sense because it doesn't match the behavior. It is much easier to just change the name. You yourself agree that it is inconsistent. It doesn't matter how few functions there are to compare it with. If it's inconsistent, it's inconsistent. That doesn't change whether it is inconsistent with 3 functions or 30 functions.

>About vectors indexing
I don't see a reason why I should avoid using numerical indexes for vectors while Garry's Mod deliberately allows it. For angles it's even more fun, you can use 1,2,3/p,y,r/x,y,z. It's handy in some cases, for example

You should avoid using them because it doesn't make sense to people who are reading the code. What the hell is ```[2]``` in the context of a vector/angle? You might know, I might know, but it isn't clear what is going on because it looks like some random number.

Those numbers [aren't even documented.](http://wiki.garrysmod.com/page/Category:Vector) It's not documented for [angles](http://wiki.garrysmod.com/page/Category:Angle) either. There is absolutely no reason to use those numbers over the letters. There is no universe in which ```[2]``` is easier to understand to a reader than ```y``` when dealing with vectors and angles.

They are functionally the same thing, except **one of them is easier for readers to understand.** Are you intentionally making your code more difficult to understand? Because that's exactly what you're doing.

Tripperful:
@sannysc Is it alright like this or should I change it to GetRounded, GetFloored, GetCeiled?

sannysc:
In my opinion, yes, but I don't want to be picky. If the core devs think it should be changed to something else, I'm sure they'll ask.

dvdvideo1234:
In fact, methods that modify the original OOP instance should **always** return `*this` !

robotboy655 (changes requested):
The new functions should be named `GetCeiled`, `GetFloored`, etc, to match other function names that return a copy of the object.

PR #1433, opened 2017-12-18
Floating point conversion functions

h3xcat:
`math.DoubleToInt32s( double )`
`math.Int32sToDouble( high, low )`
`math.FloatToInt32( float )`
`math.Int32ToFloat( int32 )`

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -265,7 +265,7 @@ function math.Int32sToDouble( high, low )
 
     if low < 0 then low = low + 0x100000000 end -- Fix sign for low bits
 
-    if exponent > -1022 then
+    if exponent ~= -1022 then
         -- Normalized
         return negative
             and -math.ldexp( (high*2^32+low)*(2^-53)+0.5, exponent ) 
@@ -273,8 +273,8 @@ function math.Int32sToDouble( high, low )
     else 
         -- Denormalized
         return negative 
-            and -math.ldexp( (high*2^32+low)*(2^(exponent+970)), exponent ) 
-            or   math.ldexp( (high*2^32+low)*(2^(exponent+970)), exponent )
+            and -math.ldexp( (high*2^32+low)*(2^-52), exponent ) 
+            or   math.ldexp( (high*2^32+low)*(2^-52), exponent )
     end
 end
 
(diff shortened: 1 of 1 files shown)


CornerPin:
What about byte order? Is it compatible with file and net functions?

edgarasf123:
These functions converts floating value numbers into integer representation and back, unrelated to net library. Manipulation of byte order can be implemented in separate code. Also net library doesn't write data in bytes, rather it writes by individual bits.

thegrb93:
Shouldn't this rather be a C impl?

edgarasf123:
Yeah, but I don't think robotboy want them to be implemented in C, since he closed the request.

edgarasf123:
There are uses for these functions, such sending and retrieving `double`s from http and sql.

neico:
All lua numbers prior to lua 5.3 are floats though, so converting them to int's isn't very effective

edgarasf123:
The conversion to integers is necessary so that you can use bit library to extract float components. Converting to strings wouldn't be as effective, because if you were to read binary data(for instance retrieved from http), they could have either big endian or little endian. If you're implementing binary reader, you probably already have functions that read integers, so you can you those functions to retrieve integers and pass it to those functions to get float.

Here's sample:
```lua
--------------------------------------------------------------------------------
-- Double
--------------------------------------------------------------------------------
function Stream:WriteDouble( double )
    local high, low = math.DoubleToInt32s( double )

    if self.big_endian then
        self:WriteUInt32(high)
        self:WriteUInt32(low)
    else
        self:WriteUInt32(low)
        self:WriteUInt32(high)
    end
end

function Stream:ReadDouble( )
    local high, low
    
    if self.big_endian then
        high = self:ReadUInt32()
        low = self:ReadUInt32()
    else
        low = self:ReadUInt32()
        high = self:ReadUInt32()
    end
    return math.Int32sToDouble( high, low )
end
--------------------------------------------------------------------------------
-- Float
--------------------------------------------------------------------------------
function Stream:WriteFloat( float )
    self:WriteUInt32(math.FloatToInt32( float ))
end

function Stream:ReadFloat( )
    return math.Int32ToFloat( self:ReadUInt32() )
end
```


 Also doubles can act as 54 bit signed integers, or 53 bit unsigned integers.

robotboy655:
This is a very backwards solution to a request like this:
https://github.com/Facepunch/garrysmod-requests/issues/1154

You can use it your code all you want, but I am not interested in adding something with such a narrow and counter-intuitive use case.

PR #1434, opened 2017-12-20
TTT: Add hook for C4 explosion

iagrib:
Add 'TTTC4Explode' hook that is called before the explosion happens


CSchulz:
Any reason why it should be called before the explosion itself?

Kefta:
Would have been nice to be able to block it, or modify the trace data.

CSchulz:
But for this scenario it has to be implemented a little bit different.

PR #1435, opened 2017-12-29
Fixed: Attempt to index local 'vars' (a nil value)

dvdvideo1234:
When the "vars" limit table is mandatory and not given, the user receives the following error:
[ERROR] lua/vgui/prop_float.lua:26: attempt to index local 'vars' (a nil value)
  1. Setup - lua/vgui/prop_float.lua:26
   2. Setup - lua/vgui/dproperties.lua:38


robotboy655:
Went for a different fix for consistency with other elements.
https://github.com/Facepunch/garrysmod/commit/af11a9c5820862a2c04fdad1c8fcf3bb947df337

PR #1437, opened 2018-01-09
Pre/Post Init Scripts

edgarasf123:
Probably it would be better to implement this on C side, but this is needed.  This would be useful for overriding functions, with a guarantee that the function hasn't been called before by different addons.  Before gmod13, we used to have `lua/includes/enum/` folder that did that exact thing, that is, execute scripts in `includes/enum/` before `autorun/` scripts. Right now, multiple anti cheats are using `init.lua` and other developers can't use that file, for any other purposes, without breaking anticheats.

### DIFF
--- a/garrysmod/lua/includes/init.lua
+++ b/garrysmod/lua/includes/init.lua
@@ -1,3 +1,10 @@
+--[[---------------------------------------------------------
+	Include init_pre scripts
+-----------------------------------------------------------]]
+
+for k, v in ipairs( file.Find("init_pre/*.lua", "LUA") ) do
+	include("init_pre/"..v)
+end
 
 --[[---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


bmwalters:
> This would be useful for overriding functions, with a guarantee that the function hasn't been called before some other addon. 

No, because now such addons will just put their scripts in `init_pre/`.

edgarasf123:
Regular addons won't use `init_pre`.

robotboy655:
@edgarasf123 And what enforces that? It's almost the same situation as priorities for hooks.

Kefta:
There's nothing you can do to enforce it - as with returning in hooks and overriding functions, you just have to trust authors to do the right thing. Doesn't mean load order features shouldn't be added.

PR #1439, opened 2018-01-14
Expand File object

edgarasf123:
Added File methods with unsigned support, in addition added aliases to existing methods that make more sense.


Kefta:
This should be done in the engine imo - maybe add methods to write ints/uints of arbitrary bit counts?

Kefta:
Maybe that would be more appropriate for a bit stream, but the methods in this PR should still be implemented internally imo

edgarasf123:
I made s request to implement unsigned number support a month ago (https://github.com/Facepunch/garrysmod-requests/issues/1059). Its now getting buried among other 500 requests. Its more likely to get pull request accepted, than someone implementing it on C.

robotboy655:
https://github.com/Facepunch/garrysmod-requests/issues/1059

PR #1442, opened 2018-01-23
Catch NaN for OnDown

Codingale:
People can spawn cameras that can't be undone by doing a console command.
```
[ERROR] lua/includes/modules/numpad.lua:180: table index is NaN
           1. OnDown - lua/includes/modules/numpad.lua:180
            2. MakeCamera - gamemodes/sandbox/entities/weapons/gmod_tool/stools/camera.lua:67
             3. LeftClick - gamemodes/sandbox/entities/weapons/gmod_tool/stools/camera.lua:96
              4. unknown - gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua:227
```

### DIFF
--- a/garrysmod/lua/includes/modules/numpad.lua
+++ b/garrysmod/lua/includes/modules/numpad.lua
@@ -173,6 +173,7 @@ end
 function OnDown( ply, key, name, ... )
 
 	if ( !key ) then ErrorNoHalt( "ERROR: numpad.OnDown key is nil!\n" ) return end
+	if ( key ~= key ) then ErrorNoHalt( "ERROR: numpad.OnDown key is NaN!\n" ) return end
 	keys_in[ key ] = keys_in[ key ] or {}
 
 	local impulse = {}


SpencerSharkey:
nice

Codingale:
Don't merge just yet, testing hasn't been completely done.

edgarasf123:
It won't get merged any time soon.

mcNuggets1:
Why shouldn't it?

Kefta:
It takes a long time for anything to get merged seemingly

thegrb93:
Yeah, they have to write a bill and get it approved by parliament before it can be merged. I think that's how Britain works.

robotboy655:
Jokes on you, its never going to be merged.

Here's the thing with PRs, they are broken more often than not. A lot of times someone finds a flaw that needs fixing. If we merge every PR almost immediately, these flaws may not get noticed, and I am not talking about the obvious shit like syntax errors and whatnot. Then we have to have more PRs.

Take this PR for example. It fixes a bug but only in a very specific condition. As in, the camera tool may be rid of the bug with this PR, but any other tool isn't.

Here's another thing - when I look into the PR, I test it and I find a problem, I have now 2 options:
- Notify the author that the code is flawed and needs fixing, in which case one of the 2 happens
  - It gets fixed, possibly with another flaw in it after sometime (>12h), then it needs to be reviewed again
  - It never gets fixed in the PR, so someone has to recreate the PR, in which case the loop restarts or the fixed PR needs to be committed manually, in which case see below
- Fix it myself, in which case people could consider it "stealing code and committing it myself"

And then there's the problem with some PRs (mostly the ones marked ```addition``` ) that they may not have a place in the base game in my mind.

Maybe I am overthinking it, maybe I am doing something wrong, but this is how I feel.
And I am not saying that all of my code is flawless. Its not.

Kefta:
I understand that Garry's Mod's standard of compatibility makes each change volatile, no matter how small, but leaving issues unanswered to pile up for years without a modicum of feedback or communication is just as bad as merging it without testing, in my opinion. Even saying "I don't believe this should be in the base game" would work wonders in having a healthy standard of developer collaboration, and would result in the best (and most agreed-upon) final product.
The dev branch was created specifically for the ability to push changes quickly while giving developers and players ample time to test and adapt their content, as well as making sure everything remains compatible. This is the perfect environment to allow others to test any flaws that may not be immediately apparent and result in requests and PRs to never being fulfilled - it does not have to be a 1-2 person burden.
I do not expect at more than a decade old for GMod to be the target of immediate change and response, but silence only furthers the chain of non-progression, and resulting frustration among many developers. I say this not to question or insult your decisions and opinions, but rather it comes from a place of care for the game, its developers, and its community.

Codingale:
Just so you know, I had the commit open for modification my maintainers which meant you could have changed it to your liking though I question why a `MsgN` instead of a `ErrorNoHalt`?

I feel like it could help overall by giving a stack trace, though I suppose the numpad module may not be used that often. Additionally the `"exploit"` did actually crash the server because it wasn't adding the entities into the `g_SBoxObjects` table wasn't being populated with the created entities and didn't get cleaned up. This applied to a few default Garry's Mod tools when setting the key via console to `nan`.

mcNuggets1:
Well, atleast it is fixed.
Thanks rubat.

Btw. cool comment, Kefta :V

robotboy655:
I feel like it making errors (and making the "Something is creating script errors" message pop up) because of convars is wrong, from the end user perspective. ErrorNoHalt does not print out a stack trace.

I wasn't aware the "edits from maintainers" thing existed, I will poke around and try to learn more it in the future, thanks.

Codingale:
It's a server-sided library that pop-up doesn't exist. Odd, when I tested it on the server it was being used on it did create one, I might not have the ErrorNoHalt as a native function, it might be overridden.

No problem, have a great day.

PR #1446, opened 2018-01-25
Create info_player_commons.lua

BreakinBenny:
I got ahead of myself back there with my commits, didn't I. Let's just do it nice and easy…

If it's possible, can the trigger_joinhuman/zombie/spectatorteam brush entities be set up like "trigger_teleport" entities? The first two teleports you to a random human and zombie/whitey spawnpoint, whereas instead of becoming an Observer you get sent to either of the former three.

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/info_player_commons.lua
+++ b/garrysmod/gamemodes/base/entities/entities/info_player_commons.lua
@@ -1,0 +1,9 @@
+-- This is just a simple point entity.
+
+-- We only use it to represent the position and angle of a spawn point
+-- So we don't have to do anything here because the baseclass will 
+-- take care of the basics
+
+-- This file only exists so that the entity is created
+
+ENT.Type = "point"


robotboy655:
https://developer.valvesoftware.com/wiki/Info_player_common

Is it common or common**s**?

Kefta:
Looks like commons:

robotboy655:
Implementation of custom entities from mountable games is up to the modding community.

We only add spawnpoint entities so that the player is always spawned inside the level, not outside of it.

robotboy655:
Come to think of it, do we even need this entity? We already have info_player_human and info_player_zombie from ZPS that worked well on all the maps in my testing.

robotboy655:
Yeah the question do we need it, if you also need other entities to escape it.

Kefta:
It at least gives a starting point for mods to expand upon. I think we should have a community discussion on how far GMod goes to support mountable games entity/weapon-wise.

robotboy655:
If you are already installing a mod, the mod might as well add the entity.

GMod will not really support mountable game entities or weapons for obvious reasons of not having the source code for one. You don't mount games to play those games in GMod. GMod shouldn't a substitute to the mountable games.

BreakinBenny:
Guess you could pretty much set up the info_player_commons and "Join team" brushes to be removed upon being spawned, then? Like a logic_auto that immediately removes itself upon triggering OnMapSpawn?

robotboy655:
That's completely pointless, if an entity class doesn't exist, the entity is never spawned.

BreakinBenny:
Then, is it better to just make the trigger_join%team% brushes act as func_teleports with the on StartTouch output that sends !activator (should be filtered to player) to a random human/zombie spawn? 

Maybe so, because you're pretty much being teleported to the spawnpoint (only without being either human or zombie in Garry's Mod). Some maps will require noclipping manually however, since the makers straight up expect players to just press F1, F2, or F3 o choose a team.

PR #1447, opened 2018-01-25
Env_fire

BreakinBenny:
Env_fire is naturally a point entity that emits a fire, it can even damage you and possibly burn you to death.


robotboy655:
Please do not create multiple PRs that do the same thing

PR #1449, opened 2018-01-25
Update weapon_alyxgun.txt

BreakinBenny:
Setting holdtype to pistol and giving it the SMG1's reload sound, just because. Might as well reference the E3 showcase when the pistol reused that.

### DIFF
--- a/garrysmod/scripts/weapon_alyxgun.txt
+++ b/garrysmod/scripts/weapon_alyxgun.txt
@@ -6,15 +6,13 @@ WeaponData
 	"printname"	"#HL2_AlyxGun"
 	"viewmodel"				"models/weapons/c_pistol.mdl"
 	"playermodel"			"models/weapons/w_alyx_gun.mdl"	//FIXME: 
-	"anim_prefix"			"alyxgun"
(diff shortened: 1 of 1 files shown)


robotboy655:
Why changing the hold type in that file necessary?
Why changing the reload sound is necessary? Is the current sound script not working?

"Just because" is not a good reasoning for anything.

BreakinBenny:
Currently, players holding the Alyxgun use the "unarmed" animation instead of aiming it like a handgun. Also no, I couldn't hear the reload sound upon reloading.

Kefta:
It's because a lot of the Weapon_Alyx_Gun.* sounds don't have sound manifest defines or the pointed sound effects don't exist. This should just be fixed by mods that restore HL2 Beta behaviour.

robotboy655:
Please also test your own changes next time. Changing "anim_prefix" doesn't affect player hold type in-game.

BreakinBenny:
I suppose changes to weapon scripts are only in effect in Singleplayer Source games and mods, like SMOD then. I promise not to hold my breath.

PR #1450, opened 2018-01-27
Update weapon_citizenpackage.txt

BreakinBenny:
I know that this, the suitcase, and harpoon aren't even weapons... but still.


robotboy655:
None of these are useful or necessary, the "weapons" are barely even usable by players, and they are not meant to be.

PR #1452, opened 2018-01-27
Update weapon_oldmanharpoon.txt

BreakinBenny:
Just in case.


mcNuggets1:
Isnt there a tranlsation string for the harpoon?

BreakinBenny:
@mcNuggets1 Nope, doesn't look like it!

PR #1453, opened 2018-01-30
Create info_player_carrier.lua

BreakinBenny:
Tonight's update added the info_player_carrier entity to Zombie Panic! Source, which acts as a spawnpoint for the whitey in that game mod. This will totally not be noteworthy in Garry's Mod though, but… yeah…


robotboy655:
I feel like this falls into the same category as https://github.com/Facepunch/garrysmod/pull/1446

PR #1455, opened 2018-02-05
Update animations.lua

cannon:
I think this should be a hook, so addons that control the mouth can work without overriding the gamemode function and screwing eachother up

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/animations.lua
+++ b/garrysmod/gamemodes/base/gamemode/animations.lua
@@ -230,7 +230,7 @@ function GM:UpdateAnimation( ply, velocity, maxseqgroundspeed )
 
 	if ( CLIENT ) then
 		GAMEMODE:GrabEarAnimation( ply )
-		GAMEMODE:MouthMoveAnimation( ply )
+		hook.Run("MouthMoveAnimation", ply)
 	end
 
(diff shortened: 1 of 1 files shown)


yupi2:
might as well do the same for `GrabEarAnimation` then too

bmwalters:
This function is called pretty frequently, right? I wonder about the performance impact of doing work inside hooks for this function. Then again, if addons are already doing it, providing an official method likely wouldn't make things worse.

cannon:
Check out how the popcorn and vape sweps deal with this. The popcorn swep overrides the global gamemode function which is obviously prone to issues, should any other code override the function. The Vape swep has an ugly solution and may break with code refreshes.

Kefta:
When calling a hook inside of a GAMEMODE function, you should use hook.Call and pass the current table (self).

cannon:
It looks like the same mistake was already in the code by calling the GAMEMODE global.

yupi, I don't agree, as I don't think addons use it. That's very subjective though.
bmwalters, I don't think the hook overhead is significant here.

mcNuggets1:
This would finally stop addon creators from overriding this.

ExtReMLapin:
how u dare using hooks in gamemode

PR #1458, opened 2018-02-18
Update kb_act.lst (LOCALIZATION)

BreakinBenny:
#Valve_Console_Toggle, whitespaces left alone.

### DIFF
--- a/garrysmod/scripts/kb_act.lst
+++ b/garrysmod/scripts/kb_act.lst
@@ -84,5 +84,5 @@
 "+lookdown"			"#Valve_Look_Down"
 "+strafe"			"#Valve_Strafe_Modifier"
 "+klook"			"#Valve_Keyboard_Look_Modifier"
-"toggleconsole"		"Toggle Console"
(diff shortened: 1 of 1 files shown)


Kefta:
You could have just updated your previous PR

BreakinBenny:
The problem is that I don't know how to edit pull requests.

neico:
Commit-push changes to the branch you made the PR in,  
you *could* even force-push it if you care about commit history,  
not that the one to accept the PR can do a squash anyway.

The PR itself is only this discussion board and a diff reference from your selected branch to the selected target branch (usually master) which you can modify outside git's system

BreakinBenny:
I still think it's kinda odd you guys didn't think to do the #Valve_Console_Toggle idea in the first place, requiring the community to help out with the issue… At least I am helping.

Kefta:
You're acting as if you just solved a major bug plaguing GMod - it was a small negligible oversight that was probably caused by Valve not including it in the localisation files when they were first copied to this game. All we suggested above and in your other pulls were ways you could improve PR usage in the future.

PR #1461, opened 2018-03-05
[TTT] Removed useless CollisionRulesChanged call

mcNuggets1:
It doesn't fix the crashes, it doesn't do anything as SetCollisionGroup already declares the collsion rule change.


Kefta:
I added it a few years ago before I knew what the functions actually did - ended up being useless

PR #1463, opened 2018-03-05
Update pickups_english.txt

BreakinBenny:
Added localization for sniper rifle rounds and the flare rounds (which don't exist in the final HL2 game's code).


Kefta:
If they don't exist, why add localisations for them?

BreakinBenny:
Because when you pick up the sniper ammo, it shows its codename for the ammo. This would also be the case for flare rounds, so I did this for completion.

Kefta:
FlareRound isn't an existing ammo type, and sniper ammo is just [one of the many ammo types without a translation](https://wiki.garrysmod.com/page/Default_Ammo_Types). The alternative is to either add translations for all the missing ammo, or leave it up to addons to add translations if they use them.

BreakinBenny:
@Kefta Oh but for the sniper rounds, that's where you are wrong. I am happy I added in the localization for them so nobody needs to do it in their own add-ons, saving them a bit of time!

There are now a few choices to make here, will you…
* Use the localization tags and implement the flare rounds, whose ammo data was omitted before the final release of Half-Life 2, even though almost no SWEP in Garry's Mod would logically wind up using them?
* Delete the programming for these two ammo types altogether, because no default weapon in this game uses them and since this is all a HL2 Beta leftover?


 Choose wisely!

Kefta:
The entities obviously won't be deleted for compatibility, but there's also no reason to add localisation when the ammo type itself doesn't exist. There's a lot of leftover entities in the HL2 code not used anywhere, and the flare round box remains one of them.

To refer to my previous post, either translate all the ammo types missing localisations or translate none.

BreakinBenny:
I have now added the other ammo types, for completion's sake. Hopefully, M9K's weapon pack will not have to deal with people seeing the localization code for the ammunition!

robotboy655:
I feel this is not a good idea because:
1) It will "break" addons (old gmod12 addons for example, before you could add custom ammo types) already using and translating these ammo types. (And so will removing the ammo types)
2) These ammo types are unobtainable without mods as far as I know.

mcNuggets1:
This servers no purpose.
Please stop criticizing the devs for denying stupid nonsense.

BreakinBenny:
@mcNuggets1 It's probably "stupid nonsense" because of some people getting their panties in a bunch over causing issues with add-ons for pre-GMOD 13. But alrighty then; sniper rounds will be left alone, but the flare round ammo will require re-implementation.

mcNuggets1:
We don't fucking need it for gods sake.

BreakinBenny:
If you don't need it, why don't you do something about it?

Here's a tip; get rid of the leftover code that makes flare rounds spawnable entities. Simple as that.
https://github.com/ValveSoftware/source-sdk-2013/blob/master/sp/src/game/server/hl2/item_ammo.cpp

Kefta:
To maintain compatibility with all other Source games that haven't removed it

BreakinBenny:
@Kefta Good, then it's confirmed the solution is to not anything about or with it.

PR #1464, opened 2018-03-05
Update game_hl2.lua

BreakinBenny:
The sniper ammo and both flare round models would require being given physics models.


Kefta:
Why even include these ammos? They can't be used with any weapons

BreakinBenny:
They can, Kefta, if you add SWEPs for them (and in case of the flare rounds, programming for the ammo type). Look at the M9K weapons, for example.

mcNuggets1:
Then do this in addons not in the main game.

PR #1466, opened 2018-03-07
Update npc-ents_english.txt

BreakinBenny:
Added npc_tf2_ghost (Test NPC included in Garry's Mod, actually) and func_tank (gun turret that you can mount


robotboy655:
I feel like Lua entities should be translated in the Lua files they are defined in.

PR #1471, opened 2018-03-18
Be nicer to people with smaller servers

Leystryku:


BlackCetha:
Have a look at #1441

Leystryku:
Playercount playing a role is nice, however, currently you need 16 players
to get your server listed high. With this its 8 which allows smaller
communities to become bigger, faster

BlackCetha <notifications@github.com> schrieb am So., 18. März 2018, 14:38:

Kefta:
I think it would be better for the playercount's effect to work as a decaying function instead of having arbitrary cutoffs.

robotboy655:
https://github.com/Facepunch/garrysmod/commit/9b99f9ea1a2d5a4c78fbde9b786910aa2b36fefe

PR #1472, opened 2018-03-19
Changing string.Comma

ThatGamingWolf:
Calculates the amount of times it needs to insert a comma instead of breaking out of a infinite loop

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -338,11 +338,12 @@ function string.Comma( number )
 		number = string.match( number, "^(.-)%.?0*$" ) -- Remove trailing zeros
 	end
 
-	local k
+	local Times = #number / 4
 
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/string.lua
```diff
@@ -338,11 +338,12 @@ function string.Comma( number )
 		number = string.match( number, "^(.-)%.?0*$" ) -- Remove trailing zeros
 	end
 
-	local k
+	local Times = #number / 4
 
-	while true do
-		number, k = string.gsub( number, "^(-?%d+)(%d%d%d)", "%1,%2" )
-		if ( k == 0 ) then break end
+	if Times < 1 then return number end
```
Kefta: The for-loop declaration already checks this

robotboy655:
Why is this necessary? Does this improve anything?

PR #1473, opened 2018-03-25
Add checks to AddBackgroundImage

ThatGamingWolf:
Adds checks to AddBackgroundImage so that only valid backgrounds get added and removed a useless parameter for ClearBackgroundImages

### DIFF
--- a/garrysmod/lua/menu/background.lua
+++ b/garrysmod/lua/menu/background.lua
@@ -71,8 +71,14 @@ function ClearBackgroundImages()
 end
 
 function AddBackgroundImage( img )
-	if ( not img or type( img ) != "string" ) then return end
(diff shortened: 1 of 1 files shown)


Kefta:
You should throw some errors instead of silently failing

robotboy655:
https://github.com/Facepunch/garrysmod/blob/0fa212127164b965d45fea781a73f4f6d762bd18/garrysmod/lua/menu/mainmenu.lua

These checks are pointless,

ThatGamingWolf:
Im aware that by default this would never be an issue but some people do run their own menu lua and this would very easily error since neither the adding or use of the backgrounds gets checked beforehand, only checked for nil

robotboy655:
Editing the menu state is not officially supported in mod form.

PR #1474, opened 2018-04-02
Fixed paint tool not clearing entity mods on reload

NO-LOAFING:
Previously, these entity mods weren't being cleared, so if you removed the decals from an entity and then duped/saved it, the copy would still have the decals applied.


robotboy655:
👍

PR #1475, opened 2018-04-03
Prevent duplicator from copying gmod_anchor

thegrb93:
Fixes https://github.com/wiremod/advdupe2/issues/127

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_anchor.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_anchor.lua
@@ -3,6 +3,7 @@ AddCSLuaFile()
 
 ENT.Type = "anim"
 ENT.PrintName = "World Anchor"
+ENT.DoNotDuplicate = true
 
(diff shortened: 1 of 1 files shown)


robotboy655:
This would make duplicated hydraulics attached to the world unstable. There's a reason this hack exists.

thegrb93:
Duped sliders create their own gmod_anchor. The duped gmod_anchor is just duped garbage. Not like I didn't test this or anything.

robotboy655:
I cannot confirm that with the built-in duplicator and ```lua_run PrintTable(ents.FindByClass("gmod_a*"))```

thegrb93:
Okay, I was testing this the 'fixed' hydraulic and it was generating garbage gmod_anchor. I'll add a fix to slider duplication so that it matches hydraulic duplication.

robotboy655:
Yeah I was testing with "fixed" enabled as well. Perhaps try without mods? Or maybe the problem lies with how advdupe2 handles it?

PR #1476, opened 2018-04-03
Make slider constraint not save its gmod_anchor and made gmod_anchor not duplicatable.

thegrb93:
'fixed' constraints (constraints that add a slider) don't save their gmod_anchor to their constraint table so any slider they make always creates a new gmod_anchor. This modifies the slider constraint to mimic this behavior and makes gmod_anchor unduplicatable to avoid it getting copied by duplicators with 'area copy' ability.


thegrb93:
This will break dupes containing gmod_anchor though so fuck.

thegrb93:
Just have to make a hack in advdupe2 that makes it ignore gmod_anchor then

PR #1477, opened 2018-04-19
Math lib update

Kefta:
- Added math.BitCount( number num ) - returns the least amount of bits needed to represent an integer in binary. This function can be used with networking indexes for shared tables in the most network-efficient manner
- Added math.IntToString( number num, number [2, 36] base, boolean caps = false ) - this is the opposite of tonumber and will convert a decimal number to a base [2, 36] string. You can now achieve the same effect as math.IntToBin (be it a little less efficient) with this. The caps argument determines if the output string's characters should be capitalised (base 11+)
- Added math.DistanceSqr( number x1, number x2, number y1, number y2 ) - returns the distance squared between two points (more efficient than math.Distance)
- math.IntToBin's string return no longer has leading zeros
- Fixed infinite loop with bad arguments to math.calcBSplineN/BSplinePoint
- Localised global/library functions
- Overall optimisations


thegrb93:
A few useful ones I use commonly; Not important, but might be useful to some. 
math.step - http://www.wolframalpha.com/input/?i=H(x)
math.sign - http://www.wolframalpha.com/input/?i=2*H(x)-1 (except x=0 should be 0)

Kefta:
Those would be more efficiently implemented inline - math.Sign was denied for this purpose: https://github.com/Facepunch/garrysmod/pull/1229

thegrb93:
There's quite a few things in there that are more efficient inline. The point is that its easier to type math.sign rather than doing the inline everytime.

Grocel:
Wouldn't math.Clamp better implemented inline as well? So why is it in again?

Kefta:
It already is inlined from LuaJIT (math.max and min at least).

Kefta:
IntToString now checks its base's bounds like tonumber. Lua 5.2 does not error for non-integer numeral representations like 5.3, so I left the unsigned int cast.

Woowz11:
Can you also add these functions? sine|cosine at the same rate
sin: https://www.desmos.com/calculator/ks1v1aqerp
cos: https://www.desmos.com/calculator/790wkrjg3r

In my project, I call them msin, mcos. (module sin, module cos)

NovaDiablox:
> Can you also add these functions? sine|cosine at the same rate sin: https://www.desmos.com/calculator/ks1v1aqerp cos: https://www.desmos.com/calculator/790wkrjg3r
> 
> In my project, I call them msin, mcos. (module sin, module cos)

He is dead...

NovaDiablox:
@robotboy655 Could you close the PR's @Kefta made in the past to avoid further confusion sir?

thegrb93:
@NovaDiablox everyone died, nothing can be done

NovaDiablox:
> @NovaDiablox everyone died, nothing can be done

Well Rubat still can close PRs.

Grocel:
> @NovaDiablox everyone died, nothing can be done

Can still merge and fix/improve.

PR #1479, opened 2018-04-21
Fixed table.Copy's global lookup_table declaration

WinterPhoenix96:
As the title says; the missing `local` on the line that declares `lookup_table` makes it be declared globally, thus defeating the entire point of `table.Copy` in scenarios where `lookup_table` is not provided as an argument.

Please let me know if you need any further information, such as example code demonstrating the bug.

Or if this is intended behavior, why that is the case, as it goes against what a developer would expect the function to do.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -30,7 +30,7 @@ function table.Copy( t, lookup_table )
 		if ( !istable( v ) ) then
 			copy[ i ] = v
 		else
-			lookup_table = lookup_table or {}
+			local lookup_table = lookup_table or {}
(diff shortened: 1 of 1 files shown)


yupi2:
The `local` isn't needed because that changes the argument's value (which doesn't make a global).

WinterPhoenix96:
Yes, but if there's no argument provided, then lookup_table is set globally.

bigdogmat:
That's not how that works, it doesn't matter if the argument is provided or not. Also, adding `local` here makes it local to the block which breaks the lookup for the rest of the loop should it be used twice.

WinterPhoenix96:
Wow. I was completely wrong in understanding how this works. Sorry guys, carry on.

PR #1480, opened 2018-04-22
Add string.Interpolate

sanny-io:
This is a useful function for having readable format strings as opposed to the normal ones.

Example usage:

```lua
print(string.Interpolate("Hello {name}!", {name = "garry"}))
```
That code would output `Hello garry!`

It only replaces valid lua identifiers between the braces. I figured there isn't really a need to have it work with anything else, considering one of this function's advantages is that it lets you read the format string to more easily understand the intent of the message.

As it stands, the function throws an error if the lookup for a variable fails. I think it's better this way, and it brings it in line with how `string.format` acts.

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -351,11 +351,10 @@ end
 
 function string.Interpolate( str, lookuptable )
 
-	return ( string.gsub( str, "{([_%a][_%w]*)}", function ( key )
-		local value = lookuptable[ key ]
-		assert( value != nil, "string.Interpolate: lookup failed for '" .. key .. "'" )
+	return ( string.gsub( str, "{([_%a][_%w]*)}", function( key )
+		local value = lookuptable[ key ] or "{" .. key .. "}"
 
 		return tostring( value )
-	end) )
+	end ) )
 
 end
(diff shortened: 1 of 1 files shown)


robotboy655 (commented):
Why not just use `string.format`?

Kefta:
Named parameters instead of ordered is useful for translated strings where the order of parameters can change based on the language.

robotboy655:
Fair enough. Not sure how I feel about the error/assert, surely it would be better to just have the token not replaced?

Kefta:
I agree.

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -348,3 +348,14 @@ function string.Comma( number )
 	return number
 
 end
+
+function string.Interpolate( str, lookuptable )
+
+	return ( string.gsub( str, "{([_%a][_%w]*)}", function ( key )
+		local value = lookuptable[ key ]
+		assert( value != nil, "string.Interpolate: lookup failed for '" .. key .. "'" )
```
robotboy655, line 356: This should not error, but should simply not replace the token.

robotboy655 (changes requested):

robotboy655 (approved):

PR #1481, opened 2018-04-25
Fix weapon_medkit

end360:
Changes:
* Toggles lag compensation on the player while performing trace in PrimaryAttack
* Allows PrimaryAttack to heal to player's max health instead of a constant 100


BlackCetha:
Reading the documentation it seems like this requires a `IsPlayer()` check.
My guess is that it would break with bots otherwise.

See here: http://wiki.garrysmod.com/page/Player/LagCompensation

mcNuggets1:
+1

Kefta:
LagCompensation does nothing for bots and can be safely run on them

PR #1484, opened 2018-05-06
Fixed holstered invisibility hack

Leystryku:
why change the FOV to get invisibility while having a crowbar as model if holstered should just be no model at all.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_unarmed.lua
@@ -7,12 +7,13 @@ if CLIENT then
    SWEP.Slot                = 5
 
    SWEP.ViewModelFOV        = 10
+   SWEP.UseHands = false
 end
 
 SWEP.Base                   = "weapon_tttbase"
 
-SWEP.ViewModel              = ""
-SWEP.WorldModel             = ""
+SWEP.ViewModel              = "models/weapons/v_crowbar.mdl"
+SWEP.WorldModel             = "models/weapons/w_crowbar.mdl"
 
 SWEP.Primary.ClipSize       = -1
 SWEP.Primary.DefaultClip    = -1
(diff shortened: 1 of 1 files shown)


Kefta:
I'm pretty sure this will spam the console with MOD_STUDIO errors because of the viewmodel not being a valid model.

mcNuggets1:
couldnt we just return true in PreDrawViewModel?

Kefta:
@mcNuggets1 That doesn't work in ``SWEP:PreDrawViewModel``. The best solution would be to run ``self:GetOwner():GetViewModel():SetWeaponModel("")`` in SWEP:Deploy to just have no viewmodel attached while the weapon is active.

mcNuggets1:
https://github.com/FPtje/DarkRP/blob/master/entities/weapons/keys/shared.lua#L59-L61

It's used in DarkRP for example and I am using it in some of my scripts and it works just fine.
I am not a source engine pro, just a hobby dev, but it seems to do something atleast.

Kefta:
It seems you're correct - the wiki doesn't have the return documented. I will correct that.

Leystryku:
there the spam should be gone now

mcNuggets1:
Just to mention it, there are a few more weapons in ttt using v_crowbar for invisibility.
Healthstation for example.

You might wanna do them too.

Side note:
World model should be able to be "" without erroring, atleast from my tests.

Also there is no PreDrawWorldModel in the wiki.

DrawWorldModel seem to be the only hook for this.

Kefta:
You can use DrawModel and DrawModelTranslucent for the world model. You can also set the models to empty strings again without spam.

FlorianLeChat:
Any news? (normally pull requests for the TTT gamemode are quickly merged but I think this one has been forgotten over the years..)

svdm:
It was left unmerged in expectation of a change from the non-existent PreDrawWorldModel to a DrawWorldModel approach etc as suggested (or just leaving the worldmodel-disabling in Deploy in place). Don't think that's about to happen so I'll close it now, new PR with a fully working solution is welcome.

PR #1485, opened 2018-05-08
Hook library rewrite (prioritization and performance enhancements)

meepen:
This is based on my rewrite from https://gist.github.com/meepen/5566c9598e0e610e5885d12fad2ec3a8

Notes: 
- This maintains almost 100% backwards compatibility with the current hook system, the only exception is that `hook.GetTable()` modifications won't add or remove hooks to run. I believe this will be fine because ULX does the same, so developers have come to expect it won't work on a majority of servers.
- Adds priority to the hook.Add function (higher number = higher priority)
- Adds `hook.Priority` (a table) with a single member `NO_RETURN`
- `NO_RETURN` is NOT enforced in code but should be followed as a standard. It will always be ran first in the priority list.
- Adds `hook.GetPriority(event, name)`
- Adds `hook.GetInternal()` so people who want to irritate the hook library can do so if they want without rewriting it or replacing it with a whole new library. Please don't document this on the wiki, or make it say it is internal and should not be used.
- Adds some type checking to arguments for sanity's sake

credits:
@ThatLing for reviewing my code to ensure everything is tip top shape and can be read

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -1,125 +1,289 @@
-local gmod                        = gmod
-local pairs                        = pairs
-local isfunction        = isfunction
-local isstring                = isstring
-local IsValid                = IsValid
+local hook_func = 1
+local hook_name = 2
+local next_hook = 3
+local real_hook_func = 4
+local last_hook = 5
+local hook_priority = 6
 
-module( "hook" )
+local event_table = {}
 
-local Hooks = {}
+-- keeps the record of how many hooks are in an event metatable to always
+-- return a value so you don't need to do checks everywhere
+local event_count = setmetatable({}, {
+	__index = function(self, k)
+		self[k] = 0
+		return 0
+	end
+})
+local gmod = gmod
+local type = type
+local math = math
+local pairs = pairs
+local assert = assert
+local IsValid = IsValid
 
---[[---------------------------------------------------------
-    Name: GetTable
-    Desc: Returns a table of all hooks.
------------------------------------------------------------]]
-function GetTable() return Hooks end
+local noop_hook = {
+	-- hook_func
+	function() return false end,
+	-- hook_name
+	nil,
+	-- next_hook
+	nil,
+	-- real_hook_func
+	nil,
+	-- last_hook
+	nil,
+	-- hook_priority
+	-math.huge
+}
 
+module "hook"
 
---[[---------------------------------------------------------
-    Name: Add
-    Args: string hookName, any identifier, function func
-    Desc: Add a hook to listen to the specified event.
------------------------------------------------------------]]
-function Add( event_name, name, func )
+-- This is not modifiable like in the past. Other hook libraries that were
+-- implemented and overrode the library in the past also did this, so I
+-- doubt many addons will break because of this.
 
-	if ( !isfunction( func ) ) then return end
-	if ( !isstring( event_name ) ) then return end
+local function _GetTable()
+	local out = {}
+	for event, hooklist in pairs( event_table ) do
+		out[ event ] = {}
 
-	if (Hooks[ event_name ] == nil) then
-			Hooks[ event_name ] = {}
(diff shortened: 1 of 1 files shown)


meepen:
sorted arrays and linked lists have no performance differences in lua.

meepen:
```
] - local T,t,nop,i,s,this={{}},SysTime,function(e)end,1 for i=1,1e7 do T[i]={{}}end this=T[i]s=t()::s:: nop(this[1]) i=i+1 this=T[i] if this then goto s end print(t()-s)
0.18326508712698
] - local this,t,nop,i,s,n={},SysTime,function(e)end,1 n=this for i=1,1e7 do n[1]={}n=n[1]end s=t()::s:: nop(this[1]) this=this[1] if this then goto s end print(t()-s)
0.13612537420715
```

meepen:
I think the main problem is the branching in the linked list, after that it's fine. If I use an iterative loop as you are doing it would be fine. I will update it.

meepen:
```local this,t,nop,i,s,n={},SysTime,function(e)end,1 n=this for i=1,1e7 do n[1]={false}n=n[1]end s=t()for i = 1, 1e7 do nop(this[1]) this=this[1] end print(t()-s)
0.041814087393362
local T,t,nop,i,s,this={{}},SysTime,function(e)end,1 for i=1,1e7 do T[i]={{}} end s=t() for i=1, #T do this=T[i] nop(this[1]) end print(t()-s)
0.0430561556039
```

mcNuggets1:
Please merge this, I'll pay with my liver!

meepen:
@thegrb93 I remember why I didn't use arrays and that is because when removing something after IsValid fails will screw up ordering. Linked lists made it a ton simpler.

meepen:
It's a LOT faster now, and still works perfectly.

thegrb93:
It looks a lot better, just one more thing. The Add function can't modify the 'event_table' because it might happen while Call is iterating, causing skipping or multi-calling of hooks. It should modify (or create if it doesn't exist) a separate staging table which 'Call' will check for update to before iterating. Of course hook.Call can be called from inside a hook as well so I think a counting sentinel variable is needed to keep track of how many hook.Call is in the stack, and only update if it's the first one. Idk, maybe you have a better solution.

meepen:
Add and Remove performance doesn't matter since it's not called 5 bajillion times a second.

FredyH:
I think I know one major improvement of this implementation.
Currently the order hooks are called is does not correspond to the order they were added in.
In your implementation this is also a problem since hooks of the same priority will also get called in any order.
I don't know if it's feasible to add this without sacrificing a lot of performance, but if it is you should go for it.

I think you should also really change these comments: [ --[[comment]] ] and rather have them be above the line or at the end of the line, because currently the code is really annoying to read.

thegrb93:
I've been testing https://github.com/Facepunch/garrysmod/pull/1485/commits/be3ec1c9c60e100f3adfe2652d9569dd8a4f4386 for a while and found a case where hook.Add wasn't working. I'll update to latest and see if its fixed.

thegrb93:
If you add a remove a hook from inside it's call and try re-adding it the add will fail

mcNuggets1:
Tomorrow this is opened for 1 month.
Can we get any reply from the devs about this?

meepen:
Hey, I'm working on this again. Currently optimizing and the fix for grb's situation was just changing the direction the update list was iterated.

mcNuggets1:
ULib goes from lowest to highest, priority wise, this is the opposite.
Thus it will break ULib compatibility.

Kefta:
ULib would still override the hook lib with their version.

mcNuggets1:
Yeah, but I guess if they want to remove their hook replacement, they have to do a lot of work in order to do so.

Kefta:
They can just reverse the priority values for their replacement since it uses enums and not raw numbers.

mcNuggets1:
Yeah, works for me.

PR #1487, opened 2018-05-17
Ammo lib update

Kefta:
- Bad tbl.name no longer silently fails.
- Made ammo names case-insensitive.

### DIFF
--- a/garrysmod/lua/includes/extensions/game.lua
+++ b/garrysmod/lua/includes/extensions/game.lua
@@ -1,6 +1,5 @@
 
 local AmmoTypes = {}
-local AmmoNames = {}
 
 --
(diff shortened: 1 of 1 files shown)


robotboy655:
This needs to be updated since one of the recent updates added its own game.GetAmmoType

Kefta:
Removed GetAmmoType from this PR.

robotboy655:
Is there any reason for `AmmoNames` to exist and to carry the ammo table which is never used?
You may argue its ~ faster ~ to check if ammo type is already registered but ` game.AddAmmoType` is not exactly something that is called often for this to matter.

Kefta:
Done

PR #1490, opened 2018-05-23
Update net.lua

KilgraveVlatkovic:


willox:
I don't think this'll be a good idea. Lag compensation can be particularly expensive and I'm confident that not every net receiver in existence requires it.

KilgraveVlatkovic:
Almost everything else in the game is lag compensated (PrimaryAttack for example), I don't see it creating too much more overhead.

KilgraveVlatkovic:
I suppose you're right, after a benchmark it shows that implementing LagCompensation ups runtime by 3x.

thegrb93:
Not to mention it could break a lot of things that don't expect this behavoir.

neico:
If anything make it opt-in, there might be use cases, but those probably are by far the minority who actually benefit of off it.

PR #1493, opened 2018-05-30
Add new HL1 NPCs

Kefta:
- Added monster_gman.
- Added monster_ichthyosaur.
- Added monster_leech.
- Added monster_flyer and monster_flyer_flock.
- Added function support to Offset and KeyValue keys in NPC list structure. This allows for determining the offset or keyvalues on NPC spawn (used for flyer random start offset and flock random size).
- Added InWater parameter to NPC list structure for underwater NPCs (ichthyosaur and leech).
- Updated duplicator NPC list.


robotboy655:
You should stop bundling multiple changes together. You should move SF and nodrop stuff to a separate PR. I'd also like to see where nodrop makes a difference. 

- The dead NPCs cannot be interacted with, so I don't think they belong in the spawnmenu. (and also, they are dead)
- monster_leech doesn't seem to do anything
- HLS gman can't be killed and we already have the HL2 one
- monster_flyer also can't be killed either
- don't put in commented out things please
- we don't need ability to use function in the NPC list

Kefta:
I usually bundle PRs with related changes together since they are dependent most times, and maintaining 20 open PRs for merge conflicts is quite toilsome. But I will separate out the NoDrop and SF changes to just leave what is necessary for the HL1 NPCs.
- I thought the dead NPCs would be nice for posing, but as they can't be moved with the physgun, I agree.
- Leeches actually do bite the player, but like the ichthyosaur, they do no damage since the HL1 sk_* convars are not set (sk_leech_dmg_bite specifically in this case). I'm working on a fix for that.
- There are already a number of HL1 NPCs that can't be killed already in the spawnlist (monster_tentacle, monster_bigmomma, monster_alien_grunt), so I don't see why gman and flyers should be excluded for that reason.
- I'll remove the comments.
- Functions are for flyer height and squad randomisation. I think it's a pretty nice feature for them to not spawn at the same height/amount every time.

BreakinBenny:
@Kefta You could use the Animation tool to animate models so they use the end of the death animation.

Kefta:
Any reason this wasn't merged alongside https://github.com/Facepunch/garrysmod/commit/00c71b02139bbaff0f242e9558ee257ae873551b?

robotboy655:
monster_flyer_flock cannot be undone once spawned through the Spawnmenu, probably have to get rid of it.

Should also drop the changes to Offset and Keyvalues. Changing NPC height feels like the job for the physgun.

Also remove the dead NPCs from duplicator support for reasons stated above.

Kefta:
What should the default height be then? It seems like flyers could easily get stuck inside of a ceiling without some kind of functional determination.

robotboy655:
Well random height is going to make the problem worse, not better. The default 32 works fine for me.

PR #1495, opened 2018-05-30
Add weapon_handgrenade to the spawnmenu

Kefta:


mcNuggets1:
What if people dont have hl1 source?

Kefta:
You're right, I totally forgot an IsMounted check.

mcNuggets1:
We might need to wait for rubats response to this.
He might be against it, as the weapon uses the "c"?-crosshair and is in the file "game_hl2".

Kefta:
Yeah, I couldn't really find a better file to put it in, but I think it's fine since the physgun is also there. Could be renamed to describe the actual contents more aptly but that could be for a larger API cleanup.

Don't see what you mean by "c" crosshair, looks like the no-crosshair default to me:

mcNuggets1:
So when not having hl1 installed, it defaults to the "c"(whats the name of the "bigger" crosshair anyways?)-crosshair, but when installed not.
Atleast it is for me.

(I dont have it installed right now)

Kefta:
Oh, that's from a missing font. That's what also causes the infamous "Q" crosshair with HL2 weapons.

mcNuggets1:
Oh when I maximize your screenshot, it actually is the crosshair I meant.
I need to take deeper looks in things, before interpreting read words.

Well, yeah the crosshair is bigger for some reason ❌
Like when you have no weapons at all.

mcNuggets1:
Just curious, didnt half life 2 use the same?

Kefta:
They're slightly different:

robotboy655:
I am not opposed, but it'll be weird with just one weapon sitting in that category.

Kefta:
Unless there are plans to add the rest of the HL1 weapons at some point, maybe this should just be thrown into Other?

mcNuggets1:
In Other there are gmod/lua entities, not c entities.
It wouldn't make much sense.

Kefta:
Well the other alternative is to make the Half-Life 2 category just Half-Life, but I don't think implementation in the engine or Lua should determine a category's contents.

mcNuggets1:
I am fine with the half life source category.

Kefta:
@robotboy655 Gonna remake this with all the new entities

PR #1496, opened 2018-05-30
Fix incorrect hand grenade translation key

Kefta:
Also changed the value to match HL1's printname definition


robotboy655:
I feel like it still should be all caps to match the HL2 weapons, the text names of weapons in HL1 never appear.

Kefta:
Fair enough

BreakinBenny:
Yeah, capitalize the strings.

PR #1497, opened 2018-05-31
Add HL1 skill cfgs

Kefta:
This PR is reliant on https://github.com/Facepunch/garrysmod-requests/issues/1148 and https://github.com/Facepunch/garrysmod-requests/issues/1151 for functionality.

Setting these convars fixes a number of HL1 NPC issues with having incorrect health or doing no damage.

skill#.cfg will first exec hl1_skill#.cfg to set HL1 skill values, then run skill.cfg to set HL2 skill values, giving precedent to HL2 while still setting necessary HL1-specific convars. skill_manifest.cfg is no longer used with this new setup but is still called by the engine. Convars for non-existing NPCs and weapons have been commented out.

https://github.com/Facepunch/garrysmod-requests/issues/1149 can allow hl1_skill#.cfg to be exec'd again at the very end of a skill level change, giving priority to HL1 values for HL1 campaign addons.

### DIFF
--- a/garrysmod/cfg/hl1_skill1.cfg
+++ b/garrysmod/cfg/hl1_skill1.cfg
@@ -1,0 +1,130 @@
+// Barney
+sk_barney_health		"35"
+
+// Apache 
+//sk_apache_health	"150"
+
+// Headcrab
+sk_headcrab_health		"10"
+sk_headcrab_dmg_bite		"5"
+
+// Zombie
+sk_zombie_health		"50"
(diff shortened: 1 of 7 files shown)


Kefta:
A Lua implementation of this PR, as well as fixes for the aforementioned issues can be found here: https://github.com/Kefta/gs_skilllevel

Kefta:
Is there a plan for these being added separately?

PR #1500, opened 2018-06-01
Add additional HL1 entity translations

Kefta:
Last batch. GitHub refuses to keep the current encoding no matter which way I commit it.


robotboy655:
grenade_spit is the antlion worker spit in gmod

Kefta:
Moved grenade_spit to #1501

PR #1501, opened 2018-06-01
Add additional HL2 entity translations

Kefta:
- prop_crane
- prop_vehicle_driveable
- prop_stickybomb
- grenade_spit
- prop_vehicle_jeep "Vehicle" -> "Jeep" (may represent the Jalopy too but it's better than Vehicle imo)

Same encoding issue as https://github.com/Facepunch/garrysmod/pull/1500


robotboy655:
I think "prop_vehicle_jeep" should stay as "Vehicle", because most of the vehicles mods are just reskins of the jeep, "Vehicle" for those would be more appropriate than "Jeep".

Kefta:
I thought people used prop_vehicle_prisoner_pod more for reskins, but fair enough.

robotboy655:
I left quite a mess here, sorry, GitHub really doesn't like the weird encoding of that file

PR #1502, opened 2018-06-01
Remove hardcoded ammo type from SWEP:ShootBullet

Kefta:
Also added additional arguments for secondary fire, force, and tracer frequency


robotboy655:
Feels to me the `ammotype` should be just an argument like `force` and `tracer` instead of the `secondary` bool argument.

PR #1503, opened 2018-06-08
Change gmod_tool's holdtype to pistol

Noiwex:
IMO this been like this since GMod 13 and it looks way better and stylish than pre-GM13 holdtypes, changing this to revolver was unnecessary.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
@@ -69,7 +69,7 @@ end
 
 function SWEP:Initialize()
 
-	self:SetHoldType( "revolver" )
+	self:SetHoldType( "pistol" )
 
 	self:InitializeTools()
 


Kefta:
I'm good with this if I can submit a version of the viewmodel with only one hand

ExtReMLapin:
comparison pic pls

Tenrys:
Just make an addon to change it back, I'm sure they won't go back on that decision just because you don't like it

thegrb93:
gm_flatgrass0005

Tenrys:
I prefer Revolver :d

Noiwex:
Virgin revolver
Chad pistol

thegrb93:
such confidence. build like pro even when using single grip

ExtReMLapin:
Okay here is what we should do

< than 5000h on gmod : Revolver mode

> than 5000h, Pistol mode

willox:
i want to agree with you

legoj15:
Wasn't the hold type changed to revolver not even a year ago? (becuase it matches the view model or something?)

sannysc:
Time to fork.

Noiwex:
Remove left hand bones and it's fixed. Can be done with Lua but it's an
ugly hack then.

PR #1504, opened 2018-06-08
cleaned up animations file

JarosLucky:
Removed double 'if client' and check for vehicle.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/animations.lua
+++ b/garrysmod/gamemodes/base/gamemode/animations.lua
@@ -54,7 +54,11 @@ function GM:HandlePlayerDucking( ply, velocity )
 
 	if ( !ply:IsFlagSet( FL_ANIMDUCKING ) ) then return false end
 
-	ply.CalcIdeal = (velocity:Length2DSqr() > 0.25) and ACT_MP_CROUCHWALK or ACT_MP_CROUCH_IDLE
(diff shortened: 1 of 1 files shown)


meepen:
if you put the words micro and optimize right next to each other, there's no point in creating a pull request really... you can call this pull request `cleaned up animations file` and it would be better.

garrysmod/gamemodes/base/gamemode/animations.lua
```diff
@@ -54,11 +54,7 @@ function GM:HandlePlayerDucking( ply, velocity )
 
 	if ( !ply:IsFlagSet( FL_ANIMDUCKING ) ) then return false end
 
-	if ( velocity:Length2DSqr() > 0.25 ) then
-		ply.CalcIdeal = ACT_MP_CROUCHWALK
-	else
-		ply.CalcIdeal = ACT_MP_CROUCH_IDLE
-	end
+	ply.CalcIdeal = (velocity:Length2DSqr() > 0.25) and ACT_MP_CROUCHWALK or ACT_MP_CROUCH_IDLE
```
BlackCetha: I heard ternary operators dont exist in GLua, so its the same for the interpreter and the old version is more readable IMO.

JarosLucky: ah. okey. i did not know.

PR #1506, opened 2018-06-13
TTT: Optimised big loops / Replaced useless loops

mcNuggets1:
* Use ipairs instead of pairs when possible (ttt works for me with these changes excellently without any errors)
* Rather big optimisation for the weapon/ammo spawning on round start (It now doesn't iterate through the weapon/ammo list x the weapon/ammo count anymore)

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
@@ -329,7 +329,7 @@ local function PlaceWeaponsAtEnts(spots_classes)
    local num = 0
    local w = nil
    for k, v in RandomPairs(spots) do
-      w = spawnables[math.random(spawnables_count)]
+      w = spawnables[math.random(1, spawnables_count)]
       if w and IsValid(v) and util.IsInWorld(v:GetPos()) then
          local spawned = PlaceWeapon(w, v:GetPos(), v:GetAngles())
 
@@ -339,7 +339,7 @@ local function PlaceWeaponsAtEnts(spots_classes)
          -- we need the spawned ent that has inherited the goods from the
          -- basegrenade swep.
          if spawned and spawned.IsGrenade then
-            w = spawnables[math.random(spawnables_count)]
+            w = spawnables[math.random(1, spawnables_count)]
             if w then
                PlaceWeapon(w, v:GetPos(), v:GetAngles())
             end
--- a/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
@@ -62,7 +62,7 @@ local function ReplaceAmmoSingle(ent, cls)
 end
 
 local function ReplaceAmmo()
-   for _, ent in pairs(ents.FindByClass("item_*")) do
+   for _, ent in ipairs(ents.FindByClass("item_*")) do
       ReplaceAmmoSingle(ent)
    end
 end
@@ -97,7 +97,7 @@ end
 
 
 local function ReplaceWeapons()
-   for _, ent in pairs(ents.FindByClass("weapon_*")) do
+   for _, ent in ipairs(ents.FindByClass("weapon_*")) do
       ReplaceWeaponSingle(ent)
    end
 end
@@ -106,7 +106,7 @@ end
 -- Remove ZM ragdolls that don't work, AND old player ragdolls.
 -- Exposed because it's also done at BeginRound
 function ents.TTT.RemoveRagdolls(player_only)
-   for k, ent in pairs(ents.FindByClass("prop_ragdoll")) do
+   for k, ent in ipairs(ents.FindByClass("prop_ragdoll")) do
       if IsValid(ent) then
          if not player_only and string.find(ent:GetModel(), "zm_", 6, true) then
             ent:Remove()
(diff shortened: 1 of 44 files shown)


garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
```diff
@@ -264,8 +264,8 @@ local SpawnableAmmoClasses = nil
 function ents.TTT.GetSpawnableAmmo()
    if not SpawnableAmmoClasses then
       local tbl = {}
-      for k,v in pairs(scripted_ents.GetList()) do
-         if v and (v.AutoSpawnable or (v.t and v.t.AutoSpawnable)) then
+      for k, v in pairs(scripted_ents.GetList()) do
```
Kefta: Missed one here

garrysmod/gamemodes/terrortown/gamemode/init.lua
```diff
@@ -207,7 +207,7 @@ function GM:Initialize()
    end
 
    local cstrike = false
-   for _, g in pairs(engine.GetGames()) do
+   for _, g in ipairs(engine.GetGames()) do
```
Kefta: This section can be changed to use https://wiki.garrysmod.com/page/Global/IsMounted

garrysmod/gamemodes/terrortown/gamemode/init.lua
```diff
@@ -524,22 +524,23 @@ end
...
    end
 
    -- This is ugly as hell, but it's kinda nice to filter out the names of the
    -- traitors themselves in the messages to them
-   for k,v in pairs(player.GetAll()) do
+   local traitornicks_count = #traitornicks
+   for k,v in ipairs(player.GetAll()) do
       if v:IsTraitor() then
-         if #traitornicks < 2 then
+         if traitornicks_count < 2 then
```
Kefta: You can cache the ``#traitornicks < 2`` check instead of just the length

garrysmod/gamemodes/terrortown/gamemode/ent_replace.lua
mcNuggets1: What did I miss?

svdm:
I can appreciate the meaningful changes like the random weapon/ammo optimisation and the cleaner mount check, but all this pairs/ipairs search and replace is change for the sake of change.

Kefta:
They are measurably more efficient, especially with JIT.

mcNuggets1:
I run this version on my TTT server right now and everything performs much more smoother.

meepen:
This is probably the top 10 worst pull requests I've seen

The performance you gain from this won't even exist since a majority of these changes affect not constantly-called functions

The only way this would ever improve any performance is if there was a function that used pairs and had no inner body and was called 100 times a frame

mcNuggets1:
Most of the optimisations are minor, besides the ttt_random_weapon optimisations which is actually a big improvement to what is was before.

Some of the ipairs changes are minor, thats true, but there are frequently called expensive loops in there also.
I just wanted to keep everything consistent, thats why I replaced NEARLY every "incorrectly used" pairs call.

It's not a big rework of TTTs performance, but it's a improvement and that is all it should be for me

BleedingOptimism:
The fix for TriggerOutput should be improved and then merged. If I attempted to use ttt_restartround too quickly, I received a nice failed timer error. I say improved because the function should still be run, just not before it's been defined.

Removing the unnecessary checks improves performance by a fraction, and should probably be merged for readability at least.

However, for pairs vs ipairs speed I found this: https://gmod.facepunch.com/f/gmoddev/mzmp/Speed-Analysis-of-pairs-and-ipairs/
I don't know how reliable these results are, but I did try out the file changes for my server. I wasn't able to notice any positive or negative performance difference (upwards of 55 weapons to process).

mcNuggets1:
On my servers I have 200 weapons and the performance improvement is significant there.

For the trigger lua error fix:
If the trigger gets removed, there is no way to call the function.
The function can't be called and wasn't being called properly before, I just fixed the error.

Alf21:
Two months ago, I also changed "pairs" into "ipairs" and got significant performance improvements (https://github.com/Alf21/TTT2/commit/bfa4cbc9bbacf5ffd92d109d18bd574d7cf4ea34#diff-43f1d4d165f1cb4a6e286c851439d551R21). The result will happen if there are many weapons. On my own, I created an addon that copies each weapon and changes their stats. So if there are 200 weapons, with the addon, there will be 400 weapons (because there is no other way to dynamically change the values). All in all, I think this is a nice pull request in case of micro-optimisation.

Kefta:
@bleedingoptimism pairs/ipairs has different results with LuaJIT, that analysis is for vanilla Lua.

BleedingOptimism:
@mcNuggets1 @Kefta Thank you for the explanations. I'm still pretty noobie.

svdm:
I've tested and committed the changes to ttt_random_weapon and _ammo (5b0c1fb), because the performance gains are pretty intuitive without having to measure them (table.Random is a hog).

I appreciate that replacing pairs with ipairs where possible is done with good intentions, but it makes the PR so large/widespread that I have no hope of testing it. There are many places where the loop is so small or called so rarely the change has no significant effect, so reviewing and testing those is a waste.

If you think performance can be improved, focus on hotspots and measure if/how much your changes help. It's a lot easier for me to spend time on a PR if you have numbers that show it makes e.g. game initialization/spawning 20% faster.

I'm closing this PR for the reasons mentioned above. If there are other changes in this PR that would be of significant benefit (like the MapSettings output, or ipairs in a hot loop somewhere), then I'd welcome them as new, smaller PRs. Thanks!

mcNuggets1:
Well then this is the part, where I maintain a different TTT version.

I am running this for 1 month on my TTT server now and it just runs fine and if there was something not working, people would have already called it out or my TTT server players would report something.

PR #1509, opened 2018-06-18
Add string.CardinalToOrdinal

DELIBURD:
This would be a useful function to have for anyone who would want to use it.

Example:
`print( string.CardinalToOrdinal( 111 ) )`

Output:
`111th`


Kefta:
You could more optimally implement this by caching the modulo results - would help in performance variant environments, like using this function multiple times in a HUD per frame.

```
function string.CardinalToOrdinal( cardinal )
	local basedigit = cardinal % 10

	if ( basedigit == 1 ) then
		if ( cardinal % 100 == 11 ) then
			return cardinal .. "th"
		end

		return cardinal .. "st"
	end

	if ( basedigit == 2 ) then
		if ( cardinal % 100 == 12 ) then
			return cardinal .. "th"
		end

		return cardinal .. "nd"
	end

	if ( basedigit == 3 ) then
		if ( cardinal % 100 == 13 ) then
			return cardinal .. "th"
		end

		return cardinal .. "rd"
	end

	return cardinal .. "th"
end
```

willox:
Is this the same as STNDRD?

DELIBURD:
@Kefta I will change it so it caches the modulo request. And @willox, it's similar to STNDRD, but it keeps the number prepended to the suffix. Kefta's way of doing it is more efficient than just doing 

`cardinal .. STNDRD(cardinal)`

Also, not too many people know about STNDRD considering its odd name.

robotboy655 (approved):

Kefta:
Believe it or not I benchmark my code before claiming efficiency.

PR #1510, opened 2018-06-18
Add function "draw.Box"

JarosLucky:
stop using draw.Round box to create a box. without check the `Round size is <= 0`

### DIFF
--- a/garrysmod/lua/includes/modules/draw.lua
+++ b/garrysmod/lua/includes/modules/draw.lua
@@ -162,6 +162,18 @@ function DrawText( text, font, x, y, colour, xalign )
 end
 
 --[[---------------------------------------------------------
+	Name: Box( x, y, w, h, color )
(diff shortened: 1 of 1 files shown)


neico:
Could have used
```lua
function Box( x, y, w, h, color )
    RoundedBox( 0, x, y, w, h, color )
end
```
but that's actually one extra function call compared to yours, but who is counting these anyway~

Kefta:
@neico Lua

JarosLucky:
@neico Your suggestion is a function that uses a function uses a fuction for check to draw a box. This just create a box.

JarosLucky:
@Kefta what do not you like about my offer?

end360:
Saves 1 comparison, very important.

ExtReMLapin:
certified darkrp coder

Etothepowerof26:
to be honest this is very useless

PR #1513, opened 2018-07-16
Optimized string.Comma.

howmanysmall:
This method is a little faster, maybe about 10% faster.


Kefta:
This was done a little more efficiently in https://github.com/Facepunch/garrysmod/pull/1407

h3xcat:
Not sure why you undid my change. I do not mind optimizing it, but it broke printing of large numbers...

image

PR #1514, opened 2018-07-18
ENT:BeingLookedAtByLocalPlayer optimisation

Kefta:
- ``ENT.MaxWorldTipDistance`` can now set the maximum distance the worldtip is displayed at when looking at an entity.
- The distance check now uses the view entity's ShootPos if it's a player, just like the local player does. This makes the result consistent between a player and spectators.
- Dist->DistToSqr
- Order of operations changed in ENT:Think and ENT:BeingLookedAtByLocalPlayer to do traces last - optimises entities too far from the player from doing an unnecessary eyetrace.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua
@@ -5,22 +5,16 @@ DEFINE_BASECLASS( "base_anim" )
 ENT.Spawnable = false
 
 if ( CLIENT ) then
-	ENT.ViewDistance = 256
+	ENT.MaxWorldTipDistance = 256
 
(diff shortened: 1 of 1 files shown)


robotboy655:
```ENT.ViewDistance``` sounds a bit vague to me, how about something like ```ENT.MaxWorldtipDistance```?

Also wouldn't it be cleaner/faster to replace the ```local pos``` variable with just 2 returns in the if statement?

Kefta:
Changed ``ENT.ViewDistance`` to ``ENT.MaxWorld**T**ipDistance`` (like ``AddWorldTip``).

Slightly faster maybe, cleaner not really since there's another half of the return that would have to be copied twice. I just changed the long if to a ternary, instead.

Kefta:
I just noticed a glaring issue with tracing, could you revert and remerge this? I've updated the original branch.

PR #1518, opened 2018-07-24
[TTT] Russian localization file fixes.

Satton2:
Very big amount of fixes and changes. To be correct, all localization file was re-translated.


Satton2:
Also see this: https://github.com/Facepunch/garrysmod-issues/issues/3544

PR #1521, opened 2018-07-30
Set entity type to "anim" by default

Kefta:
[As described on the wiki](http://wiki.garrysmod.com/page/Structures/ENT). Fixes https://github.com/Facepunch/garrysmod-issues/issues/2745.

### DIFF
--- a/garrysmod/lua/includes/modules/scripted_ents.lua
+++ b/garrysmod/lua/includes/modules/scripted_ents.lua
@@ -53,15 +53,15 @@ end
 function Register( t, name )
 
 	local Type = t.Type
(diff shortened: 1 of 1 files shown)


robotboy655:
Due to how this issue is, this fix (or really any Lua fix) cannot fix this.

Won't close just yet depending on what we decide to do about this whole thing in https://github.com/Facepunch/garrysmod-issues/issues/2745

robotboy655:
I have came to the conclusion that this fix will never work, even with c++ changes. It results in overriding of the type for a lot of base entities such as the example nextbot, because at this point the entity base table is not merged yet. ( That is done during entity creation, not registration )

PR #1523, opened 2018-08-08
Removed link for the "Next Update" thread.

apap04:
It was a pretty old link from 2017 and is useless. The last game discussion thread was in February.


Kefta:
https://github.com/Facepunch/garrysmod/pull/1507

PR #1530, opened 2018-09-15
Add utility functions for using the Hammer IO system

Lexicality:
Usage:

```lua
function ENT:KeyValue( key, value )
	if ( BaseClass.KeyValue ) then BaseClass.KeyValue( self, key, value ) end

	if ( self:SetNetworkKeyValue( key, value ) ) then
		return
	elseif ( self:AddOutputFromKeyValue( key, value ) ) then
		return
	end

	-- Your entity's logic goes here

end

function ENT:AcceptInput( name, activator, called, value )
	if ( BaseClass.AcceptInput and BaseClass.AcceptInput( self, name, activator, called, value ) ) then
		return true
	end

	if ( self:AddOutputFromAcceptInput( name, value ) ) then
		return true
	end

	-- Your entity's logic goes here

	return false

end
```

Ideally I'd like to make the base_entity functions do this for you automatically so most people won't even need to think about doing this, what do you think?

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/base_entity/outputs.lua
+++ b/garrysmod/gamemodes/base/entities/entities/base_entity/outputs.lua
@@ -23,6 +23,49 @@ function ENT:StoreOutput( name, info )
 	table.insert( self.m_tOutputs[ name ], Output )
 end
 
+-- This works like SetNetworkKeyValue, call it from ENT:KeyValue and it'll
+-- return true if it successfully added an output from the passed KV pair.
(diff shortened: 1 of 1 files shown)


TIMONz1535:
I understand that these are still different systems, but I think working with outputs directly can be more useful than making extra wrappers around the scripting entities.
https://github.com/Facepunch/garrysmod-requests/issues/1911

robotboy655:
Nothing about that request would change/replace what this PR accomplishes. In order to capture/hook outputs, they first need to be fired.

PR #1534, opened 2018-11-08
Small math changes. Enable clamping to have only one border.

dvdvideo1234:
When something needs to be greater than 0 you can do `local a = math.Clamp(b, 0)` and omit the high/low border.

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -4,8 +4,7 @@
 	Desc: Distance between two 2d points
 ------------------------------------------------------------]]
 function math.Distance( x1, y1, x2, y2 )
-	local xd = x2 - x1
-	local yd = y2 - y1
+	local xd, yd = (x2 - x1), (y2 - y1)
 	return math.sqrt( xd * xd + yd * yd )
 end
 math.Dist = math.Distance -- Backwards compatibility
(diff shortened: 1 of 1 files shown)


thegrb93:
You can disable a border by using (+/-)math.huge

dvdvideo1234:
> 
> 
> I think that's far more confusing than using math.max.

When you have `math.max(-1,-2,-3,-4)` it will return `-1`. It will not clamp it to `0`. My idea is
```lua
math.Clamp(-5, 0) -- Returns 0 as higher border is missing [0;+huge)
math.Clamp(-5, 7) -- Returns -5 as higher border is missing [7;+huge)
math.Clamp(-5, -3, 7) -- Returns -3 as both borders are present [-3;+7]
math.Clamp(10 -3, 7) -- Returns 7 as both borders are present [-3;+7]
math.Clamp(-5, nil, 7) -- Returns -5 as lower bound is missing (-huge;7]
math.Clamp(10 nil, 7) -- Returns 7 as lower bound is missing (-huge;7]
```

Kefta:
It will just make regular calls to math.Clamp slower.

willox:
> When you have math.max(-1,-2,-3,-4) it will return -1. It will not clamp it to 0. 

Then call `math.max(0,-1,-2,-3,-4)`. I'm not sure what I am missing, can't math.min and math.max replace all of your examples that omit a parameter?

dvdvideo1234:
Made my own local parameter check :D I don't need this PR anymore. Thought you guys should see it handy.

PR #1536, opened 2018-11-14
Toggling god mode in singular function

GlorifiedPig:
ply:God( !ply:HasGodMode() ) is more optimized and easier to write than if( ply:HasGodMode() ) then ply:GodDisable() else ply:GodEnable() end


robotboy655:
And break every existing mod that uses those existing functions? No.

Kefta:
I don't think we need more aliases.

PR #1538, opened 2018-11-23
Create table.IsEmpty

WilliamVenner:
Quickly returns whether a table has iterable items in it - useful for checking whether non-sequential tables are empty.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -53,6 +53,17 @@ function table.Empty( tab )
 end
 
 --[[---------------------------------------------------------
+	Name: IsEmpty( tab )
+	Desc: Returns whether a table has iterable items in it, useful for non-sequential tables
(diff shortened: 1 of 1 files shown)


Kefta:
This would be more efficiently implemented as ``return next( tbl ) == nil``

thegrb93:
lol anyone worth their salt will just do if next(t) then or if not next(t) then

AMD-NICK:
`if not next(t) then` looks better for me

Kefta:
> `if not next(t) then` looks better for me

What if ``next`` returned false?

thegrb93:
If you're using bool for your table keys then I imagine you wouldn't need to check if its empty. And if you did, you wouldn't need next.

AMD-NICK:
> > `if not next(t) then` looks better for me
> 
> What if `next` returned false?

in a case when it's possible I'll use more accurate comparison

PR #1546, opened 2018-12-17
[TTT] Loadout Bugfix

Alf21:
Here, the ID is a return of ply:EntIndex(). This should use Entity(id) (http://wiki.garrysmod.com/page/Global/Entity) instead of player.GetByID(id) (http://wiki.garrysmod.com/page/player/GetByID)


willox:
this should probably just pass the player object if i'm not missing anything

svdm:
The fix is correct, but I know I tested this when I wrote it so I wonder why it still worked. IIRC players are always entities 1 to whatever maxplayers is, so my guess is entity X is always player X (or at least was when testing).

PR #1547, opened 2018-12-17
Toolgun Think fixes

Kefta:
- Check owner validity - fixes rare edge case of the owner being invalid when Think is called and cases where the owner is an NPC
- Check last tool object validity before releasing its entity - fixes error when switching from invalid mode to another invalid mode
- Remove a return when a mode switch occurs and the last mode was invalid, causing the first Think after a mode switch to be skipped - fixes https://github.com/Facepunch/garrysmod-issues/issues/3682

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
@@ -142,38 +142,48 @@ end
 
 -- Think does stuff every frame
 function SWEP:Think()
+	local owner = self:GetOwner()
+	if ( !owner:IsPlayer() ) then return end
 
-	self.Mode = self.Owner:GetInfo( "gmod_toolmode" )
+	local curmode = owner:GetInfo( "gmod_toolmode" )
+	self.Mode = curmode
 
-	local tool = self:GetToolObject()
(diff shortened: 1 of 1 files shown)


thegrb93:
I'm seeing the toolgun ghost fail to release if there was no previous tool as well. Did you happen to see what would cause that as well?

AlphaS-code:
yep

thegrb93:
I'll try and see if this fixes it.

thegrb93:
Nope, something else is causing it

thegrb93:
Seems to be because https://github.com/Facepunch/garrysmod/blob/469a66d3c59c762b1d13119e399be4ac4737f7b8/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua#L178 isn't executed if there was no previous tool

thegrb93:
I've found the problem is that https://github.com/Facepunch/garrysmod/commit/0883499dd17f15eb2061a37bc755250a0dc67238 isn't working. Probably due to prediction and CurTime

robotboy655:
I don't like your seemingly unnecessary localization of all the variables, but I guess I will allow it this time.

Kefta:
It reduces the amount of unnecessary function calls.

PR #1550, opened 2018-12-29
Nice Ammo Reduction

MyDimeIsUp:
A much-needed feature. Stop removing the whole ammo box when someone only needs 5 ammo.

Ex.
- Pistol ammo box. Player can come up and only need 5 ammo, reduce the ammo count to 15. The other player needs 20 ammo. Player comes up and gives them 15 ammo, and then removes the ammo box.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/base_ammo_ttt.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/base_ammo_ttt.lua
@@ -86,7 +86,7 @@ function ENT:CheckForWeapon(ply)
 end
 
 function ENT:Touch(ent)
-  if (SERVER amd self.tickRemoval ~= true) and ent:IsValid() and ent:IsPlayer() and self:CheckForWeapon(ent) and self:PlayerCanPickup(ent) then
+  if (SERVER and self.tickRemoval ~= true) and ent:IsValid() and ent:IsPlayer() and self:CheckForWeapon(ent) and self:PlayerCanPickup(ent) then
     local ammo = ent:GetAmmoCount(self.AmmoType)
     -- need clipmax info and room for at least 1/4th
     if self.AmmoMax >= (ammo + math.ceil(self.AmmoAmount * 0.25)) then
--- a/garrysmod/gamemodes/terrortown/entities/entities/base_ammo_ttt.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/base_ammo_ttt.lua
@@ -31,7 +31,7 @@ function ENT:Initialize()
       self:SetTrigger(true)
    end
 
-   self.taken = false
+   self.tickRemoval = false
 
    -- this made the ammo get physics'd too early, meaning it would fall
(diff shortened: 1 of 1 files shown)


svdm:
This creates uncertainty of how much ammo is in a given box, which to people who have not seen this PR will appear random (or just buggy). In general I don't think the upside of this is worth it.

MyDimeIsUp:
Is there a better way to display the amount of ammo? I think that people should learn about this new functionality (Because currently, it already seems buggy removing a WHOLE ammo box for 5 ammo). I honestly think this looks less "buggy". If you think so, what should be implemented to let players know the amount of ammo? Again, the current system already looks buggy, so IMO this looks less "buggy", but yeah. 3d2d text to draw the ammo count? Let me know @svdm

MyDimeIsUp:
Might I saw there is also uncertainty with the ammo box because someone can press Z with the AK and drop their ammo. This is already somewhat implemented. So the thought of uncertainty already exists with the dropping of your current ammo. This isn't much farther and in fact adds more functionality to ammo boxes.

garrysmod/gamemodes/terrortown/entities/entities/base_ammo_ttt.lua
```diff
@@ -86,23 +86,23 @@ function ENT:CheckForWeapon(ply)
...
-            local given = self.AmmoAmount
-            given = math.min(given, self.AmmoMax - ammo)
-            ent:GiveAmmo( given, self.AmmoType)
-
-            self:Remove()
-
-            -- just in case remove does not happen soon enough
-            self.taken = true
-         end
+  if (SERVER amd self.tickRemoval ~= true) and ent:IsValid() and ent:IsPlayer() and self:CheckForWeapon(ent) and self:PlayerCanPickup(ent) then
```
Alf21: amd -> and ? ^^

MyDimeIsUp:
@svdm Prev. comment

Kefta:
Yeah, the behaviour is inconsistent. Either allow partial ammo boxes all together or not at all.

svdm:
Fair point, but the minimum amount you can drop is 1/4th of a weapon's clipsize. To be consistent with that, this change should remove the ammo box if it gets below 1/4th rather than at 0. That way we at least avoid boxes with a single bullet etc.

MyDimeIsUp:
> Fair point, but the minimum amount you can drop is 1/4th of a weapon's clipsize. To be consistent with that, this change should remove the ammo box if it gets below 1/4th rather than at 0. That way we at least avoid boxes with a single bullet etc.

Good Point. I can go ahead and modify this then when I get a chance in the next 2-3 days

MyDimeIsUp:
I've just pushed an update to comply with this @svdm sorry for the delay!

MyDimeIsUp:
@mcNuggets1 fixed typos and indentation

MyDimeIsUp:
> just use / 4 and also there are prints inside x3

The original code had calculation by  * 0.25. I don't want to change the formatting for how it gets a percentage. I however have removed the debug print

PR #1551, opened 2018-12-30
Added: Sign calculation

dvdvideo1234:


neico:
Wasn't this denied in #1229 already?

dvdvideo1234:
Hmm, They did not explain why it is "useless" :D

GitSparTV:
Not every function needs to be implemented to the game. Just write your not-working sign function in your addon and use it

PR #1552, opened 2019-01-03
CanProperty call changes

Maud-Pie:
Changed 'remover' to 'remove' to has same name in properties.List


mcNuggets1:
Isn't this used in addons already?
So it might break compatibility.

Kefta:
I don't think adding aliases is a good idea either

mcNuggets1:
Just leave it as it is, document it in the wiki.
That should be it.

PR #1553, opened 2019-01-04
Fixed duped effects not having AttachedEntityInfo + minor effect fixes

NO-LOAFING:
- Fixed duped prop_effects not having their AttachedEntityInfo table (meaning no entity mods, skin/bodygroup changes, bone manips, etc.) ever since the MakeEffect function was added. This value used to be copied over because duped prop_effects were being spawned with GenericDuplicatorFunction, which dumps the entire dupe data table into the new entity. Now that they have their own spawn function, we have to copy the value over manually.

- Fixed prop_effects spawned from spawnmenu not having any bodygroups other than default because it was trying to apply them to the prop_effect itself instead of its AttachedEntity

- Fixed prop spawn effect not working on prop_effects for the same reason

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/commands.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/commands.lua
@@ -132,6 +132,7 @@ function MakeEffect( ply, model, Data )
 
 	local Prop = ents.Create( "prop_effect" )
 	duplicator.DoGeneric( Prop, Data )
+	Prop.AttachedEntityInfo = table.Copy( Data.AttachedEntityInfo )
 	Prop:Spawn()
 
 	duplicator.DoGenericPhysics( Prop, ply, Data )
@@ -141,7 +142,9 @@ function MakeEffect( ply, model, Data )
 		gamemode.Call( "PlayerSpawnedEffect", ply, model, Prop )
 	end
 
-	DoPropSpawnedEffect( Prop )
+	if IsValid( Prop.AttachedEntity ) then
+		DoPropSpawnedEffect( Prop.AttachedEntity )
+	end
 
 	return Prop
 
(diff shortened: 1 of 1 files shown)


robotboy655:
I decided to go with a different fix for bodygroups, one that is consistent with retaining skins, etc

robotboy655:
https://github.com/Facepunch/garrysmod/commit/1ebf63ffb855a54f86c3fb762089de25bef526dc If you are interested

NO-LOAFING:
`self.AttachedEntity:SetBodyGroups( self:GetBodyGroups() )`

This doesn't work because SetBodyGroups takes a string and GetBodyGroups returns a table? Even if you did it the slow way like this:

```
for i = 0, self:GetNumBodyGroups() -1 do
	self.AttachedEntity:SetBodygroup(i, self:GetBodygroup(i)
end
```

it still wouldn't do anything. The prop_effect entity itself doesn't have any bodygroups because it's just an HL2 melon model, so even though you used SetBodygroup/SetBodyGroups on the same entity earlier, GetBodygroup will still return 0 for each one, because the entity doesn't seem to store bodygroup info for bodygroup IDs its model doesn't have.

That's why I had ent.AttachedEntity:SetBodyGroups directly in the spawn function the first time, because ent doesn't store them at all.

thegrb93:
Isn't self.AttachedEntity the prop_dynamic?

robotboy655:
God fucking damn it with fucked function naming conventions. Do you have a model name/link I can use to test?

NO-LOAFING:
I've been using models/props_gameplay/cap_point_base.mdl from TF2, it only has one changeable bodygroup but it gets the job done. If you want something with more bodygroups, I'd imagine you could put something like a TF2 playermodel into an addon without the .phy file (while also not mounting TF2) to make GMod spawn it as an effect.

thegrb93:
Or just call the MakeEffect function with whatever model.

PR #1555, opened 2019-01-13
Added SetSize() & GetTexture() methods in killicon.lua

timmybo5:
I tried to make some workarounds, but the only way to resize these killicons is to add another method in the module.

Tested both scenarios ( fail & pass ) and everything works like it should.

Update: Would also be nice to retrieve the texture of the killicon for custom drawing, so i added another method.

### DIFF
--- a/garrysmod/lua/includes/modules/killicon.lua
+++ b/garrysmod/lua/includes/modules/killicon.lua
@@ -48,6 +48,16 @@ function Exists( name )
 
 end
 
+function GetTexture( name )
+	
+	if (!Icons[name]) then 
+		Msg("Warning: killicon not found '"..name.."'\n")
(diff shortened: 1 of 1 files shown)


robotboy655 (changes requested):
I can understand the Get methods, but not the Set ones. When would it be logical to set font and the character of a killicon?

Also the 2 get methods could be collapsed into one `killicon.Get` method.

robotboy655:
Recent changes to the killicon library ( https://github.com/Facepunch/garrysmod/commit/29d0f42a189631788fe95f954f9ca639d2386707 ) should make this unnecessary.

And it still does not make sense to add those `Set` methods, you can just use `killicon.Add` to override existing icons.

PR #1557, opened 2019-01-25
Fix toolgun ghost being left behind

thegrb93:
I think it fixes https://github.com/Facepunch/garrysmod-issues/issues/3709 haven't seen any problems yet


mcNuggets1:
This actually could make sense in my eyes.

ZehMatt:
Still happens if you use "phys_swap"

thegrb93:
Yeah, it isn't fixed still. Just had it start happening again. Damn

PR #1558, opened 2019-02-16
Promise/A+ objects

Dragoteryx:
I wrote a Promise implementation in Lua, and thought it might be useful to other people. It's based of Javascript built-in promises so if you've already used them it shouldn't feel too different.

Use `util.Promise` to create a new Promise object, it only takes one parameter which is it's builder function, which only has two parameters that are the resolve and reject functions of the Promise.

For example this Delay function returns a Promise that is resolved after delay seconds, or rejected if delay < 0.
```lua
function Delay(delay)
    return util.Promise(function(resolve, reject)
        if delay < 0 then
            reject("delay < 0")
        else
            timer.Simple(delay, resolve)
        end
    end)
end
```

You can use `promise:Then` and `promise:Catch` just like you would in Javascript, there is also a `promise:Done` function that is called inside `promise:Then` that doesn't return a new Promise.

Finally, you can use `promise:Await`. It will call coroutine.yield until the Promise is settled, then it will either return the resolved value or error depending or whether the Promise is fulfilled or rejected.

There are also 3 utility functions:
`util.PromiseAll` which works just like Javascript's `Promise.all`.
`util.PromiseFirst` which is the same as Javascript's `Promise.race`.
`util.PromiseAsync` only has one argument which is a callback function. It will run the function inside a coroutine and return a Promise. When the coroutine stops the Promise will be resolved with the return value of the callback function, or rejected if an error happens in the coroutine. You can use this function along with `promise:Await` to simulate the async/await keywords of Javascript.

Using my Delay function from before (this is a bad example because coroutine.wait exists, but you get the idea):
```lua
util.PromiseAsync(function()
    Delay(1):Await()
    print("after 1 second")
end):Catch(print)
```

I also added a `http.Promise` function. It's just a Promise version of the HTTP function that returns a Promise instead of using callbacks. (HTTP requests are the main reason you would want to use promises so... )


thegrb93:
timer.Simple(1, function() print("after 1 second") end)

Dragoteryx:
I know don't worry, I said it's a bad example. If you don't know for what reasons Promise are used in Javascript then of course this will seem useless.

thegrb93:
I don't think I would ever use it. Maybe some will, but in the context of Garrysmod it should be a separate library imo.

One small suggestion is to only add your Think hook when there's work to do and remove it when there isn't.

Kefta:
@thegrb93 There's no extrenuous work being done. If there's no promises to simulate, the function just loops over an empty table - that is not adverse to performance.

Dragoteryx:
I just thought of a more pratical use, so...

Let's say you wrote a function called PlayerAvatar, that takes two arguments, a player and a callback function with one argument => the avatar url, this is how you would use it:
```lua
PlayerAvatar(ply, function(avatarUrl)
    -- avatarUrl is accessible here
end)
```
But what if you want to get the avatar of every player on the server? Your code will quickly start to look messy.
Now what if PlayerAvatar returns a Promise instead?
```lua
PlayerAvatar(ply):Then(function(avatarUrl)
    -- avatarUrl is accessible here
end)
``
Now using `promise.All`, accessing the avatar of every player is way easier.
```lua
local tbl = {}
for i, ply in ipairs(player.GetHumans()) do
    table.insert(tbl, PlayerAvatar(ply))
end
promise.All(tbl):Then(function(avatarUrls)
    -- avatarUrls is a table containing the avatar of every player
end)
```

sannysc:
Promises are extremely useful. This should be in the `util` library, however.

`util.Promise` instead of `promise.New`

sannysc:
@thegrb93 this is for anything that is asynchronous, of which there are many things...database queries, net messages, external API calls.

Dragoteryx:
Now that you say util.Promise does sound better than promise.New
Going to rename a few stuff.

Dragoteryx:
Done. I renamed all functions:
promise.New => util.Promise
promise.All => util.PromiseAll
promise.First => util.PromiseFirst
promise.Async => util.PromiseAsync

Now that I think of it it might be better to rename util.PromiseAsync to util.PromiseCoroutine. What do you think?

garrysmod/lua/includes/extensions/util/promise.lua
```diff
@@ -0,0 +1,193 @@
...
+      if not safe then reject(args) end
+    end)
+    coroutines[id] = co
+    id = id+1
+  end)
+end
+
+hook.Add("Think", "PromiseAsyncCoroutines", function()
+  for id, co in pairs(coroutines) do
+    if co == nil then continue end
```
Kefta: co will never be nil here - table values cannot be nil.

Dragoteryx: I put that in here just in case, I'll remove it.

NullEnt1ty:
> I don't think I would ever use it. Maybe some will, but in the context of Garrysmod it should be a separate library imo. It's really something meant for threaded applications.

I agree to a certain extent. GMod Lua programming is mostly single-threaded. But there are a few third-party libraries (for example SQL drivers) and even built-in ones (http) which use callback arguments, because they are asynchronous functions. Promises would make the usage of them more enjoyable.

ceifa:
How to print the stacktrace as a common error at the main thread?
Tried `:Catch(error)` but not works.

robotboy655 (changes requested):

PR #1560, opened 2019-02-18
Add timer.Loop

Dragoteryx:
`timer.Loop(number delay, function func)` will call the function passed as a parameter every delay second until the function returns false or nil. You can also return a number along with the boolean to change the delay.
```lua
-- prints numbers from 10 to 0 at a faster and faster rate
local value = 10
timer.Loop(10, function()
    print(value)
    value = value - 1
    return value >= 0, value
end)
```

### DIFF
--- a/garrysmod/lua/includes/extensions/timer.lua
+++ b/garrysmod/lua/includes/extensions/timer.lua
@@ -5,6 +5,7 @@
 -----------------------------------------------------------]]
 function timer.Loop(delay, callback)
   timer.Simple(delay, function()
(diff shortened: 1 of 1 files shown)


robotboy655:
timer.create with reps set to 0.

Kefta:
@robotboy655 That isn't anonymous - it requires giving the timer a name.

mcNuggets1:
This function would actually be pretty cool.

thegrb93:
You can do the same thing with timer.Simple alone.

PR #1564, opened 2019-03-06
InputDetails on Loading Screen

TheMerkyShadow:
There are many complaints that people should have a custom convar for volume for loading screen music. But this should be up to the developers of the loading screens. But thats not possible due to the limitations that web developers have. 

This allows users to left click on javascript objects. This doesn't seem much but its an extremely powerful feature. And I outlined it with for the possibility of future implementations down the road. 

Now I have posted similar pull request nearly a year ago and I took another crack at it. In no means am I disrespectful to anyone's decision or the developers. However, I felt that this solution is much better than  #1421. With the possibility of a chromium update in the future, I thought this cannot hurt anyone and could be useful.

How to test. Not really difficult just change the file and join a server that has some sort of object that is usually intractable via mouse as a webpage. I'm also posting the IP of my server which has some basic html objects. 

IP: 24.228.175.77


Kefta:
Better than https://github.com/Facepunch/garrysmod/pull/1421 but is probably still going to be denied. It'd be more advantageous to request putting the loading screen on a separate thread so that you can actually click on the screen during loading, but this is quite limited in its current state.

TheMerkyShadow:
You know I took another crack at it. And thought it can't hurt to try. If they don't want it that's fine.

Donkie:
@TheMerkyShadow I think you accidentally commited some stuff to this PR

PR #1565, opened 2019-03-20
Changed table.Count to table.IsEmpty in halos

Kefta:
based willox


willox:
i used psychohistory to predict this pull-request

PR #1566, opened 2019-03-21
Updated Stack implementation

Kefta:
- Add ``table PopMulti(number[0, inf) Count = 1)``
- Moved all data to array indicies for faster access
- Pop now errors when given invalid args
- Pop now returns the last element it popped (solves https://github.com/Facepunch/garrysmod-requests/issues/217)
- Various optimisations, biggest being changing table.remove to direct element removal in Pop

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -222,49 +222,83 @@ function util.Timer( startdelay )
 end
 
 
+local function PopStack( self, num )
+	if ( num == nil ) then
+		num = 1
+	elseif ( num < 0 ) then
+		error( string.format( "attempted to pop %d elements in stack, expected >= 0", num ), 3 )
+	else
+		num = math.floor( num )
+	end
+
+	local len = self[0]
+
+	if ( num > len ) then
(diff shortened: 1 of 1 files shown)


robotboy655:
Maybe `Push()` should return the position the new entry was put into, and I really don't like how you can remove multiple entries with `Pop()` but it will only return 1 removed entry regardless, but I am not sure what's what to do about it, or if anything should be done.

Kefta:
How about adding a boolean parameter defaulting to false which will return a table popped values instead of just the top value?

robotboy655:
That sounds extra hacky, ideally it should just return a table if it removed more than one element, but I am not sure how that affects ~ PERF ~

Kefta:
That's what I was also concerned over. Maybe two separate functions, Pop and PopTable? PopTable will always return a table no matter how many elements were popped.

robotboy655:
Perhaps something like PopMultiple or PopMulti? PopTable sounds like you are popping the first `table` type or something.

Kefta:
Added ``table Stack:PopMulti(unsigned Count)``; I chose Multi over Multiple for consistency with ``Entity:PhysicsInitMultiConvex`` and ``DModelSelectMulti``. The function will always return a table, even with ``PopMulti(0)``.

PR #1567, opened 2019-03-23
Update table.Add

Tripperful:
Allow adding more than one source table to the target using vararg, so you can do things like
```Lua
local dest = { 1, 2, 3 }
local source1 = { 4, 5 }
local source2 = { "boo", "foo" }
PrintTable( table.Add( dest, source1, source2 ) )
```

And get results like

	1	=	1
	2	=	2
	3	=	3
	4	=	4
	5	=	5
	6	=	boo
	7	=	foo


garrysmod/lua/includes/extensions/table.lua
```diff
@@ -107,16 +107,23 @@ end
...
-	-- At least one of them needs to be a table or this whole thing will fall on its ass
-	if ( type( source ) != "table" ) then return dest end
-	if ( type( dest ) != "table" ) then dest = {} end
+	local sources = { ... }
 
-	for k, v in pairs( source ) do
-		table.insert( dest, v )
+	for i, source in ipairs( sources ) do
+
+		if ( type( dest ) != "table" ) then dest = {} end
```
Kefta: Leave the dest table check outside of the loop so it doesn't unnecessarily check the type every iteration.

Kefta:
Could you benchmark the effect of adding this functionality with just a single table? Varargs can sometimes be an efficiency concern but LuaJIT might negate it in this case.

thegrb93:
I prefer using 'select'

Kefta:
@thegrb93 That's even more inefficient (tested in Lua 5.1 and LuaJIT 2.0.5) than packing the args in a table since ``select`` is an uninlined function call every time.

thegrb93:
That sucks

thegrb93:
I guess the upside is it handles nils in the args

Tripperful:
Benchmarked this. For 50 000 iterations with same parameters given the new vararg version is about 10%-15% slower. I don't think it's that much of a performance loss

Donkie:
@Tripperful your code has whitespace changes or something, affecting the entire file. This definitely won't get merged until that is fixed.

Kefta:
It's the GitHub editor's fault because he merged a conflict - the whitespace changes aren't really there.

robotboy655:
Fairly certain thats line endings changed.

Tripperful:
Oh, shit.
I certainly shouldn't have used GitHub's online code editor on my phone while I was on a toilet to resolve a conflict

AMD-NICK:
This could break some Functional Programming code. Maybe would be better to use separate function?

```lua
function table.AddMultiple(dest, ...)
	for _,t in ipairs({...}) do
		table.Add(dest, t)
	end
	return dest
end
```

PR #1571, opened 2019-03-28
Optimize table.Count comparisons

end360:
1) Remove unnecessary use of table.Count in TTT (table is sequential).

2) Remove table.Count from duplicator icon generation and keep count in the already existing loop


robotboy655:
Again, do not include TTT and non TTT changes in the same PR, but I will allow it this time.

PR #1572, opened 2019-03-28
Replace usage of type() with is* functions

aStonedPenguin:
They're faster and cleaner if just checking for a single type.


Kefta:
Why remove the parentheses around if-statements? It's inconsistent with the codebase.

aStonedPenguin:
Right, fixed.

PR #1577, opened 2019-04-09
net: OnNetMessageReceived

ItsEris:
this is cool and possibly helpful to some, more or less just want it for my particular usecase


thegrb93:
hook.Call bruh

Kefta:
bruh moment

Kefta:
I think this hook is harmful without the ability to rewind the net buffer - calling net.Read* will offset reading for the receiver.

Donkie:
Maybe the hook should be called after the receiver function so the hook is more of a notice than a place to read stuff?

thegrb93:
It's just plain bad. Have whatever addon you're trying to hook into call a hook. Don't add a hack that hooks into every net message.

PR #1578, opened 2019-04-12
Add table.Compare

Kapppa:
Will allow you to tell if two tables have the same values in the same key

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -692,3 +692,25 @@ function table.GetKeys( tab )
 	return keys
 
 end
+
+
+function table.Compare(tab1,tab2)
+
+	if #tab1 ~= #tab2 then return false end
+
+	local t1,t2 = {}, {}
+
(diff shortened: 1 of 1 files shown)


thegrb93:
I stared at this for two minutes trying to make sense of it. Why

Kefta:
Why do you need to copy all values to new tables? Also, the ``#`` operation is not defined for non-sequential tables - you could have two "equal" tables that return different len results.

11Lee1:
couldn't you do something like this
`function table.Compare(tab1,tab2)
	for k,v in pairs(tab1) do
		if(tab2[k] == nil || tab2[k] != v)then
			return false;
		end
	end
	return true;
end`

Kefta:
@11Lee1 The ``== nil`` check is necessary since a non-nil value will never be equal to nil. Also, you'll need to also loop through tbl2 since your implementation will consider ``{1, 2, 3}`` equivalent to ``{1, 2, 3, 4, 5}``.

Grocel:
The keys can also be a tables, so you need would check those to be equal as well.

PR #1581, opened 2019-04-30
fix for using the CMenu in nonvehicle entities

joeyjumper94:
improved version of my fix.
earlier description:
if you open the context menu while in a prop_vehicle_crane you will get a lua error like this
[ERROR] lua/includes/modules/properties.lua:114: attempt to call method 'GetThirdPersonMode' (a nil value)
  1. GetHovered - lua/includes/modules/properties.lua:114
   2. fn - lua/includes/modules/properties.lua:174
    3. Run - addons/ulib_557962238/lua/ulib/shared/hook.lua:109
     4. fn - lua/includes/modules/halo.lua:149
      5. unknown - addons/ulib_557962238/lua/ulib/shared/hook.lua:109

this is because LocalPlayer():GetVehicle() return the crane you are in, so the IsValid check will return true, but since it's not an actual vehicle, calling a Vehicle function will cause a lua error.

### DIFF
--- a/garrysmod/lua/includes/modules/properties.lua
+++ b/garrysmod/lua/includes/modules/properties.lua
@@ -111,7 +111,12 @@ end
 function GetHovered( eyepos, eyevec )
 
 	local filter = { LocalPlayer():GetViewEntity() }
-	if ( LocalPlayer():GetViewEntity() == LocalPlayer() && IsValid( LocalPlayer():GetVehicle() ) && !LocalPlayer():GetVehicle():GetThirdPersonMode() ) then table.insert( filter, LocalPlayer():GetVehicle() ) end
+	if ( LocalPlayer():GetViewEntity() == LocalPlayer() ) then
+		local veh=LocalPlayer():GetVehicle()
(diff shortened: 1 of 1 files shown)


Kefta:
You can improve the if-statement and surrounding code even more:
```
local ply = LocalPlayer()
local filter = ply:GetViewEntity()

if ( filter == ply ) then
	local veh = ply:GetVehicle()
	
	if ( veh:IsValid() && ( !veh:IsVehicle() || !veh:GetThirdPersonMode() ) ) then
		filter = { filter, veh }
	end
end
```

joeyjumper94:
filter is supposed to be a table listing entities the trace should ignore.

bigdogmat:
It can also just be an entity if you only want to ignore 1 thing

Kefta:
"filter" can be a single entity. From the wiki:
```
Things the trace should not hit. Can also be a table of entities or a function with one argument
```

robotboy655:
I have made the crane act as a vehicle properly, but still going to merge this.

PR #1583, opened 2019-05-06
DColorButton: Added ability to choose Tooltip

Alphaverse:
Added a boolean value to SetColor() for the DColorButton so you are able to choose to show the color string tooltip. The default value is false. 

Now the tooltip won't be reset every time the color is changed but you are still able to view it if you want to.

Fixes #Facepunch/garrysmod-issues#3836

### DIFF
--- a/garrysmod/lua/vgui/dcolorbutton.lua
+++ b/garrysmod/lua/vgui/dcolorbutton.lua
@@ -27,7 +27,7 @@ function PANEL:IsDown()
 
 end
 
-function PANEL:SetColor( color, displayTooltip = false )
+function PANEL:SetColor( color, displayTooltip )
 
 	if ( displayTooltip == true ) then
 		
--- a/garrysmod/lua/vgui/dcolorbutton.lua
+++ b/garrysmod/lua/vgui/dcolorbutton.lua
@@ -27,9 +27,9 @@ function PANEL:IsDown()
 
 end
 
-function PANEL:SetColor( color, displayTooltip )
+function PANEL:SetColor( color, hideTooltip )
 
-	if ( displayTooltip == true or displayTooltip == nil ) then
+	if ( hideTooltip == false or hideTooltip == nil ) then
 		
 		local colorStr = "R: " .. color.r .. "\nG: " .. color.g .. "\nB: " .. color.b .. "\nA: " .. color.a
 		self:SetTooltip( colorStr )
--- a/garrysmod/lua/vgui/dcolorbutton.lua
+++ b/garrysmod/lua/vgui/dcolorbutton.lua
@@ -29,7 +29,7 @@ end
 
 function PANEL:SetColor( color, hideTooltip )
 
-	if ( hideTooltip == false or hideTooltip == nil ) then
+	if ( !hideTooltip ) then
 		
 		local colorStr = "R: " .. color.r .. "\nG: " .. color.g .. "\nB: " .. color.b .. "\nA: " .. color.a
 		self:SetTooltip( colorStr )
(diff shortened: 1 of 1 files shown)


sannysc:
Default arguments do not exist in lua at a language level. That is a syntax error.

Alphaverse:
> Default arguments do not exist in lua at a language level. That is a syntax error.

Yes i fixed it... Worked with another language just before I created this..

BlackCetha:
So you are saying you haven't tested this once?

Alphaverse:
I tested it. But added this value after it because I had the idea and it hadn't changed much. Only forgot the default value thing.

garrysmod/lua/vgui/dcolorbutton.lua
```diff
@@ -27,11 +27,15 @@ function PANEL:IsDown()
 
 end
 
-function PANEL:SetColor( color )
+function PANEL:SetColor( color, displayTooltip )
 
-	local colorStr = "R: " .. color.r .. "\nG: " .. color.g .. "\nB: " .. color.b .. "\nA: " .. color.a
-
-	self:SetTooltip( colorStr )
+	if ( displayTooltip == true ) then
```
Alf21: To support the old behaviour, you should check 
```lua
if displayTooltip ~= false then
```
or use 
```lua
if displayTooltip == nil then 
    displayTooltip = true 
end
```

Alphaverse:
So everything is the same as before now. The only thing is that you can add a false after the color and the Tooltip won't display. Simply but useful in my opinion.

garrysmod/lua/vgui/dcolorbutton.lua
Alphaverse: I changed it so everything works the same as long as you don't add a false after the color.

robotboy655:
Shouldn't the argument be ```no tooltip = true```? This way the code is simpler and it is consistent with the rest of the game.

garrysmod/lua/vgui/dcolorbutton.lua
```diff
@@ -27,11 +27,15 @@ function PANEL:IsDown()
 
 end
 
-function PANEL:SetColor( color )
+function PANEL:SetColor( color, hideTooltip )
 
-	local colorStr = "R: " .. color.r .. "\nG: " .. color.g .. "\nB: " .. color.b .. "\nA: " .. color.a
-
-	self:SetTooltip( colorStr )
+	if ( hideTooltip == false or hideTooltip == nil ) then
```
fruitwasp: `hideTooltip == nil` breaks bc

garrysmod/lua/vgui/dcolorbutton.lua
```diff
@@ -27,9 +27,9 @@ function PANEL:IsDown()
 
 end
 
-function PANEL:SetColor( color, displayTooltip )
+function PANEL:SetColor( color, hideTooltip )
 
-	if ( displayTooltip == true or displayTooltip == nil ) then
+	if ( hideTooltip == false or hideTooltip == nil ) then
```
Alf21:
```lua
if ( hideTooltip ) then
```

Alphaverse: Wouldn't that change it so the tooltip is only visible when it is true? Shouldn't it he true for no Tooltip?

Kefta: I think he meant ``if ( !hideTooltip ) then``

Alphaverse: Yes that makes sense

PR #1585, opened 2019-05-15
Converted spaces to tabs

realpack:
Yes, I read CONTRIBUTING.md


thegrb93:
Spaces are bad. Keep the user from being able to choose tab size, inflate file size, and increase file process/load time. Don't.

thegrb93:
Looking at the changes, looks like you changed spaces to tabs. Your title is wrong

PR #1586, opened 2019-05-18
Do not discipline a weapon's new table type values when inheriting the BaseClass table

moatato:
Allow a weapon's BaseClass to have non-table values by using the defining weapon's new table type value instead of erroring and breaking weapons.


Kefta:
Already done in https://github.com/Facepunch/garrysmod/pull/1300/.

moatato:
eyuppp butt this pr has absolutely zero possible undesired subsequent side effects, it just fixes an error

PR #1587, opened 2019-06-06
Add ErrorNoHaltStack

Kefta:
Header: ErrorNoHaltStack(string message, number level = 1)

Like ``ErrorNoHalt`` except it adds a stack trace like ``error``. Unlike ``error``, however, the stack level argument will actually properly manipulate the stack trace as opposed to always displaying all levels. It will also not shorten the file source.


robotboy655:
Added ErrorNoHaltWithStack (...) from C side.

I know this PR technically has more options, but I prefer to keep stack related things in the same place, so if it changes, it changes everywhere. Plus its better this way for future game features I want to add.

PR #1590, opened 2019-06-21
Outline system for entities (not halo)

ShadowBonnieRUS:
Maybe you know that standard halo system greatly reduces FPS if called many times for example for other colors on entities, because of blur. And halo looks like glow. I make other system with better performance and looks like outline, not like glow (and still greatly visible on big distance).

Some screenshots:
image
image
image

I believe 28 fps is gone when drawing models to fill stencil buffer.

### DIFF
--- a/garrysmod/lua/includes/modules/outline.lua
+++ b/garrysmod/lua/includes/modules/outline.lua
@@ -1,0 +1,117 @@
+
+module("outline",package.seeall)
+
+local List = {}
+local RenderEnt = NULL
+
+local CopyMat		= Material("pp/copy")
+local OutlineMat	= CreateMaterial("OutlineMat","UnlitGeneric",{["$ignorez"] = 1,["$alphatest"] = 1})
+local StoreTexture	= render.GetScreenEffectTexture(0)
+local DrawTexture	= render.GetScreenEffectTexture(1)
+
+--[[-------------------------------
+	Modes:
+	0 - Always draw
+	1 - Draw if not visible
+	2 - Draw if visible
+--]]-------------------------------
+
+function Add(ents,color,mode)
+	if !istable(ents) then ents = {ents} end	--Support for passing Entity as first argument
+	if table.IsEmpty(ents) then return end		--Do not pass empty tables
+	if #List>=255 then return end				--Maximum 255 reference values
+	
+	table.insert(List,{
+		Ents = ents,
+		Color = color,
+		Mode = mode
+	})
+end
+
+function RenderedEntity()
+	return RenderEnt
+end
+
+local function Render()
+	local scene = render.GetRenderTarget()
+	render.CopyRenderTargetToTexture(StoreTexture)
+	
+	render.Clear(0,0,0,0,true,true)
+
+	render.SetStencilEnable(true)
+		cam.IgnoreZ(true)
+		render.SuppressEngineLighting(true)
+	
+		render.SetStencilWriteMask(255)
+		render.SetStencilTestMask(255)
+		
+		render.SetStencilCompareFunction(STENCIL_ALWAYS)
+		render.SetStencilFailOperation(STENCIL_KEEP)
+		render.SetStencilZFailOperation(STENCIL_REPLACE)
+		render.SetStencilPassOperation(STENCIL_REPLACE)
+		
+		cam.Start3D()
+			for k,v in ipairs(List) do
+				render.SetStencilReferenceValue(k)
+				
+				for k2,v2 in ipairs(v.Ents) do
+					if !IsValid(v2) then continue end
+					
+					local visible = LocalPlayer():IsLineOfSightClear(v2)
+					if v.Mode==1 and visible or v.Mode==2 and !visible then continue end
+					
+					RenderEnt = v2
+					v2:DrawModel()
+					RenderEnt = NULL
+				end
+			end
+		cam.End3D()
+		
+		render.SetStencilCompareFunction(STENCIL_EQUAL)
+		
+		cam.Start2D()
+			for k,v in ipairs(List) do
+				render.SetStencilReferenceValue(k)
+				
+				surface.SetDrawColor(v.Color)
+				surface.DrawRect(0,0,ScrW(),ScrH())
+			end
+		cam.End2D()
+		
+		render.SuppressEngineLighting(false)
+		cam.IgnoreZ(false)
+	render.SetStencilEnable(false)
+	
+	render.CopyRenderTargetToTexture(DrawTexture)
+	
+	render.SetRenderTarget(scene)
+	CopyMat:SetTexture("$basetexture",StoreTexture)
+	render.SetMaterial(CopyMat)
+	render.DrawScreenQuad()
+	
+	render.SetStencilEnable(true)
+		render.SetStencilReferenceValue(0)
+		render.SetStencilCompareFunction(STENCIL_EQUAL)
+		
+		OutlineMat:SetTexture("$basetexture",DrawTexture)
+		render.SetMaterial(OutlineMat)
+		
+		for x=-1,1 do
+			for y=-1,1 do
+				if x==0 and x==0 then continue end
+				
+				render.DrawScreenQuadEx(x,y,ScrW(),ScrH())
+			end
+		end
+	render.SetStencilEnable(false)
+end
+
+hook.Add("PostDrawEffects","RenderOutlines",function()
+	hook.Run("PreDrawOutlines")
+	
+	if #List==0 then return end
+	
+	Render()
+	
+	List = {}
+end)
--- a/garrysmod/lua/includes/modules/outline.lua
+++ b/garrysmod/lua/includes/modules/outline.lua
@@ -114,7 +114,7 @@ local function Render()
 	
 	local w, h = ScrW(), ScrH()
 	
-	render.Clear( 0, 0, 0, 0, false, true )
+	render.ClearStencil()
 	
 	render.SetStencilEnable( true )
 		render.SuppressEngineLighting( true )
@@ -126,7 +126,7 @@ local function Render()
 		render.SetStencilFailOperation( STENCIL_KEEP )
 		
 		cam.Start3D()
-			render.SetBlend(0)
+			render.SetBlend(1)
 			
 			for i = 1, ListSize do
 				
@@ -205,6 +205,8 @@ local function Render()
 		render.SetStencilCompareFunction( STENCIL_EQUAL )
 		render.SetStencilZFailOperation( STENCIL_KEEP )
 		render.SetStencilPassOperation( STENCIL_KEEP )
+		
+		render.Clear( 0, 0, 0, 0, false, false )
 		
 		cam.Start2D()
 			
--- a/garrysmod/lua/includes/modules/outline.lua
+++ b/garrysmod/lua/includes/modules/outline.lua
@@ -95,7 +95,7 @@ end
 function SetDoubleThickness( thickness )
 
 	local old_thickness = OutlineThickness == 2
-	OutlineThickness = thickness and 2 or 1
+	OutlineThickness = thickness && 2 || 1
 	
 	return old_thickness
 	
@@ -140,7 +140,7 @@ local function Render()
 				
 				if ( mode == OUTLINE_MODE_BOTH || mode == OUTLINE_MODE_VISIBLE ) then
 					
-					render.SetStencilZFailOperation( mode == OUTLINE_MODE_BOTH and STENCIL_REPLACE or STENCIL_KEEP )
+					render.SetStencilZFailOperation( mode == OUTLINE_MODE_BOTH && STENCIL_REPLACE || STENCIL_KEEP )
 					render.SetStencilPassOperation( STENCIL_REPLACE )
 					
 					for j = 1, #ents do
--- a/garrysmod/lua/includes/modules/outline.lua
+++ b/garrysmod/lua/includes/modules/outline.lua
@@ -1,128 +1,227 @@
-
-OUTLINE_MODE_BOTH	= 0
+OUTLINE_MODE_BOTH		= 0		-- Render always
 OUTLINE_MODE_NOTVISIBLE	= 1
 OUTLINE_MODE_VISIBLE	= 2
 
-local istable		= istable
-local render		= render
-local LocalPlayer	= LocalPlayer
-local Material		= Material
+OUTLINE_RENDERTYPE_BEFORE_VM	= 0		-- Render before drawing the view model
+OUTLINE_RENDERTYPE_BEFORE_EF	= 1		-- Render before drawing all effects (after drawing the viewmodel)
+OUTLINE_RENDERTYPE_AFTER_EF		= 2		-- Render after drawing all effects
+
+local istable			= istable
+local render			= render
+local Material			= Material
 local CreateMaterial	= CreateMaterial
-local hook		= hook
-local cam		= cam
-local ScrW		= ScrW
-local ScrH		= ScrH
-local IsValid		= IsValid
-local surface		= surface
+local hook				= hook
+local cam				= cam
+local ScrW				= ScrW
+local ScrH				= ScrH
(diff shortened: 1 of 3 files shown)


garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,117 @@
...
+	render.SetStencilEnable(true)
+		render.SetStencilReferenceValue(0)
+		render.SetStencilCompareFunction(STENCIL_EQUAL)
+		
+		OutlineMat:SetTexture("$basetexture",DrawTexture)
+		render.SetMaterial(OutlineMat)
+		
+		for x=-1,1 do
+			for y=-1,1 do
+				if x==0 and x==0 then continue end
```
Kefta: Why the same check twice? I think this loop should just be unrolled and you should cache ScrW and ScrH

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,117 @@
...
+	Modes:
+	0 - Always draw
+	1 - Draw if not visible
+	2 - Draw if visible
+--]]-------------------------------
+
+function Add(ents,color,mode)
+	if !istable(ents) then ents = {ents} end	--Support for passing Entity as first argument
+	if table.IsEmpty(ents) then return end		--Do not pass empty tables
+	if #List>=255 then return end				--Maximum 255 reference values
```
Kefta: Why have this limit?

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,117 @@
...
+	
+	table.insert(List,{
```
Kefta: You can use numerical/enum keys instead of string keys here for more efficient access. So you would add
```
local ENTS = 1
local COLOR = 2
local MODE = 3
```
and use those for indexing.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,117 @@
...
+				render.DrawScreenQuadEx(x,y,ScrW(),ScrH())
+			end
+		end
+	render.SetStencilEnable(false)
+end
+
+hook.Add("PostDrawEffects","RenderOutlines",function()
+	hook.Run("PreDrawOutlines")
+	
+	if #List==0 then return end
```
Kefta: Since you're already checking the length of List here, you can pass it to the Render function and use it for iteration instead of ipairs. Or have a length variable that you increment in outline.Add and use instead of generating the length every time.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,117 @@
...
+--[[-------------------------------
+	Modes:
+	0 - Always draw
+	1 - Draw if not visible
+	2 - Draw if visible
+--]]-------------------------------
+
+function Add(ents,color,mode)
+	if !istable(ents) then ents = {ents} end	--Support for passing Entity as first argument
+	if table.IsEmpty(ents) then return end		--Do not pass empty tables
```
Kefta: table.IsEmpty returns false for tables like ``{foo = 5}`` but you only use numerical keys for iteration. You should check if ents[1] is not nil, instead.

garrysmod/lua/includes/modules/outline.lua
ShadowBonnieRUS: should be "x==0 and y==0"

garrysmod/lua/includes/modules/outline.lua
ShadowBonnieRUS: As i know with write mask maximum 255 ref values (mask use 8-bit bitflag)

garrysmod/lua/includes/modules/outline.lua
ShadowBonnieRUS: that method used in halo module, i copy that.

Kefta: That is a mistake then, I will fix that.

garrysmod/lua/includes/modules/outline.lua
ShadowBonnieRUS: is it so important? string keys used in halo module

garrysmod/lua/includes/modules/outline.lua
Kefta: Actually, halos use ``pairs`` on the Ents table so it iterates over non-numerical keys. You don't do that here, you use ``ipairs``

garrysmod/lua/includes/modules/outline.lua
Kefta: Just a small optimisation since halos are already so taxing.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,120 @@
...
+local List,ListSize = {},0
+local RenderEnt = NULL
+
+local CopyMat		= Material("pp/copy")
+local OutlineMat	= CreateMaterial("OutlineMat","UnlitGeneric",{["$ignorez"] = 1,["$alphatest"] = 1})
+local StoreTexture	= render.GetScreenEffectTexture(0)
+local DrawTexture	= render.GetScreenEffectTexture(1)
+
+local ENTS,COLOR,MODE = 1,2,3
+local MODE_NOTVISIBLE,MODE_VISIBLE = 1,2
```
Kefta: I would make these global and name them ``RENDER_MODE_*`` for usage outside of this file. You should also add a ``RENDER_MODE_BOTH`` and make it the default if a mode is not specified.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,120 @@
...
+		render.SetStencilWriteMask(255)
+		render.SetStencilTestMask(255)
+		
+		render.SetStencilCompareFunction(STENCIL_ALWAYS)
+		render.SetStencilFailOperation(STENCIL_KEEP)
+		render.SetStencilZFailOperation(STENCIL_REPLACE)
+		render.SetStencilPassOperation(STENCIL_REPLACE)
+		
+		cam.Start3D()
+			for k,v in ipairs(List) do
```
Kefta: Now that you have a ListSize variable, you can make this loop
```
for i = 1, ListSize do
   local v = List[i]
   render.SetStencilReferenceValue(i)
   ...
```

for slightly more efficient iteration. Same with the cam 2D loop.

garrysmod/lua/includes/modules/outline.lua
ShadowBonnieRUS: maybe `OUTLINE_MODE_*` will be more correct

Kefta: Whoops that's what I meant

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,123 @@
...
+		render.SetStencilPassOperation(STENCIL_REPLACE)
+		
+		cam.Start3D()
+			for k=1,ListSize do
+				local v = List[k]
+				local mode = v[MODE]
+				
+				render.SetStencilReferenceValue(k)
+				
+				for k2=1,#v[ENTS] do
```
Kefta: You can also cache the v[ENTS] access instead of doing it on every entity iteration.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,124 @@
...
+					local v2 = v[ENTS][k2]
+					
+					if !IsValid(v2) then continue end
+					
+					local visible = LocalPlayer():IsLineOfSightClear(v2)
```
Kefta: Since you now have OUTLINE_MODE_BOTH, you can skip this trace if the mode is not OUTLINE_MODE_NOTVISIBLE nor OUTLINE_MODE_VISIBLE.

ShadowBonnieRUS: I think about it, but what if mode value is not equal to `both`,`notvisible`,`visible`, now there with checks used `both`, if not `notvisible` and not `visible`.

Kefta: I think you should treat all non-``nonvisible`` and ``visible`` values as both.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,125 @@
...
+		
+		cam.Start3D()
+			for k=1,ListSize do
+				local v = List[k]
+				local mode = v[MODE]
+				
+				render.SetStencilReferenceValue(k)
+				
+				local ents = v[ENTS]
+				local entssize = #ents
```
Kefta: Just nitpicking at this point, but this doesn't need it's own variable since it's only used for the loop header.

Kefta:
Thanks for all the changes, I think this will now be much more efficient than halos. The code style does still need some cleaning up, though:  it doesn't match the rest of the codebase. There should be spaces around operators (``local List, ListSize = {}, 0``), logical operators should use the C versions (``[MODE] = mode || OUTLINE_MODE_BOTH``), and there should be spaces between parentheses and arguments (``render.SetStencilPassOperation( STENCIL_REPLACE )``).

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,150 @@
...
+	if ( ListSize >= 255 ) then return end				--Maximum 255 reference values
+	if ( !istable( ents ) ) then ents = { ents } end	--Support for passing Entity as first argument
+	if ( ents[ 1 ] == nil ) then return end				--Do not pass empty tables
+	
+	local t = {
+		[ ENTS ] = ents,
+		[ COLOR ] = color,
+		[ MODE ] = mode or OUTLINE_MODE_BOTH
+	}
+	table.insert( List, t )
```
Alf21: You could improve this by

```lua
ListSize = ListSize + 1
List[ListSize] = t
```

Alf21:
How about just adding the `PostDrawEffects` hook for the outline drawing if `ListSize ~= 0`? This should reduce the number of hooks called.

garrysmod/lua/includes/modules/outline.lua
```diff
@@ -0,0 +1,150 @@
...
+				local ents = v[ ENTS ]
+				
+				render.SetStencilReferenceValue( i )
+		
+				for j = 1, #ents do
+					local ent = ents[ j ]
+					
+					if ( !IsValid( ent ) ) then continue end
+					
+					if ( ( mode == OUTLINE_MODE_NOTVISIBLE && LocalPlayer():IsLineOfSightClear( ent ) ) || ( mode == OUTLINE_MODE_VISIBLE && !LocalPlayer():IsLineOfSightClear( ent ) ) ) then continue end
```
Alf21: LocalPlayer should be cached outside the loop too

Kefta:
@Alf21 I don't think that's a good idea. Many hook library replacements put a huge load on hook.Add in order to make hook.Call faster. Calling one extra function and doing a length check is really not going to have any noticeable difference on performance.

garrysmod/lua/includes/modules/outline.lua
Kefta: You can even cache the IsLineOfSightClear function to prevent extra __index calls. Ex.

```Lua
-- Outside of loop
local pPlayer = LocalPlayer()
local IsLineOfSightClear = pPlayer.IsLineOfSightClear

-- Inside of loop
if ( ( mode == OUTLINE_MODE_NOTVISIBLE && IsLineOfSightClear( pPlayer, ent ) ) || ( mode == OUTLINE_MODE_VISIBLE && !IsLineOfSightClear( pPlayer, ent ) ) ) then continue end
```

UAVXP:
Would like to see the FPS changes after those optimizations aswell) Compared to your original halo version of course

ShadowBonnieRUS:
@UAVXP all this add 1 fps to me.

GitSparTV:
Can't relate it's significantly faster for me

FlorianLeChat:
The only small issue I've had since I've been using this library is the effect is even displayed through the player model (in comparison with the halo library). If players want to use this library as a replacement for better performance, this problem needs to be fixed.

Halo library:
image
Outline library:
image

FlorianLeChat:
The only small issue I've had since I've been using this library is the effect is even displayed through the player model (in comparison with the halo library). If players want to use this library as a replacement for better performance, this problem needs to be fixed.

Halo library:
image
Outline library:
image

UAVXP:
> the effect is even displayed through the player model
Don't think it's a much of a problem personally

FoksVHox:
> > the effect is even displayed through the player model
> 
> I personally do not think it's a much of a problem

It would still be nice if it got fixed

Kefta:
> > the effect is even displayed through the player model
> 
> I'm personally do not think it's a much of a problem

It must be fixed if this were to be merged. Better yet, make it a config option if it shows thru players.

ShadowBonnieRUS:
> > > the effect is even displayed through the player model
> > 
> > 
> > I'm personally do not think it's a much of a problem
> 
> It must be fixed if this were to be merged. Better yet, make it a config option if it shows thru players.

The problem is that used IsLineOfSightClear instead of checking the depth buffer. It does not care about where render origin is. I used IsLineOfSightClear because of problems with the depth buffer when I tried to do this first time. I will try again make this with the depth fuffer.

ShadowBonnieRUS:
So now it looks like this. Front canister (red) uses OUTLINE_MODE_VISIBLE mode. Other canister (cyan) uses OUTLINE_MODE_NOTVISIBLE mode.
image

WilliamVenner:
I can't seem to get the outlines to draw for some transparent props. For example:

`models/props_building_details/Storefront_Template001a_Bars.mdl`

`models/props_junk/PlasticCrate01a.mdl`

ShadowBonnieRUS:
> I can't seem to get the outlines to draw for some transparent props. For example:
> 
> `models/props_building_details/Storefront_Template001a_Bars.mdl`
> 
> `models/props_junk/PlasticCrate01a.mdl`

Should be fixed. Seems `render.SetBlend (0)` has a bad effect on the depth buffer when rendering transparent objects.

Andrew-Eathan:
is this going anywhere? the current halo system is just absolutely awful for performance

FlorianLeChat:
You can just include this library in your code for now, I have never had a problem in more than a year of use.

thegrb93:
Could be a performance option

CosmoKz:
New to E2, but looking for a way to do this but on MY playermodel to where everyone else can see it too. I've seen other people using one that does a color changing outline on them if anyone knows where something like that is at, because I am too awful to know how.

blobles-dev:
'Broken' as of the most recent gmod update
After using the outline.Add on an entity, the physgun halo effects can be seen through walls
https://media.discordapp.net/attachments/589120351238225940/1125558175316312124/image.png

Rubat is inclined to believe it was one of these highlighted changes
https://media.discordapp.net/attachments/589120351238225940/1125557605398482974/image.png

robotboy655 (changes requested):
I am not opposed to this, but not as a separate library. It should be part of the `halo` library.

The "render type" and "thickness" should also be per-outline, not as a global setting, global settings just will cause addons conflicting with each other on what type/thickness they each want.

ShadowBonnieRUS:
Thickness cannot be changed to be per-outline, because of how the end outline is applied (drawing colored texture with specified offsets, 'thickness' will just increase the offsets). Maybe thickness functionality is unnecessary at all?

For render type, "SetupOutlines" should be called for each render type with parameter indicating the current render type, at the moment, 3 times per frame. It may affect performance.

I'm not sure how it would be more correct to implement this in the halo library, since on the one hand, they do the same thing, and on the other hand, they work differently.

Kefta:
Why was this closed?

robotboy655:
Clearly there's an impasse in how I would expect (and I am going to assume everyone else as well) this to work, and how it actually works, in regards to the library's global settings. Given authors inability/unwillingness to conform to requested changes, I see no reason to keep this open. It will not be merged in its current state.

PR #1591, opened 2019-06-23
Optimizing the Material function

Venatuss-S:
By this way, we cache the material in a table, so we can directly use it in huge hooks as the HUDPaint one. It prevents all bad uses, especially of new developpers.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -14,9 +14,17 @@ Format = string.format
 --
 local C_Material = Material
 
+local tMaterials = {}
+
 function Material( name, words )
 
-	if ( !words ) then return C_Material( name ) end
+	if ( !words ) then
+		if not tMaterials[ name ] then
+			tMaterials[ name ] = C_Material( name )
+		end
+		
+		return tMaterials[ name ]
+	end
 
 	local str = (words:find("vertexlitgeneric") and "1" or "0")
 	str = str .. (words:find("nocull") and "1" or "0")
@@ -24,8 +32,12 @@ function Material( name, words )
 	str = str .. (words:find("mips") and "1" or "0")
 	str = str .. (words:find("noclamp") and "1" or "0")
 	str = str .. (words:find("smooth") and "1" or "0")
+	
+	if not tMaterials[ name .. str ] then
+		tMaterials[ name .. str ] = C_Material( name, str )
+	end
 
-	return C_Material( name, str )
+	return tMaterials[ name .. str ]
 
 end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
Should only cache if the material exists.

garrysmod/lua/includes/util.lua
```diff
@@ -14,18 +14,30 @@ Format = string.format
 --
 local C_Material = Material
 
+local tMaterials = {}
+
 function Material( name, words )
 
-	if ( !words ) then return C_Material( name ) end
+	if ( !words ) then
+		if not tMaterials[ name ] then
```
kotus35:
```suggestion
		local material = C_Material( name )
		if ( not tMaterials[ name ] and not material:IsError() ) then
```

garrysmod/lua/includes/util.lua
```diff
@@ -14,18 +14,30 @@ Format = string.format
...
+
 function Material( name, words )
 
-	if ( !words ) then return C_Material( name ) end
+	if ( !words ) then
+		if not tMaterials[ name ] then
+			tMaterials[ name ] = C_Material( name )
+		end
+		
+		return tMaterials[ name ]
```
kotus35:
```suggestion
		return material:IsError() and material or tMaterials[ name ]
```

garrysmod/lua/includes/util.lua
```diff
@@ -14,18 +14,30 @@ Format = string.format
...
+	end
 
 	local str = (words:find("vertexlitgeneric") and "1" or "0")
 	str = str .. (words:find("nocull") and "1" or "0")
 	str = str .. (words:find("alphatest") and "1" or "0")
 	str = str .. (words:find("mips") and "1" or "0")
 	str = str .. (words:find("noclamp") and "1" or "0")
 	str = str .. (words:find("smooth") and "1" or "0")
+	
+	if not tMaterials[ name .. str ] then
```
kotus35:
```suggestion
	local material = C_Material( name, str )
	if ( not tMaterials[ name .. str ] and not material:IsError() ) then
```

garrysmod/lua/includes/util.lua
```diff
@@ -14,18 +14,30 @@ Format = string.format
...
+		tMaterials[ name .. str ] = C_Material( name, str )
+	end
 
-	return C_Material( name, str )
+	return tMaterials[ name .. str ]
```
kotus35:
```suggestion
	return material:IsError() and material or tMaterials[ name .. str ]
```

ExtReMLapin:
Are we really talking about pushing code only because some devs are fucking retarded ?

Kefta:
This does indeed break manual reloading. It'd still be possible by calling C_Material manually or adding an argument to bypass the cache, but I think it would be better to make the cache opt-in instead of forced.

PR #1592, opened 2019-07-06
Added custom PANEL:SetPos(x,y) function

Titanic4:
The PANEL:SetPos(x,y) doesn't check if the DFrame is within screen boundaries by normal means. That was allowing setting position of the DFrame way out of bounds of the screen, making the player unable to close/move the modal window since its title bar was completely unreachable. My version of the function will actually check if the DFrame will be within screen boundaries when using SetPos function. If it isn't, it will close the DFrame, which would be otherwise out of bounds and throw the error showing the destination position of the DFrame in the error message.

### DIFF
--- a/garrysmod/lua/vgui/dframe.lua
+++ b/garrysmod/lua/vgui/dframe.lua
@@ -194,6 +194,16 @@ function PANEL:Think()
 
 end
 
+function PANEL:SetPos(x,y)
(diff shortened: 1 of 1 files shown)


robotboy655:
No, just no, especially with a solution like that (closing the window and erroring). This is a problem the coder should solve (i.e. don't create panels outside of screen bounds) in their code.

PR #1597, opened 2019-08-26
Caching Angle and Vector

ExtReMLapin:
The whole thing is a mess, the TTT code is total shit, it may need more commits to fix the whole thing


ExtReMLapin:
The ttt code is a FUCKING MESS

ExtReMLapin:
I could really use some help on it, I also need to do it with Color.

I got a regex to do it but the whole thing should be automated.

The tricky part is to find IF a vector can be set as local (or global) or not since it may be modified directly why the functions that use it.
Here is the regex I use to find the Vectors/Colors/Angles


```
Color\(\s*\d+\s*,\s*\d+\s*,\s*\d+\s*(,\s*\d+\s*)?\)

Vector\(\s*-?\d+(?:\.\d+)?\s*,\s*-?\d+(?:\.\d+)?\s*,\s*-?\d+(?:\.\d+)?\s*\)

Angle\(\s*-?\d+(?:\.\d+)?\s*,\s*-?\d+(?:\.\d+)?\s*,\s*-?\d+(?:\.\d+)?\s*\)
```

robotboy655:
Don't bundle TTT and non TTT changes into single PR.

Please read the contributing.md file:

robotboy655:
Also replacements to vector_origin, etc are easier to swallow than the addition of new globals/locals for everything.

Donkie:
I'm not entirely familiar with metatables, but wouldn't it be possible to set a custom metatable __index on the vector_origin, angle_zero etc constants to prevent them from being edited?

ExtReMLapin:
Fixed the naming a bit, but i don't see how it's really a problem, feel free to do a pull request here if you got a better idea https://github.com/ExtReMLapin/garrysmod/tree/objects-caching

ExtReMLapin:
Sorry QA guy

Kefta:
Ok

PR #1600, opened 2019-09-04
PlayerShouldTakeDamage is server-side only

CaptainPRICE:
It doesn't belong in the shared file...


ExtReMLapin:
what about future preduction maybe ?

Kefta:
Damage prediction is never getting added lol

PR #1601, opened 2019-09-04
Remove CanProperty from the base gamemode

CaptainPRICE:
I don't think it belongs there. Why is it there anyways?

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/shared.lua
+++ b/garrysmod/gamemodes/base/gamemode/shared.lua
@@ -264,10 +264,3 @@ function GM:OnViewModelChanged( vm, old, new )
 	end
 
 end
-
---[[---------------------------------------------------------
-	Disable properties serverside for all non-sandbox derived gamemodes.
------------------------------------------------------------]]
-function GM:CanProperty( pl, property, ent )
-	return false
-end


CaptainPRICE:
[It is Sandbox thing](https://wiki.garrysmod.com/page/SANDBOX/CanProperty).

Kefta:
This should be moved to Sandbox's shared.lua.

CaptainPRICE:
@Kefta 
> This should be moved to Sandbox's shared.lua.

huh wut?
But, it is already present there:
https://github.com/Facepunch/garrysmod/blob/b03cc1ed585eeb037e9f7723fc29ef4f570e1782/garrysmod/gamemodes/sandbox/gamemode/shared.lua#L201-L252

Kefta:
Huh, must have missed that.

robotboy655:
The properties system exists outside of Sandbox (In the `properties` module, and all built-in properties are loaded regardless of which gamemode is running, as well as any addon properties), so it only makes sense for clearly defined behavior for non Sandbox gamemodes to exist in the Base gamemode as well. 

It is kinda made even worse by the fact that `CanProperty` is called manually from each property and they test for "is allowed" in different ways.

PR #1604, opened 2019-09-05
Remove unused argument from the PhysgunUnfreeze call

CaptainPRICE:
[`Player.PhysgunUnfreeze`](https://wiki.garrysmod.com/page/Player/PhysgunUnfreeze) doesn't take any arguments.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/player.lua
+++ b/garrysmod/gamemodes/base/gamemode/player.lua
@@ -39,7 +39,7 @@ end
 -----------------------------------------------------------]]
 function GM:OnPhysgunReload( weapon, ply )
 
-	ply:PhysgunUnfreeze( weapon )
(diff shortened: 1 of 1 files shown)


robotboy655:
`Player.PhysgunUnfreeze` does have an unused argument: https://github.com/Facepunch/garrysmod/blob/0815ce5bff21210bea50c27c47ef53f652318a7f/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua#L57

But it is used without an argument in Sandbox:
https://github.com/Facepunch/garrysmod/blob/4a66ffa5f5836a3fc703f0df6a814ebe49758b18/garrysmod/gamemodes/sandbox/gamemode/init.lua#L67

Kefta:
The argument should be removed from PhysgunUnfreeze's header, as well. It isn't used anywhere.

PR #1605, opened 2019-09-05
Refactor net messaging for PlayerDeath (in base gamemode)

CaptainPRICE:
Fixes [1407](https://github.com/Facepunch/garrysmod-requests/issues/1407)

Note: This change has a potential of breaking client-side scripts (which rely on these specific net messages; `PlayerKilledSelf` and `PlayerKilledByPlayer`).

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_deathnotice.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_deathnotice.lua
@@ -39,38 +39,46 @@ local function PlayerIDOrNameToString( var )
 
 end
 
+local RecvPlayerDeath = {
+	-- PlayerKilled
+	[0] = function()
 
-local function RecvPlayerKilledByPlayer()
+		local victim	= net.ReadEntity()
+		if ( !IsValid( victim ) ) then return end
+		local inflictor	= net.ReadString()
+		local attacker	= "#" .. net.ReadString()
 
-	local victim	= net.ReadEntity()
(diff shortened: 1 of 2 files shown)


CaptainPRICE:
Review please.

Kefta:
I do not believe this is a beneficial change. I think saving net overhead > saving net string slots unless it becomes excessive (5-10+ for a single purpose depending on the overhead saved). In this case, saving 2 slots is not better than saving 2 bits on every stock death-related net message.

CaptainPRICE:
There're another 2 pooled strings, [for NPC death](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/base/gamemode/npc.lua).
If I were to update it to include these, it would be saving 4 slots in total.
IIRC, there's a hardlimit of 2048 slots. Imagine a server running at the edge, saving on such slots is a bless in that case.
Any opinions?

Kefta:
I think the slot limit should be raised, or better yet, the net implementation should use a shared vector that will adjust the size of the net ID at runtime, removing the limit. This adds the additional benefit of the net overhead actually being smaller if you pool less strings than the current limit.

A similar system can be made in Lua by having a synced hashtable of IDs, but it has to append an extra n bits on top of the net library's header, so it'd be better for the change to be made internally.

CaptainPRICE:
@Kefta 

I'd agree with that being a much better way if implemented internally.
Go ahead and make it a proposal.

PR #1606, opened 2019-09-05
Removed a similar condition

JarosLucky:
I think this is an unnecessary condition, and I not understand what it was for. since !nil = true and !false = true. Also I could not find a comment for what it might be. If I'm wrong, please announce this decision)


neico:
The only thing that would make sense if it were `self.GetChecked != nil` (function existence, not uncommon for multiple inheritance class systems) or `self.Button != nil`
Otherwise it does hardly make sense

PR #1607, opened 2019-09-06
DLabelEditable:GetDTextEntry()

art-gur:
With such function, you can easily edit DTextEntry without calling PANEL:GetChildren() on DLabelEditable and without extracting DTextEntry from children table.


robotboy655:
Lets talk about why you need this?

art-gur:
For customize DTextEntry, now you can

robotboy655:
Use Derma skins and https://wiki.garrysmod.com/page/GM/ForceDermaSkin

art-gur:
Okay, but how about another situation, for example, disable language indicator on DTextEntry?

PR #1608, opened 2019-09-11
cycle through ipairs from pairs

JarosLucky:
viewing an array through ipairs instead of pairs. since the table created with table.insert.


mcNuggets1:
thats one out of 100 tables which doesnt use ipairs.

JarosLucky:
Hmm, I think so too, but I usually create commits when I accidentally find
strange solutions when I search for something.

чт, 12 сент. 2019 г., 12:01 David <notifications@github.com>:

PR #1573, opened 2019-03-29
Replace usage of pairs() with ipairs()

aStonedPenguin:
ipairs is faster than pairs and it is good practice to use where applicable.

PR #1610, opened 2019-09-15
pairs -> ipairs

JarosLucky:
in places where the table is only a numeric orderly table

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_cse_proj.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_cse_proj.lua
@@ -47,9 +47,8 @@ function ENT:GetNearbyCorpses()
 
    local near_corpses = {}
 
-   local n = #near
    local ent = nil
-   for i=1, n do
+   for i=1, #near do
       ent = near[i]
       if IsValid(ent) and ent.player_ragdoll and ent.scene then
          table.insert(near_corpses, {ent=ent, dist=pos:LengthSqr()})
@@ -93,7 +92,7 @@ function ENT:ShowSceneForCorpse(corpse)
 
    if not scene then return end
 
-   for _, dummy_key in pairs({"victim", "killer"}) do
+   for _, dummy_key in ipairs({"victim", "killer"}) do
       local dummy = scene[dummy_key]
 
       if dummy then
@@ -203,5 +202,3 @@ function ENT:Explode(tr)
 
    end
 end
-
-
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_cse_proj.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_cse_proj.lua
@@ -73,6 +73,7 @@ function ENT:OnTakeDamage(dmginfo)
    end
 end
 
+local dummy_keys = {"victim", "killer"}
 function ENT:ShowSceneForCorpse(corpse)
    local scene = corpse.scene
    local hit = scene.hit_trace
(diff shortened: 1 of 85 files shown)


JarosLucky:
This TTT

garrysmod/gamemodes/terrortown/entities/entities/ttt_cse_proj.lua
```diff
@@ -93,7 +92,7 @@ function ENT:ShowSceneForCorpse(corpse)
 
    if not scene then return end
 
-   for _, dummy_key in pairs({"victim", "killer"}) do
+   for _, dummy_key in ipairs({"victim", "killer"}) do
```
ExtReMLapin: let's create a table everytime yolo

garrysmod/gamemodes/terrortown/gamemode/cl_equip.lua
```diff
@@ -310,7 +310,7 @@ local function TraitorMenuPopup()
    dinfo:StretchToParent(0, 0, 0, dih - 135)
 
    local dfields = {}
-   for _, k in pairs({"name", "type", "desc"}) do
+   for _, k in ipairs({"name", "type", "desc"}) do
```
ExtReMLapin: same here

ExtReMLapin:
Fucking hell the TTT is full of garbage code, I don't get why it's even shipped with gmod at this point, we can't fix the stupid [3 SPACES] indentation else mister TTT maintainer turns mad.

There is so much things to fix in it and yet cpt baby expect us to respect his stupid indentation.

garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
```diff
@@ -182,7 +182,7 @@ if CLIENT then
    };
 
    -- Insert all the "simple" gestures that do not need weight control
-   for _, a in pairs{ACT_GMOD_GESTURE_AGREE, ACT_GMOD_GESTURE_DISAGREE,
+   for _, a in ipairs{ACT_GMOD_GESTURE_AGREE, ACT_GMOD_GESTURE_DISAGREE,
```
ExtReMLapin: lord

JarosLucky: in this place, the local table will only take up memory, this action takes place once in 1 start of the file.

JarosLucky:
I don't know what you mean, this place needs to be done the way I did before?

ExtReMLapin:
bruh

JarosLucky:
?

svdm:
I still don't believe this is going to make a performance difference for anyone, but this was a pretty clean PR so it's merged.

JarosLucky:
ipairs much easier compared to pairs. This is confirmed by tests of performance and source code of Lua.

Donkie:
ipairs vs pairs has been tested hundreds of times throughout gmods history and the conclusion is always that the difference is extremely negligible.

JarosLucky:
Ipairs - 4 lines of code
Pairs - 3 function

вт, 17 сент. 2019 г., 2:07 Daniel Hultgren <notifications@github.com>:

Kefta:
>ipairs vs pairs has been tested hundreds of times throughout gmods history and the conclusion is always that the difference is extremely negligible.

Most of the famous pairs vs ipairs tests/FP threads were without LuaJIT. It really makes a significant difference now: https://github.com/Facepunch/garrysmod/pull/1573#issuecomment-485029892

Using numeric-for instead of ipairs would be even a little more efficient, but the difference there is truly negligible in most cases.

svdm:
My point was more that using ipairs or numeric-for in hotspots makes sense (and TTT already does in its Tick hook for example), while the changes here concern loops that are too infrequently called and too small to make a real-world difference.

Of course it can't hurt, and we're talking about this in a merged PR thread, but it's not the most rational change to make if you offset a risk of bugs against tangible improvement.

PR #1612, opened 2019-09-27
Allow numerical indexing of colors

Kefta:
- Allows color[i] for indexing elements
- Prevents any keys other than r, g, b, and a from being modified on the color object
- Prevents setting color components to non-numerical types if the key was set to nil prior. This is the best that can be done with type safety in Lua without aliasing the Color object since __newindex isn't called for keys that already exist in the table, and the object can't be aliased without breaking scripts calling ``rawget``/``rawset`` on colors.
- Don't allow negative color components, clamp between [0, 255]
- Make 255 the default for every argument of ``Color``

Relies on https://github.com/Facepunch/garrysmod/pull/1587.

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -40,9 +40,9 @@ debug.getregistry().Color = COLOR
 -----------------------------------------------------------]]
 function Color( r, g, b, a )
 	return setmetatable( {
-		r = math.max( math.min( tonumber(r), 255 ), 0 ),
-		g = math.max( math.min( tonumber(g), 255 ), 0 ),
-		b = math.max( math.min( tonumber(b), 255 ), 0 ),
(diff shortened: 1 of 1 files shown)


SimonSchick:
Doesn't this make the regular indexing slower, what's do developers gain from being able to use numerical indexing?

Kefta:
To be consistent with Vectors and Angles. Indexing color objects is already pretty uncommon from the lack of member functions, but considering https://github.com/Facepunch/garrysmod-issues/issues/2407 has existed for the longest time (prevents indexing engine colors), I don't think this will have any major negative performance consequences.

Really this is just a temporary solution until Colors are moved to C++.

Kefta:
I should also mention this does not affect the speed or regular key indexing (``color.r``, ``color.g``, ``color.b``, ``color.a``) since __index is only called if the key doesn't exist in the table.

PR #1533, opened 2018-10-19
Create markup.Escape

WilliamVenner:
Quick helper function for escaping markup text (thanks to lines 283-285!)

PR #1614, opened 2019-10-09
Create markup.Escape

WilliamVenner:
Quick helper function for escaping markup text (thanks to lines 283-285!)

_Moved from #1533 for branch change_


CornerPin:
You can use a table as a _replacement_ argument, so you don't have to call `gsub` three times.
```lua
local escapeEntities = {
	["&"] = "&amp;",
	["<"] = "&lt;",
	[">"] = "&gt;"
}

function Escape( str )
	return (tostring(str):gsub("[&<>]", escapeEntities))
end
```

PR #1616, opened 2019-10-24
GetConVarNumber deprecated

Kumo4031:
Updated to ConVar object.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/player.lua
+++ b/garrysmod/gamemodes/base/gamemode/player.lua
@@ -736,9 +736,11 @@ end
 	Name: gamemode:GetFallDamage()
 	Desc: return amount of damage to do due to fall
 -----------------------------------------------------------]]
+local mp_falldamage = GetConVar( "mp_falldamage" )
+
 function GM:GetFallDamage( ply, flFallSpeed )
 
-	if( GetConVarNumber( "mp_falldamage" ) > 0 ) then -- realistic fall damage is on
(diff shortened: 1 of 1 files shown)


Kefta:
I know the original used GetConVarNumber, but you should use GetBool here, instead, as it's only checking if the convar is enabled.

Kumo4031:
> I know the original used GetConVarNumber, but you should use GetBool here, instead, as it's only checking if the convar is enabled.

True, don't know why I didn't think of that. Fixed in https://github.com/Facepunch/garrysmod/pull/1616/commits/f8ff79a1930258bf7d87503497f6f71956d27cb5

Kefta:
You still need the outer if-statement parentheses for codebase consistency. Should be ``if ( mp_falldamage:GetBool() ) then``

Kumo4031:
> You still need the outer if-statement parentheses for codebase consistency. Should be `if ( mp_falldamage:GetBool() ) then`

Done.

robotboy655:
This is not a performance problematic place to get rid of `GetConVarNumber`  and most of performance issues with `GetConVarNumber`  were resolved by caching the internal convar objects.

I am considering undeprecating `GetConVar*` functions and adding a note about performance.

Changes like this just uglify the code for no benefit.

Kefta:
This still saves a hashtable lookup every time. I think they should stay deprecated.

robotboy655:
Does it matter in this particular case though?

Kefta:
Not particularly.

PR #1617, opened 2019-11-01
[TTT] Using table.IsEmpty

realpack:


bigdogmat:
None of these are optimizations, just a different and probably slower way of checking if a sequential table is empty.

thegrb93:
isEmpty is only good for replacing table.Count(tbl)==0

ExtReMLapin:
If it was a few % difference i wouldn't complain here but clearly the isempty call is fucking slower than using the sequential table counter.

# :	0.074
isempty: 	2.8

realpack:
I didn’t know that everything was so bad.

PR #1618, opened 2019-11-10
Kill nonexistent console variable assignments.

timschumi:
Those appear to not exist in Garry's Mod, neither in the client not in a server.


Kefta:
I requested them to be added in https://github.com/Facepunch/garrysmod-requests/issues/1150, and provided a potential solution to Rubat.

timschumi:
All I can see there is a ticket that hasn't received any activity in over a year. I thought that this solution may be more desirable, so I put it up for discussion.

PR #1619, opened 2019-11-22
Added: Function `GetOrthogonal` for easier othogonalization of a 2 ve…

dvdvideo1234:
…ctors set

Utilize: vector/angle unpack in `TypeToString`

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -126,8 +126,8 @@ function util.TypeToString( v )
 	local t = type( v )
 	t = t:lower()
 
-	if ( t == "vector" ) then
-		return string.format( "%.2f %.2f %.2f", v.x, v.y, v.z )
+	if ( t == "vector" or t == "angle" ) then
+		return string.format( "%.2f %.2f %.2f", v:Unpack() )
 	end
 
 	if ( t == "number" ) then
@@ -358,4 +358,24 @@ function util.RemovePData( steamid, name )
 	name = Format( "%s[%s]", GetUniqueID( steamid ), name )
 	sql.Query( "DELETE FROM playerpdata WHERE infoid = "..SQLStr(name) )
 	
+end
+
+--[[---------------------------------------------------------
+Recalculates the orthogonality of an up vector according to the
+other vector given as forward direction direction.
+The call changes the vector assigned as an up direction.
+ * vF   > The forward direction for the orthogonalization process
+ * vU   > The up direction being orthogonalized ( modified )
+ * bN   > When set to true, normalizes the three unit vectors
+ * vR   > The function returns the right vector on success
+-----------------------------------------------------------]]
+function util.GetOrthogonal(vF, vU, bN)
+	local vR = vF:Cross(vU)
+	vU:Set(vR:Cross(vF))
+	if(bN) then
+		vF:Normalize()
+		vR:Normalize()
+		vU:Normalize()
+	end
+	return vR
 end


dvdvideo1234:
This date commit also reduces 6 string concatenations and one table extraction.

Kefta:
1. You should not bundle unrelated changes together.
2. You should mention that this PR is reliant on your other Vector PR.
3. You should probably make the name GetOrghogonalVector or something.

thegrb93:
Garrysmod is not Matlab. Don't expect generic math function to be added to the library. They are just code bloat.

Verideth:
> Garrysmod is not Matlab. Don't expect generic math function to be added to the library. They are just code bloat.

I mean, not really. The GMod Lua library has tons of "generic math functions"

Its a good change. Makes things easier, that is a fine update for me.

thegrb93:
generic was the wrong adjective. I meant very specific use case and won't be used often.

dvdvideo1234:
@thegrb93  @Kefta

Is the date change fine then ? I am just using OS.date and it is 6 tail concatenation less ;)

dvdvideo1234:
To me atleast orthogonalizing vectors happens quite often and I have to [code stuff like this](https://github.com/dvdvideo1234/SpinnerTool/blob/master/lua/weapons/gmod_tool/stools/spinner.lua#L299-L308). instead of just calling a function :)

Verideth:
> To me atleast orthogonalizing vectors happens quite often and I have to [code stuff like this](https://github.com/dvdvideo1234/SpinnerTool/blob/master/lua/weapons/gmod_tool/stools/spinner.lua#L299-L308). instead of just calling a function :)

Some things I noticed that you should update:

Such as `Returns hour, minute, secon in a formatted string.`'
fix `secon` to `second`

Also, I don't necessarily believe that `GetOrthogonal` should be in the `util` library, thats more of just misc functions. The math library would be better, in my opinion.

Verideth:
> generic was the wrong adjective. I meant very specific use case and won't be used often.

Yea, I figured that's what you meant. 
His push req isn't bloating code or anything (at least not to my eye). It seems fine.

Kefta:
You still need to remove unrelated changes. This PR should only contain the GetOrthogonalVector addition, nothing else.

Kefta:
Also, you need to fix the code styling to match the rest of the codebase.

thegrb93:
I'm not saying it's not useful. I'm saying its not useful enough to be a core function. If it is, then it still shouldn't be in the util lib.

Kefta:
The util library is the best place for it. ``math`` is currently for numbers, and it doesn't make sense in the Vector metatable since it modifies arguments and its ambiguous which is the forward and up vector.

dvdvideo1234:
I am still going to let `function util.TypeToString( v )` exist because currently this is the right way to do it using unpack being pushed to PROD.

Kefta:
Put it into a separate PR. Also, you still need to fix the style.

dvdvideo1234:
@Kefta I am using Tabs. An I missing something else ?

Kefta:
Look around the file. There should be spaces surrounding parentheses (``if ( bN ) then``), and there should be empty lines surrounding control-statements, in this case, ``if`` and ``end``.

dvdvideo1234:
Others are requested in the PRs below:
https://github.com/Facepunch/garrysmod/pull/1621
https://github.com/Facepunch/garrysmod/pull/1622

dvdvideo1234:
@Kefta Done, Found it. thank you!

Kefta:
You still need spaces around control statements, and a space between ``if`` and the parenthesis. Ex.
```
function foo()
   somefunc()

   if ( condition ) then
      someotherfunc()
   end

   evenmorefunc()
end
```

Notice the spaces surrounding the if-statement.

Kefta:
You still need a space between ``if`` and the parenthesis...

dvdvideo1234:
@Kefta Done, thank you !

dvdvideo1234:
I wonder is there a way to this to utilize [VectorVectors](https://github.com/VSES/SourceEngine2007/blob/43a5c90a5ada1e69ca044595383be67f40b33c61/src_main/public/mathlib/mathlib.h#L695)?

Kefta:
Yes, I've requested it to be pushed.

dvdvideo1234:
@Kefta

I see it [includes the normalization of the right and up vectors](https://github.com/VSES/SourceEngine2007/blob/43a5c90a5ada1e69ca044595383be67f40b33c61/src_main/mathlib/mathlib_base.cpp#L469), but does not normalize the forward vector as it is a constant reference. It means that the mornalization to unit vectors is not optoional.

dvdvideo1234:
@Kefta

I see it [includes the normalization of the right and up vectors](https://github.com/VSES/SourceEngine2007/blob/43a5c90a5ada1e69ca044595383be67f40b33c61/src_main/mathlib/mathlib_base.cpp#L469) when only forward `Z != 0`, but does not normalize the forward vector as it is a constant reference. It means that the normalization of the unit vectors is not optional.

Kefta:
The right and up vectors are the returns: they aren't provided to the function.

dvdvideo1234:
But how does it `decide` which plane the second vector lays in. I mean if you try to create a [three-unit vector coordinate system][1] without an `UP` vector, this can have infinite solutions, which will result in a different output angle depending on the chosen `UP` direction.

1: https://en.wikipedia.org/wiki/Coordinate_system

Kefta:
0,0,1 is used as the absolute up vector.

dvdvideo1234:
@Kefta

Is is good then for [0,0,1] to be a default fallbacj value when an dedicated up direction is not provided ?

PR #1622, opened 2019-12-02
Utilized: Vector and angle `Unpack` metamethods in`util.TypeToString`

dvdvideo1234:


ExtReMLapin:
Manually unpacking is faster than unpacking with a function

thegrb93:
Have you timed it?

Kefta:
Unpack is much faster.
```
local vindex
local unpack

do
	local VECTOR = FindMetaTable("Vector")
	vindex = VECTOR.__index
	unpack = VECTOR.Unpack
end

local v = Vector(1, 2, 3)

local function pt(a, b, c)
end

local start = SysTime()

for i = 1, 10000000 do
	pt(vindex(v, 1), vindex(v, 2), vindex(v, 3))
end

print("numeric index", SysTime() - start)

local start = SysTime()

for i = 1, 10000000 do
	pt(vindex(v, "x"), vindex(v, "y"), vindex(v, "z"))
end

print("string index", SysTime() - start)

local start = SysTime()

for i = 1, 10000000 do
	pt(unpack(v))
end

print("unpack", SysTime() - start)
```
```
numeric index	4.4967768
string index	4.5139108
unpack	0.8396242
```

ExtReMLapin:
https://gitspartv.github.io/LuaJIT-Benchmarks/

Well i guess it's not using unpack() then

Kefta:
It's using Vector:Unpack which is a metatable function, not related to Lua's unpack.

Verideth:
Unpack() is definitely faster. I don't see much point in this PR honestly.

Kefta:
@Verideth His primary change is to add angle support to ``util.TypeToString``. The ``Unpack`` change is just a bonus.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -126,8 +126,8 @@ function util.TypeToString( v )
 	local t = type( v )
```
Kefta: Change this to use ``TypeID`` instead as it's more efficient to do numerical comparisons than string ones. Your enums here will be ``TYPE_VECTOR``, ``TYPE_ANGLE``, and ``TYPE_NUMBER``. You can also remove the string.lower call right below this once you change it.

Verideth:
> @Verideth His primary change is to add angle support to `util.TypeToString`. The `Unpack` change is just a bonus.

Oh my bad. Didn't even really read it honestly just kinda skimmed.

garrysmod/lua/includes/extensions/util.lua
dvdvideo1234: On it ...

dvdvideo1234: Hello, [changed here](https://github.com/Facepunch/garrysmod/pull/1622/commits/edffb502000b74e8b78392859d0ed62bef5e0254). Actually the time is relatively the same as using `if-elseif-else-end` construct.

robotboy655:
In hindsight this was a very pointless change, all it essentially does is remove 1 decimal point compared to Angle's tostring()

Kefta:
The function should probably just be deprecated.

PR #1623, opened 2019-12-04
Add hook CanUndo

WilliamVenner:
Return false to prevent undo

### DIFF
--- a/garrysmod/lua/includes/modules/undo.lua
+++ b/garrysmod/lua/includes/modules/undo.lua
@@ -380,8 +380,10 @@ end
 	Checks whether a player is allowed to undo
 -----------------------------------------------------------]]
 local function Can_Undo( ply, undo )
+
+	local call = hook.Run( "CanUndo", ply, undo )
 	
-	return hook.Run( "CanUndo", ply, undo )
+	return call == true or call == nil
 
 end
 
--- a/garrysmod/lua/includes/modules/undo.lua
+++ b/garrysmod/lua/includes/modules/undo.lua
@@ -377,6 +377,15 @@ function Do_Undo( undo )
 end
 
 --[[---------------------------------------------------------
+	Checks whether a player is allowed to undo
+-----------------------------------------------------------]]
+local function Can_Undo( ply, undo )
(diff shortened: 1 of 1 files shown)


Verideth:
Why not just make it so it's an argument to 'GM:OnUndo`?

 `bool can_undo = true`

WilliamVenner:
> Why not just make it so it's an argument to `GM:OnUndo`?
> `bool can_undo = true`

OnUndo is a clientside hook

Verideth:
> > Why not just make it so it's an argument to `GM:OnUndo`?
> > `bool can_undo = true`
> 
> OnUndo is a clientside hook

My apologies, didn't read it close enough.

robotboy655:
Why would you ever want to prevent an undo?

Donkie:
For pure building gamemodes it may not make much sense, but if you have a sandbox-derived gamemode that is for example round based it would make more sense.

In flood for example you want people to be able to build and use undo like normal during the build phase. But in the fight phase you don't want people to be able to just undo their boat once they've died.

GitSparTV:
Imagine accidentally pressing undo when you're making a poster.
:bruh:

robotboy655:
Ok, 2 things:
* Player should have feedback that their undo was blocked. Maybe some sort of "error" sound will do the trick.
* Should rename the hook to `PlayerCanUndo` to be more in line with the rest of Player related hooks such as `PlayerSpawn<whatever>`

Donkie:
CanTool does not produce any automatic error feedback. https://github.com/Facepunch/garrysmod/blob/2bc5fee85516124c98b658b450e3d6b8a9d5cfea/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua#L238

I think it's better up to the one who hooks it to also warn the user of why it was blocked if that is necessary. 

Also if we're talking about naming conventions, CanDrive, CanEditProperty and CanProperty among others was added in gmod 13 so they're relatively recent. Not many hooks are named "PlayerCanX".

robotboy655:
Fair enough.

PR #1624, opened 2019-12-31
no more freeze from setting float convars too high

joeyjumper94:
before this change was made, you cam make the server lock up when the round begins by setting ttt_traitor_pct or ttt_detective_pct to a number greater than 1
ok so the fatal logic flow was this
the table that lists the players who will participate has, say 4 players will participate in the round
the GetTraitorCount first multiplies it by 25 to give 100
then it is clamped to a value between 1 and the integer of ttt_traitor_max which dropped it to 32, which is then returned by the function.
the loop becomes infinite when it loops until it makes 32 traitors out of 4 players


Kefta:
Just clamp the loop iterations by the total amount of players. The SysTime hack is really unnecessary.

svdm:
Fixed in d5cebd3 without the weird stuff.

PR #1627, opened 2020-01-10
Added: Support for reversing a integer indexed table including holes

dvdvideo1234:
This will reverse a table if there are holes inside. It is not designed to replace `table.Reverse` in any way, but to add support for tables with vacant slots. Considering the following example:
```lua
BEFORE = {}
BEFORE[1] = 1
BEFORE[2] = 2
BEFORE[3] = 3
BEFORE[10] = 10
AFTER = {}
AFTER[1] = 10
AFTER[8] = 3
AFTER[9] = 2
AFTER[10] = 1
```

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -702,10 +702,11 @@ end
 --[[---------------------------------------------------------
 	Name: ReverseHole( t )
 	Desc: Reverses the table even if there are holes in it
+	      You can also provide a second argument as forced table length
 -----------------------------------------------------------]]
 function table.ReverseHole( t )
 	
-	local nN = table.maxn( t ) 
+	local nN = tonumber( n ) or table.maxn( t ) 
 	
 	for iD = 1, math.floor( nN / 2 ) do
 		local iK = ( nN + 1 - iD )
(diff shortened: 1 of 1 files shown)


thegrb93:
I don't get it

dvdvideo1234:
@thegrb93,

It will just revert the table whenever or not there are nils in the middle. It is used when a reversion is needed, and the middle elements are deleted, where the reversion must affect all the elements including nils.

Donkie:
Yes but when would this ever be necessary, let alone be useful enough to be included in gmod standard library

dvdvideo1234:
@Donkie,

As I said when you need to revert the list with elements including nils, like priority slots for example. You want the order to be reverted and slots to be mirrored regarding the maximim length.

Kefta:
This could be useful for vararg captured tables, but using ``select("#", ...)`` is better for that.

dvdvideo1234:
@Kefta `select("#",...` is good for obtaining the max count:
```lua
test = {1,2,3,[10]=10}
print(select("#", 1,2,3,nil,nil,nil,nil,nil,nil,10)) -- 10 arguments on stack
print(unpack(test)) -- Using unpack only 3 are visible
```

ExtReMLapin:
This PR really confuses me.

I don't think we should add any math function unless there is a real practical use for everyone.

dvdvideo1234:
@Kefta Yeah, I guess you are right ....

PR #1630, opened 2020-02-07
Added ScreenScaleX and ScreenScaleY

Fafy2801:
Adding ScreenScaleY mainly to account for diferent aspect ratios.


BlackCetha:
You can't just remove the original function. People are using it.

Donkie:
@BlackCetha Look at the bottom of the change, the original function is overridden by ScreenScaleX

BlackCetha:
Right, I missed that

robotboy655:
Definitely don't want yet another alias of an existing function.

PR #1631, opened 2020-02-10
Update hints.properties

pilot22:
Added an backslash to fix


pilot22:
Seems like this files causes this: 
https://files.lynaria.fr/hl2_lD7IRraL6I.jpg

robotboy655:
To fix what? `'` does not need to be escaped, it isn't escaped in any of the other files.

The picture you linked no longer works, but if I remember correctly it showed a missing translation for a certain hint, and I do believe this is me just forgetting to upload these strings to the translation server somehow. I will make sure to get that fixed.

pilot22:
Sorry I though the notification with the missing translation was due to the ' which isn’t escaped.

PR #1632, opened 2020-02-14
config_default.cfg reworked

Perkzitos:
- Removed `spawnmenu_icon_precache` because it is not a valid command;
- Modified the `mp_decals` to the value of `2048` to equal the value of `r_decals` (`r_decals` gets resetted everytime you launch the game, so setting `r_decals` to `5000` will get its value back to `2048` when you relaunch the game, while `mp_decals` still with the value of `5000` because `mp_decals` it is stored in the `config.cfg` file, while `r_decals` is not, and to make both equal, here is the solution);
- Modified the value of the `sensitivity` command to the value of `6` instead of `8`.

And that is it.


thegrb93:
You will anger the gods if you make whitespace changes. As for the rest, there's probably reasons commands were put there that you don't know about. I also disagree with the binding changes.

Perkzitos:
> You will anger the gods if you make whitespace changes.
Haha.

> As for the rest, there's probably reasons commands were put there that you don't know about. I also disagree with the binding changes.
Yeah, probably these commands have reasons to currently exist in the file. About the bindings, I think it's better like this.

Kefta:
I do not agree with changing default bindings, especially switching V and X.

Perkzitos:
> I do not agree with changing default bindings, especially switching V and X.

On TF2 (and I believe on many Source games) the default binding for voice chat is the letter V (except CS:GO, which is K).

Kefta:
Or Dota which is G. I think X is a better choice.

Perkzitos:
> Or Dota which is G. I think X is a better choice.

I do not disagree, but I also think the letter V is a good letter (V as if it came from the word Voice). If I see more people not agreeing to this change here, I’ll reverse it.

neico:
`~` is the default key for US keyboards, while other layouts use ```, it's how the engine translates the key codes (it's not aware of keyboard layouts)
And in that regard, about F10: https://garry.tv/more-stuff-i-did

`z` as undo is a reference to `CTRL+Z` in desktop applications, it's what feels most natural
`x` I actually use that for voice chat for various resons, so it's something I also do on other games, I don't see why you'd want to change that

I agree with most of the convars being redundant in that config, but reverting `sensitivity` back to the default 3 sounds a bit weird, at least make it 5?

And I see no reason why you'd want to lower the decal limits at all
> Removed r_decals because the default value it's already 2048;

It's set to 5000, not 2048, so this statement makes no sense
by default `r_decals = 2048` & `mp_decals = 200`

Perkzitos:
> `~` is the default key for US keyboards, while other layouts use ```, it's how the engine translates the key codes (it's not aware of keyboard layouts)
> And in that regard, about F10: https://garry.tv/more-stuff-i-did

Alright, I'll update this.

> `z` as undo is a reference to `CTRL+Z` in desktop applications, it's what feels most natural
> `x` I actually use that for voice chat for various resons, so it's something I also do on other games, I don't see why you'd want to change that

I'll revert this.

> I agree with most of the convars being redundant in that config, but reverting `sensitivity` back to the default 3 sounds a bit weird, at least make it 5?

5 seems to be a good value.

> And I see no reason why you'd want to lower the decal limits at all
> 
> > Removed r_decals because the default value it's already 2048;
> 
> It's set to 5000, not 2048, so this statement makes no sense
> by default `r_decals = 2048` & `mp_decals = 200`

I made a typo, if you see what I wrote later, you'll understand why:

> * Modified the mp_decals to the value of 2048 to equal the value of r_decals (r_decals gets resetted every time you launch the game);

Everytime you launch the game, the `r_decals` command goes back to its default value of `2048`, unless you set it in your launch options or in the `autoexec.cfg` file. However, the `mp_decals` command does not goes back to its default value, because it is stored in the `config.cfg` file, while `r_decals` is not.

Perkzitos:
> Removing stuff from the file because "it is already set to the default value" is also wrong, have you considered the file may be ran to reset stuff to their default values in case something doesn't work properly?

Did not think about that. Thanks for sharing your point of view.

> I also don't get what the decal convar changes meant to accomplish. `r_decals` is set to the value of server's `mp_decals` when joining a server. ( I imagine `r_decals` is set to client's `mp_decals` for local servers/singleplayer )

`r_decals` should be the same value of `mp_decals`, `r_decals` are for singleplayer decals (decals that appears both on singleplayer and multiplayer), and `mp_decals` are for multiplayer decals (decals that appears only on multiplayer games), and I do not understand why they are different in the `config_default.cfg` file. As I said earlier, `r_decals` gets resetted to `2048`, while `mp_decals` do not. I learned from Team Fortress 2 that they should be the same, and how Garry's Mod uses Source too, I think this knowledge should be applied here to.

PR #1634, opened 2020-02-20
Fix sorting of bodygroups and skins in spawn icon editor

WilliamVenner:
Previously the DComboBox in this GUI would sort the items alphabetically, despite the items already being in order, resulting in an order like this:

```
Skin 0, Skin 1, Skin 10, Skin 11, ...
```

 `DComboBox:SetSortItems(false)` fixes this.


robotboy655:
I'd rather switch the dropdowns to sliders so you don't have to constantly click 2 times to preview a single bodygroup/skin change

PR #1638, opened 2020-02-26
Add DButton:SetColorIcon()

WilliamVenner:
This is hacky as fuck and I hate it, but it should mean that it's quite smoothly backwards compatible (as it just kind of piggybacks off the behaviour of `DButton:SetImage()`)

Here's the cool thing you can do with it:

image

### DIFF
--- a/garrysmod/lua/vgui/dbutton.lua
+++ b/garrysmod/lua/vgui/dbutton.lua
@@ -51,6 +51,42 @@ function PANEL:SetImage( img )
 end
 PANEL.SetIcon = PANEL.SetImage
 
+local function ColorIconPaint( self, w, h )
+
+	surface.SetDrawColor( self.m_ColorIcon )
+	surface.DrawRect( 0, 0, w, h )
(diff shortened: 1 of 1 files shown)


robotboy655:
This honestly feels like something for mods to do, It seems niche and is very easy to add via panel inheritance since it doesn't require any actual changes to the panel. And as you yourself stated, it is a very hacky implementation.

WilliamVenner:
@robotboy655 What if the same method was expanded upon to allow `AvatarImage`, too?

Abstracting away the hackiness I think is justified when wanting to preserve backwards compatibility in the name of useful features.

I'm sure colour icons and avatars in DMenus would be widely used if implemented.

Alternatively, I could just add these features in without hacks, but I have no idea of the consequences to (ironically) hacky addons that may directly interface with `m_Image`.

WilliamVenner:
Example:

WilliamVenner:
Closed in favour of #1660

PR #1639, opened 2020-02-28
Replace UniqueID() with SteamID()

PotatoTheWorker:
Makes it so the PData section wont have any collisions like UniqueID() caused.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -176,7 +176,7 @@ end
 -----------------------------------------------------------]]
 function meta:GetPData( name, default )
 
-	name = Format( "%s[%s]", self:UniqueID(), name )
+	name = Format( "%s[%s]", self:SteamID(), name )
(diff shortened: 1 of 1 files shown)


robotboy655:
This will break literally everything that uses PData currently, so no go.

PotatoTheWorker:
Just a general question how would replacing UniqueID() with SteamID() break everything that PData uses? Because if it does then PData seems unfixable and if anything deprecated completely?

robotboy655:
Imagine you have an RP server that stores money currently using PData, which is stored using UniqueID. in the database on disk.

Imagine this change hits and PData now uses SteamID. But all your data in the database still obviously stores UniqueID, and you try to request by using SteamID and there's no matches, so now everyone on your server just lost their money. On every server. And not just money, but literally everything.

PotatoTheWorker:
Is there anyway to change the players UniqueID then? So a simple script could be written that just checks if any players have the same UniqueID and takes the highest UniqueID and puts it one above?

bigdogmat:
You can't change a players uniqueid. The only solution here would be an incremental one in which when information is requested it checks for data relating to steamid, but falls back to unqiueid on failure.

PR #1640, opened 2020-03-13
More fixes to the TTT Portuguese translation

Perkzitos:


Perkzitos:
@svdm

svdm:
Do you think this is finished at this point or do you expect more fixes/changes?

Perkzitos:
> Do you think this is finished at this point or do you expect more fixes/changes?

Honestly, I think so. However, it is very unlikely that it ended, at one time or another I will always end up finding new texts to improve. But, for now, it is ok. At any time I can send another Pull Request, this is something unpredictable.

PR #1641, opened 2020-03-16
Fix default skin to use self instead of SKIN

GitSparTV:
This PR fixed the bug of default skin when redefinition of the SKIN in the last skin file affects all PaintShadow and PaintFrame of current skin.

Default skin uses SKIN variable
Next skin defines itself.
SKIN variable is changed so now default skin uses the last skin


GitSparTV:
This bug was mentioned by me in https://wiki.facepunch.com/gmod/Derma_Skin_Creation

> Bug: Will be used shadow texture from last loaded skin! To fix it change SKIN in functions PaintShadow and PaintFrame to self.

PR #1642, opened 2020-03-19
New hook lib

Srlion:
Faster hook library

I have released this on [workshop](https://steamcommunity.com/sharedfiles/filedetails/?id=1907060869) few months ago to make sure that it doesn't break anything.

 - This version is slightly different, this one is faster when objects become invalid and get removed.
 - Thanks to @WilliamVenner for telling me to upload it first on workshop.
 - Big thanks to @meepen for his [hook library](https://github.com/meepen/gmod-hooks-revamped), which inspired me to make this.

benchmark (using a slightly modified version of https://github.com/meepen/gmod-hooks-revamped/blob/master/hooksuite.lua):

If you don't care about "CallInvalid" performance, I could change the code to be more readable.

```
Srlion's New Lib...
Srlion's New Lib took 0.511766100 s

Default Lib...
Default Lib took 14.408792900 s

BENCHMARK
-------------
CallInvalid (500 calls)
Srlion's New Lib (218.29% faster)
Default Lib:      0.005260700 s
Srlion's New Lib: 0.001652800 s
-------------
CallNoHooks (200000000 calls)
Default Lib (1.36% faster)
Default Lib:      0.213272500 s
Srlion's New Lib: 0.216162900 s
-------------
CallGMOnly (200000000 calls)
Default Lib (2.97% faster)
Default Lib:      0.215109900 s
Srlion's New Lib: 0.221490100 s
-------------
CallNoGM (32000000 calls)
Srlion's New Lib (19258.95% faster)
Default Lib:      6.825907900 s
Srlion's New Lib: 0.035259700 s
-------------
CallGM (32000000 calls)
Srlion's New Lib (19239.46% faster)
Default Lib:      7.149178400 s
Srlion's New Lib: 0.036966800 s
-------------
```

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -101,6 +101,11 @@ do
 		return getmetatable( value ) == functionMeta
 	end
 
+	--[[---------------------------------------------------------
+		Name: Add
+		Args: string hookName, any identifier, function func
+		Desc: Add a hook to listen to the specified event.
+	-----------------------------------------------------------]]
 	function Add( event_name, name, func )
 
 		if ( !isstring( event_name ) ) then return end
@@ -123,6 +128,11 @@ do
 	end
 end
 
+--[[---------------------------------------------------------
+    Name: Remove
+    Args: string hookName, identifier
+    Desc: Removes the hook with the given indentifier.
+-----------------------------------------------------------]]
 function Remove( event_name, name )
 
 	local event = events[ event_name ]
@@ -132,6 +142,10 @@ function Remove( event_name, name )
 
 end
 
+--[[---------------------------------------------------------
+    Name: GetTable
+    Desc: Returns a table of all hooks.
+-----------------------------------------------------------]]
 function GetTable()
 
 	local new_events = {}
@@ -146,6 +160,11 @@ function GetTable()
 
 end
 
+--[[---------------------------------------------------------
+    Name: Call
+    Args: string hookName, table gamemodeTable, vararg args
+    Desc: Calls hooks associated with the hook name.
+-----------------------------------------------------------]]
 function Call( event_name, gm, ... )
 
 	local event = events[ event_name ]
@@ -228,6 +247,10 @@ function Call( event_name, gm, ... )
 		end
 
 	end
+
+	--
+	-- Call the gamemode function
+	--
 
 	if ( !gm ) then return end
 
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -261,6 +261,11 @@ function Call( event_name, gm, ... )
 
 end
 
+--[[---------------------------------------------------------
+    Name: Run
+    Args: string hookName, vararg args
+    Desc: Calls hooks associated with the hook name.
+-----------------------------------------------------------]]
 function Run( name, ... )
 	return Call( name, gmod && gmod.GetGamemode() || nil, ... )
 end
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -1,125 +1,243 @@
-local gmod                        = gmod
-local pairs                        = pairs
-local isfunction        = isfunction
-local isstring                = isstring
-local IsValid                = IsValid
+local gmod = gmod
+local debug = debug
+local pairs = pairs
+local getmetatable = getmetatable
+local setmetatable = setmetatable
 
-module( "hook" )
+module("hook")
 
-local Hooks = {}
+local events = {}
 
---[[---------------------------------------------------------
-    Name: GetTable
-    Desc: Returns a table of all hooks.
------------------------------------------------------------]]
-function GetTable() return Hooks end
+local NewEvent
+do
+	local Event = {}
+	local EventMeta = { __index = Event }
 
-
---[[---------------------------------------------------------
-    Name: Add
-    Args: string hookName, any identifier, function func
-    Desc: Add a hook to listen to the specified event.
------------------------------------------------------------]]
-function Add( event_name, name, func )
-
-	if ( !isfunction( func ) ) then return end
-	if ( !isstring( event_name ) ) then return end
-
-	if (Hooks[ event_name ] == nil) then
-			Hooks[ event_name ] = {}
+	NewEvent = function()
+		return setmetatable({
+			n = 0,
+			keys = {},
+		}, EventMeta)
 	end
 
-	Hooks[ event_name ][ name ] = func
+	function Event:Add( name, func, object )
 
+		local pos = self.keys[ name ]
+		if ( pos ) then
+
+			--
+			-- Hook exists, just update it
+			--
+			self[ pos + 1 ] = func
+			self[ pos + 2 ] = object
+
+		else
+
+			local n = self.n
+
+			self[ n + 1 ] = name
+			self[ n + 2 ] = func
+			self[ n + 3 ] = object
+			self.keys[ name ] = n + 1 -- Keep a reference to hook position, so we can have fast hook removing/replacing
+
+			self.n = n + 3
+
+		end
+
+	end
+
+	function Event:Remove( name )
+
+		local pos = self.keys[ name ]
+		if ( pos ) then
+
+			self[ pos ] = nil --[[name]]
+			self[ pos + 1 ] = nil --[[func]]
+			self[ pos + 2 ] = nil --[[object]]
+			self.keys[ name ] = nil
+
+		end
+
+	end
+
+	function Event:GetHooks()
+
+		local hooks = {}
+		local i, n = 1, self.n
+
+		::loop::
+		local name = self[ i ]
+		if ( name ) then
+			hooks[ name ] = self[ i + 1 ] --[[func]]
+		end
+
+		i = i + 3
+		if ( i <= n ) then
+			goto loop
+		end
+
+		return hooks
+
+	end
 end
 
+do
+	--
+	-- "type" function in Garry's Mod is not jit compiled and for some reason
+	-- that idk, the hook library performs worse with it
+	--
 
---[[---------------------------------------------------------
-    Name: Remove
-    Args: string hookName, identifier
-    Desc: Removes the hook with the given indentifier.
------------------------------------------------------------]]
+	local stringMeta = getmetatable( "" )
+	local function isstring( value )
+		return getmetatable( value ) == stringMeta
+	end
+
+	local functionMeta = getmetatable( isstring ) || {}
+	if ( !getmetatable(isstring) ) then
+		debug.setmetatable( isstring, functionMeta )
+	end
+
+	local function isfunction( value )
+		return getmetatable( value ) == functionMeta
+	end
+
+	function Add( event_name, name, func )
+
+		if ( !isstring( event_name ) ) then return end
+		if ( !isfunction( func ) ) then return end
+		if ( !name ) then return end
+
+		local object = false
+		if ( !isstring( name ) ) then
+			object = name
+		end
+
+		local event = events[ event_name ]
+		if ( !event ) then
+			event = NewEvent()
+			events[ event_name ] = event
+		end
+
+		event:Add( name, func, object )
+
+	end
+end
+
 function Remove( event_name, name )
 
-	if ( !isstring( event_name ) ) then return end
-	if ( !Hooks[ event_name ] ) then return end
-
-	Hooks[ event_name ][ name ] = nil
+	local event = events[ event_name ]
+	if ( event ) then
+		event:Remove( name )
+	end
 
 end
 
+function GetTable()
 
---[[---------------------------------------------------------
-    Name: Run
-    Args: string hookName, vararg args
-    Desc: Calls hooks associated with the hook name.
------------------------------------------------------------]]
-function Run( name, ... )
-	return Call( name, gmod and gmod.GetGamemode() or nil, ... )
+	local new_events = {}
+
+	for event_name, event in pairs( events ) do
+
+		new_events[ event_name ] = event:GetHooks()
+
+	end
+
+	return new_events
+
 end
 
+function Call( event_name, gm, ... )
 
---[[---------------------------------------------------------
-    Name: Run
-    Args: string hookName, table gamemodeTable, vararg args
-    Desc: Calls hooks associated with the hook name.
------------------------------------------------------------]]
-function Call( name, gm, ... )
+	local event = events[ event_name ]
+	if ( event ) then
 
-	--
-	-- Run hooks
-	--
-	local HookTable = Hooks[ name ]
-	if ( HookTable != nil ) then
-	
-		local a, b, c, d, e, f;
+		local i, n = 2, event.n
 
-		for k, v in pairs( HookTable ) do 
-			
-			if ( isstring( k ) ) then
-				
-				--
-				-- If it's a string, it's cool
-				--
-				a, b, c, d, e, f = v( ... )
+		::loop:: -- https://github.com/Facepunch/garrysmod/pull/1508#issuecomment-398231260
+		local func = event[ i ]
+		if ( func ) then
 
-			else
+			local object = event[ i + 1 ]
+			if ( object ) then
 
-				--
-				-- If the key isn't a string - we assume it to be an entity
-				-- Or panel, or something else that IsValid works on.
-				--
-				if ( IsValid( k ) ) then
+				if ( object.IsValid && object:IsValid() ) then
 					--
-					-- If the object is valid - pass it as the first argument (self)
+					-- The object is valid - pass it as the first argument (self)
 					--
-					a, b, c, d, e, f = v( k, ... )
+					local a, b, c, d, e, f = func( object, ... )
+					if ( a != nil ) then
+						return a, b, c, d, e, f
+					end
 				else
 					--
-					-- If the object has become invalid - remove it
+					-- The object has become invalid - remove it
 					--
-					HookTable[ k ] = nil
+					event:Remove( event[ i - 1 ] --[[name]] )
+				end
+
+			else
+				local a, b, c, d, e, f = func( ... )
+				if ( a != nil ) then
+					return a, b, c, d, e, f
 				end
 			end
 
+			i = i + 3
+
+		else
+
 			--
-			-- Hook returned a value - it overrides the gamemode function
+			-- hook got removed, we gotta do some work!
 			--
-			if ( a != nil ) then
-				return a, b, c, d, e, f
+
+			local _n, _i = n, i
+
+			if ( event.n != n ) then
+				-- a new hook was added while "Call" is running, we will replace
+				-- the removed one with it and continue without calling it
+				_n = event.n
+				i = i + 3
+			else
+				-- replace the removed hook with the last hook in the table and repeat the loop
+				n = n - 3
 			end
-				
+
+			local new_name = event[_n - 2 --[[name]]]
+			if ( new_name ) then
+
+				--[[name]] --[[func]] --[[object]]
+				event[ _i - 1 ], event[ _i ], event[ _i + 1 ] = new_name, event[ _n - 1 ], event[ _n ]
+				event[ _n - 2 ], event[ _n - 1 ], event[ _n ] = nil, nil, nil
+				event.keys[ new_name ] = _i - 1 -- update hook position
+
+			end
+
+			event.n = event.n - 3
+
+			--
+			-- no hooks, so no need to keep it in the events table!
+			--
+			if ( event.n == 0 ) then
+				events[ event_name ] = nil
+			end
+
 		end
+
+		if ( i <= n ) then
+			goto loop
+		end
+
 	end
-	
-	--
-	-- Call the gamemode function
-	--
+
 	if ( !gm ) then return end
-	
-	local GamemodeFunction = gm[ name ]
-	if ( GamemodeFunction == nil ) then return end
-			
-	return GamemodeFunction( gm, ... )        
-	
+
+	local GamemodeFunction = gm[event_name]
+	if ( !GamemodeFunction ) then return end
+
+	return GamemodeFunction( gm, ... )
+
+end
+
+function Run( name, ... )
+	return Call( name, gmod && gmod.GetGamemode() || nil, ... )
 end


thegrb93:
Why are you defining your own isstring and isfunction? They already exist. Also there's already like 3 other pull requests for hook lib improvements. Get in line.

Srlion:
So hook.Add gets compiled, also other hook library fails in this test:

```
hook.Add("T", "a", function()
    print("a")
    hook.Remove("T", "a")
    hook.Add("T", "b", function()
        print("b2")
    end)
end)
hook.Add("T", "b", function()
    print("b")
end)
hook.Add("T", "c", function()
    print("c")
end)

hook.Call("T")
print("==================")
hook.Call("T")
```

Here is what yours produces:
```
-==-=-=-=--==
a
b
c
b2
==================
c
b2
```

Here is what mine produces:
```
-==-=-=-=--==
a
b2
c
==================
c
b2
```

------

Here is another test that yours and the default hook library fails in:
```
hook.Add("T", "A", function()
    print("A")
    hook.Add("T", "E", function()
        print("E")
    end)
end)
hook.Add("T", "B", function()
    print("B")
end)

hook.Call("T")
print("==================")
hook.Call("T")
```

Here is what yours produces:
```
-==-=-=-=--==
A
B
E
==================
A
B
E
```

Here is what the default lib produces:
```
-==-=-=-=--==
A
A
E
==================
B
A
E
```

Here is what mine produces:
```
-==-=-=-=--==
A
B
==================
A
B
E
```
and it's the expected output.

Mine is well tested against any case, it's even more reliable than the default hook library.
I had more tests that I don't remember where I have placed :s

thegrb93:
Considering the behavior of those cases are undefined, those tests really don't mean anything.

thegrb93:
I'm surprised the default lib prints A twice in the first hook.Call of the second test though. wtf caused that

Srlion:
I'm just showing that mine is more "reliable" and have expected behavior.
I don't really know why it prints twice and skips "B" lol

thegrb93:
That's merely an opinion if there's no definition of the behavior. I get the feeling you messed that test up somehow. The default lib is just a 'pairs' loop so I can't see how it fucked up that bad.

Srlion:
You can test it yourself lol

thegrb93:
Huh, apparently adding stuff to a table being iterated can cause it to iterate over something twice in lua 'pairs'. Did not know that.

Srlion:
I think it's only in lua5.1, tested it here https://www.lua.org/demo.html and it works fine.

thegrb93:
```lua
local t = {}
t.A = function() print("A") t.E = function() print("E") end end
t.B = function() print("B") end
for k, v in pairs(t) do
	v()
end
```

Yeah, lua demo:
B
A

gmod:
A
A
E

thegrb93:
I think this looks fine then.

Srlion:
Tested it on latest [luajit build](https://github.com/LuaJIT/LuaJIT) and lua 5.1.5 and they produce same output, so I guess it won't be fixed.

thegrb93:
Turns out that behavior too is undefined lol

https://www.lua.org/manual/5.3/manual.html#pdf-next
"The behavior of next is undefined if, during the traversal, you assign any value to a non-existent field in the table. You may however modify existing fields. In particular, you may clear existing fields. "

Kefta:
Hopefully we get on 5.2 compat or 5.3 after x64 is released. There's already a few other attempts at hook lib rewrites on here so it would be best to bench them all on one machine using a common bench lib.

ExtReMLapin:
if you want your shit to compile, don't use pairs

garrysmod/lua/includes/modules/hook.lua
```diff
@@ -41,85 +135,137 @@ end
...
+
 end
+
+--[[---------------------------------------------------------
+    Name: Run
+    Args: string hookName, vararg args
+    Desc: Calls hooks associated with the hook name.
+-----------------------------------------------------------]]
+function Run( name, ... )
+	return Call( name, gmod && gmod.GetGamemode() || nil, ... )
```
ExtReMLapin: VARG Doesn't compile

`[TRACE FAIL --- (263/stitch) @lua/includes/modules/hook.lua:270 -- NYI: bytecode VARG]`

garrysmod/lua/includes/modules/hook.lua
```diff
@@ -41,85 +135,137 @@ end
...
-	local GamemodeFunction = gm[ name ]
-	if ( GamemodeFunction == nil ) then return end
-			
-	return GamemodeFunction( gm, ... )        
-	
+
+	local GamemodeFunction = gm[event_name]
+	if ( !GamemodeFunction ) then return end
+
+	return GamemodeFunction( gm, ... )
```
ExtReMLapin: VARG Doesn't compile

`[TRACE FAIL --- (251/stitch) @lua/autorun/client/gm_demo.lua:20 -- NYI: bytecode VARG at @lua/includes/modules/hook.lua:260]`

garrysmod/lua/includes/modules/hook.lua
```diff
@@ -41,85 +135,137 @@ end
...
 					--
-					-- If the object has become invalid - remove it
+					-- The object has become invalid - remove it
 					--
-					HookTable[ k ] = nil
+					event:Remove( event[ i - 1 ] --[[name]] )
 				end
+
+			else
+				local a, b, c, d, e, f = func( ... )
```
ExtReMLapin: VARG Doesn't compile

`[TRACE FAIL --- (242/stitch) @lua/postprocess/stereoscopy.lua:44 -- NYI: bytecode VARG at @lua/includes/modules/hook.lua:198]`

Srlion:
I could have an array and hashed table to avoid using pairs, but I don’t think it’s worth it?

The test thing I’m using made me forget about VARG, maybe I should do something like “a, b, d, ..etc”? Would require @robotboy655 to know if he would like that :s

ExtReMLapin:
The JIT compilation of GetTable doesn't really matter as it's most likely only used for debug.


However, a function with more than 3 args should probably not exist.

A function with more than 3 args on a hook should 100% not exist at all.

You pretty much do what you want when calling internal functions of your addon, but if a dev decides to plug one of his functions to the hooking system, he should follow calling conventions.

Srlion:
We can’t just limit it now, would break a lot of addons :s

ExtReMLapin:
"a lot" is to define


**If you code a function that takes 6 arguments and you allow people to hook on it, you should probably reconsider your career choice.**

ExtReMLapin:
Also, while test cases doesn't mean anything without specs, it make sense in not breaking what's already in place.

Srlion:
It would depend on Garrys Mod developers saying yes or no about it (I can’t change the default behavior for anything here :s). Also I don’t know if hook.Run calls from C++ to Lua would make any difference? Could you test it, if it does’t perform better then it’s not really needed to replace the default hook library with this :c

ExtReMLapin:
Already found the first FUCKER using more than 5 args : 


```Lua
GM:CalcViewModelView( Weapon, ViewModel, OldEyePos, OldEyeAng, EyePos, EyeAng )
```

Kefta:
It's the same args as Source's CalcViewModelView. It's very arbitrary to design your entire API around a JIT limitation.

ExtReMLapin:
As I said before, the difference here is the function is exposed to other developers, this is why i'm talking about calling conventions.


We could still go full monkey mode and do something like


```lua
function Call( event_name, gm, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8 )

	local event = events[ event_name ]
	if ( event ) then

		local i, n = 2, event.n

		::loop:: -- https://github.com/Facepunch/garrysmod/pull/1508#issuecomment-398231260
		local func = event[ i ]
		if ( func ) then

			local object = event[ i + 1 ]
			if ( object ) then

				if ( object.IsValid && object:IsValid() ) then
					--
					-- The object is valid - pass it as the first argument (self)
					--
					local a, b, c, d, e, f = func( object, arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8 )
					if ( a != nil ) then
						return a, b, c, d, e, f
					end
				else
					--
					-- The object has become invalid - remove it
					--
					event:Remove( event[ i - 1 ] --[[name]] )
				end

			else
				local a, b, c, d, e, f = func( arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8 )
				if ( a != nil ) then
					return a, b, c, d, e, f
				end
			end
...
```

I mean, look at `a, b, c, d, e, f` it's pretty much the same thing.

Kefta:
Compatibility with current addons should be the number one priority:  it's too late to be limiting the number of args sent into hooks. Varadic hook returns were removed during the GM13 update as compatibility was no concern.

> As I said before, the difference here is the function is exposed to other developers,

So is CalcViewModelView for Sourcemods. It's part of the public Source API that other developers can override in their custom entities. You're still basing your entire design philosophy around JIT which would be fine if this wasn't a legacy codebase.

Srlion:
@ExtReMLapin could you measure hook.Run calls from C++, because calling Lua functions from C++ is heavy, and I don’t know if this would make any difference. (Important for things like Think, HUDPaint, ..etc)

ExtReMLapin:
Either way and even with stitching, jit compilation is kinda dead by design as there is too much Calls to C API in the lua code from addons

thegrb93:
Yeah I'm afraid you're worried about optimizing a ~10 element loop. Not something to go insane over.

DBotThePony:
https://github.com/Facepunch/garrysmod/pull/1642#issuecomment-601288451

Implementing any kind of sorting over pairs produce known to me issues with existing addons.
I know, we better had hook sorting from start of GM13, but changing it now will only break stuff

My hook's output by code provided in comment:
```
-==-=-=-=--==
b
a
c
==================
b2
c
-==-=-=-=--==
B
A
==================
B
A
E
```

Which should be very close to current "pairs undefined sorting", hence default lib, but without undefined behavior when removing hooks of event inside that event call.

Srlion:
@DBotThePony Your one fails when you modify/remove:
```lua
hook.Add("T", "a", function()
    print("a")
    hook.Add("T", "b", function()
        print("new b")
    end)
end)
hook.Add("T", "b", function()
    print("b")
end)
hook.Call("T")
print("==================")
hook.Call("T")
```
Default Lib:
```
a
new b
==================
a
new b
```
Yours:
```
a
b
==================
a
new b
```
Like I said, these tests to show that this hook library is stable and works like the default one with being faster and no undefined behaviors. (Don't talk about sorting because it's not something to worry about at all.)

Here is the default behavior you should expect with some cases:
- If you add a new hook while the even is called, you expect it to run in next call. (Default Lib has undefined behavior when adding new hooks, but most of the time it runs in next call.)
- If you modify a hook before it's called, the new function will be called not the old one. (Just like the test above.)
- If you remove a hook before it's called, then it won't be called in this event call.

Last 2 cases are not undefined behaviors according to https://www.lua.org/manual/5.1/manual.html#pdf-next

I doubt that this "CallInvalid" benchmark is correct, adding/removing hooks with your hook library could be kinda heavy because of "hook.Reconstruct"

Also addons don't relay on "pairs undefined sorting", so we don't have to worry about sorting.

DBotThePony:
I, personally, expect from normal hook library that when i call hook.Call, **all existing functions at moment of call** will be called back. Concurrent modifications of hook table while iterating it is evil, and good part of languages disallow concurrent modifications while iterating data structures.

Sorting - i tried implementing sorted hook back in 2017, it didn't end well with one of my friend's addon and with at least one workshop addon - Playable Piano (it just refused to work because it's hook got moved in iteration loop).

Srlion:
That is actually a defined behavior that someone could be depending on, so we can't just change that because it _may_ break some addons.

> Concurrent modifications of hook table while iterating it is evil (and will produce undefined result no matter what you try to do)

Well, mine has no undefined behavior when adding/removing in a call, because I control the behavior when it ever happens :D https://github.com/Facepunch/garrysmod/blob/0f26c22b9932eb32a17e0b1a464fd72acb187a2e/garrysmod/lua/includes/modules/hook.lua#L208-L222

Maybe your implementation had a problem, because it's going to be really dumb if you depend on "undefined sorting".

DBotThePony:
What would happen if we hook.Remove inside hook.Call of function that already got called?
What would happen if we hook.Add over existing hook inside hook.Call of function that already got called?

This is why it is still unreliable.

Srlion:
1. nothing
2. nothing

If you are talking about the default behavior then, like what I said, it's a defined behavior that shouldn't be changed. (even if you have to be lucky to modify/remove a hook before it's called)
But if you are talking about this lib, then you can test all the cases you want and good luck trying to break it!

Addons can't just relay on pairs sorting, name me a few addons that I can test with?

ExtReMLapin:
>Implementing any kind of sorting over pairs produce known to me issues with existing addons.

That's pretty funny coming from the guy ~spamming pony shit on github~ who is responsible of dlib.

Lua hashtables are unsorted and that's an issue here.

Gamemodes should not use hooks, only declare them so addons can hook to the said gamemode.
An for a good reason : execution order.


Having unpredictable execution order is design issue.

garrysmod/lua/includes/modules/hook.lua
```diff
@@ -1,158 +1,101 @@
-local gmod = gmod
-local debug = debug
 local pairs = pairs
-local getmetatable = getmetatable
-local setmetatable = setmetatable
+local isstring = isstring
+local isfunction = isfunction
+
+local function getGAMEMODE()
```
ExtReMLapin, line 5: overkill change for this one, wasn't required tbh

Srlion: https://github.com/Facepunch/garrysmod/blob/a254b97be82a02899bf3093f93fac277cc8a128f/garrysmod/gamemodes/base/gamemode/cl_init.lua#L73-L79
but I could revert this change

Srlion:
Here is the benchmark results after making some edits:
```
Srlion's New Lib...
Srlion's New Lib took 0.077276400 s
Default Lib...
Default Lib took 1.866785300 s

-------------
CallInvalid (500 calls)
Default Lib (176.42% faster)
Default Lib:      0.003387400 s
Srlion's New Lib: 0.009363500 s
-------------
CallNoHooks (20000000 calls)
Srlion's New Lib (3.61% faster)
Default Lib:      0.020490100 s
Srlion's New Lib: 0.019777100 s
-------------
CallGMOnly (20000000 calls)
Default Lib (0.09% faster)
Default Lib:      0.041348200 s
Srlion's New Lib: 0.041384600 s
-------------
CallNoGM (3200000 calls)
Srlion's New Lib (26580.75% faster)
Default Lib:      0.876035900 s
Srlion's New Lib: 0.003283400 s
-------------
CallGM (3200000 calls)
Srlion's New Lib (27399.49% faster)
Default Lib:      0.925467700 s
Srlion's New Lib: 0.003365400 s
-------------
```

I also tested calling hook.Call from C/C++ using https://pastebin.com/P85Fixgd & https://pastebin.com/cJABiJSL, here are the results:
```
-- DEFAULT Branch
-- Default lib: 0.5745151 ~ 0.630739
-- This lib: 0.445342 ~ 0.4860537

-- x86-64 Branch
-- Default lib: 0.559033 ~ 0.6030091
-- This lib: same as default branch
```
But when going _full monkey mode_ (love you @ExtReMLapin) on x86-64 branch and replacing varargs(...) with ```a1, a2, a3, a4, a5, a6, a7, a8``` for this lib:
```
0.2143003 ~ 0.2659884
```

ExtReMLapin:
x86-64 got LuaJIT 2.1 bytecode and it's jit compiler, it's always (or at least it should, if not, report it) faster than past versions.


>But when going full monkey mode (love you @ExtReMLapin) on x86-64 branch and replacing varargs(...) with a1, a2, a3, a4, a5, a6, a7, a8 for this lib:

garryspins:
Merge this already tf

ExtReMLapin:
Does benchmarking with JIT on really makes sense in GMod ? Because of all the C Api calls

GitSparTV:
True, but it still can help little bit

Kefta:
This won't be merged until we update to the newest LuaJIT and set up the hook battle royale with cash prizes.

thegrb93:
> 
> 
> This won't be merged until we update to the newest LuaJIT and set up the hook battle royale with cash prizes.

@Kefta requirements should be very strict particularly what behavior to take when editing hook table while iterating the hook table.

Srlion:
> This won't be merged until we update to the newest LuaJIT and set up the hook battle royale with cash prizes.

I’m ready with a new & faster one 😎

Srlion:
> > This won't be merged until we update to the newest LuaJIT and set up the hook battle royale with cash prizes.
> 
> @Kefta requirements should be very strict, particularly what behavior to take when editing hook table while iterating the hook table. Current behavior is undefined and a specific behavior will likely break addons.

I already wrote them down:

If you add a new hook while the event is called, you expect it to run in next call. (Default Lib has undefined behavior when adding new hooks, but most of the time it runs in next call.)
If you modify a hook before it's called, the new function will be called not the old one. (Just like the test above.)
If you remove a hook before it's called, then it won't be called in this event call.

FredyH:
The benchmarks you used are just insane (and I suppose meepen is to blame for this). They are probably the worst possible benchmarks that you could ever design for a hook library. I really hope I am just entirely mistaken but from what I can tell you benchmark calling a hook with 3 functions 2 billion times.

Please tell me in what world this scenario reflects anything even close to realistic? Most of the time you will have _at least_ 10 hooks per event if not way more (e.g. Think, Render, etc.). Any good benchmark should also include how performance scales with more hooks added, which is obviously not the case here.

The functions you are calling are also entirely empty - again just completely removed from reality - which means that you found out that the jit is indeed able to optimize function calls that do nothing. Congrats, not running code is faster than running code.

The issue is we already know that jit compilation will likely not work in real life because of all the calls into C++ that are required when doing basically anything in gmod, so your benchmark just does not make sense on any level.

Furthermore, I was expecting this library to guarantee that hooks are executed in the order they are added, but even that does not seem to be the case from me quickly reading through the code.

So unless you show some benchmarks that are actually realistic that demonstrate some massive performance improvement in those scenarios, I do simply not see how making code this ugly (using goto, yikes) is something we ever want to do.

Srlion:
> The benchmarks you used are extremely misleading (and I suppose meepen is to blame for this). They are not suitable for benchmarking a hook library for gmod. I really hope I am just entirely mistaken but from what I can tell you benchmark calling a hook with 3 _empty_ functions 2 billion times.
> 
> I do not believe this scenario is anywhere close to being realistic. Most of the time you will have _at least_ 10 hooks per event if not way more (e.g. Think, Render, etc.). But more importantly it would be very weird for a hook function to just be empty.
> The issue you have is that what likely happens is that luajit just optimizes away any work and at the end you get the result that not running any code is faster than running code.
> 
> The issue is we already know that jit compilation will likely not work in real life because of all the calls into C++ that are required when doing basically anything in gmod, so your benchmark just does not make sense in this regard.
> 
> Any good benchmark should also include how performance scales with more hooks and if maybe one of the two implementations does better at certain count of hooks.

I realized that like 3 weeks ago when I was making a new lib, where most events will have 10~15 hooks + most of them will have stuff that will stop the code from being jitted.

> Furthermore, I was expecting this library to guarantee that hooks are executed in the order they are added, but from what I gathered from very briefly scanning through the code, this is not the case.

Idk where you got that guarantee that hooks will be executed in the order they are added and idk why you are talkin about executing order here, there is no need to care about it because the default behavior is random.

> Unless you can show some benchmarks that are actually realistic and that demonstrate some massive performance improvement in those scenarios, I do simply not see how making code a lot uglier (using goto...) is something we ever want to do.

I tested with https://pastebin.com/P85Fixgd & https://pastebin.com/cJABiJSL but only added 10 hooks and got this:
> Mine: 0.0454218 ~ 0.050212 s
> Default: 0.0780487 ~ 0.0817999 s

That's with my new hook lib (code is ugly but there is a default behavior that I gotta follow), I'm not sure if it's gonna be a huge boost, but w/e.

FredyH:
The C++ benchmark is certainly better. However, calling CurTime() is apparently really fast (on my system ~15ns per call, idk how it is that fast) so you are still benchmarking almost empty functions (but I suppose jit does not play a role anymore?).
I have the issue with your initial benchmarks (and the updated ones), that claim to be 200(!) times faster than the default implementation in some circumstances, but that is simply not possible for what essentially just amounts to iterating over a table.

Either way, I think even with the C++ benchmark, the scenario is still not exactly realistic. I'd be interested in some real world benchmarks (darkrp or some other terrible gamemode). If you can show a meaningful improvement there (I'd say at least 10%), then it might be worth switching to the new implementation.

> Idk where you got that guarantee that hooks will be executed in the order they are added and idk why you are talkin about executing order here, there is no need to care about it because the default behavior is random.

When I read through the code I got a bit excited because at least without removing hooks you do have that guarantee in your code. This is something I'd wish for in a hook library since always having an unspecified order seems a bit bad. But it is certainly by no means required since the original does not guarantee anything either. I was just trying to make the point that if you somehow managed to improve the semantics here, it'd certainly be worth considering even if the performance improvement is just marginal.

Kefta:
By the way, you shouldn't be using CurTime for benching since it's tied to the in-game time. Use SysTime when benching from Lua.

thegrb93:
They were using it for a cheap C++ boundary call if I'm not mistaken, not to time it.

DBotThePony:
> Furthermore, I was expecting this library to guarantee that hooks are executed in the order they are added, but from what I gathered from very briefly scanning through the code, this is not the case.

This is basically sorting and

> Sorting - i tried implementing sorted hook back in 2017, it didn't end well with one of my friend's addon and with at least one workshop addon

Old addons are doing questionable things and any kind of sorting will very likely break them

Srlion:
> > Furthermore, I was expecting this library to guarantee that hooks are executed in the order they are added, but from what I gathered from very briefly scanning through the code, this is not the case.
> 
> This is basically sorting and
> 
> > Sorting - i tried implementing sorted hook back in 2017, it didn't end well with one of my friend's addon and with at least one workshop addon
> 
> Old addons are doing questionable things and any kind of sorting will very likely break them

Any kind of sorting should just work fine, if it doesn’t then you have some shit addons

ExtReMLapin:
What you guys don't know is _Kilburn is still working on gmod since all those years.
He's busy porting all of the lua c api calls to FFI calls, which will allow us to entirely JIT the code.

meepen:
The newer versions of luajit can hit most code even if there is surrounding
c calls now.

ExtReMLapin:
Stitching just "cancels" trace aborts, but isn't as fast as full jit

Denneisk:
Any plans on reintegrating your updated hook library or is this one dead in the water, too?

bruhhhandaluzxyt:
So... What will happen with this? And what happened?

PR #1644, opened 2020-03-24
Add iscolor

Heyter:
Credits: HELIX FRAMEWORK

> Returns "true" if the given input is a color table. This is necessary since the engine "IsColor" function only checks for color metatables - which are not used for regular Lua color types.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -837,11 +837,7 @@ end
 
 function SelectRoles()
    local choices = {}
-   local prev_roles = {
-      [ROLE_INNOCENT] = {},
-      [ROLE_TRAITOR] = {},
-      [ROLE_DETECTIVE] = {}
-   };
+   local prev_roles = {}
 
    if not GAMEMODE.LastRole then GAMEMODE.LastRole = {} end
 
@@ -853,10 +849,8 @@ function SelectRoles()
          -- save previous role and sign up as possible traitor/detective
 
          local r = GAMEMODE.LastRole[v:SteamID()] or v:GetRole() or ROLE_INNOCENT
-
-         table.insert(prev_roles[r], v)
-
-         table.insert(choices, v)
+		 prev_roles[v] = r
+		 choices[#choices + 1] = v
       end
 
       v:SetRole(ROLE_INNOCENT)
@@ -881,7 +875,7 @@ function SelectRoles()
       -- make this guy traitor if he was not a traitor last time, or if he makes
       -- a roll
       if IsValid(pply) and
-         ((not table.HasValue(prev_roles[ROLE_TRAITOR], pply)) or (math.random(1, 3) == 2)) then
+         ((prev_roles[pply] ~= ROLE_TRAITOR) or (math.random(1, 3) == 2)) then
          pply:SetRole(ROLE_TRAITOR)
 
          table.remove(choices, pick)
@@ -913,11 +907,7 @@ function SelectRoles()
       local pply = choices[pick]
 
       -- we are less likely to be a detective unless we were innocent last round
-      if (IsValid(pply) and
-          ((pply:GetBaseKarma() > min_karma and
-           table.HasValue(prev_roles[ROLE_INNOCENT], pply)) or
-           math.random(1,3) == 2)) then
-
+      if (IsValid(pply) and (pply:GetBaseKarma() > min_karma and prev_roles[pply] == ROLE_INNOCENT) or math.random(1,3) == 2) then
          -- if a player has specified he does not want to be detective, we skip
          -- him here (he might still get it if we don't have enough
          -- alternatives)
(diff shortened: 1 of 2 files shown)


neico:
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/util/color.lua#L33

```lua
function IsColor( tab )
    return getmetatable( tab ) == debug.getregistry().Color
    -- return tab.IsColor && tab:IsColor()
end
```
I'm actually not sure which of those 2 variants is faster (probably the commented out one...)

thegrb93:
>color tables should always be created trough Color()

One would think, but not the case unfortunately.

neico:
> One would think, but not the case unfortunately.

The only reason I could think of for that would be the missing ability to pass a table with values to the "constructor"

thegrb93:
There's glua functions that return a r,g,b,a table instead of a color.

Kefta:
That's an outstanding bug that should be resolved by making colors userdata, or setting the metatable as it comes back from C. Adding a duplicate iscolor is just adding a bandage to a long-time wound.

thegrb93:
userdata is slower so I don't recommend that. Setting metatable would be good.

Kefta:
It would actually be more efficient since the C-side wouldn't have to read 4 string keys from a table, but the issue is that it would break compatibility with code that does something like:
```Lua
function Foo(color)
   if (istable(color)) then
      -- do stuff
   end
end
```

So it appears the C-side will just have to set the metatable if this issue will ever be resolved. Just grabbing it from _R.Color seems shaky, though.

robotboy655:
I did some tests and basically, reading a C Color class would be quite a bit faster, 10 times faster on average, however creating it would be severely slower. Like **hundreds** of times slower.

Kefta:
So it's really in the same boat as Vector and Angle - caching them is a huge perf benefit.

thegrb93:
I think the cache was added just to make them not as terrible, but it's still better to just use a table.

Kefta:
I meant caching in Lua, as in caching constant objects in the file like
```
local vConstant = Vector(1, 2, 3)

function foo(v)
   return vConstant + v
end
```

Srlion:
Most of the time it’s always better to define functions in Lua for performance, Color function is JIT compiled so it’s so fast, defining it in C could make most addons that don’t cache colors rape your FPS. (plz don’t say that they should start caching them because almost all addons I have seen don’t cache them)

neico:
`Color()` could cache them internally and just return that going forward

Each Color variant only needs to exist once (and hence be created once only), the only thing I'd be worried about people actively modifying the values post-creation and possibly affecting the cached one (requiring to do copies of the static one each time)
It'd be interesting to see how many actually do that to decide if it's feasible to discourage such use and bother with making copies on each request
(a `__usednewindex` would prove very useful in that case to prevent it outright, well, one can dream...)

An interesting discussion also would be if it could be feasible to just initialize all possible color variants up front (I don't think many of us have a shortage of RAM these days) and how big such a case could even get (I mean doing it engine-wide could probably save up quite a few bytes I'd assume, but then it's a question of if the run-time creation costs outweighs the bigger memory footprint)
Maybe something rubat can play around with a bit internally and provide more interesting results.

robotboy655:
Creating every possible color combination is definitely not possible, I don't think you realize just how many combinations that is and how much memory that would take in best case scenario.

ExtReMLapin:
NEVER SAY NEVER

garrysmod/gamemodes/terrortown/gamemode/init.lua
```diff
@@ -875,7 +881,7 @@ function SelectRoles()
       -- make this guy traitor if he was not a traitor last time, or if he makes
       -- a roll
       if IsValid(pply) and
-         ((prev_roles[pply] ~= ROLE_TRAITOR) or (math.random(1, 3) == 2)) then
+         ((not table.HasValue(prev_roles[ROLE_TRAITOR], pply)) or (math.random(1, 3) == 2)) then
```
ExtReMLapin: use hashtables

Heyter: why?

ExtReMLapin: Because hashtables are faster then iterating thru the whole table.

replace `table.insert(prev_roles[r], v)` by `prev_roles[r][v] = true`

ExtReMLapin: And obviously, do the correct changes in the code

thegrb93: His improvements did use hashtables. You're referencing his revert commit.

ExtReMLapin: no u

Heyter: > His improvements did use hashtables. You're referencing his revert commit.

Yes, you're right, I returned the commit because I could not create another branch.

Heyter: > no u

Yes, I returned the commit because I could not create another branch.

robotboy655:
We already have `IsColor`. We are not going to be fixing its flaws by adding another `iscolor` function. I do believe `IsColor` is defined in Lua.

Discussion about this can move to https://github.com/Facepunch/garrysmod-issues/issues/2407

PR #1645, opened 2020-03-29
Fix GetEyeTraceNoCursor caching

Kefta:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/4430.

Also use FrameNumber instead of CurTime for caching so that it doesn't clash with prediction.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua
+++ b/garrysmod/gamemodes/base/gamemode/obj_player_extend.lua
@@ -172,13 +172,13 @@ end
 --[[---------------------------------------------------------
 	Player Eye Trace
 -----------------------------------------------------------]]
-function meta:GetEyeTrace( skipcache )
+function meta:GetEyeTrace()
 	if ( CLIENT ) then
 		local framenum = FrameNumber()
 
 		-- Cache the trace results for the current frame, unless we're serverside
 		-- in which case it wouldn't play well with lag compensation at all
-		if ( !skipcache && self.LastPlayerTrace == framenum && istable( self.PlayerTrace ) ) then
+		if ( self.LastPlayerTrace == framenum ) then
 			return self.PlayerTrace
 		end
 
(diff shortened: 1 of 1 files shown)


neico:
Weren't changes supposed to use the C-Style operators to align with the current codebase?

Kefta:
Good call.

robotboy655:
that `istable` call impacts performance, not sure we need it. Was it ever a problem? Also the change from CurTime() would also need to be applied to TTT

Kefta:
It's not necessary, I'll remove it. Where in TTT do I need to apply the change?

Donkie:
Couldn't entity parameters that's only used in one place like this easily be put in a local table next to the function, like:
```lua
local lastPlayerTrace = {}
function meta:GetEyeTrace( skipcache )
	if ( CLIENT ) then
		local framenum = FrameNumber()

		-- Cache the trace results for the current frame, unless we're serverside
		-- in which case it wouldn't play well with lag compensation at all
		if ( !skipcache && lastPlayerTrace[self] == framenum ) then
			return self.PlayerTrace
		end

		lastPlayerTrace[self] = framenum
	end

	local tr = util.TraceLine( util.GetPlayerTrace( self ) )
	self.PlayerTrace = tr

	return tr
end
```
This would be both faster and remove the need of a potential conflicting parameter.

Kefta:
Storing entities in a local table is a great idiom that none of the rest of the GMod codebase uses. It also requires an EntityRemoved to remove old entities from the table, otherwise, old references will keep sitting around (weak keys don't work). Best to leave it as a player variable for consistency.

robotboy655:
https://github.com/Facepunch/garrysmod/blob/82e842850024fa15f6ce2e20916da8b0bb2a11e4/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua#L134

Kefta:
Removed the skipcache arg since it might conflict with TTT's override, and removed the istable check.

PR #1653, opened 2020-04-08
[TTT] Additional language fixes

Kefta:
Hopefully the final one.
- Remove trailing spaces from tradchinese
- Add translated Select language to Italian

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/italian.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/italian.lua
@@ -307,7 +307,7 @@ L.set_title_lang = "Impostazioni lingua"
 -- trovare l'impostazione della lingua anche se è una lingua che non conoscono.
 -- It may be best to leave this next one english, so english players can always
 -- find the language setting even if it's set to a language they don't know.
-L.set_lang = "Language selection:"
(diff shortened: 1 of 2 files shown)


garrysmod/gamemodes/terrortown/gamemode/lang/italian.lua
```diff
@@ -307,7 +307,7 @@ L.set_title_lang = "Impostazioni lingua"
 -- trovare l'impostazione della lingua anche se è una lingua che non conoscono.
 -- It may be best to leave this next one english, so english players can always
 -- find the language setting even if it's set to a language they don't know.
-L.set_lang = "Language selection:"
+L.set_lang = "Selezionare lingua (Select language):"
```
svdm, line 310: My Italian is not fluent but the infinitive for selezionare seems weird here (vs imperative). I know that came up for Portuguese as well, and the right choice probably depends on the language. Maybe @PinoMartirio has an opinion?

Kefta: The right choice doesn't really depend on the language for Romance languages, it's apparently standard in the modern age for Spanish and Portuguese to use unconjugated forms instead of the singular they form when you don't have a specific target. I decided to do the same for Italian for consistency.

PR #1654, opened 2020-04-09
[TTT] italian.lua hotfix

Perkzitos:
Probably L.body_fount_t will work correctly now.


Kefta:
It should already work correctly. Is this necessary?

Perkzitos:
> It should already work correctly. Is this necessary?

I think so.

svdm:
This is not necessary, language files are just Lua and this is not a syntax error. Just to confirm I loaded up the language and tested it, it works fine without this change.

I appreciate the contributions but please don't submit PRs without verifying the problem (and testing the fix).

PR #1655, opened 2020-04-11
net.WriteArray and net.ReadArray

GitSparTV:
These functions help to reduce the net cost for arrays.

Writing arrays by net.WriteTable costs `(8 + 64 + 8 + v)` per value. Which is `(80 + v)` per value. Where `v` is the value's cost.

Writing arrays by net.WriteArray costs `(8 + v)` per value. This is 72 bits less.

The reason why net.ReadArray uses `tab[n] = v; n = n + 1` because external counter is faster than `tab[#tab + 1] = v`

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -127,6 +127,34 @@ function net.ReadTable()
 
 end
 
+function net.WriteArray(tab)
+
+	for k, v in ipairs(tab) do
+
+		net.WriteType(v)
+
+	end
+
+	-- End of table
+	net.WriteType(nil)
+
+end
+
+function net.ReadArray()
+
+	local tab, n = {}, 1
+
+	while true do
+
+		local v = net.ReadType()
+		if (v == nil) then return tab end
+
+		tab[n] = v
+		n = n + 1
+
+	end
+end
+
 net.WriteVars = 
 {
 	[TYPE_NIL]			= function ( t, v )	net.WriteUInt( t, 8 )								end,


Kefta:
We should encourage people to not be writing randomly-typed containers over the net stream for how much data it wastes. This and net.Write/ReadTable should not exist.

GitSparTV:
This is why I wrote some tips in https://wiki.facepunch.com/gmod/Net_Library_Usage#Improving

__But__, this rule doesn't apply to dynamic data transferring. The are still some cases where you have to use net.WriteTable.

Btw, I found one limitation of my PR.
If you have table in the array, it will write it as a table.

Kefta:
There is no case in which you have to use dynamic data transferring - when would you not know the types you're sending over the data stream?

GitSparTV:
Ok I can't think for anything right now. Everything can be made to known-types, but sometimes it's too complex, like if you have interchangeable types

thegrb93:
pac3 uses it extensively. starfall uses its own implementation but same concept.

GitSparTV:
@thegrb93 Do you know the case when it's used?

thegrb93:
https://github.com/CapsAdmin/pac3/blob/97ab99e9e8f5f16063ee5480ee33d21970822b8c/lua/pac3/libraries/netx.lua

thegrb93:
For transmitting users' outfits in pac3, and starfall uses it to transmit the generated documentation.

gspetrou:
Was talking about this before on the Discord and going to throw it here for the sake of further discussion. I feel that net.Write/Read**Array** should imply each element is of the same type in the (numeric) table and if thats the case we can save some bits like this:
```lua
function net.WriteArray( arr, bitMax, writeFunc, ... )
	net.WriteUInt( #arr, bitMax )
	for i = 1, #arr do
		writeFunc( arr[i], ... )
	end
end

function net.ReadArray( bitMax, readFunc, ... )
	local arrMaxize = net.ReadUInt( bitMax )
	local arr = {}
	for i = 1, arrMaxize do
		arr[i] = readFunc( ... )
	end
	return arr
end

-- Example:
-- Server Code:
net.WriteArray({5, 2, 6, 9}, 2, net.WriteUInt, 4)
-- Client code:
local arr = net.ReadArray(2, net.ReadUInt, 4)
```
But I feel like this might get denied since expecting people to pass another function and vararg isn't entirely idiot proof.

GitSparTV:
It's easier to have such functions in your addon without passing the function. Just make a know structure and that's it.
I think I will close this PR by the same reason.

PR #1656, opened 2020-04-20
[TTT] associative array for prev_roles

Heyter:


Kefta:
TTT uses three spaces for indentation, not tabs.

Heyter:
> TTT uses three spaces for indentation, not tabs.

No problem, fixed.

garrysmod/gamemodes/terrortown/gamemode/init.lua
```diff
@@ -893,7 +889,7 @@ function SelectRoles()
    -- traitor, so becoming detective does not mean you lost a chance to be
    -- traitor
    local ds = 0
-   local min_karma = GetConVarNumber("ttt_detective_karma_min") or 0
+   local min_karma = GetConVar("ttt_detective_karma_min"):GetInt() or 0
```
Kefta, line 892: Cache the GetConVar call outside of the function if you're gonna change it.

garrysmod/gamemodes/terrortown/gamemode/init.lua
```diff
@@ -889,7 +889,6 @@ function SelectRoles()
    -- traitor, so becoming detective does not mean you lost a chance to be
    -- traitor
    local ds = 0
-   local min_karma = GetConVar("ttt_detective_karma_min"):GetInt() or 0
```
Kefta: Keep this cached outside the loop so it doesn't get called multiple times uselessly.

Heyter: > Keep this cached outside the loop so it doesn't get called multiple times uselessly.

?

svdm:
SelectRoles runs once per round. The tables this PR is making associative contain maybe 10 entries on a typical 16 player server (for ROLE_INNOCENT, fewer for the traitor one). On default settings we're picking 4 traitors and 2 detectives, which will involve only a handful of `table.HasValue` calls on those very small tables. There's just not much to optimize here.

As a consequence I doubt this change will make a measurable difference on any real-world server. So why mess with code that has worked correctly for ages and risk introducing bugs, for literally no benefit? I know this type of zero-impact optimization is the gmod lua community's holy crusade, but really, why?

PR #1658, opened 2020-04-25
dlabel: make white default text color

NathanX-S:
why not


PatrickRatzow:
You'll be breaking a lot of menus that doesn't set the colour of DLabel.

NathanX-S:
because gmod basically has to support cavemen, closed

PR #1659, opened 2020-05-01
Add DComboBox:OnMenuOpened(menu)

WilliamVenner:
Simple override to allow for advanced custom DMenu changes once it has been opened by the DComboBox.


WilliamVenner:
Use case example:

Say I have a DComboBox populated with choices of sound files. I want to allow the player to preview those sound files before they select the one they'd like to choose. I can create a function which plays the sound file of a hovered DMenuOption. I can hook into DComboBox.OnMenuOpened, iterate through the DMenuOptions and apply that hovered function to each option.

PR #1661, opened 2020-05-05
Dtree node - self.Expanded validity check

ravo10:
In my gamemode I edited the default spawnmenu. I derived the gamemode from Sandbox and deleted some `Panels` at start. I got an error telling me that `:SetVisible ` could not be used on a nil value. `self.Expanded` was nil. This pull request removed that error for me.


ravo10:
Ah, shoot, I renormalized all files... The change was atleast at `dtree_node.lua` line `237`.

neico:
You should probably squash your changes so the UI properly removes the unwanted changes

ravo10:
@neico You mean close this pull-request and open a new one with not all files renormalized ? I can do that.

Donkie:
That won't be necessary. If you "squash" multiple commits together into one commit, the problem should be fixed. https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

ravo10:
> That won't be necessary. If you "squash" multiple commits together into one commit, the problem should be fixed. https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

Ah, yes! I will check it out. I appreciate the help both!

neico:
Squashing describes a method of git rebase, where you select a commit that you want to be squashed (merged) into the previous commit.

I didn't provide a proper explanation because if you use git you should be able to know about it or at least know where to get the info about it.
Not to mention that the instructions differ depending on which tool you use for git.

The `gist` of it is this basic git chain of commands: https://gist.github.com/longtimeago/f7055aa4c3bba8a62197
most git tools provide rebasing, and that's where you can find it.
I'll provide an example for TortoiseGit because that's what I usually use as it also showcases how it could roughly look like: https://tortoisegit.org/docs/tortoisegit/tgit-dug-rebase.html

ravo10:
> Squashing describes a method of git rebase, where you select a commit that you want to be squashed (merged) into the previous commit.
> 
> I didn't provide a proper explanation because if you use git you should be able to know about it or at least know where to get the info about it.
> Not to mention that the instructions differ depending on which tool you use for git.
> 
> The `gist` of it is this basic git chain of commands: https://gist.github.com/longtimeago/f7055aa4c3bba8a62197
> most git tools provide rebasing, and that's where you can find it.
> I'll provide an example for TortoiseGit because that's what I usually use as it also showcases how it could roughly look like: https://tortoisegit.org/docs/tortoisegit/tgit-dug-rebase.html
> 
> Make sure you use a force flag with the push command afterwards,
> as a rebase basically rewrites git history, which a normal push doesn't allow by default (but isn't really relevant for a pull request)
> https://git-scm.com/docs/git-push#Documentation/git-push.txt---force-with-leaseltrefnamegt

Thanks! OK, I will try and fix this issue now!

PR #1662, opened 2020-05-05
Precise lerp

ravo10:
Could add better performance in milliseconds.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -246,6 +246,15 @@ function Lerp( delta, from, to )
 end
 
 --[[---------------------------------------------------------
+	Precise lerp - Basically the same as `Lerp( ... )`, but will always guarantee that the return value = `to` when `delta` = 1.
+-----------------------------------------------------------]]
+function LerpPrecise( delta, from, to )
+
+	return ( 1 - delta ) * from + delta * to;
+
+end
+
+--[[---------------------------------------------------------
 	Convert Var to Bool
 -----------------------------------------------------------]]
 function tobool( val )


garrysmod/lua/includes/util.lua
```diff
@@ -1,460 +1,469 @@
...
+	if ( delta < 0 ) then return from end
+
+	return from + ( to - from ) * delta
+
+end
+
+--[[---------------------------------------------------------
+	Precise lerp - Basically the same as `Lerp( ... )`, but will always guarantee that the return value = `to` when `delta` = 1.
+-----------------------------------------------------------]]
+function LerpPrecise( delta, from, to )
```
ravo10, line 251: This is the change!

garrysmod/lua/includes/util.lua
```diff
@@ -1,460 +1,469 @@
...
+
+	return (1 - delta) * from + delta * to;
+
+end
```
ravo10, line 255: This is the change!

Donkie:
This will not get merged until you properly fix the line endings or whatever is causing all the changes

ravo10:
> This will not get merged until you properly fix the line endings or whatever is causing all the changes

I have no idea why that happend... I only staged the lines I wanted. I have to look into it.. I have not used git allot before.

ravo10:
OK fixed!!

ravo10:
> OK fixed!!

@neico @Donkie I used `git add --renormalize -p` and the `git rebase -i HEAD~[n]` and that fixed it for me. Thanks for the help.

Donkie:
Why is this needed? Your code should rarely if ever depend on a float number being an exact value.

thegrb93:
I have seen this issue in the professional world so there are cases. Personally I'd prefer this replace the original.

Kefta:
The original should just be moved to Lua.

robotboy655:
The original is in Lua.

Kefta:
I'm blind btw. Sorry it had to come out like this.

robotboy655:
Surely we can just change the existing `Lerp` to do this and not add a duplicate function?

PR #1663, opened 2020-05-05
Update getmaps.lua

MoskalykA:


robotboy655:
No thank you. Gamemode maps get their own categories when their gamemode is installed, we don't want to be hardcoding all possible gamemode variations into getmaps.lua

What exists there is there because those gamemodes didn't come from GMod itself or are from the mountable games.

Kefta:
If that's the case, all the custom gamemodes at the bottom should have their categories removed. This is inconsistent.

Kefta:
Though I think they should just all be allowed - who has all of these gamemodes installed?

robotboy655:
Huh, how did some of those get in there? 

But honestly, do we really want to start aggregating gamemodes there like that? Why not just starting hardcoding gamemode names for everything like in the server list. At what point do we just ship all gamemodes with the game?

Kefta:
I think harcoding them in the map categories is fine since the prefixes themselves are hardcoded: ``scp_`` will never be anything but ``Secure. Contain. Protect.``, ``dr_`` will never be anything but ``Deathrun``, etc. It's for user convenience and it's updated like once a year at most.

robotboy655:
I don't mind having very popular things there - RP, deathrun, surf, bhop, etc, but I don't think custom obscure things have a place there.

PR #1664, opened 2020-05-05
Do_Undo 'forceUpdateClientUndoList' method option

ravo10:
I found that in some cases I need to update on the client side as well (the user undo list). This change allows that. It is tested multiple times, and I found no bugs.

### DIFF
--- a/garrysmod/lua/includes/modules/undo.lua
+++ b/garrysmod/lua/includes/modules/undo.lua
@@ -336,7 +336,7 @@ end
 --[[---------------------------------------------------------
 	Undos an undo
 -----------------------------------------------------------]]
-function Do_Undo( undo )
+function Do_Undo( undo, forceUpdateClientUndoList )
 
 	if ( !undo ) then return false end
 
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/modules/undo.lua
```diff
@@ -1,451 +1,486 @@
...
+						end
+	
+					end
+	
+				end
+			end
+
+		end
+	end
+
```
ravo10, line 389: The logic

garrysmod/lua/includes/modules/undo.lua
```diff
@@ -1,451 +1,486 @@
...
+	Current_Undo = nil
+
+end
+
+--[[---------------------------------------------------------
+	Undos an undo
+-----------------------------------------------------------]]
+function Do_Undo( undo, forceUpdateClientUndoList )
```
ravo10, line 339: The parameter

garrysmod/lua/includes/modules/undo.lua
```diff
@@ -1,451 +1,486 @@
...
+			func[ 1 ]( undo, unpack( func[ 2 ] ) )
+			count = count + 1
+
+		end
+	end
+
+	-- Force-update the client list
+	-- In some instances, this is usefull. When you also want the
+	-- client to get their list updated for whatever reason.
+	if ( undo.Entities && forceUpdateClientUndoList ) then
```
ravo10, line 358: The method

ravo10:
I will try fix that weird issue I am having where git adds the whole file, but not locally..

ravo10:
Jeesuz. I am learning

robotboy655:
What about just using `gmod_undo` /  `gmod_undonum <num>` console commands instead?

ravo10:
> What about just using `gmod_undo` / `gmod_undonum <num>` console commands instead?

Yes, that works also! I am just tried making a lazy way.

robotboy655:
Closing this as well then, using the console commands is a better solution. I have also added this info to the wiki.

ravo10:
> Closing this as well then, using the console commands is a better solution. I have also added this info to the wiki.

I don't use GitHub much. Yes!

PR #1665, opened 2020-05-05
Dtree node - self.Expanded validity check

ravo10:
In my gamemode I edited the default spawnmenu. I derived the gamemode from Sandbox and deleted some `Panels` at start. I got an error telling me that `:SetVisible ` could not be used on a nil value. `self.Expanded` was nil. This pull request removed that error for me.

### DIFF
--- a/garrysmod/lua/vgui/dtree_node.lua
+++ b/garrysmod/lua/vgui/dtree_node.lua
@@ -234,9 +234,9 @@ end
 
 function PANEL:PerformRootNodeLayout()
 
-	self.Expander:SetVisible( false )
-	self.Label:SetVisible( false )
-	self.Icon:SetVisible( false )
+	if ( IsValid(self.Expander) ) then self.Expander:SetVisible( false ) end
+	if ( IsValid(self.Label) ) then self.Label:SetVisible( false ) end
+	if ( IsValid(self.Icon) ) then self.Icon:SetVisible( false ) end
 
 	if ( IsValid( self.ChildNodes ) ) then
 


Donkie:
Of course it will throw errors at places when you delete components that the panel depends on. This isn't how you fix your issue.

Kefta:
The issue should be fixed by adding validity checks everywhere.

ravo10:
On start-up of the Sandbox gamemode, there is a category called SProps which has the icon of a brick ("icon16/brick.png") which got populated automatically on start. I manged to find and hinder everything else from populating into the spawnmenu except that. I needed to remove it, so I had to delete it on startup.. But what I did for my gamemode now, is that I added that part (this change) to my gamemode, so it will be used if the active gamemode is my gamemode. I thought I would be usefull, but it won't probably if you don't mod. it like I did! I run this code inside `custom.lua` inside the `AddCustomizableNode` function.

```
if !haveDeletedOldPanelDropdowns then
	haveDeletedOldPanelDropdowns = true
	local panel = parent:GetChildren()[1] -- [name:Panel][class:Panel][0,0,64,24]

	for _,DTree_Node in pairs(panel:GetChildren()) do
		for _,childrenOfDTree_Node in pairs(DTree_Node:GetChildren()) do
			--[[ 
				1	=	Panel: [name:DTree_Node_Button][class:Label][0,0,64,22]
				2	=	Panel: [name:DExpandButton][class:Label][0,0,15,15]
				3	=	Panel: [name:DImage][class:Panel][0,0,16,16]
				4	=	Panel: [name:DListLayout][class:Panel][0,0,64,68]
				]]
			childrenOfDTree_Node:Remove()
		end
	end
end
```

robotboy655:
This definitely feels like the wrong way to go about things, you should simply not delete DExpandButton & co

If you want to simply delete all nodes of a node, use Node:Clear() which should properly delete all nodes of a node.

You should never edit base game files, in addons or otherwise. You addons will get broken by an update to that file, or your addon will break game's update if a file that is also shipped in an addon was changed.

ravo10:
> This definitely feels like the wrong way to go about things, you should simply not delete DExpandButton & co
> 
> If you want to simply delete all nodes of a node, use `Node:Clear()` which should properly delete all nodes of a node.
> 
> You should never edit base game files, in addons or otherwise. You addons will get broken by games' update to that file, or your addon will break games' update if a file that is also shipped in an addon was changed.

Okay, yes... I will check `Node:Clear()`. Hmm, I could have asked on Discord maybe. I did change the file name to something unique and only allowed usage if the active gamemode was the correct one.

ravo10:
> This definitely feels like the wrong way to go about things, you should simply not delete DExpandButton & co
> 
> If you want to simply delete all nodes of a node, use `Node:Clear()` which should properly delete all nodes of a node.
> 
> You should never edit base game files, in addons or otherwise. You addons will get broken by games' update to that file, or your addon will break games' update if a file that is also shipped in an addon was changed.

`Node:Clear()` Did the job!! I searched everywhere for something under `DTree`, but should have looked elsewhere.

robotboy655:
So closing this then.

ravo10:
> So closing this then.

Yep

PR #1666, opened 2020-05-07
Update users.txt

MoskalykA:


robotboy655:
```Player.IsUserGroup``` is a valid function.

PR #1671, opened 2020-05-08
Create ping categories to reduce impact of miniscule ping differences

Leystryku:
Tl;dr:
If this gets added anycast is much less of an advantage.

Long desc:
If there's a miniscule ping difference in the lower bound then it should not make a difference.

I propose this change to make the serverlist a fairer place to everyone, since some serverhosts spoof to get super low pings which are only 10 to 20 lower than the other servers actually within that region. It gives them a ranking advantage which you can not get with real servers. This counters that.

Meanwhile gameplay wise for the players there's basically no difference.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -262,7 +262,9 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 
 	data.hasmap = DoWeHaveMap( data.map );
 
-	data.recommended = data.ping;
+	data.recommended = 40;
+	if(data.ping >= 60) data.recommended = data.ping - (data.ping % 50)
+	
 	if ( data.players == 0 ) data.recommended += 75; // Server is empty
 	if ( data.players >= data.maxplayers ) data.recommended += 100; // Server is full, can't join it
 	if ( data.pass ) data.recommended += 300; // Password protected, can't join it
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -263,7 +263,7 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	data.hasmap = DoWeHaveMap( data.map );
 
 	data.recommended = 40;
-	if(data.ping >= 60) data.recommended = data.ping - (data.ping % 50)
+	if(data.ping >= 60) data.recommended = data.ping;
 	
 	if ( data.players == 0 ) data.recommended += 75; // Server is empty
 	if ( data.players >= data.maxplayers ) data.recommended += 100; // Server is full, can't join it


Donkie:
Wouldn't a fitted curve to this data be better to prevent any obvious grouping of servers. With fixed categories you still have the issue you describe because a 101 ping server would be twice as "bad" as a 99 ping server.

Leystryku:
I'd be up for alternatives tbh, not sure whether something like this will even get implemented.

The idea basically is to avoid people with e.g. pingspoofing having more popular servers because their ping is 20 while the other real servers are having e.g. 40.

BlackCetha:
A linear function as it is right now seems fine to me. "Miniscule" ping differences will result in small differences in rating.
The example you give is not a small difference.

Leystryku:
@BlackCetha 
The issue with the current way is that the main factor for servers showing up is ping.
That's why there's an entire host which focuses on spoofing ping. If you buy a server from them (which costs ~400 to 1000$ depending on your community),
they'll ensure that your ping is basically lower than any others.

Even people actually in that location will have ping variation due to the way source works going from e.g. ~30 to max ~60.

However, when using GMC you consistently only get ~20.
Pings which are not achieveable by normal servers.

And that's what this PR is supposed to avoid, so that regardless whether you have a ping spoofed server, or a actual server located there, the ranking's equivalent.

willox:
is there any benefit to this vs. clamping the ping to be 60 or higher?

Donkie:
Yeah I agree, pings under 60 shouldn't be punished at all, beyond that I believe the punishment should follow an exponential growth.

Leystryku:
@willox 
Can you be a little more specific? If you mean putting all over 60 in one group  and that's the only thing done I think that'd be a bad idea. It's specificly aimed at putting the lower ping counts in one group.
Tbh, perhaps even just doing something like (ping <= 60) { ping = 0; } might be enough.

Ah, I get it. yeah exactly, alright, I'll make the PR do that then.

@Donkie 
Actually as said in my big post, this is mainly aimed at the servers with low ping (i.e. under 60):

While at ping 20 vs ping 60 there's barely any difference playability wise (especially in a game such as gmod compared to e.g. csgo), yet it affects ranking so much that people pay thousands a year to spoof their ping.

willox:
>  if you mean putting all over 60 in one group and that's the only thing done I think that'd be a bad idea.

i mean the exact opposite

Leystryku:
@willox yeah I figured, I've changed the PR do to that now.

willox:
ok I think we'll probably try this some when, but it won't be today

garrysmod/html/js/menu/control.Servers.js
```diff
@@ -262,7 +262,9 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 
 	data.hasmap = DoWeHaveMap( data.map );
 
-	data.recommended = data.ping;
+	data.recommended = 40;
```
MoskalykA, line 265: I don't think 40 is necessarily ideal. I'd say more like 20.

Leystryku: data.recommended default value 20 vs 40 will have no difference. As long as the ping is under 60 all the servers will have data.recommended pre-playercount etc. calculation set to 40

Whether the number being used is 40 or 20 does not matter.

robotboy655:
Let's try it.

PR #1673, opened 2020-05-16
Fix toolgun ghost not being cleaned up in dedicated server when tool is switched to another weapon

thegrb93:
Finally got annoyed and attempted this again. This seems to be the most reliable fix I can think of.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/ghostentity.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/ghostentity.lua
@@ -11,11 +11,6 @@ function ToolObj:MakeGhostEntity( model, pos, angle )
 	if ( SERVER && !game.SinglePlayer() ) then return end
 	if ( CLIENT && game.SinglePlayer() ) then return end
 
-	-- The reason we need this is because in multiplayer, when you holster a tool serverside,
-	-- either by using the spawnnmenu's Weapons tab or by simply entering a vehicle,
-	-- the Think hook is called once after Holster is called on the client, recreating the ghost entity right after it was removed.
-	if ( !IsFirstTimePredicted() ) then return end
-
 	-- Release the old ghost entity
 	self:ReleaseGhostEntity()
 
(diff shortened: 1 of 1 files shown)


Kefta:
Please do not add hooks dynamically. hook.Add is made as the expensive function in hook library updates. I don't think this fix is appropriate anyway.

thegrb93:
Well it's the best fix currently. It's only called each time the ghost is spawned so don't be a performance nerd.

Kefta:
It doesn't take a nerd to see this is a pretty hacky fix with potentially awful performance as the icing on top. The additional Think call after Holstering should just be removed from the engine, but I have submitted a better Lua fix in https://github.com/Facepunch/garrysmod/pull/1674.

thegrb93:
Yea it's a hacky fix, but the performance complaint is idiotic. 10 nanoseconds. Bout fucking time there was a working fix.

Kefta:
Performance is more than just the runtime of one addition like this, it's cumulative from a mountain of inefficient solutions. Anyone who runs a 128 slot server will tell you that once all the major bottlenecks have been resolved, Lua's footprint is still significant, and micro-optimisations become increasingly important. Not that 128 player servers would use much stock GMod code anyway in this environment.

More than that, GMod should just be moving away from ``hook.Add``ing dynamically so that any hook lib changes eek out the most performance possible.

GitSparTV:
What if you make send a network message from the server to release ghost props?
Toolgun is a SWEP, so you can call `CallOnClient`

Kefta:
That's unnecessary net traffic for something that's already fixed semi-properly in https://github.com/Facepunch/garrysmod/pull/1674. The best fix would be removing the additional Think call after Holstering.

thegrb93:
Whatever man. Sounds like you're stuck with a really bad hook library if adding/removing is more expensive than the function being called.

GitSparTV:
Well, because you're creating it in the runtime. I see nothing that prevents from localizing the function

thegrb93:
No that isn't why... and the `self` prevents your suggestion

GitSparTV:
Do you know you can use an object in the hook name and it will be passed as the first argument?

thegrb93:
I guess that works, but that's still like half a nanosecond savings. People act like lambdas recompile a whole new function, but that's not what it does. That's besides the point he was making too.

Kefta:
> Sounds like you're stuck with a really bad hook library if adding/removing is more expensive than the function being called.
That's not what I said at all, I'm saying hook.Add is made to be the expensive function with priority sorting in hook lib updates instead of putting the weight on hook.Call, making dynamic hook addition less preferred/efficient over static hooks. See https://github.com/Facepunch/garrysmod/pull/1485, https://github.com/Facepunch/garrysmod/pull/1642, or even your own hook lib PR.

ExtReMLapin:
Is that the bugs that exists since 2008 ?

Verideth:
Hooking a function dynamically with a static function definition..
nice.. /s

There is absolutely no need for a hook in the first place.

thegrb93:
Criticize it all you like, but it works. I see you're continuing your reputation of adding nothing to the conversation.

Kefta:
Yeah stop continuing to bump issues for no reason, it's understandable why he approached the issue with a hook.
@ExtReMLapin It's always been present to my knowledge.

thegrb93:
@Kefta any progress on this one?

PR #1674, opened 2020-05-16
Fixed lingering toolgun ghost when using Player:SelectWeapon in multiplayer

Kefta:
See https://github.com/Facepunch/garrysmod/pull/1673. Consider fixing the extra clientside Think call after Holster in the engine instead. Also cached the constant trace mask in Primary/SecondaryAttack since I was already editing the file.


mcNuggets1:
This should be fixed in the engine and I do not know, why it hasnt yet.
There were attempts by the Devs itself to fix it for the tool gun ghost, so why not fix the bug in the engine?

robotboy655 (approved):

PR #1676, opened 2020-05-18
Add duplicator.Disallow

Lexicality:
Mainly this is because I want to be able to duplicate doors in code without letting players do it, but I'm sure there are other uses around blocking entities introduced by workshop addons.

Usage:
```lua
duplicator.Allow( "prop_door_rotating" )
local tab = duplicator.CopyEnts( Ents )
duplicator.Disallow( "prop_door_rotating" )
```


eLeCtrOssSnake:
yes

PR #1672, opened 2020-05-13
TTT: French translation!

Niilyx:
In addition to that, the score panel will be wider **only if the French language is selected**.

PR #1677, opened 2020-05-20
[TTT] Update CLSCORE

Kefta:
- The CLSCORE panel now adjusts its width to the win text, fixing clipping with some languages (https://github.com/Facepunch/garrysmod/pull/1672). This required moving the win calculations to ``CLSCORE:ShowPanel`` from ``CLSCORE:BuildHilitePanel``

Before:

After:

- ``CLSCORE:BuildHilitePanel(Panel base)`` -> ``CLSCORE:BuildHilitePanel(Panel base, WinTitle /*See definition at the bottom*/ title, Number starttime, Number endtime)``
- Changed ``Error`` calls to ``error`` as ``Error`` is no longer halting (https://github.com/Facepunch/garrysmod-issues/issues/2113)
- ``pairs`` loops of ``CLSCORE.Events`` are now numeric-for loops to make sure no stray string keys are considered
- Changed a usage of ``type`` to ``TypeID`` - using ``type`` is not safe for checking userdata types since the ``MetaName`` key of its metatable affects the output
- ``ValidAward`` now checks the types of its members instead of verifying they are simply non-nil
- ``CLSCORE:ShowPanel`` now deletes the last panel instead of creating a duplicate
- ``CLSCORE:ClearPanel`` now properly checks panel validity before trying to remove it
- Merged two loops over the event table in ``CLSCORE:Init``. Events with no traitor or detective tables no longer error from there
- Some hardcoded tables and colors are now members of ``CLSCORE`` so addons can modify them without overriding the whole system. These new members include:
-- Array<values = String> ``CLSCORE.ScorePanelNames``
-- Color ``CLSCORE.ScorePanelColor``
-- Table<keys = Number, values = WinTitle> ``CLSCORE.WinTypes``. This must have a string key ``Default`` pointing to a default WinTitle with all keys provided. WinTitle is defined as a Table<keys = String> {[String] ``Text`` /* Will be translated */, [Color] ``BoxColor``, [Color] ``TextColor``, [Color] ``BackgroundColor``}. All keys are optional and will fall back to the ``Default`` WinTitle table if not specified

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
@@ -64,16 +64,16 @@ function CLSCORE.DeclareEventDisplay(event_id, event_fns)
    -- basic input vetting, can't check returned value types because the
    -- functions may be impure
    if not tonumber(event_id) then
-      error("Event ??? display: invalid event id")
+      error("Event ??? display: invalid event id", 2)
    end
    if (not event_fns) or not istable(event_fns) then
-      error(string.format("Event %d display: no display functions found.", event_id))
+      error(string.format("Event %d display: no display functions found.", event_id), 2)
    end
    if not event_fns.text then
-      error(string.format("Event %d display: no text display function found.", event_id))
+      error(string.format("Event %d display: no text display function found.", event_id), 2)
    end
    if not event_fns.icon then
-      error(string.format("Event %d display: no icon and tooltip display function found.", event_id))
+      error(string.format("Event %d display: no icon and tooltip display function found.", event_id), 2)
    end
 
    CLSCORE.EventDisplay[event_id] = event_fns
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
@@ -20,10 +20,13 @@ CLSCORE.EventDisplay = {}
 
 local skull_icon = Material("HUD/killicons/default")
 
-surface.CreateFont("WinHuge", {font = "Trebuchet24",
-                               size = 72,
-                               weight = 1000,
-                               shadow = true})
+surface.CreateFont("WinHuge", {
+   font = "Trebuchet24",
+   size = 72,
+   weight = 1000,
+   shadow = true,
+   extended = true
+})
 
 -- so much text here I'm using shorter names than usual
 local T = LANG.GetTranslation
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -350,11 +371,40 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+   if title == nil then
+      title = self.WinTypes.Default
+   end
+
+   -- size the panel based on the win text w/ 88px horizontal padding and 44px veritcal padding
+   surface.SetFont("WinHuge")
+   local w, h = surface.GetTextSize(title.Text or self.WinTypes.Default.Text)
+
+   -- w + padding (100) + DPropertySheet padding (8) + winlbl padding (30) + offset margin (margin * 2) + size margin (margin)
+   w, h = math.max(700, w + 138 + margin * 3), 500
```
Kefta, line 405: This is the text clipping fix.

Niilyx:
Excellent work ! Just what I needed

svdm:
Cool, I'll go over this (and the other scoreboard PR) as soon as I get a chance, but that may not be for another week or two due to circumstances.

Kefta:
Stay safe @svdm

garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -447,7 +499,7 @@ function CLSCORE:SaveLog()
 
    log = log .. string.format("%s | %-25s | %s\n", " TIME", "TYPE", "WHAT HAPPENED") .. string.rep("-", 50) .."\n"
 
-   for _, e in pairs(self.Events) do
+   for i = 1, #events do
```
saibotk, line 502: You're missing an `local e = events[i]` in the next line.

garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -474,33 +525,48 @@ function CLSCORE:Reset()
...
-      if e.id == EVENT_GAME and e.state == ROUND_ACTIVE then
-         starttime = e.t
+   -- Get start time, traitors, detectives, scores, and nicks
+   local starttime = 0
+   local traitors, detectives
+   local scores, nicks = {}, {}
+   
+   local game, selected, spawn = false, false, false
+   for i = 1, #events do
+      e = events[i]
```
saibotk, line 535: Is the missing `local` intentional?

Kefta:
Thanks @saibotk, fixed those two.

garrysmod/gamemodes/terrortown/gamemode/cl_scoring.lua
```diff
@@ -350,11 +371,40 @@ function CLSCORE:BuildHilitePanel(dpanel)
...
+      end
+   end
+
+   if title == nil then
+      title = self.WinTypes.Default
+   end
+
+   -- size the panel based on the win text w/ 88px horizontal padding and 44px veritcal padding
+   surface.SetFont("WinHuge")
+   local w, h = surface.GetTextSize(title.Text or self.WinTypes.Default.Text)
```
saibotk, line 402: Also, wouldn't this need to be calculated on the translated string instead of the translation identification string? E.g. `T(title.Text or self.WinTypes.Default.Text)` instead?

Kefta: I already fixed this in another diff but it appears I copied over an older version. Fixed now.

PR #1678, opened 2020-05-20
[TTT] Optimised SCORE:StreamToClients implementation

Kefta:
- ``SCORE:StreamToClients`` now compresses its contents before sending it to the client
- Optimised sending the event payload in multiple parts, specifically by separating the part-splitting net message from the final arrival message. This introduces a new net message ``TTT_ReportStream_Part``.
- Event table sorting now happens after it's read on the client instead of before it's written on the server
- Raised the part separation limit to 65529 from 65500

Some additional small changes to the surrounding file:
- ``GetDamageType`` is no longer converted to a string in kill events as ``TableToJSON`` no longer has an issue encoding large integers
- ``CurTime`` is no longer rounded in events, causing events happening at near the same time to be ordered properly
- Changed ``pairs`` loop of ``SCORE.Events`` to numeric-for

New net message specs:
- ``TTT_ReportStream``: UInt-16 length, BinaryString[length] data
- ``TTT_ReportStream_Part``: BinaryString[65529] data


svdm:
Maybe move that MaxStreamLength value to scoring_shd.lua instead of duplicating it. Other than that this looks good, thanks.

Kefta:
Should it be a member of SCORE?

svdm:
SCORE's not shared (there's CLSCORE on the client, don't ask me why) and it seems excessive to change that. I'd treat it like the other pseudo-constants in TTT and make it a capitalised global (see shared.lua and some in scoring_shd.lua).

Kefta:
I'm not a fan of that since it leaks into the global space. Maybe define it in both SCORE and CLSCORE in one line based on a SERVER check ternary check? Ex. ``SERVER and SCORE or CLSCORE.EventMaxPartLength = ...``. I'll conform to what you want but adding additional globals could conflict with addons unintentionally.

svdm:
Defining it on both SCORE and CLSCORE in one place works for me too (doesn't have to be a oneliner).

Kefta:
I made it a one-liner for simplicity's sake, but I did have to include scoring_shd from the actual realm-specific scoring files so that the tables would be created in time.

PR #1680, opened 2020-05-24
net.Cancel()

SupinePandora43:
cancels net message.
##### actually sends message to noone

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -2,6 +2,10 @@ TYPE_COLOR = 255
 
 net.Receivers = {}
 
+function net.Cancel()
+	net.Send(RecipientFilter())
+end
+
 --
 -- Set up a function to receive network messages
 --


GitSparTV:
This should be done on C side instead, the code I made is the current walkaround

Kefta:
Why be done at all? This seems really niche and that idiom could just be done outright.

GitSparTV:
See #lua

Earu:
Its useful when you work with net messages on a more global scale than just your own addon, I was dealing earlier with some script that was building a net message but ended up never sending it, I didnt have control over the code of that script. Thats the kind of case where it could be useful.

robotboy655:
@Earu That is the sort of situation the author of that code/addon should fix. Adding stupid workarounds like this is not the solution or a good way to handle issues in general.

Earu:
And while I really understand what you mean, I cannot wait indefinitely for an addon to be fixed or start forking/modifying every project that has the same issues. I do agree that yes the author should be fixing it, and this is generally bad code. I see this as a potential workaround for bad code.

Kefta:
Just use the ``net.Send({})`` idiom then. net.IsActive() would be much more useful anyway.

PR #1682, opened 2020-05-24
Fix cvars.AddChangeCallback adding duplicate callbacks

FeldrinH:
Fix cvars.AddChangeCallback adding a new callback instead of replacing the existing one, when called multiple times with the same identifier.


GitSparTV:
Duplicate https://github.com/Facepunch/garrysmod/pull/1679

robotboy655:
Duplicate of https://github.com/Facepunch/garrysmod/pull/1679

PR #1683, opened 2020-05-26
Add table.Shuffle

WilliamVenner:
Performs an inline Fisher-Yates shuffle on the table in O(n) time

https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle#The_modern_algorithm

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -172,6 +172,17 @@ function table.Random( t )
 	end
 end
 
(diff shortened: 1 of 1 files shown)


GitSparTV:
Similar but O(n/2) with ignoring `# < 2`
```lua
    for i = #list, 2, -1 do
        local j = math.random(i)
        list[i], list[j] = list[j], list[i]
    end
```

WilliamVenner:
That's still O(n), and to be honest I think it would be preferential to stick to standard.

PR #1684, opened 2020-05-27
New achievements sending method

GitSparTV:
This PR changes achievements sending method from SendLua to networking 1-2 numbers.
The problem was recently mentioned in https://github.com/Facepunch/garrysmod/pull/1375

6 achievements are sent by server from Lua. Server calls `achievements.Call` with player, achievement ID (note it's not internal achievement ID, IDs are written in `idToAchievement` table lookup.) and additional number for Remover.

`achievements.Call` exists clientside as well but does nothing.

Net cost: 3 bits per call, except Remover achievement: 19 bits per call.

Remover sends additional number for massive removing, for the count it sends 16 bits becuase it's a maximum EntIndex.

### DIFF
--- a/garrysmod/lua/includes/extensions/achievements.lua
+++ b/garrysmod/lua/includes/extensions/achievements.lua
@@ -1,13 +1,13 @@
 if SERVER then
 	util.AddNetworkString("achievements")
+	achievements = { }
 end
 
 function achievements.Call(ply, aID, removerCount)
-	if CLIENT then return end
-	net.Start("achievements")
+	if (CLIENT) then return end
 	net.WriteUInt(aID, 3)
 
-	if removerCount then
+	if (aID == 2) then
 		net.WriteUInt(removerCount, 16)
 	end
 
@@ -17,14 +17,20 @@ end
 if CLIENT then
 	local idToAchievement = {
 		[0] = achievements.BalloonPopped,
-achievements.EatBall, achievements.Remover, achievements.SpawnedNPC, achievements.SpawnedProp, achievements.SpawnedRagdoll
+		achievements.EatBall,
+		achievements.Remover,
+		achievements.SpawnedNPC,
+		achievements.SpawnedProp,
+		achievements.SpawnedRagdoll
 	}
 
 	net.Receive("achievements", function()
 		local aID = net.ReadUInt(3)
-		if aID == 2 then
+
+		if (aID == 2) then
 			local func = idToAchievement[aID]
-			for k=1, net.ReadUInt(16) do
+
+			for k = 1, net.ReadUInt(16) do
 				func()
 			end
 		else
--- a/garrysmod/lua/includes/extensions/achievements.lua
+++ b/garrysmod/lua/includes/extensions/achievements.lua
@@ -30,7 +30,7 @@ function achievements.Register(name,func)
 	end
 
 	achievementsN = achievementsN + 1
-	bitsize = math.ceil(math.log(achievementsN) / math.log(2)) + 1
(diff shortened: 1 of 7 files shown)


robotboy655:
Have you actually tested this? I mean does this actually progress the achievements?

Kefta:
He hasn't as achievements.Call doesn't net.Start its net message. This should be marked as a draft.

GitSparTV:
For now Remover achievement will always send 16 bits even for one deleting.
This can be fixed but the clientside code should be changed:

```diff
- achievements.Register("Remover",function(count) for i = 1, count do achievements.Remover() end end)
+ achievements.Register("Remover", function(count)
+	if (count > 0) then
+		for i = 1, count do
+			achievements.Remover()
+		end
+	else
+		achievements.Remover()
+	end
+end)
```

Tell me if I should commit this

robotboy655:
I also feel like this is trying to solve a non existent problem.

Kefta:
Using SendLua all over the codebase is an existent problem. I have a better approach though.

garrysmod/lua/includes/extensions/achievements.lua
```diff
@@ -0,0 +1,52 @@
...
+	if CLIENT and not isfunction(func) then error("Second argument must be a function") end
+
+	if (SERVER) then
+		achievementsLookup[name] = achievementsN
+	else
+		achievementsLookup[achievementsN] = func
+	end
+
+	achievementsN = achievementsN + 1
+	bitsize = math.ceil(math.log(achievementsN) / math.log(2)) + 1
```
CornerPin, line 33: - After `achievementsN` becomes 3, `bitsize` is larger than it should be by 1;
- `math.log(x) / math.log(base)` can be simplified to `math.log(x, base)`.

```suggestion
	bitsize = math.ceil(math.log(achievementsN + 1, 2))
```

GitSparTV:
Still relevant? @Kefta

garrysmod/lua/includes/extensions/achievements.lua
```diff
@@ -30,7 +30,7 @@ function achievements.Register(name,func)
 	end
 
 	achievementsN = achievementsN + 1
-	bitsize = math.ceil(math.log(achievementsN) / math.log(2)) + 1
+	bitsize = math.ceil(math.log(achievementsN + 1, 2))
```
thegrb93, line 33: Should be math.Round I'd imagine.

Kefta: Definitely not. It should be ``math.floor(math.log(achievementsN, 2)) + 1``, but I am working on an internal solution.

Kefta:
A better internal solution will be made, but I'll leave this open so I don't forget.

GitSparTV:
I can convert it to garrysmod-request

garrysmod/lua/includes/extensions/achievements.lua
thegrb93: Oh ye lol. 4am brain

CornerPin: `math.ceil(math.log(achievementsN + 1, 2))` gives the same result as `math.floor(math.log(achievementsN, 2)) + 1` except when `achievementsN` is 0 because `log 0` is undefined.

Kefta:
I'd rather not have duplicates.

GitSparTV:
I just want to delete the fork and recreate it D:

GitSparTV:
> A better internal solution will be made, but I'll leave this open so I don't forget.

You? Oops

PR #1686, opened 2020-06-16
Add draw.MultiColorText()

EmperorSuper:


GitSparTV:
`pairs`?
`{...}` appears twice???

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,58 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+	for k, v in pairs{ ... } do
+		if IsColor( v ) then
+			curColor = v
+			continue
+		end
+		local text = tostring( v )
+		surface.SetTextColor( curColor )
+		surface.SetTextPos( curX, y )
+		surface.DrawText( text )
+		curX = curX + surface.GetTextSize( text )
```
GitSparTV, line 164: TextPos is already shited by `surface.DrawText`

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,58 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+	xAlign	= xAlign	or TEXT_ALIGN_LEFT
+	yAlign	= yAlign	or TEXT_ALIGN_TOP
+
+	surface.SetFont( font )
+
+	local curX = x
+	local curColor = color_white
+	local fullText = ""
+
+	for k, v in pairs{ ... } do
```
GitSparTV, line 135: `pairs`?

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,58 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+		curX = x - w
+	end
+
+	if ( yAlign == TEXT_ALIGN_CENTER ) then
+		y = y - h/2
+	elseif ( yAlign == TEXT_ALIGN_BOTTOM ) then
+		y = y - h
+	end
+
+	for k, v in pairs{ ... } do
```
GitSparTV, line 155: Why `pairs` and second `{...}`. Localize it

EmperorSuper:
What? What do you mean?

Kefta:
You pack the table ({...}) twice instead of caching it, and you use pairs instead of ipairs.

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,58 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+		if IsColor( v ) then
+			curColor = v
+			continue
```
GitSparTV, line 158: `continue` is not recommended, but I don't know what Rubat thinks

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,58 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+		end
+		local text = tostring( v )
+		surface.SetTextColor( curColor )
```
Kefta, line 161: Instead of calling surface.SetTextColor every time, you can set it once before the loop starts than only when the color is changed. There's no need to have a curColor variable at all.

garrysmod/lua/includes/modules/draw.lua
Kefta: It's just not necessary here

GitSparTV:
There is also a markup library that does the same, maybe not with best perfomance, but still

EmperorSuper:
@GitSparTV @Kefta Thanks for helping me with optimizing it, I'm not nearly as experienced at that as I should be.

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,63 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+	end
+
+	surface.SetTextColor( color_white )
+	surface.SetTextPos( x, y )
+
+	for k, v in ipairs, table, origin do
+		if IsColor( v ) then
+			surface.SetTextColor( v )
+		else
+			local text = tostring( v )
```
Kefta, line 167: Don't localise here, just call tostring inline in the surface.DrawText call.

EmperorSuper: Yeah I realized that too, but I figure I should wait for Rubato to review the code and make any change suggestions instead of adding small commits.

GitSparTV: Why, we already gave you suggestions

EmperorSuper:
Why does it still say changes requested? I already resolved them....

viral32111:
> Why does it still say changes requested?

It's probably GitHub just acting up :P 
It can still be merged by anyone with write access so no need to worry.

Kefta:
Oh I got write perms to the repo and I didn't "approve" the changes, so it was still flagged as needing changes.

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,62 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+
+		w, h = surface.GetTextSize( fullText )
+
+		if ( xAlign == TEXT_ALIGN_CENTER ) then
+			x = x - w/2
+		elseif ( xAlign == TEXT_ALIGN_RIGHT ) then
+			x = x - w
+		end
+	else
+		h = draw.GetFontHeight( font )
```
GitSparTV, line 151: Would be great if it won't use other draw functions.
surface.GetTextSize works as well and use in FontHeight

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,62 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+
+	local ipairs, table, origin = ipairs({ ... })
+	local w, h = 0, 0
+
+	if ( xAlign != TEXT_ALIGN_LEFT ) then
+		local fullText = ""
+		
+		for k, v in ipairs, table, origin do
+			if not IsColor( v ) then
+				fullText = fullText .. tostring( v )
```
GitSparTV, line 139: This would be slow, using a table will be better, and even better if the table is localized outside the function.
But gmod is not that modern to use such things, so idk

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -113,6 +113,62 @@ function SimpleTextOutlined(text, font, x, y, colour, xalign, yalign, outlinewid
...
+	end
+
+	surface.SetTextColor( color_white )
+	surface.SetTextPos( x, y )
+
+	for k, v in ipairs, table, origin do
+		if IsColor( v ) then
+			surface.SetTextColor( v )
+		else
+			surface.DrawText( tostring( v ) )
```
GitSparTV, line 167: Needs testing, doesn't it already tostring'ing them?

garrysmod/lua/includes/modules/draw.lua
EmperorSuper: Using a table for what? This is used to find the length of the full text...
Do you mean for caching the lengths? Yeah it would be better but would that be worth it?

garrysmod/lua/includes/modules/draw.lua
EmperorSuper: Just tested, it doesn't.

garrysmod/lua/includes/modules/draw.lua
EmperorSuper: I could use `select(2, surface.GetTextSize("W"))` which is what the draw function uses, but `draw.GetFontHeight()` caches the results so it's better.

GitSparTV: Fine, don't use select.
I was suggesting `local _, h = surface.GetTextSize(text)`

garrysmod/lua/includes/modules/draw.lua
GitSparTV: table.concat that is

GitSparTV: separate concats are super slow

EmperorSuper: table.concat doesn't `tostring()` the values, are you sure it'd be better to use it instead?

EmperorSuper: Wait actually it'd be impossible to use it instead because we're expecting table values for the colors to be in that table. Or am I missing something here?

garrysmod/lua/includes/modules/draw.lua
EmperorSuper: Performance-wise, Wouldn't `draw.GetFontHeight()` still be better because it caches the values?

garrysmod/lua/includes/modules/draw.lua
Kefta: He means create a separate table and add each tostringed entry, than table.concat it.

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -132,15 +134,15 @@ function MultiColorText( font, x, y, xAlign, yAlign, ... )
 	local w, h = 0, 0
 
 	if ( xAlign != TEXT_ALIGN_LEFT ) then
-		local fullText = ""
-		
+		table.Empty( MultiColorText_StringTable )
```
GitSparTV, line 137: table.Empty uses pairs

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -132,15 +134,15 @@ function MultiColorText( font, x, y, xAlign, yAlign, ... )
...
 
 	if ( xAlign != TEXT_ALIGN_LEFT ) then
-		local fullText = ""
-		
+		table.Empty( MultiColorText_StringTable )
+
 		for k, v in ipairs, table, origin do
 			if not IsColor( v ) then
-				fullText = fullText .. tostring( v )
+				table.insert( MultiColorText_StringTable, tostring( v ) )
```
GitSparTV, line 141: the best way to do this if you started out is making a local `i` and using it instead of `tbl[#tbl + 1] = `

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -132,15 +134,15 @@ function MultiColorText( font, x, y, xAlign, yAlign, ... )
...
 			end
 		end
 
-		w, h = surface.GetTextSize( fullText )
+		w, h = surface.GetTextSize( table.concat( MultiColorText_StringTable, "" ) )
```
GitSparTV, line 145: Since you decided to upvalue the table this wont work anymore.
Imagine this, you had lots of strings before, but now less. This will merge previous strings with the new one, use `i` counter

garrysmod/lua/includes/modules/draw.lua
Kefta: Just create a new table every time.

garrysmod/lua/includes/modules/draw.lua
GitSparTV: Ignore, see the first conversation

garrysmod/lua/includes/modules/draw.lua
GitSparTV: Creating a new table would be faster. Clearing the table is stuitable for large table

garrysmod/lua/includes/modules/draw.lua
EmperorSuper: Ok I can't tell if I'm stupid or not because I have no idea what you mean.

GitSparTV: Make a local and count it in the loop separately

robotboy655:
We already have a more flexible solution to this problem:
https://wiki.facepunch.com/gmod/markup.Parse

PR #1687, opened 2020-06-16
Adds draw.Circle

connorh315:
Adds the function draw.Circle to draw a filled circle

### DIFF
--- a/garrysmod/lua/includes/modules/draw.lua
+++ b/garrysmod/lua/includes/modules/draw.lua
@@ -299,6 +299,25 @@ function TexturedQuad( tab )
 
 end
 
+--[[---------------------------------------------------------
+	Name: Circle( x, y, radius, segments )
+	Desc: pawrapper
+-----------------------------------------------------------]]
+function Circle( x, y, radius, seg )
+	local cir = {}
+
+	table.insert( cir, { x = x, y = y, u = 0.5, v = 0.5 } )
+	for i = 0, seg do
+		local a = math.rad( ( i / seg ) * -360 )
+		table.insert( cir, { x = x + math.sin( a ) * radius, y = y + math.cos( a ) * radius, u = math.sin( a ) / 2 + 0.5, v = math.cos( a ) / 2 + 0.5 } )
+	end
+
+	local a = 0
+	table.insert( cir, { x = x + math.sin( a ) * radius, y = y + math.cos( a ) * radius, u = math.sin( a ) / 2 + 0.5, v = math.cos( a ) / 2 + 0.5 } )
+
+	surface.DrawPoly( cir )
+end
+
 function NoTexture()
 	surface.SetTexture( tex_white )
 end


garrysmod/lua/includes/modules/draw.lua
```diff
@@ -299,6 +299,25 @@ function TexturedQuad( tab )
...
+	Name: Circle( x, y, radius, segments )
+	Desc: pawrapper
+-----------------------------------------------------------]]
+function Circle( x, y, radius, seg )
+	local cir = {}
+
+	table.insert( cir, { x = x, y = y, u = 0.5, v = 0.5 } )
+	for i = 0, seg do
+		local a = math.rad( ( i / seg ) * -360 )
+		table.insert( cir, { x = x + math.sin( a ) * radius, y = y + math.cos( a ) * radius, u = math.sin( a ) / 2 + 0.5, v = math.cos( a ) / 2 + 0.5 } )
```
Kefta, line 312: Do not use table.insert for loop insertion. Insert as ``cir[i + 2] = ...``.

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -299,6 +299,25 @@ function TexturedQuad( tab )
...
 end
 
+--[[---------------------------------------------------------
+	Name: Circle( x, y, radius, segments )
+	Desc: pawrapper
+-----------------------------------------------------------]]
+function Circle( x, y, radius, seg )
+	local cir = {}
+
+	table.insert( cir, { x = x, y = y, u = 0.5, v = 0.5 } )
```
Kefta, line 309: Instead of calling table.insert here, make it the first key of the table during creation.

garrysmod/lua/includes/modules/draw.lua
```diff
@@ -299,6 +299,25 @@ function TexturedQuad( tab )
...
+	for i = 0, seg do
+		local a = math.rad( ( i / seg ) * -360 )
+		table.insert( cir, { x = x + math.sin( a ) * radius, y = y + math.cos( a ) * radius, u = math.sin( a ) / 2 + 0.5, v = math.cos( a ) / 2 + 0.5 } )
+	end
+
+	local a = 0
+	table.insert( cir, { x = x + math.sin( a ) * radius, y = y + math.cos( a ) * radius, u = math.sin( a ) / 2 + 0.5, v = math.cos( a ) / 2 + 0.5 } )
```
Kefta, line 316: Same here. Insert as ``seg + 3``.

GGG-KILLER:
Generating the circle every single time is extremely inefficient. There should be a way to generate the points then just reuse that value.

Having the segment count as an input is also totally unnecessary. That input should be the max distance between the points with a sane default.

SneakySqwuid:
my dude let me introduce you to [Circles!](https://github.com/SneakySqwuid/Circles/)

CornerPin:
This is very unoptimized and should be named `draw.TexturedCircle`.

Kefta:
Or even not be implemented at all in favour of circle objects.

PR #1688, opened 2020-06-16
Revert "Create ping categories to reduce impact of miniscule ping dif…

ThatDashiell:
This reverts the disastrous change in  #1671 . The change in #1671 makes it so any server with a ping less than 60 gets the result of a 40 ping in the ranking.

First and foremost, the assertion that anycast based routing is "ping spoofing" is simply obtuse. Latency, through anycast, is genuinely decreased as there are reduced hops and therefore minimized RTT. Also, why are we punishing a valid DDoS mitigation method that any host could implement?

Second, #1671 indiscriminately punishes servers with low pings. The change in no way addresses the concerns in #4501 (https://github.com/Facepunch/garrysmod-issues/issues/4501) and only serves to worsen it. As seen in the screenshot of the aforementioned issue, Russia based servers had significantly lower pings then other servers outside of the region, even with superior routing. With the changes of #1671, there is no bias towards those low ping Russian servers, and player count is the only thing that matters. Thus, they will be buried by larger US/EU based servers. A proper solution for concerns such as this would be the ability to sort by region.

Finally, this harms smaller servers. Smaller servers that used hosts that utilized superior routing were able to compete in ranking with larger servers. However, by making the server list ping agnostic, and therefore, the top just based in player count: smaller servers will be hurt the most.

For several reasons, #1671 is a rushed change with more unintended consequences then benefit. It's clear that people want something to blame for the success of larger servers and their personal failings, so latency has become a scapegoat. Perhaps the blame should be shifted to hosts with horrible routing and high pings who can take simple steps to give their customers a better experience. The disaster of disregarding pings is clear in mods such as FiveM, and it's truly a shame to kill the quality of the server list in Garry's Mod. Therefore, #1671 must be reverted.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -265,8 +265,7 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 
 	data.hasmap = DoWeHaveMap( data.map );
 
-	data.recommended = 40;
-	if ( data.ping >= 60 ) data.recommended = data.ping;
+	data.recommended = data.ping;
 
 	if ( data.players == 0 ) data.recommended += 75; // Server is empty
 	if ( data.players >= data.maxplayers ) data.recommended += 100; // Server is full, can't join it


Kefta:
> It's clear that people want something to blame for the success of larger servers and their personal failings, so latency has become a scapegoat.

From someone who has never own a community, I'll be the first to say that it clearly hasn't. Anyone who has browsed the server list when redirect anycast servers were a thing 2 years ago, to today in the TTT category can tell you it's a complete sham of servers that spoof their ping. Anycast may be a legitimate method of mitigating DDoS attacks, but the benefits of appearing at the top of anyone's regions are clear and the primary selling point for many hosts.

Really, the fault is on the devs for not implementing a system to report the actual ping in the A2S request yet, meaning servers are allowed to Anycast with false ping as they have no alternative. This change was hasty and a product of the serverlist we are currently dealing with, but until an alternative is added, reducing the impact of ping up to a reasonable amount will help non-US regions greatly once the player count has redistributed.

The change was intended to be experimental and there is no need to revert it until we see how it works in the next update. This is a learning experience for everyone.

ThatDashiell:
> today's TTT category will tell you it's a complete sham of servers that spoof their ping. Anycast may be a legitimate method of mitigating DDoS attacks, but the benefits of appearing at the top of anyone's regions are clear and the primary selling point for many hosts.

I'll reiterate that anycast is a valid form of DDoS mitigation any host could implement. But I do agree, A2S would simply solve it instead of these rash changes.

> reducing the impact of ping up to a reasonable amount will help non-US regions greatly once the player count has redistributed.

The issue with that is the change does not reduce the impact, only magnify it. As I said, if you look at the pings from the screenshot, on the issue, Russian based servers had massively lower pings. The change removes the advantage of those lower pings and makes them compete almost purely on player count, thus harming them. Also, to assume player count will be "redistributed" is ridiculous considering that the change will push them farther down.

>The change was intended to be experimental and there is no need to revert it until we see how it works in the next update. This is a learning experience for everyone.

I believe it's completely valid to revert it to prevent damage to smaller communities who will be harmed by its implementation.

Kefta:
I'm sure the devs will have no problem reverting the change once we see the results so that it can be used as evidence in future changes to priority sorting. No better evidence than the actual game.

SnowboiTheGr8:
> Latency, through anycast, is genuinely decreased as there are reduced hops and therefore minimized RTT.

This is an obtuse assertion to make. Simply using Anycast doesn't necessarily reduce the amount of hops, and even if it did, reducing the amount of hops doesn't inherently mean the latency was lower. This totally comes down to how the provider manages their network, which is often restricted by financial overhead in some capacity.

> Also, why are we punishing a valid DDoS mitigation method that any host could implement?

Once again, this is an obtuse assertion to make. I'm not sure why you think this is a punishment, but from everything I could read on the other issue, this seems to merely level the playing field, and minimize the gain from something like A2S caching, and A2S caching at an anycast level.

If you want to run a gmod server, you **shouldn't** need to understand nuanced stuff about how A2S works, to be on a level playing field with the large server owners.

> As seen in the screenshot of the aforementioned issue, Russia based servers had significantly lower pings then other servers outside of the region, even with superior routing.

This is probably because GMC doesn't have a POP near Russia, however for locations like the Europe, this is a pretty big issue.

> With the changes of #1671, there is no bias towards those low ping Russian servers, and player count is the only thing that matters. Thus, they will be buried by larger US/EU based servers. A proper solution for concerns such as this would be the ability to sort by region.

Thanks for the original idea, sadly afaik Garry doesn't want any changes like this being made to the server browser - things like this (and even better ideas) have been suggested for years.

> Finally, this harms smaller servers. Smaller servers that used hosts that utilized superior routing were able to compete in ranking with larger servers. However, by making the server list ping agnostic, and therefore, the top just based in player count: smaller servers will be hurt the most.

That's a pretty misleading statement, but **first** I think it's worth acknowledging the fact smaller servers **shouldn't** need to be doing L3 network shit in order to compete with larger servers.
**Secondly**, this doesn't make "the server list ping agnostic". You're not even "stretching the truth" there, that's just a lie. The server browser just normalizes" any ping below a certain threshold; of which is a reasonable response time you'd get from a standard srcds instance in a average dc, for the average person.
**Finally**, I don't see an issue with "larger servers" getting a higher ranking because they have more players - as far as I'm aware, that's how the system **should work** - _to an extent_ - there are limitations on how that logic works, still, and it's more-so about **how full** your server is, not simply **how many players** your server has.

> For several reasons, #1671 is a rushed change with more unintended consequences then benefit. It's clear that people want something to blame for the success of larger servers and their personal failings, so latency has become a scapegoat. 

I think you're getting a bit emotional, and it's clouding your judgement. I don't think the people making these changes have any vendetta against the "larger servers", I think they have an issue with latency playing such a huge role in the server browser, which lends itself to manipulation. Just like manipulating any other data regarding your server query (player count, etc.) isn't typically allowed, people shouldn't have an advantage because of networking/"ddos protection"/"black magic" bullshit

SnowboiTheGr8:
> I'll reiterate that anycast is a valid form of DDoS mitigation, and genuinely reducing latency, any host could implement.

Also because no one seems to be doing this excluding me -- I'll reiterate, that anycasting the a2s queries isn't needed for DDoS mitigation, and I think the claim that it's needed for ddos mitigation is just that. Any host could implement DDoS mitigation without the need for anycast.

ThatDashiell:
I'll close this as it's experimental and I believe the negative results will show in the future--and I don't have a desire to argue this further as it will fall on deaf ears.

SnowboiTheGr8:
> as I'd be arguing against people with a financial gain by its implementation.

That's a very cheap way to discount all my points. But if we're going there, from what you've said it seems like you're only against this change is because you think you stand to lose from it's implementation:

> Smaller servers that used hosts that utilized superior routing were able to compete in ranking with larger servers.

ThatDashiell:
> That's a very cheap way to discount all my points. But if we're going there, from what you've said it seems like you're only against this change is because you think you stand to lose from it's implementation:


I edited that out right when you replied, I agree it was a cheap shot

Kefta:
Again, it's just an experiment. Not speaking from the company, but Facepunch seems to be totally disconnected from GMod. Rubat and Willox are the devs that are assigned to maintain the game and FP as a company seems to (want to) have little say in how that goes about. In that respect, this is an experiment for the few of us left who care about serverlist sorting affecting the masses.

ThatDashiell:
> Again, it's just an experiment. Not speaking from the company, but Facepunch seems to be totally disconnected from GMod. Rubat and Willox are the devs that are assigned to maintain the game and FP as a company seems to (or want to) have little say in how that goes about. In that respect, this is an experiment for the few of us left who care about serverlist sorting affecting the masses.

I respect that, we'll see how it goes, hopefully, I'm wrong and it helps.

viral32111:
This is probably going to be a *very* unpopular opinion and I'm sorry in advance, but why is there server ranking in the first place? It just introduces unnessasary competition and has kept this whole anycast ping spoofing conversating been going on for years. Big communities with money keep getting bigger, and small communities without money keep dying out.

I don't know exactly know the steam master server list works, but I'm pretty sure it just returns server IP addresses at random, so if that's the case then why can't the server list just be populated in the order the IPs are returned, but along with some basic sorting in this order:
* Servers with players and GSLTs appear at the first.
* Servers with players and without GSLTs appear second.
* Servers without players and with GSLTs appear third.
* Servers without players and without GSLTs appear last.

Then it would be nobody's competition and maybe all servers would have a chance for once. Hopefully, finally ending the advantage that servers with lower pings get. It might also help to kill out A2S spoofing in general.

If players would rather see servers with the lowest ping at the top, or the highest player count at the top, then simply let them sort the colums like that?

(Again, I'm probably contributing nothing to the conversation and I haven't been around here for long, but I'd just like to know *why* we have to have server ranking in the first place when players can easily sort the serverlist how they want.)

PR #1690, opened 2020-06-20
Added undo for weapons

viral32111:
This adds an undo entry for weapons spawned via the Creator tool or `gm_spawnswep`. [Before](https://youtu.be/j_XEN6Q0Zrs), [after](https://youtu.be/wASzpnQTDGg).

I'm not aware of any reliable method to fetch a language phrase on the server, so I just used the weapon's class name as a fallback for when the weapon's print name is the key to a language phrase.

Tested on the dev branch, version 2020.06.19.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/commands.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/commands.lua
@@ -855,17 +855,10 @@ function Spawn_Weapon( ply, wepname, tr )
 
 	gamemode.Call( "PlayerSpawnedSWEP", ply, entity )
 
-	local PrintName = swep.PrintName
-
-    	-- We haven't got a nice name :/
-	if ( PrintName:sub( 0, 1 ) == "#" ) then
-		PrintName = swep.ClassName
-	end
-
 	undo.Create( "SWEP" )
 		undo.SetPlayer( ply )
 		undo.AddEntity( entity )
-		undo.SetCustomUndoText( "Undone " .. PrintName )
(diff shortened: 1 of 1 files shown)


VladUndying:
Undo is mainly supposed to be for props, I don't understand why would you even undo weapons 🤔.

Kefta:
Because it was an entity that was created by the player with a sandbox tool. It makes sense.

viral32111:
> Undo is mainly supposed to be for props, I don't understand why would you even undo weapons thinking.

Considering that every single other thing spawned via the Creator tool can be undone, it doesn't make any sense why weapon entities spawned via the same thing can't be undone.

garrysmod/gamemodes/sandbox/gamemode/commands.lua
```diff
@@ -855,6 +855,19 @@ function Spawn_Weapon( ply, wepname, tr )
 
 	gamemode.Call( "PlayerSpawnedSWEP", ply, entity )
 
+	local PrintName = swep.PrintName
+
+    	-- We haven't got a nice name :/
```
Kefta, line 860: language.GetPhrase on the client

viral32111: Yeah...? But isn't `commands.lua` a serverside-only file?

Kefta: I had no idea the undo module just networked the string straight to the client, and now I wish I didn't.

PR #1691, opened 2020-06-20
"Legal wall hack" fix

Wani4ka:
Recently we received a bug report from one of our community members. You can enter ANY value in "Post Process / Stereoscopy" Text Entry and it didn't clamp. This way you could see through walls:

These commits fix clamping for DNumSlider and pp_stereoscopy_size ConVar.

### DIFF
--- a/garrysmod/lua/postprocess/stereoscopy.lua
+++ b/garrysmod/lua/postprocess/stereoscopy.lua
@@ -15,7 +15,7 @@ function RenderStereoscopy( ViewOrigin, ViewAngles )
 	local w = ScrW() / 2.2
 	local h = ScrH() / 2.2
 
-	local Right = ViewAngles:Right() * pp_stereoscopy_size:GetFloat()
+	local Right = ViewAngles:Right() * math.Clamp( pp_stereoscopy_size:GetFloat(), 0, 10 )
 
 	local view = {}
 
--- a/garrysmod/lua/vgui/dnumslider.lua
+++ b/garrysmod/lua/vgui/dnumslider.lua
@@ -11,15 +11,7 @@ function PANEL:Init()
 	self.TextArea:SetWide( 45 )
 	self.TextArea:SetNumeric( true )
 	self.TextArea:SetUpdateOnType( true )
-	self.TextArea.OnValueChange = function( textentry, val )
-		local cPos = self.TextArea:GetCaretPos()
-		local clamped = math.Clamp( tonumber( val ) || 0, self:GetMin(), self:GetMax() )
-		if clamped != val then
-			self.TextArea:SetText( clamped )
-			self.TextArea:SetCaretPos( cPos )
-		end
-		self:SetValue( clamped )
-	end
+	self.TextArea.OnValueChange = function(textentry, val) self:SetValue( val ) end
 	-- Causes automatic clamp to min/max, disabled for now. TODO: Enforce this with a setter/getter?
 	--self.TextArea.OnEnter = function( textarea, val ) textarea:SetText( self.Scratch:GetTextValue() ) end -- Update the text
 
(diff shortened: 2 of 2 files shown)


garrysmod/lua/postprocess/stereoscopy.lua
```diff
@@ -15,7 +15,7 @@ function RenderStereoscopy( ViewOrigin, ViewAngles )
 	local w = ScrW() / 2.2
 	local h = ScrH() / 2.2
 
-	local Right = ViewAngles:Right() * pp_stereoscopy_size:GetFloat()
+	local Right = ViewAngles:Right() * math.Clamp( pp_stereoscopy_size:GetFloat(), 0, 10 )
```
Kefta, line 18: Perhaps this should reference a new convar (with FCVAR_CHEAT of course) instead of hardcoding the min/max.

garrysmod/lua/vgui/dnumslider.lua
```diff
@@ -10,7 +10,16 @@ function PANEL:Init()
 	self.TextArea:SetPaintBackground( false )
 	self.TextArea:SetWide( 45 )
 	self.TextArea:SetNumeric( true )
-	self.TextArea.OnChange = function( textarea, val ) self:SetValue( self.TextArea:GetText() ) end
+	self.TextArea:SetUpdateOnType( true )
+	self.TextArea.OnValueChange = function( textentry, val )
```
Kefta, line 14: This whole callback is unnecessary as the numslider's value is already clamped in its SetValue. Instead, just clamp the textentry's value in that function alongside the numslider so it isn't performed duplicitously.

garrysmod/lua/postprocess/stereoscopy.lua
neico: Why not write min/max into the convar creation above?

`local pp_stereoscopy_size = CreateClientConVar( "pp_stereoscopy_size", "6", true, false, "", 0, 10 )`

Kefta: Yeah, pp_stereoscopy_min and pp_stereoscopy_max will suffice. It will have to scale the AddControl call below as well, though.

Wani4ka: Declaring min and max value in CreateClientConVar won't clamp the values if they have been set before introducing those ranges...

garrysmod/lua/vgui/dnumslider.lua
Wani4ka: So do you advice just not to set slider's value in that function? Seems to work.
```lua
self.TextArea.OnValueChange = function( textentry, val )
	local cPos = self.TextArea:GetCaretPos()
	local clamped = math.Clamp( tonumber( val ) || 0, self:GetMin(), self:GetMax() )
	if clamped != val then
		self.TextArea:SetText( clamped )
		self.TextArea:SetCaretPos( cPos )
	end
end
```

Kefta: It works but it does the clamping twice, once in OnValueChange and another time in SetValue. You can optimise this by using the NumSlider's clamping from https://github.com/Facepunch/garrysmod/blob/057e56ed4d56fa6b177a44d3028473fbedcdc033/garrysmod/lua/vgui/dnumslider.lua#L111-L121 , ex.
```Lua
function PANEL:SetValue( val )

	val = math.Clamp( tonumber( val ) || 0, self:GetMin(), self:GetMax() )

	if ( self:GetValue() == val ) then return end

	self.Scratch:SetValue( val ) -- This will also call ValueChanged

	val = tostring( val )
	self.TextArea:SetText( val )
	self.TextArea:SetCaretPos( #val )

	self:ValueChanged( self:GetValue() ) -- In most cases this will cause double execution of OnValueChanged

end
```

robotboy655:
Just put `hook.Remove( "RenderScene", "RenderStereoscopy")` somewhere? Convars for min/max sounds ridiculous for this.

robotboy655:
https://github.com/Facepunch/garrysmod/commit/85d5ea04dc1264e6ec101b45aeebc2e43d73811a

Kefta:
Forgot convars had min/maxs binded now tbh.

PR #1697, opened 2020-07-11
remove server is full rating penalty

meepen:
Since servers just abuse this and find ways to make it not matter, should just remove it.


meepen:
To add onto this, it only affects servers that aren't lying and it can also change pretty quickly

robotboy655:
What is this meat to achieve? How are people abusing this?

Kefta:
They falsify having at least one open slot even when they don't. See https://twitter.com/garrynewman/status/1282582528453378049

mcNuggets1:
It would be cool to enter the queue to join the server, once a slot is free.
Like the old browser has the ability to.

robotboy655 (approved):

PR #1700, opened 2020-07-25
Add DButton:SetMaterial & DImageButton:SetMaterial

aStonedPenguin:
Quality of life change.

### DIFF
--- a/garrysmod/lua/vgui/dbutton.lua
+++ b/garrysmod/lua/vgui/dbutton.lua
@@ -51,6 +51,27 @@ function PANEL:SetImage( img )
 end
 PANEL.SetIcon = PANEL.SetImage
 
+function PANEL:SetMaterial( mat )
+
+	if ( !mat ) then
+
+		if ( IsValid( self.m_Image ) ) then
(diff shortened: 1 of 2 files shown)


GitSparTV:
Add binding for DImageButton as well.
It doesn't have SetMaterial as well

GitSparTV:
Epic

robotboy655:
You can just do `PANEL.SetMaterial = PANEL.SetImage` for the DButton, but do we really need a 3rd function to do the same thing?

aStonedPenguin:
[DButton:SetImage](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/dbutton.lua#L32) calls [DImage:SetImage](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/dimage.lua#L91) which only accepts a string.

DButton:SetImage could be made to call DImage:SetMaterial but if SetMaterial and SetImage are different functions in DImage shouldn't they be different in all for cohesiveness?

robotboy655:
Missed that difference, my bad.

PR #1702, opened 2020-08-06
Added PANEL:PropSelect() to replace PANEL:AddControl()

dvdvideo1234:


dvdvideo1234:
Control presets convar information array can be generated via `local gtConvarList = TOOL:BuildConVarList()`

PR #1704, opened 2020-08-07
Fix table.RemoveByValue throwing an error when key is not a number

peter-r-g:
Title says it all, backwards compatibility is still maintained and when the key is not a number it just makes the value to the key nil.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -651,7 +651,12 @@ function table.RemoveByValue( tbl, val )
 	local key = table.KeyFromValue( tbl, val )
 	if ( !key ) then return false end
 
-	table.remove( tbl, key )
+	if ( isnumber( key ) ) then
+		table.remove( tbl, key )
+	else
+		tbl[ key ] = nil
+	end
+
 	return key
 
 end


ExtReMLapin:
table.RemoveByValue itself sound pretty dumb to be honest

peter-r-g:
Yeah, it has very rare use cases. Gotta maintain the precious backwards compatibility though I suppose

ExtReMLapin:
I'm curious to see it's a real good reason to use this kind of shit.

When you got the value of a table and you know the table has it's value you're supposed to already have the key somewhere.
It's exactly like table.HasValue being used instead of hashtables.

robotboy655:
```
> local t= {fuck = "da", [2] = "net"} table.RemoveByValue(t, "net") PrintTable( t )...
2	=	net
fuck	=	da
```
This function is pretty fucked regardless

robotboy655:
Also what is it supposed to do with cases like this: 
```
> local t= {"a", "b", "c", "b"} table.RemoveByValue(t, "b") PrintTable( t )...
1	=	a
2	=	c
3	=	b
```

Kefta:
Just deprecate this terrible function, no fixes needed.

dvdvideo1234:
@peter-r-g Why don't you use `pairs()` or directly setting ti to `nil` ?

PR #1705, opened 2020-08-11
Create draw.OutlinedBox

Down-s:
A far more convenient way to draw an outlined box


mcNuggets1:
Code style is different.

robotboy655:
```
 for i=0,thickness do
        surface.DrawOutlinedRect( x + i, y + i, w - i * 2, h - i * 2 )
    end
```

No! Very no!

PR #1706, opened 2020-08-11
Network Entity.Set/GetCreator

FeepsDev:
This will add a useful simple owning system to Garry's Mod. I know that CPPI already exists but it is not used everywhere...

This modification will avoid workshop & gmodstore addon from creating their own ownabiliy functions to support gamemode that hasn't CPPI.

The only downside I see is that addons which already used this function will unnecessarily network the owner.

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -49,6 +49,7 @@ function meta:GetVar( name, default )
 end
 
 function meta:SetCreator( ply )
+	self.m_PlayerCreator = ply -- Backward compatibility
 	self:SetNWEntity( "PlayerCreator", ply )
 end
 
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -48,16 +48,12 @@ function meta:GetVar( name, default )
 
 end
 
-if ( SERVER ) then
+function meta:SetCreator( ply )
+	self:SetNWEntity( "PlayerCreator", ply )
+end
 
-	function meta:SetCreator( ply )
-		self:SetNWEntity( "PlayerCreator", ply )
-	end
-
-	function meta:GetCreator()
-		return self:GetNWEntity( "PlayerCreator" )
-	end
-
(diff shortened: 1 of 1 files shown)


viral32111:
Imo, CPPI works fine and this is exactly what modding is for. The whole point is that the gamemode or some sort of addon is meant to implement the spec, then all other addons or the gamemode itself uses the spec. Almost every single prop protection addon and alike implement it these days. If a gamemode or addon doesn't implement/use CPPI than that's their fault. Making GetCreator networked wouldn't change the fact that said gamemode/addon isn't using some sort of ownership system.

Also SetCreator couldn't be networked because the client can't be trusted with stuff like that, GetCreator on the other hand though should be fine.

viral32111:
It should probably still set `.m_PlayerCreator` to maintain backwards compatibility with any addons that may be directly accessing that property of an entity.

viral32111:
Also perhaps SetCreator should still be in `if SERVER then` statement to prevent clients from arbitrarily taking ownership. I know SetNWEntity will only affect the local client when ran from the client realm, but it's probably better to not expose it to the client at all.

GitSparTV:
image

Kefta:
Yeah GMod doesn't need to step on addons with using NW stringtable entries in the Lua state. It also makes an official function fragile from Lua overrides. I'll keep this open so that I won't forget to do this.

FeepsDev:
@Kefta Maybe add a hook like OnEntityCreatorChanged ?

Kefta:
What uses would that have?

viral32111:
Perhaps in a custom gamemode if a player disconnected, all their spawned props/entities could have their "Creator" player set to another player or NULL, then this hook could be used to inform other addons or scripts that the creator has changed, or the creator is no longer on the server/has disconnected (so NULL).

I suppose there aren't any immedite good uses, but that shouldn't be a reason to not add it as in the future more creative developers might find a use for it in their addons/gamemode.

GitSparTV:
SetCreator is also detoured by FPP

robotboy655:
Essentially duplicate of https://github.com/Facepunch/garrysmod-requests/issues/1086

PR #1708, opened 2020-08-15
Update dalphabar.lua

MoskalykA:
Switch from Color(...) to color_white (Micro cache optimization)

### DIFF
--- a/garrysmod/lua/vgui/dalphabar.lua
+++ b/garrysmod/lua/vgui/dalphabar.lua
@@ -9,7 +9,7 @@ AccessorFunc( PANEL, "m_BarColor", "BarColor" )
 
 function PANEL:Init()
 
-	self:SetBarColor( Color( 255, 255, 255, 255 ) )
+	self:SetBarColor( color_white )
 	self:SetSize( 26, 26 )
 	self:SetValue( 1 )
 


GitSparTV:
Please always provide description for your PRs

FoksVHox:
This PR are point less, yes it'll save a very minor percentage of resources, but it'll be so minimal that you don't feel it or will see a difference.

nodgear:
I would approve this PR just for the sake of consistency across files.
It's a good practice recommended by the wiki itself, so that's how the code in the game should be, but i don't see the reason why he's doing it in separated requests. Gather all code with inconsistency and push all at once...

MoskalykA:
> I would approve this PR just for the sake of consistency across files.
> It's a good practice recommended by the wiki itself, so that's how the code in the game should be, but i don't see the reason why he's doing it in separated requests. Gather all code with inconsistency and push all at once...

Unfortunately, I don't know how to make a request and put all my suggestions into one.

viral32111:
Just commit everything to the same fork then make a PR for that fork. If you're doing this directly from the GitHub website, you can make a single fork of the repo, make all changes there, then come back to this repo and make the PR. I think by default GitHub automatically creates a "patch" PR if you edit a file in a repository you don't own.

Kefta:
Getting rid of constant Vector, Angle, Matrix, Color, etc. allocs in the GMod base code is the goal. This PR is fine.

PR #1709, opened 2020-08-15
Update dhtml.lua

MoskalykA:


GitSparTV:
Please provide the description

GitSparTV:
Note that 2 MsgC doesn't have white color, it's little bit different

nodgear:
I don't think these micro optimizations are going to make any difference to be honest.
Also, like Spar mentioned, the MsgC's are not using the white colors.

FoksVHox:
This is so minor...

CornerPin:
It should be `self.JS[#self.JS + 1] = j`

mcNuggets1:
This isn't even minor anymore.
This is something new.

viral32111:
I'd prefer if the `[HTML]` bits stayed that purpleish color, it makes it easier to read.

MoskalykA:
I mostly make this proposal for the equivalent of the table.insert because it's a bit faster (According to the Zaros benchmark)

FoksVHox:
> This isn't even minor anymore.
> This is something new.

How can it not be minor? The changes that this PR adds are so small and they wont be noticed in regards om the performance.

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -53,8 +53,8 @@ function PANEL:QueueJavascript( js )
 	end
 
 	self.JS = self.JS or {}
-
-	table.insert( self.JS, js )
+	self.JS[self.JS + 1] = js
```
viral32111, line 56:
```suggestion
	self.JS[#self.JS + 1] = js
```

Just making https://github.com/Facepunch/garrysmod/pull/1709#issuecomment-674363812 more obvious

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -76,8 +76,8 @@ function PANEL:ConsoleMessage( msg, file, line )
 			file = string.sub( file, 1, 64 ) .. "..."
 		end
 
-		MsgC( Color( 255, 160, 255 ), "[HTML] " )
-		MsgC( Color( 255, 255, 255 ), file, ":", line, ": ", msg, "\n" )
+		MsgC( color_white, "[HTML] " )
```
viral32111, line 79:
```suggestion
		MsgC( color_purple, "[HTML] " )
```

Define `color_purple = Color( 255, 160, 255 )` somewhere else - https://github.com/Facepunch/garrysmod/pull/1709#issuecomment-674349833

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -99,8 +99,8 @@ function PANEL:ConsoleMessage( msg, file, line )
 	--
 	-- Plain ol' console.log
 	--
-	MsgC( Color( 255, 160, 255 ), "[HTML] " )
-	MsgC( Color( 255, 255, 255 ), msg, "\n" )
+	MsgC( color_white, "[HTML] " )
```
viral32111, line 102:
```suggestion
	MsgC( color_purple, "[HTML] " )
```

Define `color_purple = Color( 255, 160, 255 )` somewhere else - https://github.com/Facepunch/garrysmod/pull/1709#issuecomment-674349833

viral32111:
> I mostly make this proposal for the equivalent of the table.insert because it's a bit faster (According to the Zaros benchmark)

I've made some code comments mentioning other comments so that this PR would actually work and keep the original colours, regardless of if it will be merged or not 😛

viral32111:
You haven't actually defined `local color_purple = Color( 255, 160, 255 )` anywhere else like I said to do in the code comments.

MoskalykA:
> local color_purple = Color( 255, 160, 255 )

Yes i'm sorry

viral32111:
At least it's fine for merging now, just up to devs & others if this is worth it.

Kefta:
It's not fine, it's still ``self.JS[self.JS + 1] = js``. Instead just do
```Lua
self.JS = { js }
```

garrysmod/lua/vgui/dhtml.lua
robotboy655, line 56: This needs to be changed

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -1,4 +1,4 @@
-
+local color_purple = Color( 255, 160, 255 )
```
robotboy655, line 1: Minor, but this should be moved closer to where it is used i.e. before `PANEL:ConsoleMessage`

robotboy655 (changes requested):

MoskalykA:
Excuse me, I didn't mean to do that.

garrysmod/lua/vgui/dhtml.lua
CornerPin: This is wrong, it must be `self.JS[#self.JS + 1] = js`

PR #1712, opened 2020-08-22
VGUI Optimized

MoskalykA:
Optimization of a lot of GUIs.


robotboy655:
I am not merging those horrible `color_190190190230` changes. You could make a case of things like that in render/think hooks, but not in code that is ran once.
There also hasn't been any activity with table.insert replacement problems pointed out by @GitSparTV in almost a month, so I am going to close this.

PR #1713, opened 2020-08-22
Simplified tobool

mastopgunaf:


ExtReMLapin:
bruh

mastopgunaf:
> bruh
> 

If I omit the "not not", it will return nil if it's given a nil instead of false, so not not has to be there. No bruh.

Kefta:
Use @neico's solution, it's the easiest to read.

mastopgunaf:
> Use @neico's second solution, it's the easiest to read.

Imagine using C operators in GMod in 2020... Why would you?

ExtReMLapin:
Hello,
As I feel like we will not agree on what to do, I decided to compile it to Lua 5.1 bytecode.
You can now load the bytecode using this interpreter : 

https://raw.githubusercontent.com/Rerumu/Rerubi/master/Source.lua
[bytecode.out](https://github.com/Facepunch/garrysmod/files/5119679/bytecode.txt)

Kefta:
@mastopgunaf Because it's the accepted style of the codebase at the moment. We can evaluate changing it all in the future, but for now, stick with the common style or this won't be accepted.

mastopgunaf:
> @mastopgunaf Because it's the accepted style of the codebase at the moment. We can evaluate changing it all in the future, but for now, stick with the common style or this won't be accepted.

I changed it to those terrible operators, should be ok now.

Kefta:
The double ``!`` is still unreadable, but whatever I guess.

robotboy655:
What is this meant to achieve?

PR #1714, opened 2020-08-25
Remove odd concat from PrintTable

darkjacky:
Seperate Msg instead of concat, also add Msg as local in the function to prevent checking the global all the time.
The odd concat is ":" .. "\n" why not just go ":\n" then? While we are add it just patch the function up a little.


CornerPin:
You can just put everyting in `Msg` parameters because it's a vararg function, you also don't have to call `tostring` since `Msg` already converts each argument to a string.

darkjacky:
Good point leme try that

PR #1715, opened 2020-08-31
util.SteamIDToAccountID and util.SteamIDFromAccountID

mastopgunaf:
Added functions to convert between SteamID and AccountID (aka SteamID3)


Kefta:
This will be added in the engine if it is at all.

robotboy655:
I am going to deny this request solely based on the fact that `AccountID` should be avoided altogether. It contains less information that a `STEAM_0:0:0` SteamID and theoretically could have collisions.

You should be using the 64bit SteamIDs.

See this page for more information:
https://developer.valvesoftware.com/wiki/SteamID

PR #1716, opened 2020-09-05
Remove string.__index error

GitSparTV:
We don't need this exclusive error.
Let it be `attempt to call method foo` instead of `attempt to index a string value with bad key ('foo' is not part of the string library)`.

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -293,8 +293,6 @@ function meta:__index( key )
 		return val
 	elseif ( tonumber( key ) ) then
 		return self:sub( key, key )
-	else
-		error( "attempt to index a string value with bad key ('" .. tostring( key ) .. "' is not part of the string library)", 2 )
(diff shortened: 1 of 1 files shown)


SneakySquid:
thanks spar very good :+1:

SneakySquid:
thanks gs very good 👍

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -289,12 +289,10 @@ local meta = getmetatable( "" )
 
 function meta:__index( key )
 	local val = string[ key ]
-	if ( val ) then
+	if ( val ~= nil ) then
```
dvdvideo1234, line 292: Is it more accurate when `key` is nil to raise error ? This way `("test")[nil]` will do nothing.

Kefta: All this change does is allow
```Lua
string.foo = false
print(string.foo)
```
to not error.

PR #1722, opened 2020-09-18
Add More Arguments to derma.SkinHook

wowm0d:
Also removes useless concatenation of hookType and hookName arguments that will break thirdparty implementations of derma.SkinHook & Derma_Hook

`derma.SkinHook( "Paint", "Frame", panel, width, height )` turns into `derma.SkinHook( "PaintFrame", panel, width, height )`
`Derma_Hook( PANEL, "Paint", "Paint", "Frame" )` turns into `Derma_Hook( PANEL, "Paint", "PaintFrame" )`

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/controlpanel.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/controlpanel.lua
@@ -51,7 +51,7 @@ end
 function PANEL:MatSelect( strConVar, tblOptions, bAutoStretch, iWidth, iHeight )
 
 	local MatSelect = vgui.Create( "MatSelect", self )
-	Derma_Hook( MatSelect.List, "Paint", "Paint", "Panel" )
+	Derma_Hook( MatSelect.List, "Paint", "PaintPanel" )
 
 	MatSelect:SetConVar( strConVar )
 
@@ -198,7 +198,7 @@ function PANEL:AddControl( control, data )
 		ctrl:ControlValues( data ) -- Yack.
 		self:AddPanel( ctrl )
 
-		Derma_Hook( ctrl.List, "Paint", "Paint", "Panel" )
+		Derma_Hook( ctrl.List, "Paint", "PaintPanel" )
 
 		return ctrl
 
@@ -360,7 +360,7 @@ function PANEL:AddControl( control, data )
 		ctrl:SetNumRows( data.rows or 4 )
 		ctrl:SetConVar( data.convar or nil )
 
-		Derma_Hook( ctrl.List, "Paint", "Paint", "Panel" )
+		Derma_Hook( ctrl.List, "Paint", "PaintPanel" )
 
 		for name, tab in pairs( data.options ) do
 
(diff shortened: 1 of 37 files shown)


SneakySquid:
I don't really see the point of this pr. It'll break any addons that use Derma_Hook/derma.SkinHook but for no benefit. What arguments even need passing to the callbacks that aren't already or can't just be stored on the panel itself?

wowm0d:
> I don't really see the point of this pr. It'll break any addons that use Derma_Hook/derma.SkinHook but for no benefit. What arguments even need passing to the callbacks that aren't already or can't just be stored on the panel itself?

Well then perhaps we should also just remove the Width and Height arguments because anything at all can be put into the panel table, also it has a benefit it removes a string concatenation which is called every frame for no reason

SneakySquid:
> > I don't really see the point of this pr. It'll break any addons that use Derma_Hook/derma.SkinHook but for no benefit. What arguments even need passing to the callbacks that aren't already or can't just be stored on the panel itself?
> 
> Well then perhaps we should also just remove the Width and Height arguments because anything at all can be put into the panel table, also it has a benefit, it removes a string concatenation which is called every time for no reason

Having the panel's width and height passed to derma.SkinHook is fine since pretty much every callback is going to use them, but if you're just adding arguments for the sake of adding them it'd be better just to store the stuff on the panel itself if you need anything extra. This also has a bonus of not being limited to the function's arguments so you can have as many as you want.

And removing a single string concat per function call doesn't have much benefit, especially when you're going to sacrifice any addons that use either Derma_Hook or derma.SkinHook for it.

wowm0d:
> > > I don't really see the point of this pr. It'll break any addons that use Derma_Hook/derma.SkinHook but for no benefit. What arguments even need passing to the callbacks that aren't already or can't just be stored on the panel itself?
> > 
> > 
> > Well then perhaps we should also just remove the Width and Height arguments because anything at all can be put into the panel table, also it has a benefit, it removes a string concatenation which is called every time for no reason
> 
> Having the panel's width and height passed to derma.SkinHook is fine since pretty much every callback is going to use them, but if you're just adding arguments for the sake of adding them it'd be better just to store the stuff on the panel itself if you need anything extra. This also has a bonus of not being limited to the function's arguments so you can have as many as you want.
> 
> And removing a single string concat per function call doesn't have much benefit, especially when you're going to sacrifice any addons that use either Derma_Hook or derma.SkinHook for it.

Think if you wanted to draw something specific using a SkinHook at a random X value, sure you could set Panel.IWantToDrawThisAt = RandomXPos then index that value in the skin function but that is just extra hassle adding more arguments doesn't hurt anything.

"sacrifice" is a strong word, things get improved/changed and if an addon creator or maintainers are unwilling to update then so be it. Backwards compatibility should not be the most important thing to look for

GitSparTV:
The idea with making less concatenations is good but it has problems with compatibility.
I can't think of better method to save compatibility rather than isstring on `panel` parameter.
I would leave this PR for adding 6 parameters but remove concat.

PR #1723, opened 2020-09-18
SmoothDamp function for math.* and vectors.

SneakySquid:
Critically damped smoothing function for the math library and Vector metatable. Useful for camera motion, path following, and as an alternative to `Lerp`/`LerpVector`.

Produces smoother results than `Lerp`/`LerpVector` with a constant fraction argument.

Based on `Game Programming Gems 4` chapter 1.10.


dvdvideo1234:
Yay, my [Catmull-Rom](https://en.wikipedia.org/wiki/Centripetal_Catmull%E2%80%93Rom_spline) cure was not merged :D. At Least I hope yours mod will

PR #1724, opened 2020-09-20
Reload spawn menu when gmod_language changes

WilliamVenner:
The spawn menu does not reload when the user changes the language of Gmod, resulting in strings in the wrong language that do not update until rejoining the server (or running spawnmenu_reload), this fixes it.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/spawnmenu.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/spawnmenu.lua
@@ -231,6 +231,7 @@ end
 -- Hook to create the spawnmenu at the appropriate time (when all sents and sweps are loaded)
 hook.Add( "OnGamemodeLoaded", "CreateSpawnMenu", CreateSpawnMenu )
 concommand.Add( "spawnmenu_reload", CreateSpawnMenu )
+cvars.AddChangeCallback( "gmod_language", CreateSpawnMenu, "spawnmenu_reload" )
 
 function GM:OnSpawnMenuOpen()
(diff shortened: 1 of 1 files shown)


robotboy655:
I am making changes to my spawnlists and I decided to change my language, poof all my changes are gone.

TomDotBat:
Who the hell actually changes their spawnlists though? I've never seen anyone actually utilise that feature.

viral32111:
I do :P

robotboy655:
Force saving isn't really an acceptable option either.

WilliamVenner:
I think people's spawn menus being the wrong language until they rejoin the server is more of an issue of someone losing the changes they were making to their spawn list. To be honest I don't really know anyone who actually changes their spawn list and I'm sure those who do have already got their language set.

Kefta:
You could ask if they want to save their menu changes before a language change takes place, but that would require direct communication between the client and menu states.

robotboy655:
>  I'm sure those who do have already got their language set.
So what's the point of this at all then? We can't just ignore a game's feature because "you didn't see anyone using it".

WilliamVenner:
Fear not, I come bearing a solution.

dvdvideo1234:
@TomDotBat Me atleast. Very useful.

GitSparTV:
Set language and then call `spawnmenu_reload`, what's the deal.

WilliamVenner:
> Set a language and then call `spawnmenu_reload`, what's the deal.

Do you really expect the average player to know this command

robotboy655 (approved):

PR #1726, opened 2020-09-25
Fix DImageButton depression, add DImageButton:SetDepressImage()

WilliamVenner:
DImageButton always had a depression problem.

Upon depressing the image after being clicked, it would immediately PerformLayout and therefore forget about its depression in the next frame. _If only it were that simple._

I have also added the `DImageButton:SetDepressImage( m_bDepressImage )` accessor function so that developers can enable or disable this behaviour.

### DIFF
--- a/garrysmod/lua/vgui/dimagebutton.lua
+++ b/garrysmod/lua/vgui/dimagebutton.lua
@@ -1,12 +1,14 @@
 
 local PANEL = {}
 AccessorFunc( PANEL, "m_bStretchToFit", "StretchToFit" )
+AccessorFunc( PANEL, "m_bDepressImage", "DepressImage" )
 
 function PANEL:Init()
 
 	self:SetPaintBackground( false )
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dimagebutton.lua
```diff
@@ -1,19 +1,21 @@
...
 	self:SetStretchToFit( true )
+	self:SetDepressImage( true )
 
 	self:SetCursor( "hand" )
 	self.m_Image = vgui.Create( "DImage", self )
 
 	self:SetText( "" )
 
-	self:SetColor( color_white )
+	self:SetColor( Color( 255, 255, 255, 255 ) )
```
FlorianLeChat, line 18: Why not using `color_white` constant ?

WilliamVenner: Ah didn't notice that - my local file was not up-to-date.

PR #1728, opened 2020-10-01
Improved DAdjustableModelPanel movement

theomonnom:
- Removal of hard-coded keys
- The cursor is no longer displayed when moving
- Reduced move speed when holding shift key
- Fixed a bug where the pitch did not stay between -90 and 90 degrees, which resulted in an inverted yaw
- Space and control keys are no longer relative to viewing angle

### DIFF
--- a/garrysmod/lua/vgui/dadjustablemodelpanel.lua
+++ b/garrysmod/lua/vgui/dadjustablemodelpanel.lua
@@ -1,4 +1,3 @@
-
 local PANEL = {}
 
 AccessorFunc( PANEL, "m_bFirstPerson", "FirstPerson" )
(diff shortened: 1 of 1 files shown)


GitSparTV (approved):

robotboy655 (commented):
The problem with the removal of hardcoded keybinds is that if you have `+forward` bound to more than 1 key, only one of those keys will work in `DAdjustableModelPanel` now. Maybe check if either `+forward` keybind or the W keys are being held? Altho that can use issues too, if you rebind your move keys to ESDF or something.

Also I do not see the point in reducing the SHIFT speed.

FeepsDev:
There is not any solution to this since input.LookupBinding return the first found key. But I think that it's still better than having hardcoded keys. (And that cannot necessarily be a problem because there are still the arrow keys)

I reduced the shift speed because the value is too big compared to most models

nodgear:
I don't think changing values like the shift speed is benefitial, everyone is used to it like that, i have used DModelPanel with large models before and the shift helped a lot!

PR #1738, opened 2020-10-24
[TTT] Change TTT defaults

Tiagoquix:
Changes:

- Increased map time limit to 90 minutes;
- Decreased postround time to 10 seconds;
- Doubled round limit (now it is 12);
- Enabled postround DM;
- Detectives now start with two credits instead of only one;
- Enabled detective hats;
- Enabled discombobulator jump;
- Blocked grenade throwing during preround;
- Enabled teleporter fragging;
- Increased name change ban time from 10 minutes to 20 minutes.

Any feedback is appreciated. The changes are based mainly on what the community servers use and on my personal opinion too.

### DIFF
--- a/garrysmod/gamemodes/terrortown/terrortown.txt
+++ b/garrysmod/gamemodes/terrortown/terrortown.txt
@@ -28,7 +28,7 @@
 			"name"		"ttt_time_limit_minutes"
 			"text"		"Map Time Limit"
 			"type"		"Numeric"
-			"default"	"75"
+			"default"	"90"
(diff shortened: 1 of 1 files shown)


svdm:
I've implemented some of these in feb0767, primarily those that I didn't even realise were off by default because of how commonly they have always been enabled. Some other suggestions I don't agree with, for example discombob jump can mess up certain maps, whether it can be enabled therefore is server-dependent. Round times and limits are similarly up to the server.

PR #1739, opened 2020-10-26
Toolgun bug fixes

WilliamVenner:
I noticed that there is a bug where the toolgun does not call Deploy on the first tool that is selected (both realms, so it's not a prediction thing), it's a simple fix. The bug is caused by SWEP:GetToolObject() being called before SWEP:Think(), which sets SWEP.Mode.

Also fixed Holster being called twice in the name of releasing ghost entities.

And finally, I have also re-enabled ToolObj:Allowed() on the client as FCVAR_REPLICATE now works for Lua ConVars.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
@@ -176,7 +176,7 @@ function SWEP:Think()
 
 		if ( lastmode_obj ) then
 			-- We want to release the ghost entity just in case
-			lastmode_obj:Holster()
+			lastmode_obj:ReleaseGhostEntity()
 		end
 	end
 
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua
@@ -171,18 +171,13 @@ function SWEP:Think()
 		return
 	end
 
-	if ( lastmode ) then
-		if ( lastmode != curmode ) then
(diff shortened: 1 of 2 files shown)


Kefta:
This is not a proper fix for this, the tool objects should just be initialised in Deploy.

WilliamVenner:
> This is not a proper fix for this, the tool objects should just be initialised in Deploy.

https://github.com/Facepunch/garrysmod/blob/34d2ccc460688750cb81c0ca031c75635aacfbaf/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stool.lua#L132-L150

https://github.com/Facepunch/garrysmod/blob/40b05fe9957e896a6d5b68b6db5cb1b9a07932d3/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua#L334

This should be called long before SWEP:Deploy() on the toolgun - assuming that, shouldn't the tool objects already be initialized when Deploy is called?

WilliamVenner:
https://github.com/Facepunch/garrysmod/blob/eb05b5cebd10e476e9f2f7ba5b0a8e1b4ff2a7b1/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua#L180

This line is also problematic, if the intention is to release the ghost entity, why call Holster twice? I encountered a race condition because of it.

WilliamVenner:
I should note that the ghost entities still stay behind when switching to another weapon in the spawn menu. https://github.com/Facepunch/garrysmod-issues/issues/3709

I fixed this in my own addon by implementing https://github.com/Facepunch/garrysmod-requests/issues/1739 and using a variable set by `TOOL:Deployed()` and `TOOL:Holstered()` that blocks MakeGhostEntity.

PR #1741, opened 2020-11-07
Fix parent IsModal check

ceifa:
After the last commit in this file, this error happens constantly:
```
[murder] lua/vgui/dmenu.lua:223: attempt to call method 'IsModal' (a nil value)
  1. Open - lua/vgui/dmenu.lua:223
   2. DoScoreboardActionPopup - addons/murder/gamemodes/murder/gamemode/cl_scoreboard.lua:161
    3. DoClick - addons/murder/gamemodes/murder/gamemode/cl_endroundboard.lua:131
     4. unknown - lua/vgui/dlabel.lua:237
```

With this check we ensure there are no breaking changes on `dmenu`.


FlorianLeChat:
I encountered the same issue, but I think it's now fixed in the Dev branch (?).
https://commits.facepunch.com/374359

PR #1743, opened 2020-11-12
Including some missing entities from CS:S and the ZPS mod

BreakinBenny:
The footstep control one (http://src-ents.shoutwiki.com/wiki/Func_footstep_control) seems the most interesting because it modifies footstep sounds, whereas the ones named [B]random[/B] are for randomly generated ammo and weaponry in ZPS maps.

Basically, ladders you can strafe on can be set up to make other sounds. Or you can set up to auto-remove them with self.Remove().


robotboy655:
This belongs on Workshop, not in the base game.

PR #1744, opened 2020-11-23
Use mask_shot for player id trace

thegrb93:


robotboy655:
Please do not submit Pull Requests with no description.

PR #1745, opened 2020-11-23
Allow null ply in MakeVehicle

thegrb93:
Lets advdupe2 load vehicles from a dupe on map start with no owner.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/commands.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/commands.lua
@@ -916,7 +916,7 @@ end )
 
 local function MakeVehicle( ply, Pos, Ang, model, Class, VName, VTable, data )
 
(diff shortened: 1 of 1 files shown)


thegrb93:
Fyi this is how makeProp does it so should be fine. https://github.com/Facepunch/garrysmod/blob/42721ec20cfc0e212401bd803d10ea4b699a8f7f/garrysmod/gamemodes/sandbox/gamemode/commands.lua#L135

robotboy655:
Introduced in https://github.com/Facepunch/garrysmod/commit/00c71b02139bbaff0f242e9558ee257ae873551b#diff-b7a357bf1c9ee1f589fa300430bf6c5c1331cb4fcb1ddc81d7a9bca0b36fbd8b due to a copypaste mistake.

PR #1746, opened 2020-11-23
Use mask_shot for player id trace

thegrb93:
Use mask_shot for player id traces

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_targetid.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_targetid.lua
@@ -6,6 +6,7 @@
 function GM:HUDDrawTargetID()
 
 	local tr = util.GetPlayerTrace( LocalPlayer() )
+	tr.mask = MASK_SHOT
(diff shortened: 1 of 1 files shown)


thegrb93:
Lets you see player names through fence props.

Kefta:
I'm iffy on this because I think the point of the name display is to show the name of players you can **see.** If someone is hiding thru opaque props that can be shot thru, this would allow you to soft-wallhack. The question is if any gamemodes that relies on this behaviour derives from Sandbox and doesn't override the function.

thegrb93:
Well you can see players through translucent props but not their name so it's equally a problem for that case. This is at least a better middle-ground.

Kefta:
I think it's a worse middle-ground since you still can't see them thru translucent props that can't be shot thru. Perhaps MASK_VISIBLE would be better but edge cases would need to be tested.

thegrb93:
I honestly don't know of any MASK_SHOT props you can't see through... that'd be really bad game design.

PR #1747, opened 2020-11-24
Make Bokeh DOF effect function callable from anywhere

FlorianLeChat:
The `DrawBokehDOF` function is now global to allow it to be called everywhere (like the other built-in effects). This can be useful to create some scripts using these effects without having to copy/paste these functions.


robotboy655:
The Bokeh DOF is broken which is why its hidden like that.

robotboy655 (approved):

PR #1748, opened 2020-11-28
Allow hiding of unwanted creation tabs in spawnmenu

KennyHarrer:


JustPlayerDE (changes requested):
This works fine for me, but also giving the key value seems useless to me ^^
image
[1748.zip](https://github.com/Facepunch/garrysmod/files/5697457/1748.zip)

robotboy655:
Same as #1298, this implementation does not have an effect live, which may be unwanted in case of things like hiding the tools/tabs for non admins only, which can happen mid game, not just on loading time.

ExtReMLapin:
Wish you could do that with your kids, uh ?

PR #1750, opened 2020-12-03
Implement correct hook behaviour for ENTITY.CanTool and ENTITY.CanProperty

WilliamVenner:
If a hook returns `nil` it should not affect any others

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/shared.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/shared.lua
@@ -247,7 +247,10 @@ function GM:CanProperty( pl, property, ent )
 
 	-- Give the entity a chance to decide
 	if ( ent.CanProperty ) then
-		return ent:CanProperty( pl, property )
+		local entCanProperty = ent:CanProperty( pl, property )
+		if ( entCanProperty != nil ) then
+			return entCanProperty
+		end
 	end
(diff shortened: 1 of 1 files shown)


WilliamVenner:
Example of uber-popular addon expecting this behaviour:

thegrb93:
Gamemode functions are not hooks though. They are also always called last as long as all hooks returned nil.

WilliamVenner:
> Gamemode functions are not hooks though. They are also always called last as long as all hooks returned nil.

ENTITY.CanTool and ENTITY.CanProperty are not gamemode functions. They are not always called in the context of a gamemode function, as shown in the example above.

thegrb93:
I think https://github.com/FPtje/Falcos-Prop-protection/blob/master/lua/fpp/server/core.lua#L496 should be `if val==false then return end`

thegrb93:
ENTITY.CanTool and ENTITY.CanProperty are meant to only return true or false, not nil.

WilliamVenner:
Yes I think you are right,

> Return false to disallow using that property, return true to allow.
> You must return a value. Not returning anything can cause unexpected results.

(From CanProperty)

I'm not really sure what "unexpected" results whoever wrote this are talking about, but yeah this seems like an FPP problem.

PR #1752, opened 2020-12-03
Conditionally prevent drawing voice panels

vakkendwarf:
In an event where you might want to only conditionally replace them with something else.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_voice.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_voice.lua
@@ -97,6 +97,8 @@ function GM:PlayerStartVoice( ply )
 
 	if ( !IsValid( ply ) ) then return end
 
+	if hook.Run("ShouldDrawPlayerVoicePanel", ply) == false then return end
+
 	local pnl = g_VoicePanelList:Add( "VoiceNotify" )
 	pnl:Setup( ply )
 	


garrysmod/gamemodes/base/gamemode/cl_voice.lua
```diff
@@ -97,6 +97,8 @@ function GM:PlayerStartVoice( ply )
 
 	if ( !IsValid( ply ) ) then return end
 
+	if hook.Run("ShouldDrawPlayerVoicePanel", ply) == false then return end
```
Kefta, line 100: Use hook.Call and pass ``self`` as the gamemode table. Replace GAMEMODE with self above.

Kefta (changes requested):

garrysmod/gamemodes/base/gamemode/cl_voice.lua
Kefta: Also needs to match the codebase style. Spaces around between parentheses and body, and needs parentheses around the if-condition.

vakkendwarf: Done

robotboy655:
```lua
hook.Add( "PlayerStartVoice", "myStuff", function( ply )

-- Do your stuff here

return false  -- prevent the default GM:PlayerStartVoice from running

end )
```

Kefta:
There are tons of listener PlayerStartVoice hooks (non-drawing, non-returning) that would be affected by that.

vaqxai:
Yeah the solution provided by @robotboy655 is hardly a good workaround. Much less a solution.

robotboy655:
I mean, technically you could call the hooks manually, but either way, do we really need a hook to disable every single small thing?

What are you replacing the voice panels with?

Kefta:
Another HUDShouldDraw ID could here instead of a unique hook. It won't be called if there's a custom voice panel that overrides PlayerStartVoice and the hook is generally pretty optimised on every server already by directing it into a straight lua hashtable search.

PR #1754, opened 2020-12-13
Use Brackets for keys in PrintTable

JustPlayerDE:
To avoid confusion when there is [357] and ["357 "] for example.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -73,7 +73,7 @@ function PrintTable( t, indent, done )
 	for i = 1, #keys do
 		local key = keys[ i ]
 		local value = t[ key ]
-		key = (type(key) == "number") && "[" .. key .. "]" || "[\"" .. key .. "\"]"
+		key = (type( key ) == "number") && "[" .. key .. "]" || "[\"" .. tostring( key ) .. "\"]"
 		Msg( string.rep( "\t", indent ) )
 
 		if  ( istable( value ) && !done[ value ] ) then
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -73,6 +73,7 @@ function PrintTable( t, indent, done )
 	for i = 1, #keys do
 		local key = keys[ i ]
 		local value = t[ key ]
+		key = (type(key) == "number") && "[" .. key .. "]" || "[\"" .. key .. "\"]"
 		Msg( string.rep( "\t", indent ) )
 
 		if  ( istable( value ) && !done[ value ] ) then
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/util.lua
```diff
@@ -73,6 +73,7 @@ function PrintTable( t, indent, done )
 	for i = 1, #keys do
 		local key = keys[ i ]
 		local value = t[ key ]
+		key = (type(key) == "number") && "[" .. key .. "]" || "[\"" .. key .. "\"]"
```
WilliamVenner, line 76: I think `isnumber` is faster than `type == "number"`?

WilliamVenner: Also, I don't really see the need for brackets around numeric keys.

JustPlayerDE: isnt `isnumber` just a shortcut for `type == "number"` (which basicly means a function call more somewhere)?

with the brackets i think it would look more consistent when both have them (maybe they can even be removed on both but that way it looks more like a table) ^^

WilliamVenner: I don't think it is internally.

The point of adding the brackets around the strings is so that they're not consistent with each other though.

CornerPin:
You are assuming that keys are always either numbers or strings, the concatenation will throw an error with other key types.

garrysmod/lua/includes/util.lua
JustPlayerDE: i mean it just looks nicer to have
```
["string"] = {
 ...
}
[2] = "number"
```

than 
```
["string"] = {
 ...
}
2 = "number"
```

or
```
"string" = {
 ...
}
2 = "number"
```

so for this i may wait for a maintainer to decide which one would suite more for gmod

JustPlayerDE:
> You are assuming that keys are always either numbers or strings, the concatenation will throw an error with other key types.

you are right i totally missed that, i will try fix/comply with that as soon as i get to my computer again (thanks for the info tho)

garrysmod/lua/includes/util.lua
WilliamVenner: OK nevermind, I agree with you.

robotboy655:
image

Maybe this is a bit cleaner?
```if ( isstring( key ) && key != key:Trim() ) then key = "[\"" .. key .. "\"]" end```

JustPlayerDE:
> image
> 
> Maybe this is a bit cleaner?
> `if ( isstring( key ) && key != key:Trim() ) then key = "[\"" .. key .. "\"]" end`

I agree that this will fully fix the issue while adding less stuff to the console

_but_ i may not be able to answer the question if it is cleaner as i also prefer consistency a bit (in that case the key could also be 
 `Entity(1)["[\"Fuck  \"]"]` as there is no reference tho)

robotboy655:
A fair point. I am not sure what to do about it, either way it's gonna be either confusing or just increase amount of irrelevant symbols on each line.

ExtReMLapin:
Also there is the same kind of issue with table to string, it doesn't correctly print keys

robotboy655 (approved):

PR #1755, opened 2020-12-18
New library: ease

WilliamVenner:
The ease library contains 30 different standard easing functions from https://github.com/ai/easings.net

Each easing function can be previewed here: https://easings.net/

#### Easing Function Source Code

https://github.com/ai/easings.net/blob/master/src/easings.yml

#### YML -> JSON

https://onlineyamltools.com/convert-yaml-to-json

#### Conversion + Testing Script

https://gist.github.com/WilliamVenner/9714e9ec3d16a9baef4daf372501a275

_This converter is not 100% future proof nor accurate, and does require some human intervention_

#### Lua Testing Data

https://gist.github.com/WilliamVenner/52b4218e830d0759bdd322b549a1bf11

#### Test Results

https://gist.github.com/WilliamVenner/0a9fbba8fd6739d4af016645acdffb25

`0` = no difference between JavaScript function's return value and Lua function's return value

### DIFF
--- a/garrysmod/lua/includes/init.lua
+++ b/garrysmod/lua/includes/init.lua
@@ -59,6 +59,7 @@ end
 
 if ( CLIENT ) then
 
+	require ( "ease" )			-- Easing function library
 	require ( "draw" )			-- 2D Draw library
 	require ( "markup" )		-- Text markup library
 	require ( "effects" )
--- a/garrysmod/lua/includes/init_menu.lua
+++ b/garrysmod/lua/includes/init_menu.lua
@@ -12,6 +12,7 @@ include( "util/sql.lua" ) -- Include sql here so it's available at loadtime to m
 require( "concommand" )
 require( "list" )
 require( "hook" )
+require( "ease" )
 require( "draw" )
 require( "http" )
 require( "cvars" )
--- a/garrysmod/lua/includes/modules/ease.lua
+++ b/garrysmod/lua/includes/modules/ease.lua
@@ -1,0 +1,183 @@
+local math = math
+
+module( "ease" )
+
+--[[---------------------------------------------------------
+	Source code of functions
+-----------------------------------------------------------]]
+
+-- https://github.com/Facepunch/garrysmod/pull/1755
+-- https://web.archive.org/web/20201212082306/https://easings.net/
+-- https://web.archive.org/web/20201218211606/https://raw.githubusercontent.com/ai/easings.net/master/src/easings.yml
+
+--[[---------------------------------------------------------
+	Easing constants
+-----------------------------------------------------------]]
+
+local c1 = 1.70158
+local c3 = c1 + 1
+local c2 = c1 * 1.525
+local c4 = ( 2 * math.pi ) / 3
+local c5 = ( 2 * math.pi ) / 4.5
+local n1 = 7.5625
+local d1 = 2.75
+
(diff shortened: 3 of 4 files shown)


Donkie:
This is something I often use in many different projects. Would definitely be nice to have it as a standard library.

WilliamVenner:
@robotboy655 I would be happy to document this myself on the wiki if you do not have the time.

garryspins:
good job billy, keep it up, im proud of you *dunn da dun dunandun da dundundundundun dun dun dun dun dunnn dun dun*

WilliamVenner:
@robotboy655 Pls merge, there's no reason not to merge this

thegrb93:
99 other prs but a bitch ain't one.

dvdvideo1234:
How exactly this eases stuff ?

WilliamVenner:
> How exactly this eases stuff ?

There is an example at the bottom of the PR...?

WilliamVenner:
@Kefta Changed it from ease -> math.ease

I'm too much of a mathlet to change all the `x * x ...` to `x ^ ...` accurately. I don't think it's really that important though.

Edits for maintainers is on if it does bother you though.

WilliamVenner:
@Kefta Those `x * x ...` are actually intentional, they are a different operation to `x ^ ...` because `(x * x) * x) != x ^ 3` as `x` changes during the expression

CornerPin:
> @Kefta Those `x * x ...` are actually intentional, they are a different operation to `x ^ ...` because `(x * x) * x) != x ^ 3` as `x` changes during the expression

That is not true. `x^n`, where `n` is a positive integer, is a repeated multiplication, so `(x * x) * x` **is** `x^3`.

The difference is that `(x * x) * x` may be more efficient, unless LuaJIT already optimizes `x^3` to repeated multiplication.

WilliamVenner:
Yes, thank you for the elementary math lesson, but nevermind it was just a 🅱️avascript thing:

```js
function easeInQuintx(x) {
return x ^ 5;
}
undefined
function easeInQuint(x) {
return x * x * x * x * x;
}
undefined
easeInQuint(.5);
0.03125
easeInQuintx(.5);
5
```

Lua thankfully doesn't evaluate expressions like this lol

WilliamVenner:
Fixed, and I updated the tests, too. Everything appears to be in order.

WilliamVenner:
Dumping a review gs did here:

> - math.ease: I'm not convinced on the constants from the original codebase. Please source them as being derived purposefully or make them default configureable values that are bounded

thekillertoaster:
Honestly, this would make life a lot easier with a lot of my projects

WilliamVenner:
Honestly I think this should just be merged. If someone wants to change the specifics of these functions, they should just make their own easing functions; Gmod should just provide some helpful, common easing functions.

MyDimeIsUp:
So whats the hold up?

Kefta:
The arbitrary constants.

WilliamVenner:
> The arbitrary constants.

If someone wants to change the specifics of these functions, they should just make their own easing functions; Gmod should just provide some helpful, common easing functions.

WilliamVenner:
That being said, I've localized the `math.*` functions (and `math.pi`) that the extension uses so that it won't have to lookup from _G multiple times per frame.

PR #1757, opened 2020-12-20
Add ent:AddHealth()

Down-s:
This would be a big time saver

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -118,6 +118,14 @@ function meta:CallOnRemove( name, func, ... )
 	mytable.OnDieFunctions[ name ] = { Name = name, Function = func, Args = { ... } }
 
 end
+			
+--[[---------------------------------------------------------
+	Name: AddHealth
+	Desc: Adds health to the entity.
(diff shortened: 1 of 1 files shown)


WilliamVenner:
The clamping functionality of this makes it useful. +1

FlorianLeChat:
When it will be added, maybe do the same thing with `Player:AddArmor()`.

Down-s:
> When it will be added, maybe do the same thing with `Player:AddArmor()`.

Thanks, created a PR for it

TomDotBat:
Probably should've just updated this one and changed the title instead of making another.

Kefta:
As discussed in the Discord, there's the whole problem of a player having over the max health and this function reverting it to the max. I believe this should be implemented selectively in people's code to deal with this behaviour in their own way.

Down-s:
> As discussed in the Discord, there's the whole problem of a player having over the max health and this function reverting it to the max. I believe this should be implemented selectively in people's code to deal with this behaviour in their own way.

I can add a check if the player's health is above max health, then add to that without clamping

Kefta:
The behaviour of only adding if the health is already above the max health when the max health can change seems very specific and addon-implementable.

PR #1758, opened 2020-12-21
Add ent:AddArmor()

Down-s:


FlorianLeChat:
Nice work but I think it's better to put this function only for the players and not the entities (like **Player**:SetArmor).

Kefta:
See my comment on https://github.com/Facepunch/garrysmod/pull/1757.

PR #1759, opened 2020-12-28
Add PANEL:GetValues function

EclipseCantCode:
Allows for easy extraction of values from  a DListView.

Nested table's keys are the same as the related column's text.


GitSparTV:
I don't think getting values using header name is convenient, I would use integers.

PR #1760, opened 2020-12-28
Add Panel:EnableCursorOnRelease()

Down-s:
Currently when you release a dragndrop panel, your mouse will hidden on release. Adding this function will give people the option for disabling this.

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel/dragdrop.lua
+++ b/garrysmod/lua/includes/extensions/client/panel/dragdrop.lua
@@ -443,6 +443,12 @@ function meta:DragClick( mcode )
 
 end
 
+function meta:EnableCursorOnRelease( enable )
+
+	self.ReleaseCursor = enable
+
+end
+
 function meta:DragMouseRelease( mcode )
 
 	if ( IsValid( dragndrop.m_DropMenu ) ) then return end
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/client/panel/dragdrop.lua
```diff
@@ -461,8 +467,7 @@ function meta:DragMouseRelease( mcode )
 
 	dragndrop.Drop()
 
-	-- Todo.. we should only do this if we enabled it!
-	if ( gui.EnableScreenClicker ) then
+	if ( gui.EnableScreenClicker && !self.ReleaseCursor ) then
```
GitSparTV, line 470: Maybe it should check `if CLIENT` or `if not MENU_DLL`, not `if gui.EnableScreenClicker`?

Kefta: I think the former is more appropriate.

robotboy655:
This needs further evaluation. What code you are running where this is a problem? Why is it even trying to hide the cursor? Maybe saving the screen clicker state before drag'n'drop and restoring it after is a better solution?

robotboy655:
Closing due to no activity.

PR #1762, opened 2021-01-04
undo: PreUndo hook

NathanX-S:


VeXan39231:
This would be nice

robotboy655:
Do either of you have an actual use case for this or is this a "just because" proposition?

NathanX-S:
@robotboy655 refunding props

WilliamVenner:
Isn't the point of Garry's Mod to be as moddable as possible?

GitSparTV:
> refunding props
Well what if undo didn't happen? In this case you need PostUndo

NathanX-S:
What? PreUndo gets called right before the entities are removed but the undo is valid, this is what I need?

WilliamVenner:
Maybe it should be moved into the `if ( count > 0 ) then` block, or better yet just make it shared and also call it in `Undo_FireUndo` on the server.

WilliamVenner:
While we're here, can we bring some attention to: https://github.com/Facepunch/garrysmod/pull/1749 https://github.com/Facepunch/garrysmod/pull/1685

NathanX-S:
@robotboy655

Yogpod:
could be used to add a 'are you sure?' thing before undoing

NathanX-S:
@Kefta

GitSparTV:
> could be used to add a 'are you sure?' thing before undoing

This is a serverside hook tho.
Would you network the confirmation?
I used PlayerBindPress on gmod_undo to make Undo Lock

NathanX-S:
@robotboy655

Kefta:
Your use case wouldn't even work with this because of where it's placed. Nor would

> could be used to add a 'are you sure?' thing before undoing

because you don't have a return value. Change the hook to have a bool return value that can block the undo, add PostUndo before the return where it passed the undo object and count, then I'll merge it.

Yogpod:
> Your use case wouldn't even work with this because of where it's placed. Nor would
> 
> 
> 
> > could be used to add a 'are you sure?' thing before undoing
> 
> 
> 
> because you don't have a return value. Change the hook to have a bool return value that can block the undo, add PostUndo before the return where it passed the undo object and count, then I'll merge it.

guess that's on me for not actually reading the code and just looking at the name

NathanX-S:
paging @Kefta

garrysmod/lua/includes/modules/undo.lua
```diff
@@ -356,7 +356,9 @@ function Do_Undo( undo )
 
 	if ( !undo ) then return false end
 	
-	hook.Run( "PreUndo", undo )
+	local call = hook.Run( "PreUndo", undo )
+
+	if ( !call ) then return end
```
Kefta, line 361: This should be ``if ( call == false ) then return end``, othewise, all undos would be blocked. You should also add default GM funcs for PreUndo and PostUndo.

Kefta (changes requested):

garrysmod/lua/includes/modules/undo.lua
NathanX-S: @Kefta What do you mean by default GM funcs?

Kefta: > @Kefta What do you mean by default GM funcs?

Add GM:PreUndo (returning true) and GM:PostUndo to the base gamemode.

NathanX-S:
paging @Kefta

ThatLing:
more bloat hooks for no reason, just what we needed

NathanX-S:
> more bloat hooks for no reason, just what we needed

a) irrelevant comment good shit
b) usecase was given, i personally only wanted one hook and was told to add another, cry somewhere else

ThatLing:
you seem awfully upset about my single comment, maybe some timeout will do you some good

garrysmod/lua/includes/modules/undo.lua
```diff
@@ -355,7 +355,11 @@ end
 function Do_Undo( undo )
 
 	if ( !undo ) then return false end
+	
+	local call = hook.Run( "PreUndo", undo )
```
Kefta, line 359: Inline the if-statement instead of making this a local since it's only used in one place.

Kefta (changes requested):

NathanX-S:
@Kefta please accept

Kefta (approved):

VeXan39231:
^ this is big

PR #1763, opened 2021-01-06
Fix Disconnect button width with other languages (russian, etc.)

osiris-plus:
Fixes this:


GitSparTV:
Already fixed

osiris-plus:
No it's not, the button will have default size.

PR #1766, opened 2021-01-22
Update colour.lua

MeteorTheLizard:
Constant Color tables should not be re-created over and over again.


CornerPin:
#1740

robotboy655:
Even though I could not reproduce the said corruption, better not risk it breaking again.

PR #1769, opened 2021-02-10
Proposing table.Where method

Lexa013:
Example:

local admins = table.Where(
    player.GetAll(),
    function(i, player)
        return player:isAdmin()
    end,
    ipairs
)

*Shitty GitHub code identation*

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -698,3 +698,22 @@ function table.GetKeys( tab )
 	return keys
 
 end
+
+--[[---------------------------------------------------------
+	Name: table.Where( table a, boolean condition, *optional ipairs/pairs )
+	Desc: A simple Where method like c#
+-----------------------------------------------------------]]
+
+function table.Where(a, b, iterator)
+    local c = {}
+
+    iterator = iterator or pairs
+    for i, v in iterator(a or {}) do
+        local res = b(i, v)
+        if (res) then
+            c[i] = v
+        end
+    end
+
+    return c
+end


thegrb93:
Why do people hate typing a loop so much. It's not hard at all.

thegrb93:
My only problem with it is the way it uses iterator isn't universal

Also it's not saving much space
```lua
local admins = table.Where(player.GetAll(), function(i, player) return player:isAdmin() end, ipairs)
local admins = {} for _, v in ipairs(player.GetAll()) do if v:IsAdmin() then admins[#admins+1] = v end end
```

FlorianLeChat:
I'm pretty sure it's faster to do a traditional loop than to use this function.

GitSparTV:
No, just no

garrysmod/lua/includes/extensions/table.lua
```diff
@@ -698,3 +698,22 @@ function table.GetKeys( tab )
...
+-----------------------------------------------------------]]
+
+function table.Where(a, b, iterator)
+    local c = {}
+
+    iterator = iterator or pairs
+    for i, v in iterator(a or {}) do
+        local res = b(i, v)
+        if (res) then
+            c[i] = v
```
thegrb93, line 714: If you use this in your own code, make sure you correct this to `c[#c+1]` otherwise, you'll end up with nil holes in your array.

garrysmod/lua/includes/extensions/table.lua
```diff
@@ -698,3 +698,22 @@ function table.GetKeys( tab )
...
+--[[---------------------------------------------------------
+	Name: table.Where( table a, boolean condition, *optional ipairs/pairs )
+	Desc: A simple Where method like c#
+-----------------------------------------------------------]]
+
+function table.Where(a, b, iterator)
+    local c = {}
+
+    iterator = iterator or pairs
+    for i, v in iterator(a or {}) do
```
GitSparTV, line 711: Why do we need to iterate empty table?

CornerPin:
Does it really need a third parameter? I think it should just always use `pairs`.

GitSparTV:
Instead of this function, table.HasValue should return key as well

garrysmod/lua/includes/extensions/table.lua
CornerPin: > If you use this in your own code, make sure you correct this to `c[#c+1]` , otherwise you'll end up with nil holes in your array.

It basically just filters the table, it's not supposed to return a sequential array. Maybe `table.Filter` will be a better name for it.

thegrb93: It is supposed to return a sequential array

CornerPin: I took a glance at the C# method. For a dictionary, the closest type to a Lua table, `Where` returns an iterator over a collection of key-value pairs. https://rextester.com/QDJN89112

Kefta:
Inefficient and confusing for marginally less verbosity.

PR #1773, opened 2021-03-03
Penalize massive servers (>80) heavily 

karmakarmakarmakarmakarmakarmakarma:
This PR will penalize massive servers in the server list and drop them much further down than small to medium servers. My reasoning for this change is that massive servers effectively suffocate smaller servers by not allowing them to get any players which doesn't allow them to gain a community. It's very difficult to start a successful server due to how massive servers have become and the rise of player + ping spoofing (including anycast). This PR attempts to alleviate this by penalizing servers that grow above 43-128 population by dropping them further below.
Owners of the massive servers who are targeted will not be affected too much as they have already developed a community. The only people who WILL be affected are the people who don't have a community, and they will be affected in a good way (being brought up to the top of the server list)

This will obviously not address the rise of certain hosting providers that provide ping spoofing disguised as anycast with their servers, nor player spoofing. 
This can also be improved by a player to max player ratio system.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -286,8 +286,8 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	if ( data.players >= 4 ) data.recommended -= 20;
 	if ( data.players >= 8 ) data.recommended -= 15;
 	if ( data.players >= 16 ) data.recommended -= 15;
-	if ( data.players >= 32 ) data.recommended += 20;
-	if ( data.players >= 64 ) data.recommended += 30;
+	if ( data.players >= 32 ) data.recommended -= 5;
+	if ( data.players >= 43 ) data.recommended += 30;
 	if ( data.players >= 80 ) data.recommended += 60;
 	if ( data.players >= 128 ) data.recommended += 90;
 
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -283,11 +283,13 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	if ( data.isAnon ) data.recommended += 1000; // Anonymous server
 
 	// The first few bunches of players reduce the impact of the server's ping on the ranking a little
-	if ( data.players >= 4 ) data.recommended -= 10;
+	if ( data.players >= 4 ) data.recommended -= 20;
 	if ( data.players >= 8 ) data.recommended -= 15;
 	if ( data.players >= 16 ) data.recommended -= 15;
-	if ( data.players >= 32 ) data.recommended -= 10;
-	if ( data.players >= 64 ) data.recommended -= 10;
+	if ( data.players >= 32 ) data.recommended += 20;
+	if ( data.players >= 64 ) data.recommended += 30;
+	if ( data.players >= 80 ) data.recommended += 60;
+	if ( data.players >= 128 ) data.recommended += 90;
 
 	data.listen = data.desc.indexOf( '[L]' ) >= 0;
 	if ( data.listen ) data.desc = data.desc.substr( 4 );
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -282,7 +282,7 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 	if ( data.pass ) data.recommended += 300; // Password protected, can't join it
 	if ( data.isAnon ) data.recommended += 1000; // Anonymous server
 
-	// The first few bunches of players reduce the impact of the server's ping on the ranking a little
+	// Penalize massive servers while giving bonuses to small to medium servers
 	if ( data.players >= 4 ) data.recommended -= 20;
 	if ( data.players >= 8 ) data.recommended -= 15;
 	if ( data.players >= 16 ) data.recommended -= 15;


karmakarmakarmakarmakarmakarmakarma:
image
What it looks like after in the DarkRP section

artemking4:
Doubt that this is a good idea, idk why
but on the other hand, monopolization is always bad for any sort of community

KingofBeast:
Letting players choose their own sorting scheme will always be better.
Many players choose high density servers _because_ they're high density. Being populated isn't inherently a bad thing and therefore shouldn't inherently be a reason to lower their "ranking"

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> Letting players choose their own sorting scheme will always be better.
> Many players choose high density servers _because_ they're high density. Being populated isn't inherently a bad thing and therefore shouldn't inherently be a reason to lower their "ranking"

It'll always be better, but right now you either sort by most players or no players, there is no in between. This is a good middle ground almost in a literal sense.

KingofBeast:
It forces people into that middle ground, which is a net negative overall. Why not filter options above the server list to hide empty/full servers, or to set a range of active players and only show servers falling into that zone

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> It forces people into that middle ground, which is a net negative overall. Why not filter options above the server list to hide empty/full servers, or to set a range of active players and only show servers falling into that zone

Nobody is forced into it. It is very simple to sort by players. People are currently unable to search for a middle ground because there is no option to, as you only get extremes. I would argue it's not a net negative at all, but bigger servers getting more and more new players is when small servers suffocate and die from a lack of incoming players.

KingofBeast:
> because there is no option to

Then something like what I proposed gives them greater granularity - let them choose a range of player counts, or at least a toggle to prioritize modestly populated servers.

I agree that the only sorting options being on the extreme ends of either player count/ping is a bad thing. You're right, people who do want the middle ground have no easy way to get there. I don't see it as a reason to punish the larger servers that are popular for a reason though.

Sorting presets like what I mentioned above seems like it would be a much better solution than having several different factors playing into a single global sorting method with the only other options being looking at empty servers, full servers, or the laggiest servers.

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> > because there is no option to
> 
> Then something like what I proposed gives them greater granularity - let them choose a range of player counts, or at least a toggle to prioritize modestly populated servers.
> 
> I agree that the only sorting options being on the extreme ends of either player count/ping is a bad thing. You're right, people who do want the middle ground have no easy way to get there. I don't see it as a reason to punish the larger servers that are popular for a reason though.
> 
> Sorting presets like what I mentioned above seems like it would be a much better solution than having several different factors playing into a single global sorting method with the only other options being looking at empty servers, full servers, or the laggiest servers.

This is all out of scope for this PR. I will probably work on this later but it will be confusing for new players. To address your point, I'm not punishing large servers so much as not rewarding them for being big. Most big servers are big simply because they're big and new players click on the first server they see, which is what I'm aiming to change so that new players join servers that need a community. Large servers will NOT be punished because they have a community that already exists so they will not really be affected.

t00ny:
good idea

Kefta:
> Letting players choose their own sorting scheme will always be better.

That's what the player sorting tab is for. The default sorting alg SHOULD be more weighted to small/medium servers since there is no tab you can easily click to put them at the top. We could add an option menu to configure the penalties/weights each property has, but it's niche and people would still argue what the default values should be. I think this is a could change but perhaps it should be >64 since that's the cap in other Source games - >80 seems a bit arbitrary.

mcNuggets1:
I dont believe this is a great idea.
Servers with high player count usually have a much higher quality standard than other servers.

I just dont believe this isn'th the right approach for balance.
Rather dont penalize servers with high players, but display all servers above a certain player count as EQUAL.

thegrb93:
Score servers by their framerate. The only true metric of whether its terrible or not.

Kefta:
> Servers with high player count usually have a much higher quality standard than other servers.

That is a completely false standard. Servers with more players already have the benefit of appearing closer to the top when sorting by player count which is something many players instinctively do.

mcNuggets1:
> 
> 
> > Servers with high player count usually have a much higher quality standard than other servers.
> 
> That is a completely false standard. Servers with more players already have the benefit of appearing closer to the top when sorting by player count which is something many players instinctively do.

I said "usually".
Servers with bad content obivously have lower player counts, because they earned a negative reputation.

I am hosting servers for 7 years now and the only problem I have with the server browser for me is the gamemodes sorting.
Unpopular gamemodes never will be popular, because they are on the bottom of the list.

Kefta:
> > > Servers with high player count usually have a much higher quality standard than other servers.
> > 
> > 
> > That is a completely false standard. Servers with more players already have the benefit of appearing closer to the top when sorting by player count which is something many players instinctively do.
> 
> I said "usually".
> Servers with bad content obivously have lower player counts, because they earned a negative reputation.
> It's hard to get popularity when having 0 players, but after exceeding like 10 players, it get's a lot easier.
> 
> I am hosting servers for 7 years now and the only problem I have with the server browser for me is the gamemodes sorting.
> Unpopular gamemodes never will be popular, because they are on the bottom of the list.

There are plenty of "bad"/unoptimised/tired servers with high player counts. The marketplace of ideas is not dictated by population density when the entry level is not even. This is what changes in this vein aim to solve.

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> Score servers by their framerate. The only true metric of whether its terrible or not.

It's pretty hard to determine tickrate from a server without fully connecting

ExtReMLapin:
> 
> 
> image
> What it looks like after in the DarkRP section


I remember in 2008 when there was literally 3 french server, one with 24 slots, two with 32.

aStonedPenguin:
So instead of having a couple 128 slot servers the new meta would be to have no less than 8 32 slot servers. This would be incredibly annoying for small and large server owners alike.

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> So instead of having a couple 128 slot servers the new meta would be to have no less than 8 32 slot servers. This would be incredibly annoying for small and large servers alike.

How is this annoying for small servers at all? This greatly benefits them.

thegrb93:
Replace it with an inverted gaussian centered at 16 players and sigma 8. `data.recommended -= 60*exp(pow(16-x, 2)*-0.125)` :)

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> Replace it with an inverted gaussian centered at 16 players and sigma 8. `data.recommended -= 60*exp(pow(16-x, 2)*-0.125)` :)

Thanks, I'll give it a shot.

KingofBeast:
What stops larger communities from splitting up their max cap servers into several smaller servers that cap out right in your arbitrary sweet spot, servers which will without doubt fill up and stay full, probably with even more demand than they do at 100 player caps. Now, rather than the top few listings being "massive servers", the same handful of communities are taking up the top few pages instead?

In another light, that gaussian method is begging you to add a slider above the list for people to choose their own center and sigma :)

thegrb93:
I like the slider idea.

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> What stops larger communities from splitting up their max cap servers into several smaller servers that cap out right in your arbitrary sweet spot, servers which will without doubt fill up and stay full, probably with even more demand than they do at 100 player caps. Now, rather than the top few listings being "massive servers", the same handful of communities are taking up the top few pages instead?
> 
> In another light, that gaussian method is begging you to add a slider above the list for people to choose their own center and sigma :)

It's harder to run 8 smaller servers than one large server and even then I would argue that higher tickrate lower player servers provide a better experience to the players while also giving smaller servers a chance to blossom. As it stands, they stand NO chance right now as shown by the current state of the server list.

karmakarmakarmakarmakarmakarmakarma:
> 
> 
> I like the slider idea.

Yeah this would be a nice feature, I would have to learn more about HTML and templating to get it working though

aStonedPenguin:
> How is this annoying for small servers at all? This greatly benefits them.

A large community turns their 2 servers into 8 to get that sweet sweet favorable ranking. How many additional servers do you have to scroll past to find one of these small servers?

GitSparTV:
That would better for playability though. :vee:

MrDiaboloz:
So the goal is to show to player random darkrp full gmodstore leak workshop server at 20-25 player instead of the bigger server which work to optimize and run 128 slots/player decently?

ExtReMLapin:
> random darkrp full gmodstore leak workshop server at 20-25 player 

imagine paying for lua lmao

>instead of the bigger server which work to optimize and run 128 slots/player decently

Implying gmod server owner know anything about optimizing

Remove all the physical entities and the whole server will be so much faster and even if it's not, the only thing the high ping will affect are moving players and doors.

Source : hl2rp clockwork

MrDiaboloz:
> Implying gmod server owner know anything about optimizing
> 
> Remove all the physical entities and the whole server will be so much faster and even if it's not, the only thing the high ping will affect are moving players and doors.

Balancing gameplay & fun at 128 players is not just delete some props it's hundreds of hours to dev whole custom infrastructure to bring the best experience.

And optimizing isn't just run any random lua scanner to find hooks

robotboy655:
As discussed above, this will essentially backfire if the bigger servers will start to split their servers to multiple smaller ones pushing your new and small servers even further down.

Kefta:
They should split, higher tickrates on the most popular servers would be nice.

ExtReMLapin:
Didn't garry backdoor gmod with chinese statistic garbage ? Couldn't this stastic stuff be used to report the "measured" tickrate of the current server to a master server and then rank the server based on the said reports.

I wouldn't trust a server self-reporting it's tickrate

PR #1774, opened 2021-03-03
Update .gitignore to allow players to use their Garry's Mod install to suggest changes easier.

meepen:
This also adds some untracked files and adds protections for cfg files to not be pushed.


robotboy655:
What's up with the 2 added files?

robotboy655 (approved):

PR #1775, opened 2021-03-05
Fix theme issues with DMenuOptions foreground text color

CapsAdmin:
This removes the call to SetTextColor in DMenuOption since it will override theme colors here:
https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/dlabel.lua#L68

Before: 
image

After: 
image

Left side is the default skin and right side is this skin https://steamcommunity.com/sharedfiles/filedetails/?id=1596913191

You can also see that the default skin seems more accurate according to its design.


DBotThePony:
Actually fixes https://github.com/Facepunch/garrysmod-issues/issues/2605

PR #1783, opened 2021-03-26
code smell

7aninhos:
it calls vector on every frame


thegrb93:
Name the local better

robotboy655 (approved):

PR #1784, opened 2021-03-26
Add GetConVarCache

garryspins:
Pretty self explanatory on how this would be useful

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -430,7 +430,7 @@ end
 --[[---------------------------------------------------------
 	Convar access functions
 -----------------------------------------------------------]]
-
+-
 local ConVarCache = {}
 
 function GetConVar( name )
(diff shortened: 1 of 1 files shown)


Kefta:
It's not self-explanatory, the cache is an implementation detail. Why do you need this?

garryspins:
since theres no built-in way of getting a table of convars this seemed like the lightest way to get a semi-complete list of them without detouring

this is 100% useless if a method of getting every convar is added

Kefta:
> since theres no built-in way of getting a table of convars this seemed like the lightest way to get a semi-complete list of them without detouring
> 
> this is 100% useless if a method of getting every convar is added

Not only does this not contain every convar (and constantly changes), I don't see a reason why you'd need such a list. What use is having all convars?

robotboy655:
https://github.com/Facepunch/garrysmod-requests/issues/341

This is not the answer.

PR #1785, opened 2021-04-01
Replace backslashes with forward slashes in Hammer Sprinkle Example

Jesse-Culver:
Using backslashes instead of forward slashes in a sprinkle script works fine until the map is reloaded at which point Hammer can not load the model due to the path truncating to just "models". I added a warning comment in as well as this is an example file others will work off of.


robotboy655:
I fixed in Hammer itself so it replaces backslashes for models when using Sprinkle tool, which I think it's a better solution so people don't have to worry about this.

PR #1790, opened 2021-04-21
New askconnect dialog

WilliamVenner:
https://i.venner.io/TgAwZfl7BM.mp4

https://i.venner.io/PmpMr7hKev.mp4

### DIFF
--- a/garrysmod/lua/menu/askconnect.lua
+++ b/garrysmod/lua/menu/askconnect.lua
@@ -1,0 +1,202 @@
+local ASK_CONNECT_WAIT_TIME = 5
+
+local MAT_GMOD_LOGO = Material( "../html/img/gmod_logo_brave.png", "smooth" )
+local MAT_BACKGROUND = Material( "../html/img/bg.png", "smooth" )
+
+local ASK_CONNECT_HEIGHT = 340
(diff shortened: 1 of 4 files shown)


NoSharp:
Looks really cool!

TomDotBat:
Much prefer this solution over the current :)

shadow7483147:
Hard pass. Client connects should never be automatic. The askconnect dialog was added for this exact reason, because servers were abusing automatic connections.

WardenPotato:
My main concern regarding this is that if the redirect is automatic it will get abused by larger servers to put all their afks in their less popular servers, which is unfair to other communities and misleading in the server browser

WilliamVenner:
How about pausing the timer if the mouse hasn't moved since the dialog has been opened?

Some kind of hack would have to be implemented to stop input.SetCursorPos and stuff from triggering it.

robotboy655:
As already mentioned, this defeats the purpose of the askconnect dialog. The only solution to the "IT RUINS EVERYTHING!!!11" argument is ability for the player to consciously save their choice for future popups from the same server.

WilliamVenner:
Was hoping to have an actual discussion, but alright.

MrCreamie:
hot, like always billy.

PR #1793, opened 2021-04-24
Call net.Receive "editvariable" once outside InstallDataTable

TIMONz1535:
**util.AddNetworkString( "editvariable" )** and **net.Receive( "editvariable", ...** called several times inside **meta:InstallDataTable()**.
It's wrong because all entities use the same net.Receive. I just moved it shared function outside.

p.s. Github shows a complex diff due to indentation changes, but I didn't change the code itself.


TIMONz1535:
I tested this case, indeed it is called several times

```lua
	...
	if ( SERVER ) then

		print("AddNetworkString editvariable", self, "good work 47")
		local callerInfo = debug.getinfo(1, "lnS")
		print("", callerInfo.name, "-", callerInfo.short_src .. ":" .. callerInfo.currentline)
		local callerInfo = debug.getinfo(2, "lnS")
		print("", callerInfo.name, "-", callerInfo.short_src .. ":" .. callerInfo.currentline)

		util.AddNetworkString( "editvariable" )

		net.Receive( "editvariable", function( len, client )
	...
```

Sandbox:
изображение

PR #1794, opened 2021-04-24
Added Whitelist Feature for Server connects/hopping (via LUA)

catSIXe:
See https://github.com/Facepunch/garrysmod-requests/issues/1751

We would only need to optimize/handle the saving of the whitelist, to be write protected against malicious code writing to it.
But it should be clear (according to the Server Operator Rules) to not write to this file. So its not a must-have to merge this.

(Maybe a Path where only the `MenuLua` can write to?).

It adds a Checkbox to the connect requests, and if you check it. It will save the IP/Domain, to never ask again.
(everything can be cleared with the concmd `connect_clearwhitelist`).
  
IPv4 Address will be matched exactly, where as Domain Names will only match to the `Domain.TLD`, so you can have multiple Servers under "yourdomain.tld" like "server1.yourdomain.tld" and "server2.yourdomain.tld".. Therefore it will only ask you once if you want to allow "yourdomain.tld".
  
Does ask you to whitelist the entire domain:
image

Does match the whitelist Entries accordingly(no trickery possible):
image

Also works for IPs:
image

If whitelisted, it just directly connects.

### DIFF
--- a/garrysmod/lua/menu/openurl.lua
+++ b/garrysmod/lua/menu/openurl.lua
@@ -39,6 +39,105 @@ function GMOD_OpenURLNoOverlay( url )
 end
 
 ----------------------------------------------
+-- patch by @catSIXe - can be used in orig. gmod code if wanted
+----------------------------------------------
+
+-- better code, but we should try to save this somewhere where only the menu context can write to(read should be no problem lol)
+local _whitelistFilePath = "_connectwhitelist.json"
+local _whitelistDir = "DATA"
+local function parseAddress(serverip)
+	local Host, Port = "", 27015
+	if string.find(serverip, ":") then
+		local parts = string.Split(serverip, ":")
+		if #parts != 2 then return -1 end
+		Host = parts[1]
+		Port = tonumber(parts[2])
+	else
+		Host = serverip
+	end
+
+	local a, b, c, d = Host:match("^(%d%d?%d?)%.(%d%d?%d?)%.(%d%d?%d?)%.(%d%d?%d?)$")
+	a = tonumber(a)
+	b = tonumber(b)
+	c = tonumber(c)
+	d = tonumber(d)
+	if not a or not b or not c or not d then return 0, Host, Port end
+	if a<0 or 255<a then return 0, Host, Port end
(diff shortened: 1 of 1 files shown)


ExtReMLapin:
>We would **only** need to [...] be write protected against malicious code writing to it.

lol

catSIXe:
> > We would **only** need to [...] be write protected against malicious code writing to it.
> 
> lol

Yeah, rn. its a file which every game lua code can write to. Basicly you could write your domain to it, and then prevent the dialog from asking again... But there are the Server Operator Rules and such behaivour is not allowed in general. 
And there is not really a danger from joining a server. 

Sooo... its prob not necessary, but would be nice to have.

ExtReMLapin:
>But there are the Server Operator Rules and such behaivour is not allowed in general [...] sooo... its prob not necessary,

I'm pretty sure gmod isn't canada

garrysmod/lua/menu/openurl.lua
```diff
@@ -38,6 +38,100 @@ function GMOD_OpenURLNoOverlay( url )
...
 end
 
+----------------------------------------------
+-- patch by @catSIXe - can be used in orig. gmod code if wanted
+----------------------------------------------
+
+-- better code, but we should try to save this somewhere where only the menu context can write to(read should be no problem lol)
+local _whitelistFilePath = "_connectwhitelist.json"
+local _whitelistDir = "DATA"
+local function parseAddress(serverip)
```
swampservers, line 48: your code is cringe, make it more like this
```
local function ParseAddress(ip)
    local host, port = unpack(ip:Split(":"))
    port = tonumber(port) or 27015

    if port ~= math.Clamp(math.floor(port), 1, 65535)) then return end

    local ipbytes = host:Split(".")
    if #ipbytes ~= 4 then return end
    for _,v in ipairs(ipbytes) do
        v = tonumber(v)
        if v ~= math.Clamp(v or 0, 0, 255) then return end
    end

    return host, port
end
```

ExtReMLapin: i hope that was satire

robotboy655:
This was added into the game with a better, more future proof system.

PR #1795, opened 2021-05-01
Adding an animation tab

OfficialFreline:
New additional tab "Animations", which allows you to run certain animations on any model.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
@@ -121,7 +121,7 @@ list.Set( "DesktopWindows", "PlayerEditor", {
 		local animcontrolspanel = animPanel:Add( "DScrollPanel" )
 		animcontrolspanel:Dock( FILL )
 
-		for v, n in pairs( LocalPlayer():GetSequenceList() ) do
+		for v, n in ipairs( LocalPlayer():GetSequenceList() ) do
 			local an = animcontrolspanel:Add( "DButton" )
 			an:Dock( TOP )
 			an:SetText( n )
--- a/garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
@@ -129,7 +129,7 @@ list.Set( "DesktopWindows", "PlayerEditor", {
 			"pose_standing_02",
 		}
 
-		for v, n in pairs( animsTable ) do
+		for v, n in ipairs( animsTable ) do
 			local an = animcontrolspanel:Add( "DButton" )
 			an:Dock( TOP )
 			an:SetText( n )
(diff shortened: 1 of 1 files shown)


OfficialFreline:
Screenshot: https://i.imgur.com/sjzdVq9.jpeg

garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
```diff
@@ -115,6 +115,33 @@ list.Set( "DesktopWindows", "PlayerEditor", {
...
+
+		local animsTable = {
+			"idle_all_scared",
+			"menu_gman",
+			"menu_zombie_01",
+			"menu_combine",
+			"pose_standing_02",
+		}
+
+		for v, n in pairs( animsTable ) do
```
FlorianLeChat, line 132: You can use ipairs here to be more efficient.

OfficialFreline: > You can use ipairs here to be more efficient.

Thank you

mcNuggets1:
Very cool and innovative.
But wouldn't there be a way to retrieve all working animations and make this far more extensive?

garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
ExtReMLapin: You're wrong @FlorianLeChat 
It's not more efficent here, as ipairs is only faster when JIT-compiled, here is not going to be JIT-compiled
On literally each line there is a NYI(closure)/C api call.
When not JIT-compiled, `ipairs` is 3.29 times slower than `pairs`


It is WRONG to use pairs, but not slower.

ExtReMLapin:
Didn't gmod already have in the icon editor an existing sequence lister with preview ?
You could also use `ent:GetSequenceList()` instead of hardcoding the sequences strings.

I see no point of this PR as you can already see the animations in the icon editor.
Playermodel menu is meant to be used by players, there have no need to see this kind of menu, icon editor is meant to be used by "advanced users".

garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
FlorianLeChat: Good point. I didn't notice that.

OfficialFreline:
> Didn't gmod already have in the icon editor an existing sequence lister with preview ?
> You could also use `ent:GetSequenceList()` instead of hardcoding the sequences strings.
> 
> I see no point of this PR as you can already see the animations in the icon editor.
> Playermodel menu is meant to be used by players, there have no need to see this kind of menu, icon editor is meant to be used by "advanced users".

This addition will allow the same players who make the models to fully finish and test their model. Also, weapon makers wouldn't mind using this either.

OfficialFreline:
Demonstration: https://youtu.be/u_Wb5003Mo0

ExtReMLapin:
why did you move back to `pairs` are you even reading

OfficialFreline:
> why did you move back to `pairs` are you even reading

Small banter) The world is not without fun

OfficialFreline:
And yes, I spoke to some content creators today. The idea of ​​the tab was appreciated

ExtReMLapin:
already exists for fuck sake

OfficialFreline:
> 
> already exists for fuck sake

This is a little different

PR #1796, opened 2021-05-05
Optimize player:__index

cannon:
If the value in the entity table is nil, it would return nil anyway, so we don't need to check it.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -27,8 +27,7 @@ function meta:__index( key )
 	--
 	local tab = entity.GetTable( self )
 	if ( tab ) then
-		local val = tab[ key ]
-		if ( val != nil ) then return val end
+		return tab[ key ]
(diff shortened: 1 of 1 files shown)


ExtReMLapin:
Please don't post this kind of PR just to get your name in the contributors list.

And in Lua, returning nil and nothing isn't exactly the same.
Even if here I doubt it breaks anything.

Kefta:
> And in Lua, returning nil and nothing isn't exactly the same.

That has nothing to do with this PR, he's just removing a useless operation from an often called function.

thegrb93 (approved):

sanny-io (approved):

WilliamVenner (approved):

ceifa (approved):

robotboy655 (approved):

ceifa:
> Please don't post this kind of PR just to get your name in the contributors list.

Please don't post this kind of comment just to be pedantic.

PR #1797, opened 2021-05-07
Allow for color arg to be nil on RoundedBoxEx

Down-s:
Having to create a Color metatable every time you use a RoundedBox is a pain.
Adding this will allow the `color` argument to be nil, enabling people to use `surface.SetDrawColor` instead.

Example:
```
surface.SetDrawColor(0, 0, 0)
draw.RoundedBox(16, 50, 50, 200, 200)
```
image
Not a single `Color` object to deal with!

### DIFF
--- a/garrysmod/lua/includes/modules/draw.lua
+++ b/garrysmod/lua/includes/modules/draw.lua
@@ -179,7 +179,9 @@ end
 -----------------------------------------------------------]]
 function RoundedBoxEx( bordersize, x, y, w, h, color, tl, tr, bl, br )
 
-	surface.SetDrawColor( color.r, color.g, color.b, color.a )
+	if ( color ) then
+		surface.SetDrawColor( color.r, color.g, color.b, color.a )
+	end
 
 	-- Do not waste performance if they don't want rounded corners
 	if ( bordersize <= 0 ) then


NoSharp:
Pointless.

Down-s:
> Just don't create color every time?

That's not always an option, what if you wanted to have a flashing color? You would be forced to do `Color(math.sin(CurTime(), 0, 0)`, creating a color every frame.

NoSharp:
There are many ways for you to do this in a more optimized fashion. (Render targets if you really want to)
From a few preliminary benchmarks, this actually slower, baring in mind this is a fairly basic benchmark and it might be flawed, if so point it out.
```lua
local benchmark = os.clock()

for i = 1, 2e5 do
    draw.RoundedBox(8,0,0,100,100,Color(i % 255, i % 255, i % 255))
end
print("Benchmark 1): ",os.clock() - benchmark)

benchmark = os.clock()

for i = 1, 2e5 do
    surface.SetDrawColor(i % 255, i % 255, i % 255)
    draw.RoundedBoxChanged(8,0,0,100,100)
end
print("Benchmark 2) (Updating Color each frame):",os.clock() - benchmark)

benchmark = os.clock()
surface.SetDrawColor(255, 255, 255)
for i = 1, 2e5 do
    draw.RoundedBoxChanged(8,0,0,100,100)
end
print("Benchmark 3) (Updating Color once before the loop.):", os.clock() - benchmark)

benchmark = os.clock()
surface.SetDrawColor(255, 255, 255)
for i = 1, 2e5 do
    draw.RoundedBoxChanged(8,0,0,100,100,Color(i % 255, i % 255, i % 255))
end

print("Benchmark 4) (Original benchmark ran on the new function):", os.clock() - benchmark) 

benchmark = os.clock()
for i = 1, 2e5 do
end
print("Benchmark 5) (Baseline):", os.clock() - benchmark) 
```
Output: 
```
Benchmark 1): 	2.3580000000001
Benchmark 2) (Updating Color each frame):	4.1079999999999
Benchmark 3) (Updating Color once before the loop.):	4.116
Benchmark 4) (Original benchmark ran on the new function):	4.284
Benchmark 5) (Baseline):	0
```
So it can be seen this is wildy inefficient.

GitSparTV:
Don't forget the margin error.

GitSparTV:
Also, using os.clock is cringe.
Use SysTime.

NoSharp:
> Also, using os.clock is cringe.
> Use SysTime.
Thanks :)

TomDotBat:
Not worth the performance hit for a "feature" this minor.

Down-s:
image
image
Results from my own benchmark ^

TomDotBat:
> image
> image
> Results from my own benchmark ^

You haven't benchmarked using a colour object with the new function. You've got to remember that this function will be called hundreds of times per frame with a colour object still.

thegrb93:
You could make your own helper draw function instead of editing this one.

aStonedPenguin:
> 
> 
> > Just don't create color every time?
> 
> That's not always an option, what if you wanted to have a flashing color? You would be forced to do `Color(math.sin(CurTime(), 0, 0)`, creating a color every frame.

`color.r = math.sin(CurTime()`

PR #1799, opened 2021-05-22
[TTT] Changes to the Brazilian Portuguese translation

Tiagoquix:
I also changed "Português (Brasil)" to "Português brasileiro" because the letter B of "Brasil" doesn't remain uppercase in-game.

@svdm Also, I want to report a bug here: letters with accents aren't uppercased in the round's end highlights. They aren't treated like normal letters.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/brazilian_portuguese.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/brazilian_portuguese.lua
@@ -1,6 +1,6 @@
 ---- Brazilian Portuguese language strings
 
-local L = LANG.CreateLanguage("Português brasileiro")
+local L = LANG.CreateLanguage("Português (Brasil)")
 
 --- General text used in various places
 L.traitor    = "Traidor"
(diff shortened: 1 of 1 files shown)


svdm:
I have just fixed the issue with the language name in the menu, so you can change that back to "(Brasil)".

The problem with the uppercasing in the highlights is difficult to fix, it's a Lua limitation to an extent. One workaround would be to make the translations uppercase for all the `aw_*_title` keys. For example: `L.aw_nkt5_title = "OPERAÇÃO CONTRA-TERRORISTA"`.

Tiagoquix:
@svdm Done. And thanks for the fix.

Kefta:
> I have just fixed the issue with the language name in the menu, so you can change that back to "(Brasil)".
> 
> The problem with the uppercasing in the highlights is difficult to fix, it's a Lua limitation to an extent. One workaround would be to make the translations uppercase for all the `aw_*_title` keys. For example: `L.aw_nkt5_title = "OPERAÇÃO CONTRA-TERRORISTA"`.

You can use the utf8 library.

svdm:
If there was a `utf8.upper` I would, but implementing that from scratch seems like too much work compared to capitalising a few translations.

FlorianLeChat:
The [Nutscript](https://github.com/NutScript/NutScript/blob/1.2.2-wip/gamemode/core/libs/thirdparty/sh_utf8.lua)/[Helix](https://github.com/NebulousCloud/helix/blob/master/gamemode/core/libs/thirdparty/sh_utf8.lua) gamemodes have already implemented this functionality with a public library. Perhaps you could do the same. But it would be cool if in the future this kind of stuff is directly included in the Gmod UTF8 library.

Kefta:
> If there was a `utf8.upper` I would, but implementing that from scratch seems like too much work compared to capitalising a few translations.

Oops, I thought that function already existed.

PR #1805, opened 2021-06-09
Tool ghost entity improvements

Down-s:
I have fixed the jittering ghost entities for tools when you move them about, this happens because it's being updated in a TOOL:Think function, which isn't called at the client's FPS.
I have moved the update ghost entity function for each tool into a TOOL:DrawHUD function, which is called every frame.
I have also made balloons align with the surface normal, which feels more natural.

Before:
https://user-images.githubusercontent.com/53242610/121425487-5da8ec00-c96a-11eb-89f6-5ae62ad4ab38.mp4

After:
https://user-images.githubusercontent.com/53242610/121425559-731e1600-c96a-11eb-84ee-0a4c8ea8d8a1.mp4

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/balloon.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/balloon.lua
@@ -16,6 +16,8 @@ TOOL.Information = {
 
 cleanup.Register( "balloons" )
 
+local hitangle_offset = Angle( 90, 0, 0 )
+
 function TOOL:LeftClick( trace, attach )
 
 	if ( IsValid( trace.Entity ) && trace.Entity:IsPlayer() ) then return false end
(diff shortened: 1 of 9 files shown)


garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/balloon.lua
```diff
@@ -84,8 +86,10 @@ function TOOL:LeftClick( trace, attach )
 	local Offset = CurPos - NearestPoint
 
 	local Pos = trace.HitPos + Offset
+	local Ang = trace.HitNormal:Angle() + hitangle_offset
```
thegrb93, line 89: If you want to 'add' angles, you should use `local _, Ang = LocalToWorld(Vector(), hitangle_offset, Vector(), trace.HitNormal:Angle())`

thegrb93 (changes requested):

Kefta:
This breaks the ghost in single-player due to it existing as a serverside entity, not a clientside one.

Down-s:
> This breaks the ghost in single-player due to it existing as a serverside entity, not a clientside one.

Ah you are right, is there any reason it's a serverside entity when in singleplayer?

Kefta:
It could have been a choice made before Garry added the hack to Think to be called shared in single-player, or maybe some CSEnt behaviour was broken in SP at one point. Either way it's too late to change now since some custom tools access the variable directly in a specific realm. Also, SWEP:DrawHUD is not going to be called if cl_drawhud is 0. You could instead use a Think hook which I believe is called every frame in single-player as opposed to server ticks.

thegrb93:
I wouldn't use self as the key since that has a higher chance of conflicting, but that is a valid solution.

robotboy655:
Closing due to lack of activity.

PR #1807, opened 2021-06-14
New serverlist ranking system

Synkstar:
A ranking system based on which server has the most players but with some randomness to give smaller servers a chance because gmod players like to join populated servers and servers want to get players.
The current ranking system is biased towards servers with enterprise/ datacenter internet connections with optimized routes etc. 

I also made it so if a server has more than 150 ping then it would be ranked lower to avoid servers with really bad latency.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -334,7 +334,7 @@ function FormatVersion( ver )
 function getRandomInt(min, max) {
 	min = Math.ceil(min);
 	max = Math.floor(max);
-	return Math.floor(Math.random() * (max - min) + min); //The maximum is exclusive and the minimum is inclusive
+	return Math.floor(Math.random() * (max - min) + min); 
   }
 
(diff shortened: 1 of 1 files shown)


YohSambre:
the idea is good but it will probably never be implemented

Synkstar:
I thought for a sec before that the server list ranking system was different in 2015. But its just really the gmod player population. Most of the time there is more servers than players running so this probably honestly wouldn't really matter. Back in 2015 and before your server was considered dead when it had an average of 10 people. I would pull an average of like 20 on a 1 core laptop( no advertising or anything  and thats on sandbox). Now yeah I might have a server that can run 40-100 players depending on the tickrate used etc but it never gets higher than 10 and is usually empty. I only really run sandbox though atm and the player population in that gamemode is even worse then some other ones ... https://www.battlemetrics.com/servers/gmod/stats  just look at that

robotboy655:
I do not see a point in making the server ranking random. If people want to join populated servers, they can sort by player count.

PR #1810, opened 2021-06-18
 Fix hook.Add with invalid types causing lua errors

Leystryku:
since all the other parameters are validated, the name should be validated too


GitSparTV:
This is used not just for entities, also for panels.

thegrb93:
Any table with an IsValid function

Leystryku:
I wonder whether this can be done more elegantly, I think I got a simple idea

Leystryku:
> This is used not just for entities, also for panels.



> Any table with an IsValid function

Both concerns should be addressed now

GitSparTV:
Isn't IsValid errors if you pass numbers?

thegrb93:
Error is good

GitSparTV:
Ok I would prefer :IsValid method rather than global version, anyway

Kefta:
```
if (!isstring(name) && !pcall(function(v) return v.IsValid end, name)) then
```

Obviously cache the pcall func in the file. This will not prevent errors with improperly implemented IsValid methods, but it's not feasible to properly test as Lua has no sense of function contracts.

GitSparTV:
Pepega. Let's bloat JIT even more

thegrb93:
That's disgusting, Kefta. No

Kefta:
The alternative is not much better.
```Lua
if (!isstring(name)) then
   if (istable(name)) then
      if (!isfunction(name.IsValid)) then
         -- ErrorNoHaltWithStack
         return
      end
   else
      local meta = getmetatable(name)

      if (!istable(meta)) then
         -- ErrorNoHaltWithStack
         return
      end

      local index = meta.__index

      if ((!istable(index) && !isfunction(index)) || !isfunction(name.IsValid)) then
         -- ErrorNoHaltWithStack
         return
      end
   end
end
```

The downsides of this approach:
1. Much more bloated, confusing code regardless of how many comments you add.
2. Errors thrown by __index are not caught, so improperly implemented __index methods will still throw constant errors if an object of that type is passed as the identifier.

The first is really not that "disgusting" in comparison, and if either solution of properly checking for the presence of IsValid is not taken, then there is no point in this PR:  it'd just be wholly incomplete coverage.

> Pepega. Let's bloat JIT even more

That doesn't matter in this case since hook.Add/Remove should never be called during active runtime and should contain the slowest operations in order to make hook.Call as fast as possible. Every hook lib rewrite worth discussing has taken this philosophy to varying degrees.

Kefta:
Both solutions are still a regression in the case that obj.IsValid is not a function but an object with a __call metamethod, but that can easily be fixed.

thegrb93:
Just let it error if something bad goes into it. You don't need to treat lua like E2.

thegrb93:
The only purpose of this is to prevent bad stuff from going into the hook.Call loop

thegrb93:
that said, I would change the check to `isbool(name:IsValid())` since something invalid should technically be allowed.

thegrb93:
Scratch that last comment I deleted.

Kefta:
> The only purpose of this is to prevent bad stuff from going into the hook.Call loop

Which is what the solutions are doing. It's making sure an object with no IsValid method doesn't go into the hook.Call loop.

thegrb93:
Just call it. name:IsValid()

Kefta:
> Just call it. name:IsValid()

That will error for numbers, functions, threads, bools, nil, tables without an IsValid method, and usersata without an __index or IsValid method. I've provided two solutions that make sure no value you pass will error, the first of which is concise and easily readable. The only arguments against it thus far are that it is not JITted which I've already stated why that doesn't really matter in this case, and that it looks "disgusting" which I entirely disagree. Even if it did, all solutions that don't properly check an object's IsValid method will just lead to user confusion and regression. For example,

```Lua
hook.Add("HookThatWillNotBeCalled", 5, function() end)
```

With the two solutions I provided, this will throw a meaningful, non-halting error. With yours, this will throw a halting error about attempting to index a number which will cause the rest of the file to not be executed.

If type checking is going to be done at all here, it should be done properly.

thegrb93:
> If type checking is going to be done at all here, it should be done properly.

None of the other lua functions in gmod waste time doing this. Waste of cycles and time. Quit being pedantic

Kefta:
> > If type checking is going to be done at all here, it should be done properly.
> 
> None of the other lua functions in gmod waste time doing this. Waste of cycles and time. Quit being pedantic

Then this PR should be closed by that logic. Not that the cycles matter since these are not functions being called in active runtime.

thegrb93:
Yes they are, many mods do. And no, this will made debugging such a mishap very easy even if the typecheck errors.

thegrb93 (approved):

Kefta (changes requested):
Prevent errors when name is a non-string object that cannot be indexed

Leystryku:
The point of this PR is to solve hook.Call literally spamming errors and freezing servers when someone adds a hook with a number as name. Plus it erroring in an entirely different place making it a pain to debug. Something I have had to do for one of my customers. This fixes the issue and does so in a concise manner as well.

image

The point of this PR is not to argue about whether we should wrap every single function in Garry's Mod library in loads of validation, or whether we should try to set up a contract-esque system in GLua.

Put simply, as @thegrb93 said, it's just being pedantic and overlooking the point of this bug fix.

GitSparTV:
I think you just need `!isnumber` 🧠

Kefta:
> I think you just need `!isnumber` 🧠

```Lua
hook.Add("foo", function() end, function() end)
```

Or probably more common in the wild, accidentally omitting the second arg for just the function and having it give an unhelpful error.

PR #1812, opened 2021-06-25
A few string lib optimizations

thegrb93:


mcNuggets1:
Didn't know string.gsub was faster than string:gsub

TomDotBat:
yeah can we see some benchmarks on that?

Kefta:
> yeah can we see some benchmarks on that?

It's intuitive. str:x will call __index then the method as opposed to str.x which accesses the method directly (though both do string table lookups every time as opposed to localising which is even more efficient). You can see this difference clearly in the output bytecode, even with JIT. It's too few cycles for benchmarks to be as useful of a tool in measuring gain, but I'm sure at enough indexes it will become noticeable enough.

These changes are all included in https://github.com/Facepunch/garrysmod/pull/1407 too.

thegrb93:
@Kefta only thing missing is the regex updates here. I have them in starfall at least until whenever devs decide to accept PR code

PR #1816, opened 2021-07-07
Fix gmod_camera being dependent on tickrate

DBotThePony:
This bothers me for insanely long time

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_camera.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_camera.lua
@@ -108,8 +108,8 @@ function SWEP:Tick()
 
 	if ( !cmd:KeyDown( IN_ATTACK2 ) ) then return end -- Not holding Mouse 2, bail
 
-	self:SetZoom( math.Clamp( self:GetZoom() + cmd:GetMouseY() * 0.1, 0.1, 175 ) ) -- Handles zooming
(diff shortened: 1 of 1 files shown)


thegrb93:
Multiply the existing magic numbers with 66. Would look better

thegrb93:
`* 6.6`
`* 1.65`

DBotThePony:
If you are so nervous about performance of Division vs Multiplication even in non performance critical function then ok

robotboy655:
It's less about performance and more about readability/"niceness" of code.

GitSparTV:
Not to mention, such expressions are evaluated at compile time, so it doesn't matter

robotboy655:
Also is it really necessary to replace tabs with spaces for no reason?

thegrb93:
> Not to mention, such expressions are evaluated at compile time, so it doesn't matter

Lua doesn't optimize division by constant unless you do `* (1/constant)`

GitSparTV:
> `* 6.6`
> 
> `* 1.65`

Ah I thought you mentioned this because the code had `smth * 6.6 * 1.65`.
Nvm then

DBotThePony:
> Also is it really necessary to replace tabs with spaces for no reason?

because it is done automatically on save to avoid obvious github editor and other editors issues when tab width is not 4 spaces (which is obviously the case with Garry's code editor preferences)

Kefta:
I think the constants should be multiplied by 66 to conserve their initial values.

thegrb93 (approved):

PR #1817, opened 2021-07-08
Create util.BinaryToString and util.StringToBinary

Down-s:
## **String -> Binary -> String**
```lua
local bin = util.StringToBinary("string 2 bin")
print("Binary: " .. bin)
print("Back to string: " .. util.BinaryToString(bin))
```

Output:
```
Binary: 01101110 01101001 01100010 00100000 00110010 00100000 01100111 01101110 01101001 01110010 01110100 01110011
Back to string: string 2 bin
```

## **Binary-> String -> Binary**
```lua
local str = util.BinaryToString("01100111 01101110 01101001 01110010 01110100 01110011 00100000 01101101 01101111 01110010 01100110 00100000 01101001 01001000")
print("String: " .. str)
print("Back to binary: " .. util.StringToBinary(str))
```

Output:
```
String: Hi from string
Back to binary: 01100111 01101110 01101001 01110010 01110100 01110011 00100000 01101101 01101111 01110010 01100110 00100000 01101001 01001000
```

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -365,3 +365,53 @@ function util.RemovePData( steamid, name )
 	sql.Query( "DELETE FROM playerpdata WHERE infoid = " .. SQLStr( name ) )
 
 end
+
+--[[---------------------------------------------------------
+	Name: StringToBinary( str )
+	Desc: Converts a string to a binary string
+-----------------------------------------------------------]]
+
+function util.StringToBinary( str )
+
+	local bin = ""
+	str = str:reverse()
+
+	for i = 1, str:len() do
+		local num = str[ i ]:byte()
+		for j = 7, 0, -1 do -- 8 bits in each char
+			bin = bin .. math.floor( num / 2 ^ j )
+			num = num % 2 ^ j
+		end
+
+		bin = bin .. " "
+	end
+
+	return bin:TrimRight()
+
+end
+
+--[[---------------------------------------------------------
+	Name: BinaryToString( binary )
+	Desc: Converts a binary string to a readable string
+-----------------------------------------------------------]]
+
+function util.BinaryToString( binary )
+
+	binary = binary:reverse()
+	local str = ""
+
+	for byte in binary:gmatch("%S+") do
+		local dec = 0
+
+		for i = 1, 8 do
+			if ( byte[ i ] == "1" ) then
+				dec = dec + math.pow( 2, i - 1 )
+			end
+		end
+
+		str = str .. string.char( dec )
+	end
+
+	return str
+
+end
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -373,20 +373,19 @@ end
 
 function util.StringToBinary( str )
 
-	local bin = ""
-	str = str:reverse()
+	local bin = {}
 
-	for i = 1, str:len() do
-		local num = str[ i ]:byte()
+	for i = str:len(), 1, -1 do
+		local num = str:byte(i)
 		for j = 7, 0, -1 do -- 8 bits in each char
-			bin = bin .. math.floor( num / 2 ^ j )
+			bin[#bin + 1] = math.floor( num / 2 ^ j )
 			num = num % 2 ^ j
 		end
 
-		bin = bin .. " "
+		bin[#bin + 1] = " "
 	end
 
-	return bin:TrimRight()
+	return table.concat(bin)
 
 end
 
@@ -397,21 +396,20 @@ end
 
 function util.BinaryToString( binary )
 
-	binary = binary:reverse()
-	local str = ""
+	local str = {}
 
 	for byte in binary:gmatch("%S+") do
 		local dec = 0
 
 		for i = 1, 8 do
-			if ( byte[ i ] == "1" ) then
-				dec = dec + math.pow( 2, i - 1 )
+			if ( byte[ 9 - i ] == "1" ) then
+				dec = dec + 2 ^ (i - 1)
 			end
 		end
 
-		str = str .. string.char( dec )
+		str[#str + 1] = string.char( dec )
 	end
 
-	return str
+	return table.concat(str):reverse()
 
 end


garrysmod/lua/includes/extensions/util.lua
```diff
@@ -365,3 +365,53 @@ function util.RemovePData( steamid, name )
...
+	str = str:reverse()
+
+	for i = 1, str:len() do
+		local num = str[ i ]:byte()
+		for j = 7, 0, -1 do -- 8 bits in each char
+			bin = bin .. math.floor( num / 2 ^ j )
+			num = num % 2 ^ j
+		end
+
+		bin = bin .. " "
```
WilliamVenner, line 386: I believe using a table to build strings is faster than repeated concatenation(?)

```lua
local str = {}
str[#str + 1] = "..."
...
str = table.concat(str)
```

CornerPin:
What is the use case?
There are several problems with this implementation:
- Concatenating multiple strings in a for loop with the concatenation operator. In this case, filling a table and using `table.concat` will be much more efficient in terms of memory usage and CPU time.
- No need to reverse the string.
- `string.byte` has a `startPos` parameter. Indexing the string is unnecessary.
- Is using math operations faster than using the `bit` library?

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -365,3 +365,53 @@ function util.RemovePData( steamid, name )
...
+
+--[[---------------------------------------------------------
+	Name: StringToBinary( str )
+	Desc: Converts a string to a binary string
+-----------------------------------------------------------]]
+
+function util.StringToBinary( str )
+
+	local bin = ""
+	str = str:reverse()
```
WilliamVenner, line 377: Why not just iterate backwards?

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -365,3 +365,53 @@ function util.RemovePData( steamid, name )
...
+
+	binary = binary:reverse()
+	local str = ""
+
+	for byte in binary:gmatch("%S+") do
+		local dec = 0
+
+		for i = 1, 8 do
+			if ( byte[ i ] == "1" ) then
+				dec = dec + math.pow( 2, i - 1 )
```
WilliamVenner, line 408: Why not just `2 ^ (i - 1)`?

TomDotBat:
Adding spaces between each byte seems a bit unnecessary. If you look at the implementation of `util.Base64Encode()` we don't space things out there, so it only makes sense to keep without spacing - to me at least.

garrysmod/lua/includes/extensions/util.lua
Down-s: What's the difference?

thegrb93:
I don't see any use case for this function

GitSparTV:
Lol, you still didn't remember table.concat, you had this problem few weeks ago with the same concat loop

GitSparTV:
Also useless non-efficient function.
I personally only used byte concatenation but I would never suggest it to gmod codebase though

garrysmod/lua/includes/extensions/util.lua
GitSparTV: Function call overhead

GitSparTV:
Who in the world still uses :len()??? Your knowledge is so mixed tbh

GitSparTV:
`bin[#bin + 1] = " "`
Introducing 2nd argument of table.concat

Down-s:
> `bin[#bin + 1] = " "`
> Introducing 2nd argument of table.concat

That would just cause each bit to be separated by spaces

GitSparTV:
Have you used it at least once?

Down-s:
`return table.concat( bin, " " )`
image

GitSparTV:
Ok got it

NoSharp:
useless feature, just have it as a repo on your own account, no need to bloat the garrysmod code base.

Kefta:
This should be generalised to all bases and put in the string lib, calling it something it like ``BitRepresentationToString`` to prevent any assumption of it turning a string of a binary sequence into a lua number.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -365,3 +365,51 @@ function util.RemovePData( steamid, name )
...
+--[[---------------------------------------------------------
+	Name: BinaryToString( binary )
+	Desc: Converts a binary string to a readable string
+-----------------------------------------------------------]]
+
+function util.BinaryToString( binary )
+
+	local str = {}
+
+	for byte in binary:gmatch("%S+") do
```
Kefta, line 401: This is reliant on each segment of the string being exactly 8 chars which silently errors/gives a garbage result if it is not. Instead, read the string in segments of 8 (use a manually incremented i in a while loop instead of a for loop to be able to skip whitespaces). If the final length without whitespaces is not divisible by 8 or a char other than 0 or 1 show up, error.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -365,3 +365,51 @@ function util.RemovePData( steamid, name )
...
+--[[---------------------------------------------------------
+	Name: StringToBinary( str )
+	Desc: Converts a string to a binary string
+-----------------------------------------------------------]]
+
+function util.StringToBinary( str )
+
+	local bin = {}
+
+	for i = str:len(), 1, -1 do
```
Kefta, line 378: str:len() -> #str

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -365,3 +365,51 @@ function util.RemovePData( steamid, name )
...
+		local num = str:byte(i)
```
Kefta, line 379: str:byte -> string.byte, it skips the string meta's __index call.

Kefta (changes requested):

ExtReMLapin:
Binary is a format used by computer, string is for humans.
The guy who decides to mix up format used by computer and by humans is making a mistake.
Here it doesn't make sense beside very low level debuging and maybe learning how things works.
As others said before, it doesn't make sense to add this to gmod.

robotboy655:
I have to agree that I don't see any uses for this besides learning how computers work which you can do elsewhere, and nobody so far given any.

ExtReMLapin:
This is half off topic, but I see no point in optimizing this type of code that should not be used at runtime in a production context/mode. @Kefta 

I'm talking about the cpu optimizations like # instead of :len() etc

Kefta:
> This is half off topic, but I see no point in optimizing this type of code that should not be used at runtime in a production context/mode. @Kefta
> 
> I'm talking about the cpu optimizations like # instead of :len() etc

There should be a consistent standard across the GMod codebase where the most efficient operation should be used. This isn't specific nor particularly useful to this PR.

PR #1818, opened 2021-07-12
[TTT] Updated german language

SenseiHiraku:
I´ve updated the german language and improved some older translation.

Old idle notification: 

New idle notification:


SenseiHiraku:
@svdm

BlackCetha:
Nice to see you take care of it. However, some suggestions:
- The pronoun in the last sentence of your screenshot should be "kannst _ihn_ auch", referring to "Modus"
- "jetzt sofort" implies haste, I think just "jetzt" would suffice
- L.equip_tooltip_xfer could maybe be "senden" or "schicken", "transferieren" is uncommon
- L.knife_instant should be either "SOFORTIGE TÖTUNG" or "SOFORT TÖTEN", but not a mix of the two
- L.sb_playervolume could simply be "Lautstärke", because it is used directly in the context of the scoreboard

SenseiHiraku:
Thanks for raising these suggestions, I do agree with you in most terms.

> * L.equip_tooltip_xfer could maybe be "senden" or "schicken", "transferieren" is uncommon

"transferieren" may be uncommon, but in the given context of the following one, I've thought it would fit.
https://github.com/Facepunch/garrysmod/blob/0c0d1e2e30b34a7074424fe78fc48918223dea19/garrysmod/gamemodes/terrortown/gamemode/lang/german.lua#L110-L120
Even so, I would still agree that the tooltip should be more clear about the purpose, so I've changed it to `übertragen`.

PR #1819, opened 2021-07-12
Fix an obscure usergroup bug with `IsUserGroup`

WilliamVenner:
On servers that remove the `PlayerAuthSpawn` hook which sets your usergroup to `user` if you don't have one, `PLAYER:IsUserGroup("user")` will return false despite `PLAYER:GetUserGroup()` returning `user` _if_ the server doesn't call `PLAYER:SetUserGroup("user")` at initial spawn.

Namely FAdmin (literally every DarkRP server) is vulnerable to this, unless ULX is installed alongside.

### DIFF
--- a/garrysmod/lua/includes/extensions/player_auth.lua
+++ b/garrysmod/lua/includes/extensions/player_auth.lua
@@ -33,7 +33,7 @@ function meta:IsUserGroup( name )
 
 	if ( not self:IsValid() ) then return false end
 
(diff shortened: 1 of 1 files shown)


robotboy655:
Feels to me that the bug is with FAdmin, but current implementation `meta:IsUserGroup` should be the proposed one, I agree.

robotboy655 (approved):

PR #1820, opened 2021-07-14
TTT PT-BR translation: more updates

Tiagoquix:
@svdm


ceifa:
Honestly, I didn't see much difference, I'm not the most suitable person for grammar rules, but from my understanding you only changed to synonyms and didn't make any corrections. And what is purpose to change things like `OLOKINHO MEU` para `PIMBA`?

Tiagoquix:
> Honestly, I didn't see much difference, I'm not the most suitable person for grammar rules, but from my understanding you only changed to synonyms and didn't make any corrections. And what is purpose to change things like `OLOKINHO MEU` para `PIMBA`?

Most changes were made to adapt the translation to grammar rules. And yes, I also changed some words for some others (synonyms), because they are better that way.

About the change from "OLOKINHO MEU" to "PIMBA": it's a purely arbitrary change. In the original English file, it says "OMGBANBANBAN". So to adapt in a good way that makes sense, I used a meme of the television presenter Fausto Silva. A while later I realized that "PIMBA" would be better because it is more emphatic, and it comes from the Davy Jones channel, a Brazilian YouTuber. Translation is a Brazilian meme, but in general, this change doesn't make all that much difference.

PR #1821, opened 2021-07-14
Update front view settings

Wgis:
I guess this real front view for props and charaters like models


Wgis:
image
image

garrysmod/gamemodes/sandbox/gamemode/gui/iconeditor.lua
```diff
@@ -320,11 +320,19 @@ function PANEL:AboveLayout()
 
 	local ent = self.ModelPanel:GetEntity()
 	local pos = ent:GetPos()
-	local campos = pos + Vector( 0, 0, 200 )
+	local ang, campos
+	
+	if (ent:GetSequence() != 0) and (#ent:GetSequenceList() > 1) then
```
robotboy655, line 325: I don't think this is a good idea. Not all models with sequences are character models.

PR #1822, opened 2021-07-16
Fix exploit to spam chat

MyDimeIsUp:
This can be used to spam chat or cause lag for those that use custom chatboxes


svdm:
Thanks, I ended up implementing this fix with the delay on a per-corpse basis instead of per-player (I'm now thinking it could perhaps have been once total per corpse, oh well).

MyDimeIsUp:
> Thanks, I ended up implementing this fix with the delay on a per-corpse basis instead of per-player (I'm now thinking it could perhaps have been once total per corpse, oh well).

Yea once I pushed this I remember that there was a delay from when you click click the button to request again.

PR #1823, opened 2021-07-22
replaced all self.Owner with self:GetOwner

lionbryce:


CornerPin:
#1417

Kefta:
Good. Hopefully Rubat won't close this one.

lionbryce:
Call it an optimization, I also localized Owner in some of the functions instead of calling it 3+ times

GitSparTV:
Nice job

robotboy655:
I will not merge PRs that change both TTT and base game.

lionbryce:
I have reverted the changes to gamemodes\terrortown files

jorjic:
Line 36-37 of `includes/extensions/entity.lua` states:
```
	-- Legacy: sometimes use self.Owner to get the owner.. so lets carry on supporting that stupidness
	-- This needs to be retired, just like self.Entity was.
```

The key name is compared against "Owner" as the last check in the __index metamethod of the Entity metatable, after "searching the table" in compiled C++. I ran a benchmark test against `self.Owner` vs `self:GetOwner()` to see what the speed difference was between them. The mean of 100 calls is given for each check.

```
self.Owner
  Server: 0.00000190 seconds
  Client: 0.00000199 seconds
  
self:GetOwner()
  Server: 0.00000123 seconds
  Client: 0.00000128 seconds
```

`self.Owner` comes out clearly slower than `self:GetOwner`. That's not good, but it's not really going to impact performance even if used 100 times per frame with 120 frames per second (0.00000199 * 100 = 0.000199; 1 / ( (1/120) + 0.000199) = 117.2 frames per second, or about 3 frames per second lost).

For curiosity sake, I ran a check against an unrelated variable stored on the SWEP that was initialized with Lua, as well as a simple getter implementation that returned that variable.

```lua
function SWEP:GetTestVariable()
	return self.__test
end
```

```
self.__test
  Server: 0.00000181 seconds
  Client: 0.00000169 seconds
  
self:GetTestVariable()
  Server: 0.00000256 seconds
  Client: 0.00000296 seconds
```

Yuck. `GetTestVariable` seems as slow, if not slower, than the special case `self.Owner`. Still, the direct `self.__test` variable lookup seems as fast, if maybe a bit slower, than `self:GetOwner()`. I'm not sure why `self:GetOwner()` is significantly faster than the direct variable lookup here. The reason `GetTestVariable` is so slow is likely because it has to perform two __index lookups, on top of the overhead for executing a function (even as simple as it is). This got me curious about the function call overhead:

```
anonymous in-scope function
  Server: 0.00000065 seconds
  Client: 0.00000061 seconds
```

Overall, `self:GetOwner()` is considerably faster than `self.Owner`, but it seems to be only about as slow as another function call which returns an instance variable. Both of which should not typically cause any performance problems, as long as they're used even somewhat performance-consciously. I came into this PR strongly in favor of it, expecting the performance gain to be significant, but now I'm neutral. 

I would appreciate it if someone replicated my benchmarks. If the findings are consistent, then this change might be more philosophical than practical. I'd still argue that since `self.Owner` is officially deprecated, it would make sense to update the few cases where it's used in the base game. Many new coders copy and modify the base SWEPs, so perhaps leading by example is the way to go if we really want people to switch to `self:GetOwner()`.

GitSparTV:
You are correct and even too long and correct.

Here is the reason: https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/entity.lua

jorjic:
Thanks @GitSparTV, I cut the cruft.

Kefta:
More important than it being slower, it's been deprecated since garry added it.

garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stool.lua
```diff
@@ -103,7 +103,7 @@ function ToolObj:Init() end
 
 function ToolObj:GetMode()		return self.Mode end
 function ToolObj:GetSWEP()		return self.SWEP end
-function ToolObj:GetOwner()		return self:GetSWEP().Owner or self.Owner end
+function ToolObj:GetOwner()		return self:GetSWEP():GetOwner() or self:GetOwner() end
```
robotboy655, line 106: This looks like a bug. `self.Owner` in this case would be correct.

robotboy655 (changes requested):

GitSparTV:
> it's been deprecated since garry added it.

Very epic

garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stool.lua
jorjic: That's right. Should `self:GetSwep():GetOwner()` ever return `nil`, this will cause a stack overflow.

```suggestion
function ToolObj:GetOwner()		return self:GetSWEP():GetOwner() or self.Owner end
```

PR #1824, opened 2021-07-22
Attribute CS:GO lobby map to CS:GO

EntranceJew:
makes "Other" empty by default again


robotboy655:
Is that even like a real map? Does it have anything? Maybe it's better to just hide it?

EntranceJew:
It's not a real map like any other map or even the training room, It's a map used with the lobby system for CS:GO usually I only ever see it when preparing for a 2player mission, full of all these identical rooms and blank cubes.

 It does have notable dioramas of the Ts and CTs chilling around that I don't think I've seen elsewhere:
CT
T
one of many phone rooms

You do start in a pitch black cube, though (pictured: 3 of many), so kind of not useful for general play.

cubes in space

I wouldn't object to hiding it either but it is somewhat novel.

garryspins:
i know this doesnt relate to the pr but goddamn that huds epic

robotboy655 (approved):

PR #1609, opened 2019-09-15
Optimize some points

JarosLucky:
- Removed some unnecessary local variables before loops.
- Replaced pairs with ipairs for places where the order table is not broken down.
- Localized some vectors so as not to create them different for each operation.
- Faster function variations are used to combine strings.
- A more effective solution to calculate the end position for the beam for the Widgets.
- A more effective way to indicate which panels in DTree are the last ones.
- Faster "MakeNiceName" function in the Player Editor.

PR #1825, opened 2021-07-23
Change pairs to ipairs

ceifa:
A micro optimization to save some nanosecs of cpu

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_init.lua
@@ -186,7 +186,7 @@ function GM:OnChatTab( str )
 
 	if ( LastWord == nil ) then return str end
 
-	for k, v in pairs( player.GetAll() ) do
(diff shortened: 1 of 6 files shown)


ExtReMLapin:
It's better but not faster as it will not jit compile and outside or jit compiled mode, pairs is faster than ipairs

Kefta:
This will be faster in x86_64's JIT.

ceifa:
> It's better but not faster as it will not jit compile and outside or jit compiled mode, pairs is faster than ipairs

Didn't know that, does it have a reason to jit compile pairs but not ipairs?

GitSparTV:
It wouldn't compile in both cases.
But pairs **apprently** is faster in non-JITted code.
The possible reason is the implementation, I didn't look deeper but I assume this.

JarosLucky:
#1609

robotboy655:
Superseded by https://github.com/Facepunch/garrysmod/pull/1609

PR #1826, opened 2021-07-29
Derma_StringRequest children access from obj

BullyHunter32:
made the children accessible from the returned object, used this for the AllowInput on the text entry, twas pretty handy.

### DIFF
--- a/garrysmod/lua/derma/derma_utils.lua
+++ b/garrysmod/lua/derma/derma_utils.lua
@@ -191,7 +191,7 @@ end
 
 --]]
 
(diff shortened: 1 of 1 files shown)


GitSparTV (changes requested):
pls don't do that this way, lua doesn't have optimizations in VM for detecting you are using the same value.
Revert them back to `local`, at the end of this code save the local `x` to `Window.x`

WilliamVenner:
Have needed this many times before, with the exact same use case :v

robotboy655:
Probably should only do this for the text entry itself, not for everything, or have AllowInput be a callback argument.

PR #1829, opened 2021-08-19
TTT PT-BR translation: fix inconsistencies

Tiagoquix:


vitordsouza:
Owner of the current only brazilian ttt server. The translation fixes proposed here are either not correct or completely not necessary. I ask you guys to not make a pull request for this update.

Tiagoquix:
> Owner of the current only brazilian ttt server. The translation fixes proposed here are either not correct or completely not necessary. I ask you guys to not make a pull request for this update.

I disagree. They are correct and I can justify each one.

``L.search_armor``: in the English translation, the written word is "nonstandard". At first, I translated it as "clandestino" (clandestine), since in Brazilian Portuguese, the word "atípico" (atypical) has a similar meaning to "clandestino" (clandestine). However, after playing TTT for a while after this change, I realized that this translation did not make much sense, as the same line that shows a Traitor was wearing a vest is also shown when a Detective was wearing a vest. And it would not make any sense for a Detective to wear a clandestine vest. So I translated it to "atípico" (atypical), as it makes more sense and is more suited for TTT.

``L.disg_hud``: Changing from "ativado" to "habilitado" was to actually correct an inconsistency. Everywhere, with the exception of ``L.disg_hud``, disguising was "habilitado" and not "ativado". To correct this inconsistency, then, I just changed it so that all the lines use the same word.

``L.tip17``: the current translation is weird. It is not wrong, but the way it was written looks like an automatic translation. I changed it to improve writing and also make it easier to understand. The current translation is not wrong, but it could be improved, so I made the change.

``L.confgrenade_name``: I just removed the uppercase there. It is unnecessary (yes, I have capitalized it myself, but now I am removing it, as "Granada de Impulso" is not exactly a proper noun to necessarily be capitalized).

All other changes that I have not mentioned here refer to prominal placement. You can read more here, in Brazilian Portuguese: https://www.educamaisbrasil.com.br/enem/lingua-portuguesa/colocacao-pronominal

vitordsouza:
The changes are mostly either not necessary or wrong:

- L.disg_hud: These words have equal meaning and use in the portuguese language. The supposed reason for the change is for 'consistensy purposes', but from a literary point of view this makes no sense what so ever. The more variety of words and synonyms you use in a text, the richer it gets. You don't stop yourself from using the word 'bug' because you already used 'insect' before. So this 'fix' changes nothing at meaning and only make the translation text poorer in value.

- L.tip17: The change distorced the meaning of the sentence. The english meaning goes as follows: "Are the innocents all grouped up and hard to pick off?". The meaning is that the the innocents are all togheter and are hard to kill. It isn't suggesting that the entire group is hard to kill, but that it is hard for traitors to even make a single kill. The change suggests that it is only hard to kill all of them at once. The proposed change in english is "Are the innocents all grouped up preventing a massacre?". It just makes the text more confusing without adding anything to the understanding of the meaning. The original is closer to the actual translation.

- L.tip22 : Is incorrect. This is a case of 'próclise' because of the interrogative pronoum "quem". So the original form "quem se curou" is the one correct and the change "quem curou-se" is incorrect. https://www.conjugacao.com.br/verbos-transitivos-e-intransitivos/

- L.set_avoid_det: Also incorrect. This is a case of 'próclise' because of the negative form of 'não' that comes before. The original form "não lhe selecionar" is the one correct, the change is wrong. https://www.soportugues.com.br/secoes/sint/sint73.php.

-L.confgrenade_name: is a bad change, the rest of the translation file in all languages puts all nouns in entity names of ammo, weapons and player states in uppercase. This happens both because these names are often used as titles and because it makes it easier to read them at the logs. However for some reason the grenades in the pt-br file were changed to only uppercase at the initial, away from the standard from the rest of the translation files.

In L.search_armor the author of the commit tries to fix a mistake he made when again changed the translation files without necessity and with little attention to the actual benefits of the changes.

Tiagoquix:
Thanks for your feedback. I made some changes based on it.

About ``L.search_armor``: the written word in English is "nonstandard", which when translated is the word "atypical", that has a similar meaning to "clandestine" in Portuguese. I did not translate wrong, but I ended up choosing the word that was not the most suitable.

PR #1831, opened 2021-08-23
Added: Grid quantization snapper to math.lua

dvdvideo1234:


dvdvideo1234:
Hi, see it was added and the implementation is better, Woks like a charm and good enough for me ;P

PR #1832, opened 2021-08-28
Removed useless variable

BullyHunter32:
image


Nyaaaa:
This should be left alone, Because it's working fine, It just saves us having to put in the alpha value.

Kefta:
> This should be left alone, Because it's working fine, It just saves us having to put in the alpha value.

The Color constructor automatically sets the alpha value if left to nil, so this check is quite literally useless.

TheKingOfKong:
> 
> 
> > This should be left alone, Because it's working fine, It just saves us having to put in the alpha value.
> 
> The Color constructor automatically sets the alpha value to 255 if left to nil, so the check being removed is quite literally dead code.

Exactly what i just said, But in different context.

ceifa (approved):

GitSparTV (approved):

PR #1833, opened 2021-08-30
Added Guns for the jeep and airboats

Pootisp:
and an Magnusson carrier


robotboy655:
The guns are disabled on purpose, so they will not be accepted.
I don't know how wanted/useful is the Magnusson carrier to be present by default.

GitSparTV:
Maybe we should have same vehicles with these guns but separate and with AdminOnly?

PR #1836, opened 2021-09-03
PData using SteamID64 to prevent collisions

ghost:
Changed PData from UniqueID to SteamID64 to prevent any collision between players.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -161,7 +161,7 @@ end
 -----------------------------------------------------------]]
 function meta:GetPData( name, default )
 
-	name = Format( "%s[%s]", self:UniqueID(), name )
+	name = Format( "%s[%s]", self:SteamID64(), name )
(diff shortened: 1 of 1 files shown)


sanny-io:
This is a breaking change.

FlorianLeChat:
See #1639 for more information.

Kapsulon:
Then why not either, make so the function uses SteamID64() automatically if the value does not exist for the UniqueID(), or create a separate function like GetPData64() and SetPData64() as a new way to store data quickly ?

Kefta:
The functions should be separate, not use an argument to change this altering behaviour.

Kapsulon:
Here i just created 3 separate functions with alternative naming so old code doesn't get screwed

robotboy655:
I have decided to go in a different direction in resolving the issue of using `UniqueID` for PData that is more transparent to existing addons, see here:
https://github.com/Facepunch/garrysmod/commit/8fc4e6d064ba249343dd22d24ab9e753c986d6d0

PR #1841, opened 2021-09-23
Add argument to net.Read/WriteColor to skip alpha

Kefta:
More network efficient if you aren't using the alpha. The value between the read and write calls must match and should be a warning on the wiki page if this is merged.

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -69,24 +69,28 @@ end
 --
 -- Read/Write a color to/from the stream
 --
-function net.WriteColor( col )
+function net.WriteColor( col, writeAlpha )
+	if ( writeAlpha == nil ) then writeAlpha = true end
 
 	assert( IsColor( col ), "net.WriteColor: color expected, got ".. type( col ) )
 
 	net.WriteUInt( col.r, 8 )
 	net.WriteUInt( col.g, 8 )
 	net.WriteUInt( col.b, 8 )
-	net.WriteUInt( col.a, 8 )
 
+	if ( writeAlpha ) then
+		net.WriteUInt( col.a, 8 )
+	end
 end
 
-function net.ReadColor()
+function net.ReadColor( readAlpha )
+	if ( readAlpha == nil ) then readAlpha = true end
 
 	local r, g, b, a = 
 		net.ReadUInt( 8 ),
 		net.ReadUInt( 8 ),
 		net.ReadUInt( 8 ),
-		net.ReadUInt( 8 )
+		readAlpha and net.ReadUInt( 8 ) or 255
 
 	return Color( r, g, b, a )
 


BlackCetha:
I don't think the added parameter is needed. Changing your check to `if (alpha != 255)` should be equivalent.

CornerPin:
> I don't think the added parameter is needed. Changing your check to `if (alpha != 255)` should be equivalent.

This can't be done for the `net.ReadColor` function, it still needs the parameter.

Kefta:
Easy merge, please consider this for next update.

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -69,24 +69,31 @@ end
...
-	net.WriteUInt( col.r, 8 )
-	net.WriteUInt( col.g, 8 )
-	net.WriteUInt( col.b, 8 )
-	net.WriteUInt( col.a, 8 )
+	local r, g, b, a = col:Unpack()
+	net.WriteUInt( r, 8 )
+	net.WriteUInt( g, 8 )
+	net.WriteUInt( b, 8 )
 
+	if ( writeAlpha ) then
```
GitSparTV, line 82: Can we do `not wroteAlpha` and remove line 73?

Kefta: Since @robotboy655 doesn't like default args in comments in the header, I like the way the first line looks at implicitly signaling the arg defaults to true. If changing it will get this merged faster/at all, though, I'll do it.

GitSparTV: Depends on what rubat's opinion is right now

robotboy655: > Since @robotboy655 doesn't like default args in comments in the header

I am not sure what you mean by that.

Kefta: > 

``function net.WriteColor( writeAlpha --[[= true]] )``

robotboy655: Ah. Yeah not a fan, and its is not done like so anywhere else in the codebase.

I think its fine in its current form, the only alternative I see is changing the arg to `noAlpha`, but I am not sure which once would be more convenient.

GitSparTV: NoAlpha sounds fine.
Since truth of this value will mean "no alpha". Or "dontWriteAlpha"

Kefta: I prefer the writealpha so that false == dont write and true == do write. Seems more intuitive.

robotboy655 (approved):

garrysmod/lua/includes/extensions/net.lua
GitSparTV: Ok then maybe explicitly expect true to be passed?

Kefta: That is what is done currently :) It defaults to true to maintain backwards compatibility/write alpha by default.

PR #1843, opened 2021-09-30
Modern links to the site + redrawing of ping icons for a better design

OfficialFreline:


OfficialFreline:
As for links, this is done so that there is no redirection. As for the icons, they look more aesthetically pleasing, more beautiful like this (I conducted a survey with my people, everyone approved)

WilliamVenner:
that looks way nicer, just need to split the pr.

PR #1848, opened 2021-10-14
Add SafeIndex, SafeCallMetaFunction, SafeCopy, and SafeSerialize

Kefta:
Adds:
- `bool success, ... rets SafeIndex(any var, any key)` - attempts to index the var with the given key. Returns true and the ret(s) if the index could occur without error even if it returns nil. If false is returned, also returns an error as the second argument.
- `bool success, any ret SafeCallMetaFunction(any var, any funcname, table mappings = nil)`
- `bool success, any ret SafeCopy(any var, any funcname, table mappings = {/*See impl*/})`
- `bool success, string serial SafeSerialize(any var, any funcname = "Serialize", table mappings = {/*See impl*/})` - attempts to run var:Serialize(). The mappings table can be used to call a custom function instead of Serialize on a per-type basis - this allows bools, strings, and numbers to be serialized. Returns true and the string result of the function if successful, or false and the error if failed.

See https://github.com/Facepunch/garrysmod/pull/1849 for use cases.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -458,3 +458,113 @@ function GetConVarString( name )
 	local c = GetConVar( name )
 	return ( c and c:GetString() ) or ""
 end
+
+
+--[[---------------------------------------------------------
+	Safe functions
+-----------------------------------------------------------]]
+
+local type = type
+local pcall = pcall
+local rawget = rawget
+local TypeID = TypeID
+local istable = istable
+local tostring = tostring
+local string_format = string.format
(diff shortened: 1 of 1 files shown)


Kefta:
Don't have a computer that can run GMod to test, won't have one for the foreseeable future. Leaving this as a draft until it's verified.

GitSparTV:
Are you sure using goto is good here? Could be just an if/else

thegrb93:
You've lost your mind

Kefta:
Yeah I can't believe I missed that string.format call. I'm losing it.

PR #1852, opened 2021-10-21
Don't stop refreshing server list when joining a server

aStonedPenguin:
Canceling the server list refresh when connecting to a server may lead people to believe there's less servers than there really is. 

This is especially problematic after @willox's change to load populated servers first was reverted as servers can now take quite some time to load.


aStonedPenguin:
> If I join a server, I wouldn't want my client to continue searching.

You’re also not the average user who’s going to gloss over the fact they have to refresh to find more servers.

storm37000:
> > This was added for people whose routers cannot handle that many pings which results in not being able to join any server if the server list is still refreshing after connecting to a server.
> 
> Steam should handle that.
> 
> img

The issue is that steam doesnt automatically handle it properly at all as I get many reports of servers not showing up because people's junk ISP routers soft crash all the time when loading the list.  I tested it myself with a modern wireless AC spectrum router and it soft crashed while loading the server list.

PR #1312, opened 2016-12-29
Added additional color constructors

Kefta:
Finale to the critically acclaimed color-hex PR series (https://github.com/garrynewman/garrysmod/pull/1089); I ended up going with the implementation recommended by @SimonSchick. Although @sannysc made some very good points on the side of having a single polymorphic constructor, with Lua being weakly-typed, it would be very messy, unclear, and inefficient. Color is now a global metatable that acts as a contructor and a calltable. Supports constructing by integer, hex string (0x112233, #112233, #123), X11/web color name, string ("123456", "123 213 231"), vector of floats between 0 and 1, color (copy-constructor), and table with numerical or string (rgba) keys.

Testing:
```
local testcol = Color(255, 255, 255)
assert(testcol == Color.FromInt(0xFFFFFFFF))
assert(testcol == Color.FromInt(0xFFFFFF))
assert(testcol == Color.FromHexString("0xFFFFFFFF"))
assert(testcol == Color.FromHexString("#FFFFFF"))
assert(testcol == Color.FromHexString("#FFF"))
assert(testcol == Color.FromName("white"))
assert(testcol == Color.FromString("255 255 255"))
assert(testcol == Color.FromString("255 255 255 255"))
assert(testcol == Color.FromString(tostring(0xFFFFFFFF)))
assert(testcol == Color.FromString("foo bar"))
assert(testcol == Color.FromVector(Vector(1, 1, 1)))
assert(testcol == Color.FromColor(testcol))
assert(testcol == Color.FromTable({r = 255, g = 255, b = 255}))
assert(testcol == Color.FromTable({255, 255, 255, 255}))

local testcol = Color(199, 21, 133)
assert(testcol == Color.FromInt(0xc71585))
assert(testcol == Color.FromHexString("0xc71585"))
assert(testcol == Color.FromHexString("#c71585"))
assert(testcol == Color.FromName("mediumvioletred"))
assert(testcol == Color.FromString("199 21 133"))
assert(testcol == Color.FromString(tostring(0xFFc71585)))
assert(testcol == Color.FromVector(testcol:ToVector()))
assert(testcol == Color.FromColor(testcol))
assert(testcol == Color.FromTable({r = 199, g = 21, b = 133, a = 255}))
assert(testcol == Color.FromTable({199, 21, 133}))
```

PR #1855, opened 2021-10-31
Default the color creation to white when no arguments are provided

dvdvideo1234:


dvdvideo1234:
These will produce the same result:
```lua
local c = Color(1,2,3,4)

Color(1,2,3,4)
Color(c)
Color({1,2,3,4})
```

Kefta:
I do not see the convinence of the default args, especially with an already-present color_white const, and the constructor additions have been done more efficiently and cohesively here: https://github.com/Facepunch/garrysmod/pull/1312

dvdvideo1234:
@Kefta 

What should the 2-nd, 3-rd and 4-th option return then ?

Kefta:
The 4th already defaults to 255, the 1st thru 3rd should not be optional.

GitSparTV:
Can we just have a separate function for this? Color function is already called many times

dvdvideo1234:
@GitSparTV  The idea is to be able to be called like `Vector()` and defaulting to `255` will not change anything as right now everyone calls it with `Color(0,0,0)` for example by providing at least 3 arguments.

Kefta:
Vector() works as a single constructor because it's a C func which can handle multiple Lua types. Adding this same functionality to Color in Lua is going to significantly slow the func down.

dvdvideo1234:
@Kefta Will wait for the other patch implementation then. Will be nice to have `Color()` ;)

thegrb93:
color_white >:v

ExtReMLapin:
racist

dvdvideo1234:
Well, using `Color(rgba)` is the way to go to create one and use it for whatever

PR #1859, opened 2021-11-12
[TTT PT-BR translation] Final update with lots of revisions

Tiagoquix:
"equipe" is used every time in the file, instead of "time".

@svdm

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/brazilian_portuguese.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/brazilian_portuguese.lua
@@ -780,7 +780,7 @@ L.col_role   = "Cargo"
 L.col_kills1 = "Inocentes mortos"
 L.col_kills2 = "Traidores mortos"
 L.col_points = "Pontos"
-L.col_team   = "Bônus de time"
+L.col_team   = "Bônus de equipe"
 L.col_total  = "Total de pontos"
 
 -- Name of a trap that killed us that has not been named by the mapper
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/brazilian_portuguese.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/brazilian_portuguese.lua
@@ -396,9 +396,9 @@ L.hstation_help   = "{primaryfire} posiciona a Estação de Cura."
 L.hstation_desc   = [[
 Permite que as pessoas se curem quando posicionada.
 
-Seu tempo de recarga é lento. Qualquer um pode usá-la,
-e ela pode ser danificada. Pode ser usada para
-analisar as amostras de DNA dos seus utilizadores.]]
+Seu tempo de recarga é lento. Qualquer um pode
+usá-la, e ela pode ser danificada. Pode ser usada
+para analisar o DNA de seus utilizadores.]]
 
 -- Knife
 L.knife_name      = "Faca"
@@ -496,7 +496,7 @@ do jogador ao qual a amostra selecionada pertence.
 Rastrear alvos distantes consome mais energia.]]
 
 L.dna_menu_scan   = "Rastrear"
-L.dna_menu_repeat = "Rastrear automaticamente"
+L.dna_menu_repeat = "Automatizar"
 L.dna_menu_ready  = "PRONTO"
 L.dna_menu_charge = "CARREGANDO"
 L.dna_menu_select = "SELECIONAR AMOSTRA"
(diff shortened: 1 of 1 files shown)


svdm:
So TTT's texts do not change often and as a result most translations get 1-2 updates a year, but we're now closing in on 20 PRs just for PT-BR. I appreciate the effort but this seems excessive, I don't think there's a good reason why a language would need a constant stream of tiny changes.

I would suggest going through the translation as a whole and working on a final set of improvements, after which we can consider it complete and only add new texts as they are added to the game. Much like what happens with every other translation.

Tiagoquix:
> So TTT's texts do not change often and as a result most translations get 1-2 updates a year, but we're now closing in on 20 PRs just for PT-BR. I appreciate the effort but this seems excessive, I don't think there's a good reason why a language would need a constant stream of tiny changes.
> 
> I would suggest going through the translation as a whole and working on a final set of improvements, after which we can consider it complete and only add new texts as they are added to the game. Much like what happens with every other translation.

The point is that much of the current translation is already good, and it's rare for me to see something that I choose to suggest some change.

That's why it's possible to observe, for example, that the time between a translation suggestion is quite long (two months or even more).

I didn't translate TTT to Brazilian Portuguese originally, I just make new pull requests as soon as I see that something should be changed.

Tiagoquix:
Done @svdm.

I have completely revised the translation and I believe it will no longer need to be updated.

Tiagoquix:
Please wait. I'll fix some inconsistencies I noticed with "corpse" and "body" in the translation.

Tiagoquix:
Done. This pull request is ready for review.

vitordsouza:
> So TTT's texts do not change often and as a result most translations get 1-2 updates a year, but we're now closing in on 20 PRs just for PT-BR. I appreciate the effort but this seems excessive, I don't think there's a good reason why a language would need a constant stream of tiny changes.
> 
> I would suggest going through the translation as a whole and working on a final set of improvements, after which we can consider it complete and only add new texts as they are added to the game. Much like what happens with every other translation.

Hello, I'm the owner of one of the only brazilian TTT servers up and running and i completely agree with you. As a native portuguese speaker i can attest that the vast majority of these changes in the last years were either irrelevant to the final quality of the translation (switching words for equal value synonyms for instance) or just introduced gramatic and lexical mistakes that the autor himself had to correct on future pull requests, amounting to more pulls for no good reason. We tried talking to this guy about not updating the translations this much without need, but he just stated that since he could, he would keep updating because he wanted to. So, in the past months we actually have our server running a local file for the TTT BR translation that is an old pull request to avoid getting these unwanted changes all the time from github. 

Even though there were some improvements in a few of updates, basically every pull request there is at least one or two mistakes or unwanted changes and it is so frequent that we can't keep track to help filter everything. Usually we only see it after they are already merged, and therefore need another pull request to be corrected. I would really appreciate if you could help us to mitigate these updates in the future. In my personal opinion the brazilian translations are already fine as they are, there is no good reason to keep changing them this much.

Tiagoquix:
> > So TTT's texts do not change often and as a result most translations get 1-2 updates a year, but we're now closing in on 20 PRs just for PT-BR. I appreciate the effort but this seems excessive, I don't think there's a good reason why a language would need a constant stream of tiny changes.
> > I would suggest going through the translation as a whole and working on a final set of improvements, after which we can consider it complete and only add new texts as they are added to the game. Much like what happens with every other translation.
> 
> Hello, I'm the owner of one of the only brazilian TTT servers up and running and i completely agree with you. As a native portuguese speaker i can attest that the vast majority of these changes in the last years were either irrelevant to the final quality of the translation (switching words for equal value synonyms for instance) or just introduced gramatic and lexical mistakes that the autor himself had to correct on future pull requests, amounting to more pulls for no good reason. We tried talking to this guy about not updating the translations this much without need, but he just stated that since he could, he would keep updating because he wanted to. So, in the past months we actually have our server running a local file for the TTT BR translation that is an old pull request to avoid getting these unwanted changes all the time from github.
> 
> Even though there were some improvements in a few of updates, basically in every pull request there is at least one or two mistakes or unwanted changes and it is so frequent that we can't keep track to help filter everything. Usually we only see it after they are already merged, and therefore need another pull request to be corrected. I would really appreciate if you could help us to mitigate the frequency of these updates in the future. In my personal opinion the brazilian translations are already fine as they are, there is no good reason to keep changing them this much.

There will be no more updates in the future, unless there are new additions.

And, with all due respect, there is another Brazilian TTT server. Yours is not the only one.

PR #1860, opened 2021-11-23
Fix TTTCanPickupAmmo hook not allowing false return

MyDimeIsUp:
If you attempted to return false when adding TTTCanPickupAmmo, it would still allow players to pick up ammo. Instead of checking if the hook returns true, it now checks if the hook returns anything.


Kefta:
true and false returns functioning the same seems confusing to document and no other hook does it.

MyDimeIsUp:
> true and false returns functioning the same seems confusing to document and no other hook does it.

You worded that a little odd for me to understand.

svdm:
This seems correct: true/false is passed on to ENT:PlayerCanPickup, nil gives default TTT behaviour.

PR #1861, opened 2021-11-28
The code has become simpler and better.

johngetman:


WilliamVenner:
This unnecessarily creates new `Color` objects that the GC must free.

GitSparTV (changes requested):
This is in no way better.
Creating a colour structure is an overhead. SetTextColor also has an overhead to read table, you also made not a white color. Why is it 255,0,255?

`color` is unpacked on purpose, you packed it back.

GitSparTV:
You also create new Color's, not even localising default white colour

robotboy655:
Yeah, and this is a code style change at best which is pointless.

PR #1862, opened 2021-12-01
Correction of a typing error in the comments

MaaxIT:
This is a problem of huge importance: it was written "interger" instead of "integer" in the math.lua file.
The game will surely be better with this fix ;)


GitSparTV:
Such a low quality pass to become a contributor tbh.

MaaxIT:
> Such a low quality pass to become a contributor tbh.

I don't care about being a contributor honestly, I just saw it during a dev session.
and it cost nothing to make a PR about it ^^

PR #1863, opened 2021-12-03
Fix entity spawning out of bounds

jorjic:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/5028

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/commands.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/commands.lua
@@ -774,11 +774,31 @@ function Spawn_SENT( ply, EntityName, tr )
 		if ( !EntTable ) then return end
 
 		PrintName = EntTable.PrintName
+		
+		-- If the player is aimed at an entity, do not place below the entity
+		if ( !tr.HitWorld ) then
+			tr.HitNormal.z = 0
+		end
 
 		local SpawnPos = tr.HitPos + tr.HitNormal * 16
 		if ( EntTable.NormalOffset ) then SpawnPos = SpawnPos + tr.HitNormal * EntTable.NormalOffset end
+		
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/sandbox/gamemode/commands.lua
```diff
@@ -774,11 +774,31 @@ function Spawn_SENT( ply, EntityName, tr )
 		if ( !EntTable ) then return end
 
 		PrintName = EntTable.PrintName
+		
+		-- If the player is aimed at an entity, do not place below the entity
+		if ( !tr.HitWorld ) then
+			tr.HitNormal.z = 0
```
thegrb93, line 780: Modifying the trace data could screw up other things using the CanTool hook. Try this
```lua
local SpawnPos = tr.HitPos
if ( tr.HitWorld ) then
    SpawnPos  = SpawnPos + tr.HitNormal * 16
end
```

jorjic:
This PR should mostly fix entities spawning out of bounds. With the changes, I can't easily make entities clip through the ground, and entities are now positioned better in tight spaces. It should be very close to the old behavior, otherwise.

I removed the part about moving the spawn position up by half of the entity's bounding box, because the return value for OBBCenter is 0 before spawning the entity. Maybe that can be re-added after Spawn(), but I didn't need it to fix the issue and I'm not confident that it won't cause problems for some other entities.

garrysmod/gamemodes/sandbox/gamemode/commands.lua
jorjic: Thank you!

jorjic:
I found a simpler way to fix this that is also closer to the old behavior. Unless something else comes up, this should be the final commit. Here is a demonstration

novyy

PR #1864, opened 2021-12-05
Add SWEP.HoldType

swampservers:
This would be nice to set this instead of having to define SWEP:Initialize on trivial sweps


Malivil:
Probably should copy how its done in wespon_tttbase.lua and then remove it from there: https://github.com/Facepunch/garrysmod/blob/88a247aaecd6b9b0977f0c5df0f0e766824e2f76/garrysmod/gamemodes/terrortown/entities/weapons/weapon_tttbase.lua#L486

May also have to call the base class initialize from the tttbase initialize too

Kefta:
It is far too late in the game's history for this imo.

WilliamVenner:
Amazingly, this is probably a breaking change for some addons.

PR #1865, opened 2021-12-06
TTT: Replace usage of voicerecord with permission check

Malivil:
Matching logic changed in ad636ea0569228b3656e72637a76570db488856c


robotboy655:
I do not feel this necessary, I have removed `-voicerecord` from blocked concommands. Should be fine as is.

Malivil:
Alright, I can just close this then. I was just making the change to keep it in sync with the mentioned commit =)

PR #1866, opened 2021-12-06
Add sequential parameter to net.Read/WriteTable

Down-s:
Writing a bunch of useless keys when networking a sequential table is a waste of bits. This will allow people to network sequential tables without networking all that useless data.

Using a table of 100 Colors;
`11208` bits are networked when writing as a non-sequential table.
`4032` bits for a sequential table.

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -115,7 +115,7 @@ function net.WriteTable( tab, seq )
 
 		for i = 1, len do
 
-			net.WriteType( tab[i] )
+			net.WriteType( tab[ i ] )
 
 		end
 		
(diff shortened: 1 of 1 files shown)


GitSparTV:
This should be as separate function net.WriteArray. And it's possible duplicate since someone already made this

Kefta:
I think using an argument is actually a better approach since Arrays (and sequential tables) are not distinguished anywhere else in the codebase, and is easier to document and explain to people who don't write non-Lua languages.

GitSparTV (approved):

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -106,32 +106,61 @@ end
...
+			tab[i] = net.ReadType()
+
+		end
+
+	else
+
+		while true do
+
+			local k = net.ReadType()
+			if ( k == nil ) then return tab end
```
Kefta, line 155: Change this to break. Makes the return statement unified between the two conditional branches which eases reading.

Kefta (changes requested):

garrysmod/lua/includes/extensions/net.lua
GitSparTV: Had the same feeling

GitSparTV (approved):

robotboy655 (approved):

PR #1868, opened 2021-12-22
Add util.ModuleExists

Down-s:
This would be a much easier way to quickly check if somebody has a binary module installed, instead of having to manually check the operating system/architecture for file.Exists.

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -94,6 +94,20 @@ function util.LocalToWorld( ent, lpos, bone )
 
 end
 
+--[[---------------------------------------------------------
+	Name: ModuleInstalled( name )
+	Desc: Checks if a binary module is installed
+-----------------------------------------------------------]]
+function util.ModuleInstalled( name )
+    
+    local realm = CLIENT and "cl" or "sv"
+    local arch = string.sub( jit.arch, 2 )
+    local ops = system.IsWindows() and "win" or ( system.IsOSX() and "osx" or "linux" )
+    local f = string.format( "lua/bin/gm%s_%s_%s%s.dll", realm, name, ops, arch )
+
+    return file.Exists( f, "GAME" )
+
+end
 
 --[[---------------------------------------------------------
 	Returns year, month, day and hour, minute, second in a formatted string.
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/util.lua
```diff
@@ -94,6 +94,20 @@ function util.LocalToWorld( ent, lpos, bone )
 
 end
 
+--[[---------------------------------------------------------
+	Name: ModuleInstalled( name )
+	Desc: Checks if a binary module is installed
+-----------------------------------------------------------]]
+function util.ModuleInstalled( name )
+    
+    local realm = CLIENT and "cl" or "sv"
```
GitSparTV, line 103: Nice, this also covers menu realm.

GitSparTV (approved):

GitSparTV:
One thing, maybe for consistency it should be named as ModuleExists

WilliamVenner:
Doesn't handle `gmsv_module_linux.dll`

WilliamVenner:
Edge case: `gmsv_module_linux32.dll` works on x86-64, but not on main branch. Lol

(main branch has to be `gmsv_module_linux.dll`)

GitSparTV:
Branch inconsistency?

Down-s:
> Edge case: `gmsv_module_linux32.dll` works on x86-64, but not on main branch. Lol
> 
> (main branch has to be `gmsv_module_linux.dll`)

So if `BRANCH` is main on linux it should just ignore the arch?

GitSparTV:
The thing is BRANCH is not reliable

CornerPin:
This should also handle `gmsv_module_osx.dll`/`gmcl_module_osx.dll`.
https://wiki.facepunch.com/gmod/Creating_Binary_Modules#naminglocation

WilliamVenner:
BRANCH doesn't even work on the server.

Kefta:
I think this would be better as a global function since we don't have a module/lua library. Would be consistent with `require`. The name is also a bit ambiguous with `module` in the API but I think that's unavoidable.

GitSparTV:
What if we just add second argument into require to not throw  error and instead return bool?

`require(str)`
`bool = require(str, true)` (nothrow version)

swampservers:
seems like good idea

garryspins:
that would be a far better solution to this

Kefta:
Modifying more built-in functions is a step in the wrong direction.

WilliamVenner:
still needs fixing to work for `gmsv_module_linux32.dll` on `x86-64` branch 32-bit :(

Down-s:
Closing for #1896

PR #1871, opened 2022-01-07
Added: Editable entity combo box can now use default selected and hav…

dvdvideo1234:
…e icons

### DIFF
--- a/garrysmod/lua/vgui/prop_combo.lua
+++ b/garrysmod/lua/vgui/prop_combo.lua
@@ -10,13 +10,12 @@ function PANEL:Setup( vars )
 
 	self:Clear()
 
+	local bicon = istable( vars.icons )
 	local combo = vgui.Create( "DComboBox", self )
 	combo:Dock( FILL )
 	combo:DockMargin( 0, 1, 2, 2 )
 	combo:SetValue( vars.text or "Select..." )
-	
-	local bicon = istable( vars.icons )
-	
+
 	for id, thing in pairs( vars.values or {} ) do
 		local icon = bicon and vars.icons[id] or vars.icons
 		combo:AddChoice( id, thing, vars.select, icon )
--- a/garrysmod/lua/vgui/prop_combo.lua
+++ b/garrysmod/lua/vgui/prop_combo.lua
@@ -18,8 +18,8 @@ function PANEL:Setup( vars )
 	local hasIcons, icon = istable( vars.icons )
 	for id, thing in pairs( vars.values or {} ) do
 
-		if ( hasIcons) then
-			icon = vars.icons[id]
+		if ( hasIcons ) then
+			icon = vars.icons[ id ]
 		else
 			icon = vars.icons
 		end
--- a/garrysmod/lua/vgui/prop_combo.lua
+++ b/garrysmod/lua/vgui/prop_combo.lua
@@ -18,7 +18,7 @@ function PANEL:Setup( vars )
 	local hasIcons, icon = istable( vars.icons )
 	for id, thing in pairs( vars.values or {} ) do
 
-		if(hasIcons) then
+		if ( hasIcons) then
 			icon = vars.icons[id]
 		else
 			icon = vars.icons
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/prop_combo.lua
```diff
@@ -10,13 +10,15 @@ function PANEL:Setup( vars )
 
 	self:Clear()
 
+	local bicon = istable( vars.icons )
```
GitSparTV, line 13: Why should it be there?

It's used only in loop

dvdvideo1234: I put it outside so it shold not be cheched every iteration to generate the boolean control. It is for performance.

GitSparTV: The question is why did you moved even further in other commit?

dvdvideo1234: Put it with the other declarations. Wherw do you perer it to be moved?

garrysmod/lua/vgui/prop_combo.lua
```diff
@@ -10,13 +10,15 @@ function PANEL:Setup( vars )
 
 	self:Clear()
 
+	local bico = istable( vars.icons )
```
robotboy655, line 13: This should be moved to just above the loop.

```
	local bico = istable( vars.icons )
	for id, thing in pairs( vars.values or {} ) do
```

should also probably have a better name like `hasIcons`.

robotboy655 (changes requested):
```lua
local Panel = vgui.Create( "DFrame" )
Panel:SetSize( 500, 500 )
Panel:MakePopup()

local DP = vgui.Create( "DProperties", Panel )
DP:Dock( FILL )

local choice = DP:CreateRow( "Choices", "Combo #1: Default" )
choice:Setup( "Combo", {} )
choice:AddChoice( "Allow", true )
choice:AddChoice( "Disallow", false )

local choice = DP:CreateRow( "Choices", "Combo #2: Custom default text" )

local vals = {}

vals[ "Table" ] = {}
vals[ "Function" ] = function() end
vals[ "String" ] = "Hello world"

local ics = {}
ics[ "Table" ] = "icon16/cross.png"

choice:Setup( "Combo", { text = "Select type...", values = vals, icons = ics } )
choice.DataChanged = function( self, data )
    print( "You selected: ", data )
end
```

```
[ERROR] lua/vgui/dimage.lua:99: bad argument #1 to 'Material' (string expected, got table)
  1. Material - [C]:-1
   2. SetImage - lua/vgui/dimage.lua:99
    3. SetIcon - lua/vgui/dbutton.lua:47
     4. OpenMenu - lua/vgui/dcombobox.lua:205
      5. DoClick - lua/vgui/dcombobox.lua:272
       6. unknown - lua/vgui/dlabel.lua:237
```

Not acceptable.

garrysmod/lua/vgui/prop_combo.lua
```diff
@@ -15,8 +15,15 @@ function PANEL:Setup( vars )
 	combo:DockMargin( 0, 1, 2, 2 )
 	combo:SetValue( vars.text or "Select..." )
 
+	local hasIcons, icon = istable( vars.icons )
 	for id, thing in pairs( vars.values or {} ) do
-		combo:AddChoice( id, thing )
+		if(hasIcons) then
```
FlorianLeChat, line 20: Spacing is required here

dvdvideo1234: Fixed in fbf1e00debbf39c8e81097dc351453f2ff98d30d

Kefta: You did not fix it, there is still no space between if and the parenthesis.

Kefta (changes requested):

garrysmod/lua/vgui/prop_combo.lua
Kefta: Still missing the space on the other side of the if..

PR #1872, opened 2022-01-22
Implement some improvements for material select

dvdvideo1234:
Instead of doing this manually:
```lua
function wipeMaterials(pMat)
  for k, v in pairs(pMat.Controls) do
    v:Remove(); pMat.Controls[k] = nil
  end; pMat.List:CleanList()
  pMat.SelectedMaterial = nil
end
```

### DIFF
--- a/garrysmod/lua/vgui/matselect.lua
+++ b/garrysmod/lua/vgui/matselect.lua
@@ -39,8 +39,8 @@ end
 function PANEL:SetCustomPaintOver(sMat, nSize, nOrig, cPaint)
 
 	local rpnto = cPaint or color_white
-	local rorig = tonumber(nOrig) or border
-	local rsize = tonumber(nSize) or border_w
+	local rorig = tonumber( nOrig ) or border
+	local rsize = tonumber( nSize ) or border_w
 	local rmhov = sMat and Material( sMat, "nocull" ) or matHover
 	local rboxp = GWEN.CreateTextureBorder( rorig, rorig, 64 - rorig * 2, 64 - rorig * 2,
 	                                        rsize, rsize, rsize, rsize, rmhov )
--- a/garrysmod/lua/vgui/matselect.lua
+++ b/garrysmod/lua/vgui/matselect.lua
@@ -45,7 +45,7 @@ function PANEL:SetAutoHeight( bAutoHeight )
 
 end
 
-function PANEL:UpdatePaintOver( Mat )
+function PANEL:UpdatePaintOver( mat )
 
 	-- Restore the current overlay
 	if ( self.SelectedMaterial ) then
@@ -53,9 +53,9 @@ function PANEL:UpdatePaintOver( Mat )
 	end
 
 	-- Add the overlay to this button
-	self.OldSelectedPaintOver = Mat.PaintOver
-	Mat.PaintOver = HighlightedButtonPaint
-	self.SelectedMaterial = Mat
+	self.OldSelectedPaintOver = mat.PaintOver
+	mat.PaintOver = HighlightedButtonPaint
+	self.SelectedMaterial = mat
 
 end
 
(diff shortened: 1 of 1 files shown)


GitSparTV:
Why do you change thing that are not the subject of PR? Your PR adds Clear function, why do you change other functions?

Also pro tip: first push all commits then post PR.

dvdvideo1234:
@GitSparTV Thought it would be more ordered and convenient when I factorize it, but oh, well. [Turns out I need Find selected method as well and a method that will actually update the `PaintOverlay` when the convar is changed or `DImageButton` is clicked, so I added it][ref-laser-custom]. This PR will not change the panel behavior in any way. It will just offer methods that were unavailable until now. If you insist that I only add the clear method, will just force push. And delete the other commits. That's why I renamed the PR name also I am still having problems with `MatSelect:SetPos(x,y)` and `MatSelect:SetSize(sx,sy)` when [I clearly call it with intended parameters but my `MatSelect` panel is cut in half an I cannot figure out why it does that][ref-mat-cut-panel]... 👍 

[ref-laser-custom](https://github.com/dvdvideo1234/LaserSTool/blob/out-medium-refraction/lua/autorun/laserlib.lua#L262)
[ref-mat-cut-panel](https://github.com/dvdvideo1234/LaserSTool/blob/out-medium-refraction/lua/weapons/gmod_tool/stools/laseremitter.lua#L129)

robotboy655:
I don't see why `UpdatePaintOver` is necessary. Your link is dead.

robotboy655:
But you already have a method to select a given material? (`FindAndSelectMaterial`) What's wrong with using that?

dvdvideo1234:
It runs a loop to select the material panel. I already know what panel I need to update, so running a loop to find it is pointless in my case. It will not harm the way the panel works in every way. It just consolidated the loop in one method `FindByValue` and updating the actual panel in the other `UpdatePaintOver`.  Calling `MatSelect:UpdatePaintOver(DImageButton)` can replace the code I've sent in my last comment. It adds an extra layer of factorization and exposes the direct material update part `UpdatePaintOver`, so I can avoid coding a local version of the following and utilize the build in panel selector switch:
```lua
local border = 0
local border_w = 8
local matHover = Material( "gui/ps_hover.png", "nocull" )
local boxHover = GWEN.CreateTextureBorder( border, border, 64 - border * 2, 64 - border * 2, border_w, border_w, border_w, border_w, matHover )

-- This function is used as the paint function for selected buttons.
local function HighlightedButtonPaint( self, w, h )
	boxHover( 0, 0, w, h, color_white )
end
```
I override the `DImageButton` `DoClick` method so every time I click on material I directly switch to it and update the convar accordingly without the need. I already know that to update and where, so I am just calling it to avoid the loop :+1: 
```lua
 local tRow, pnImg = sort[iD]
  if(tRow.Draw) then -- Drawing is enabled
    local sCon = LaserLib.GetSequenceInfo(tRow, sort.Info)
    local sInf, sKey = sCon.." "..tRow.Key, tRow.Key
    pnMat:AddMaterial(sInf, sKey); iC = iC + 1; pnImg = tCont[iC]
    function pnImg:DoClick()
      LaserLib.SetMaterialPaintOver(pnMat, self)
      LaserLib.ConCommand(nil, sort.Sors, sKey)
      pnFrame:SetTitle(sort.Name.." > "..sInf)
    end
    function pnImg:DoRightClick()
      -- Sort the list of materials or do something other with it
```
The library function does the exact same thing the method `SetMaterialPaintOver` does :+1:

robotboy655:
Why are you so afraid of an extra loop on selection like it's gonna explode your PC?

Why not just update the base panel itself so it doesn't perform that loop on selection instead of trying to add pointless methods to the base game so that only your addon then can avoid the loop?

dvdvideo1234:
That is because I grow that hash table with a rate that every player will eventually contribute to adding stuff it will just keep growing to be able to preform fast hash lookup-s and retrieve proper medium data. Currently it has about 30 elements, but with every beam that hits this will grow larger. The problem is that I like how the selected material overrides the paint over method and I do not want to define my own, so I aim to use the one already defined ones. I could add my own methods to the panel, but I have to rewrite the active image button as well because I will not have access to `boxHover` or `matHover`. I also need to be able to clear the panel so I can rebuild it afterwards. This actually makes the context menus manipulation quite neat and allows sorting.
Besides this panel does not have a `Clear()` method to begin with.

robotboy655 (approved):

PR #1873, opened 2022-01-23
Fix DIconLayout

swampservers:
x starts at m_iBorder, not zero, so it should be compared to max width including the left side border

### DIFF
--- a/garrysmod/lua/vgui/diconlayout.lua
+++ b/garrysmod/lua/vgui/diconlayout.lua
@@ -39,7 +39,7 @@ function PANEL:LayoutIcons_TOP()
 	local x = self.m_iBorder
 	local y = self.m_iBorder
 	local RowHeight = 0
-	local MaxWidth = self:GetWide() - self.m_iBorder * 2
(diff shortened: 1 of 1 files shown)


2048khz-gachi-rmx:
this doesn't sound right; spacing should be present on both sides so spacing * 2 will be the available width

BullyHunter32:
The max width should still be `Width - (m_iBorder * 2)` to include the left & right margin, no? otherwise you only take into account one of the margins.

2048khz-gachi-rmx:
as already mentioned, `x` is compared to `MaxWidth`; since `x` is already `m_iBorder`, the expression
`startX + curX + panelWidth > MaxWidth` ==
`m_iBorder + curX + panelWidth > MaxWidth - m_iBorder` ==
`curX + panelWidth > MaxWidth - m_iBorder` ==
`curX + panelWidth > iconLayoutWidth - m_iBorder * 2`

without this fix, it's `curX + panelWidth > iconLayoutWidth - m_iBorder * 3`

PR #1874, opened 2022-01-31
Flash the window if the game is ready (and not focused)

brandonsturgeon:
This is a small QoL feature that will alert players when they've fully loaded into a server.

Typically players tab out of the game before joining a server - this would give them a "heads up" that the game is (nearly) ready to play.

**Note**: `system.FlashWindow()` only works on Windows, but appears to have no negative side effects when run on Linux, OSX.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_init.lua
@@ -23,6 +23,9 @@ end
 	Desc: Called as soon as all map entities have been spawned
 -----------------------------------------------------------]]
 function GM:InitPostEntity()
+
+	if ( !system.HasFocus() ) then system.FlashWindow() end
+
 end
 
 --[[---------------------------------------------------------


GitSparTV:
I don't think this has to be in-game by default.
There is at least 2 addons that have this feature:
https://steamcommunity.com/sharedfiles/filedetails/?id=2717140231
https://steamcommunity.com/sharedfiles/filedetails/?id=1131903966

garryspins:
Theres really no reason _not_ to have this, this would be one of the more useful modern additions lol

djsime1:
In the mean time I wrote a menu plugin that can do this for both when the menu is loaded and when you load into a game. Is this self-promo? Yes, yes it is.
https://github.com/djsime1/menu-plugins-redux/blob/dev/lua/menu_plugins/pling.lua
(Unfortunately you need to install the rest of the plugin framework to load it, sorry.)

wrefgtzweve:
This is awesome, I manually added it to my menu state a while ago and I can say it's awesome. Having this in every server would definitely be a QoL improvement for everyone. However one small issue might be the fact that it only works on windows not on mac or Linux, but other than that I wouldn't see any reason not to add it!

djsime1:
> However one small issue might be the fact that it only works on windows not on mac or Linux, but other than that I wouldn't see any reason not to add it!

I can see why it's only supported on Windows. First, Gmod on Mac is a practical joke at this point. Second, unlike Windows, Linux users can choose one of numerous desktop environments/window managers, making it tedious to come up with a solution that makes all of them flash.

Periapsises:
How many times have I forgotten I was joining a server because of how long my computer takes to load anything, only to find out hours later that I was kicked for being AFK. I really don't see why this couldn't be implemented by default.
I definitely support this 👍

robotboy655:
This is best done from somewhere in the engine, rather then relying on `InitPostEntity` which relies on servers having the code and gamemodes/addons not overwriting that hook.

brandonsturgeon:
I agree - is it going to happen?

PR #1876, opened 2022-02-08
Add a check for a nil parameter

Deadbeetle:
If `name` is not passed to `hook.Remove`, the initialization of `notValid` indexes a nil value and raises a generic Lua error instead of one more readable by addon authors. This patch raises the correct error if the parameter is omitted.


Kefta:
This would make `false` as the name not error. The check should be changed to `name != nil` so that booleans still error in some way, even if obscured by the IsValid call.

Fixing this to raise an error for all improper types would require a lot more code that should be in an external type check lib if implemented at all, but the above is at least a quick fix.

robotboy655:
I will do it properly.

PR #1879, opened 2022-02-21
Localizing g_SkyPaint to retrieve data

JarosLucky:
There is no need to access global variables every time to get data. And if it is necessary, something is wrong.


robotboy655:
It's going to be accessing the global either way so what does this achieve really?

Kefta:
It saves accessing it more then 10+ more times than it needs to be.

robotboy655 (approved):

PR #1880, opened 2022-03-05
Add util.MakeColoredTextMessage

Kefta:
Usage
```Lua
-- Shared outright in a file
local HelloWorldMessage = util.MakeColoredTextMessage( "HelloWorldMessage", Color(255, 0, 255), "Hello World!" )

-- Inside of a serverside function
HelloWorldMessage(ply)
-- or
HelloWorldMessage(player.GetAll())

-- Or in a clientside function
HelloWorldMessage()
```

No more PrintMessage, no more serverside chat.AddText. Should have done this years ago. Not a great function name so I'm open to suggestions.

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -365,3 +365,24 @@ function util.RemovePData( steamid, name )
 	sql.Query( "DELETE FROM playerpdata WHERE infoid = " .. SQLStr( name ) )
 
 end
+
+if ( SERVER ) then
+	function util.MakeColoredTextMessage( name )
+		util.AddNetworkString( name )
+
+		return function( filter )
+			net.Start( name )
+			net.Send( filter )
+		end
+	end
+else
+	function util.MakeColoredTextMessage( name, ... )
+		local func = function()
+			chat.AddText( ... )
+		end
+
+		net.Receive( name, func )
+
+		return func
+	end
+end


WilliamVenner:
Just encourages network string bloat, they're not infinite

It would be better if it used a single net message and networked the identifier in its own way

Kefta:
65535 is a lot, but I do agree it would be better to network a string table index instead. Problem is that it would require realm syncing the indicies which should be done in the engine; otherwise there will need to be PlayerFullyConnected hacks in base game code.

WilliamVenner:
The network string table isn't only just used for net messages though, it's used for networking stuff like sub materials. Imagine how bloated it gets in servers with skins

Kefta:
If you assume the server had 1000 models with the max 32 submaterials each plus the previously assumed worst case 10000 message count, you'd still be left with ~20000 open net message slots. I do agree this should not contribute to the bloat if it would be added to the engine but it's not a deal breaker since there's plenty of headroom.

WilliamVenner:
The limit is actually 4096. I am not sure why the network ID is 16 bits considering.

`dumpstringtables_new networkstring`

Kefta:
What the fuck

thegrb93:
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/sflib.lua#L1189-L1201

thegrb93:
Every addon has to do this because gmod devs never added a way to do it.

thegrb93:
Same goes for the fucking AddNotify function

mcNuggets1:
Oh only 10 years late

WilliamVenner:
I think a better implementation would be registering the coloured chat message shared, assign & agree on a unique identifier for each registered coloured message, and then simply network down the ID instead.

You could take it a bit further by allowing for dynamic messages as well (basically just a specialized net.WriteTable for coloured chat messages) or using format strings.

thegrb93:
> I think a better implementation would be registering the coloured chat message shared, assign & agree on a unique identifier for each registered coloured message, and then simply network down the ID instead.
> 
> You could take it a bit further by allowing for dynamic messages as well (basically just a specialized net.WriteTable for coloured chat messages) or using format strings.

We've heard this same thing and this solution is not dynamic enough for 99% of cases

WilliamVenner:
Yeah, the second suggestion is probably the most useful one.

Kefta:
> > I think a better implementation would be registering the coloured chat message shared, assign & agree on a unique identifier for each registered coloured message, and then simply network down the ID instead.
> 
> We've heard this same thing 1000x before and this solution is not dynamic enough for 99% of cases

I'd say it's the reverse. Most text/colors being networked is static while the dynamic messages are the minority case. But regardless, this PR is attempting to only solve \#1. I will submit something to tackle \#2 separately.

This PR is going to be in limbo until either PlayerFullyConnected is added, the net limit is raised to 65535, or it's implemented in the engine.

robotboy655:
I don't think this is how it should work - it returning a function you'd then call later.

Kefta:
glua programmers are never making it out of the trenches

PR #1881, opened 2022-03-07
Add IsTableColor

Kefta:
Checks if the table can be interpreted as an rgb(a) color since engine functions that return a color do not set the metatable. The alternative would be to add a Lua override to every function which returns a color setting the metatable, however, this does not scale for custom module functions.

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -36,6 +36,17 @@ function IsColor( obj )
 
 end
 
+local function IsColorComponent( v )
+	return isnumber( v ) and v >= 0 and v <= 255 and math.floor( v ) == v
+end
+
+function IsTableColor( obj, alpha )
+	return istable( obj ) and
+		IsColorComponent( obj.r ) and
+		IsColorComponent( obj.g ) and
+		IsColorComponent( obj.b ) and
+		( not alpha or IsColorComponent( obj.a ) )
+end
 
 --[[---------------------------------------------------------
 	Returns color as a string


ExtReMLapin:
I don't understand why you would actually need this.

How would you at some point end up with a variable and now knowing what this variable contains or where it comes from.

thegrb93:
> I don't understand why you would actually need this.
> 
> How would you at some point end up with a variable and now knowing what this variable contains or where it comes from.

Serialization most likely

ExtReMLapin:
Specify the structure type before sending it then ?

thegrb93:
No one is going to maintain a table of types with their table. That would be a buggy disaster.

Kefta:
It's because all engine functions return colors without the metatable. This can be used in a lot of other places to add proper type checking to the codebase, including [here](https://github.com/Facepunch/garrysmod/blob/c1cb9adb57bbccc2212eb4f155a0895170a1e375/garrysmod/lua/includes/extensions/net.lua#L76), [here](https://github.com/Facepunch/garrysmod/blob/c1cb9adb57bbccc2212eb4f155a0895170a1e375/garrysmod/lua/includes/extensions/net.lua#L155), and [here](https://github.com/Facepunch/garrysmod/blob/a9f131b9b2e2e3f8970778a1570df0fed696ea6d/garrysmod/lua/includes/util.lua#L212).  Currently, `net.WriteColor(ent:GetColor)` errors.

ExtReMLapin:
You're missing my point Kefta, YOU as a developer control what messages you send to the client or to the server.
It's not some guy writing the requests by hands, so you control how you build the network message and control how you read it on the other realm.

I understand it's "cool 😎" to be able to send a message and at the end end of the water pipe it "guess" how to handle it but that's just a nasty practice.

It always has to be `[COLOR->TEXT; COLOR->TEXT;]`

not 

`[COLOR->TEXT; TEXT; COLOR, TEXT]`


I might understand what you're trying to do ? You're tring to mimic what `chat.AddText` does , it support that ?

Well don't do it, because it's a mistake, even if garry is behind it, and it's not the first/last thing we need to get rid of (cf : table.HasValue)

Kefta:
@ExtReMLapin I really don't understand what you are trying to say. chat.AddText is going nowhere and it behaves how it does. I am mimicking it's exact behaviour - there is no "guessing." Also, `[COLOR->TEXT; TEXT; COLOR, TEXT]` behaves the same as `[COLOR, TEXT, COLOR, TEXT]` where the two texts side-by-side are concated.

ExtReMLapin:
Yes I answered you in the other PR, I mixed things up, just don't allow people to incorrectly call your function

Kefta:
They can't. That's what the type checking is for.

thegrb93:
I downvoted btw only because this should be `IsColor`'s job

robotboy655:
I do not think this is a good idea. If there's a problem with engine functions retuning non color objects, that should be addressed, not adding more workarounds.

Kefta:
Then please set the metatable of the color tables pushed from the engine.

PR #1882, opened 2022-03-07
Add player.SendMsg

Kefta:
An alternative of https://github.com/Facepunch/garrysmod/pull/1880 for dynamic text. Relies on https://github.com/Facepunch/garrysmod/pull/1881. Supersedes https://github.com/Facepunch/garrysmod/pull/1333 - I closed that one because people should really be using a function like that added in #1880 but I have now accepted that addons will not stop adding their own serverside chat.AddText variants until one is provided for them. This one is also far more network efficient.

Marked as WIP until I fix the styling and have a computer to test.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -87,6 +87,108 @@ function player.GetBySteamID64( ID )
 	return false
 end
 
+if (SERVER) then
+	player.AddNetworkString("GMOD_player_SendMsg_Text")
+	player.AddNetworkString("GMOD_player_SendMsg_Color")
+
+	local function SendMsgError(uArg, v)
+		error(string.format("bad argument #%u to player.SendMsg (string or Color expected, got %s)",
+			uArg + 1, type(v)), 3)
+	end
+
+	function player.SendMsg(filter, ...)
+		local tParts = {...}
+		local bLastString = false
+
+		local bFirstColor = false
+		local tWriteParts = {}
+		local uWriteParts = 0
+
+		for i = 1, select("#", ...) do
+			local part = tParts[i]
+
+			if (bLastString) then
+				if (IsTableColor(part, false)) then
+					uWriteParts = uWriteParts + 1
+					tWriteParts[uWriteParts] = part
+					bLastString = false
+				elseif (isstring(part)) then
+					tWriteParts[uWriteParts] = tWriteParts[uWriteParts] .. part
+				else
+					SendMsgError(i, part)
+				end
+			elseif (isstring(part)) then
+				uWriteParts = uWriteParts + 1
+				tWriteParts[uWriteParts] = part
+				bLastString = true
+			elseif (IsTableColor(part, false)) then
+				if (tWriteParts == 0) then
+					bFirstColor = true
+					uWriteParts = uWriteParts + 1
+				end
+
+				tWriteParts[uWriteParts] = part
+			else
+				SendMsgError(i, part)
+			end
+		end
+
+		if (bFirstColor) then
+			-- Don't proceed if there's nothing to send
+			if (uWriteParts == 1) then return end
+
+			net.Start("GMOD_player_SendMsg_Color")
+			net.WriteUInt(uWriteParts, 32)
+		else
+			if (uWriteParts == 0) then return end
+
+			net.Start("GMOD_player_SendMsg_Text")
+			net.WriteUInt(uWriteParts, 32)
+
+			-- Align for the loop below
+			net.WriteString(tWriteParts[1])
+		end
+
+		for i = bFirstColor and 1 or 2, uWriteParts, 2 do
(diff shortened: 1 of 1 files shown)


Kefta:
The current implementation ignores the alpha since the default chatbox doesn't use it. Are there any custom ones that override chat.AddText and use the alpha?

Kefta:
The current implementation ignores the alpha since the default chatbox doesn't use it. Are there any custom ones that override chat.AddText and use the alpha?

garrysmod/lua/includes/extensions/player.lua
```diff
@@ -87,6 +87,108 @@ function player.GetBySteamID64( ID )
...
+		for i = 2, uPartCount, 2 do
+			tParts[i] = net.ReadColor(false)
+			tParts[i + 1] = net.ReadString()
+		end
+
+		chat.AddText(unpack(tParts))
+	end)
+
+	net.Receive("GMOD_player_SendMsg_Color", function()
+		local uPartCount = net.ReadUInt(32)
```
ExtReMLapin, line 177: With LuaJIT default settings, you can't unpack more than 8000 values, you can stick to 13 bits.

garrysmod/lua/includes/extensions/player.lua
```diff
@@ -87,6 +87,108 @@ function player.GetBySteamID64( ID )
...
+
+		local bFirstColor = false
+		local tWriteParts = {}
+		local uWriteParts = 0
+
+		for i = 1, select("#", ...) do
+			local part = tParts[i]
+
+			if (bLastString) then
+				if (IsTableColor(part, false)) then
```
ExtReMLapin, line 111: This is where it's wrong, stick to a format, don't "guess"/detect if it's STRING->COLOR->STRING->COLOR etc or STRING->STRING->COLOR->STRING

ExtReMLapin (changes requested):
Please stick to a format don't send something and guess/detect on the fly what's inside

garrysmod/lua/includes/extensions/player.lua
Kefta: It's not guessing. Instead of networking STRING->STRING->COLOR->STRING, the strings are concated so it sticks to a consistent format of alternating strings and colors. Aka what the engine already does with chat.AddText.

garrysmod/lua/includes/extensions/player.lua
Kefta: Good to know, this should probably be a global/const somewhere.

ExtReMLapin: There is nothing in luajit allowing you to get const from the .h file, however, you can do thing hacky thing (don't do it actually) to detect the limit.

```lua
local tbl = {}

local i = 0
while i < 2^32 do
	table.insert(tbl,"")
	unpack(tbl)
	i = i + 1
end
```

garrysmod/lua/includes/extensions/player.lua
ExtReMLapin: Yes my bad on this, I mixed up the client and the server.
Anyway from the start the function should not be called this way.

`assert` table and assert string, fuck him if he doesn't call it correctly

garrysmod/lua/includes/extensions/player.lua
Kefta: It can just be pushed from the C side, maybe into the jit.util table. No need for hacks. I will make a request for it.

garrysmod/lua/includes/extensions/player.lua
Kefta: There should be no "fuck him," there should just be proper type checking like what's done here. You can send a string first or a color first. What's the actual issue here?

garrysmod/lua/includes/extensions/player.lua
thegrb93: 13 bits is up to 8192. Needs to be 12

garrysmod/lua/includes/extensions/player.lua
ExtReMLapin: I think this is where we disagree, I don't think it's a good idea to allow multiple way to call a function.

Kefta: Well this is how chat.AddText behaves. It's too late to change it, and this is not just some garry-ism we can just ignore as it is the only way to add coloured text (or add any text clientside).

ExtReMLapin: Sure it's too late to change chat.AddText or to remove it, but I don't think we can justify everything by "that's how it is".

for example, table.HasValue exists, yet, thing kind of stuff should not be "repeated".

Kefta: There is an easy alternative to using table.HasValue - you can even structure your tables around it. There is no alternative chat.AddTexf alternative and it's fine to maintain this consistency.

PR #1885, opened 2022-04-07
Added support for Color's with table sanitization

SnisnotAl:
# Added support for Color's with table sanitization
Also added 0's to the first argument of these, due to the fact that if the first argument of a angle/? is set to NIL, then ALL the values will be 0.
This is probably due to the check where it tries to convert a string/function/? to a angle/?
This is quite important because if you had Angle(0,1,2) and you sanitised it, the sanitisation replaces all 0's with nil, and your angle will break since the first value is 0 (nil now):

and
```lua
tbl[ k ] = Angle( v.p or 0, v.y, v.r )
```

Have also made the nilled values of Color to 255 instead of 0, as 255 is going to be the most common number in a Color, especially in the alpha channel, and will save on data

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -282,7 +282,7 @@ function table.Sanitise( t, done )
 
 	for k, v in pairs ( t ) do
 
-		if ( istable( v ) and !done[ v ] ) then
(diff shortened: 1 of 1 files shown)


SnisnotAl:
image

robotboy655 (approved):

PR #1886, opened 2022-04-10
Added string.NiceNumber() to abbreviate large numbers

SnisnotAl:
Added string.NiceNumber() to abbreviate large numbers
Examples:
```lua
string.NiceNumber( 69420 ) --> "69.42K"
string.NiceNumber( -123456789 ) --> "-123.46M"
```

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -268,6 +268,47 @@ function string.NiceSize( size )
 
 end
 
+--[[---------------------------------------------------------
+	Name: NiceNumber( num )
+	Desc: Abbreviates numbers and adds a suffix	
+	Examples: string.NiceNumber( 69420 ) ==> "69.42K"
+			string.NiceNumber( -123456789 ) ==> "-123.46M"
+-----------------------------------------------------------]]
+local nice_number_suffixes = {
+    "K","M","B","T","Qa",
+    "Qi","Sx","Sp","Oc","No",
+    "Dc","Ud","Dd","Td","Qad",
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/extensions/string.lua
```diff
@@ -345,3 +345,44 @@ function string.Comma( number )
...
+        return string.format( "%sInfinity", isNegative and "-" or "" )
+    end
+
+    for i = 1, #nice_number_suffixes do
+        if absNum >= ( 10 ^ ( i * 3 ) ) then
+            suffixid = i
+        else
+            break
+        end
+    end
```
CornerPin, line 381: Perhaps this is more efficient, not nested:
```suggestion
    if ( absNum >= 1000 ) then
        suffixid = math.min( #nice_number_suffixes, math.floor( math.log10( absNum ) / 3 ) )
    end
```
Also, you should indent using tabs to be consistent with the code base.

GitSparTV:
Are you sure double can represent the number with the largest suffix?

SnisnotAl:
im not sure i understand?
it works up to 1e+89 until it starts just adding 0's before the last prefix
image

NoSharp:
It seems feature-less, perhaps being able to define the character(s) used etc. for it might be a good idea.

Ffyvjgnmmm:
> It seems feature-less, perhaps being able to define the character(s) used etc. for it might be a good idea.

those characters are the well-known standard which almost everything uses

Grocel:
> > It seems feature-less, perhaps being able to define the character(s) used etc. for it might be a good idea.
> 
> those characters are the well-known standard which almost everything uses

This does not work well together with localized content.

PR #1887, opened 2022-04-19
[TTT] Prevent weapons disappearing during round cleanup

Yashirmare:
During round restarts, living players will be stripped of their weapons for 1 tick before they get respawned, during which they can pick up weapons that will immediately be lost as soon as they respawn.
This is the reason why a weapon will sometimes be missing from it's usual spawn point.
This fixes the issue by preventing players from picking up any weapon during the 1-tick period before respawning.

Credit goes to @wgetJane for the fix.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/init.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/init.lua
@@ -389,6 +389,10 @@ local function CleanUp()
    for k,v in ipairs(player.GetAll()) do
       if IsValid(v) then
          v:StripWeapons()
+            if v:IsTerror() then
+               -- don't let players pick up weapons until they respawn
+               v.prevent_pickup_until = CurTime() + 1
+            end
       end
    end
 
--- a/garrysmod/gamemodes/terrortown/gamemode/player.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player.lua
@@ -71,6 +71,8 @@ function GM:PlayerSpawn(ply)
    ply.spawn_nick = ply:Nick()
    ply.has_spawned = true
 
+   ply.prevent_pickup_until = nil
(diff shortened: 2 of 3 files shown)


Kefta:
Why not wait one tick of length instead of a full second?

mcNuggets1:
CurTime() + engine.TickInterval() would work?

robotboy655:
I would imagine the 1 second is just a safety net, in reality it would last less than that as the "timer" is deleted when player spawns.

Kefta:
I think it should just be a boolean then since it's reset upon Spawn anyway.

Kefta:
I think it should just be a boolean then since it's reset upon Spawn anyway.

svdm:
Seems like the restriction is not related to the individual player, but to the game state. The reason no weapons should be picked up is that the map cleanup/respawning is in progress, so perhaps it makes more sense to add a flag for that on `GAMEMODE` that's true between cleanup and player respawning, and check that in `PlayerCanPickupWeapon`.

I'll see if I can repro this issue.

mgetJane:
i guess such a gamemode flag can be set in the CleanUp function in init.lua and then removed in the SpawnEntities function which is run with a 1-tick timer after CleanUp is called
(i still prefer putting the "can't pick up" condition on the player)

the issue is essentially caused by living players still being at their round-end positions for 1 tick when the map is cleaned up
so another possible fix would be to immediately put living players in spec as soon as the map is cleaned up

mgetJane:
@svdm btw is the timer the only reason why this was closed?

if so, i'll open a new PR with the requested changes

svdm:
It was closed by fa4ee61 which implements a fix of the underlying issue. The only reason to do a new PR would be if the problem still exists in some form.

PR #1889, opened 2022-04-26
Tiny fix - validation of panel in timer

bewflast:
# Problem:

```
[ERROR] lua/postprocess/super_dof.lua:123: Tried to use a NULL Panel!
  1. SetVisible - [C]:-1
   2. unknown - lua/postprocess/super_dof.lua:123

Timer Failed! [Simple][@lua/postprocess/super_dof.lua (line 123)]
```

# How to reproduce:

https://user-images.githubusercontent.com/68069182/165396278-3ae5ea54-4078-495b-a0d1-304655685122.mp4

### DIFF
--- a/garrysmod/lua/postprocess/super_dof.lua
+++ b/garrysmod/lua/postprocess/super_dof.lua
@@ -120,7 +120,11 @@ function PANEL:Init()
 	local THIS = self
 	function Break:DoClick()
 		THIS:SetVisible( false )
-		timer.Simple( 5, function() THIS:SetVisible( true ) end )
+		timer.Simple( 5, function()
+			if IsValid(THIS) then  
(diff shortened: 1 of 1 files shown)


garrysmod/lua/postprocess/super_dof.lua
```diff
@@ -120,7 +120,11 @@ function PANEL:Init()
 	local THIS = self
 	function Break:DoClick()
 		THIS:SetVisible( false )
-		timer.Simple( 5, function() THIS:SetVisible( true ) end )
+		timer.Simple( 5, function()
+			if IsValid(THIS) then
```
brandonsturgeon, line 124:
```suggestion
			if IsValid( THIS ) then  

```
nit: Looks like this file uses spaces inside parens

PR #1890, opened 2022-04-27
Replaced util.TraceLine by util.TraceHull for tool traces. Fixes trace bypassing parented props

MartyX5555:
## Whats this?

Its a too little workaround about traces that tools use (TOOL:LeftClick( trace ), TOOL:RightClick( trace ) and TOOL:Reload( trace ))

## What does?

Allows tools to work properly with props that under very special cases (mainly being parented) cannot, or does it barely. Fails are like trace simply ignoring them.

## Why?

Because the used trace function (util.Traceline()) tends to fail most of the time with parented stuff. Here's an issue related to that if you want details: https://github.com/Facepunch/garrysmod-issues/issues/4142. So while this bug is presented, we could use tracehull for now.

## How can i reproduce this bug if this workaround is not present?

Simple, you can try to edit props which were parented to another prop (note props like prop_physics). You will note that the tool cannot edit them, since the trace can fail to detect them. Heres a tool for parenting: https://steamcommunity.com/sharedfiles/filedetails/?id=111929524&searchtext=parent

Steps if you need it

1.- select a few props which you want to test with the parent tool using left click (or another tool which you have)
2.- parent them to another prop of the same class (note: can be a different class) using right click
3.- try to use a tool on the parented prop, like using material tool or color tool

## Where is this bug more present?

Mostly in the building community, where due to optimization purposes, we parent the stuff we put. This doesnt discard for anyone using parented entities.

## Expected bugs?

From my testing, tools work as expected, including parented props now. The only change was replacing the trace function, which still keeps the ray and not a volume ( Vector(0,0,0) for both mins and maxs ), so it should not break stuff.

## Also

Apologies if the end lines are not fine (they look good on github though), my first PR here, if they need change, i will do.


Kefta:
I think this is a workaround to an engine but that should be fixed first, and considering the state of merges recently, there's probably an equal chance of this being merged to the root cause being fixed.

CornerPin:
You can replace `Vector(0,0,0)` with `vector_origin` so it doesn't have to create a new `Vector` every time a ray needs to be traced.

MartyX5555:
@Kefta if things are as you state, prob this pr could be used just in case the fix is harder than it seems (actually no idea how traceline and tracehull are built into engine, so idk about its complexity).

@CornerPin i changed the `vector(0,0,0)` to `vector_origin`. I forgot about those global variables.

robotboy655 (approved):

PR #1895, opened 2022-06-10
Add HL1 357 and shotgun to NPCUsableWeapons

grande1900:
Sure, some might argue that the animations are broken and the names can cause confusion, but the Glock does as well! so?


Kefta:
The list should really have a whitelist/blacklist for NPC classes that have proper anims with it. Adding weapons which (almost?) no NPCs have anims for is addon territory.

PR #1897, opened 2022-07-07
Added: Support for typing disable for text boxes

dvdvideo1234:


robotboy655:
Why?

dvdvideo1234:
So you can be able to forbid the players to edit the text box, but to be allowed only to copy its contents.

robotboy655 (changes requested):
It should at least be named `readonly` then, and affect all the other property types as well. 

Also please refrain from pointlessly modifying whitespace and comments.

dvdvideo1234:
My IDE is automatically set to trim the trailing spaces.

Kefta:
They aren't trailing though, they're separators

dvdvideo1234:
Hello,  @robotboy655... Done ;)

robotboy655 (changes requested):
I have tested it and it simply won't work this way.

I think however you can simply add
```
-- If the field is read only, disable it
if ( vars && vars.readonly ) then
	self:SetEnabled( false )
end
```

to `DProperties.lua` on line 65 and that should cover it for all property types, even addon ones.

dvdvideo1234:
Should I remove the `prop_*.lua` changes then. I mean revert all the changes in prop_*.lua files ?

robotboy655:
Yeah

PR #1898, opened 2022-07-17
fix infinite panel resizing recursion

PrikolMen:


robotboy655:
This code results in things being cutoff on the right:
imageimage

PR #1900, opened 2022-07-22
TTT: Spectate previous player when left clicking

actuallyNothing:
Current behaviour when left clicking is spectating a random player:

```lua
-- snap to random guy
ply:Spectate(OBS_MODE_ROAMING)
ply:SetEyeAngles(angle_zero) -- After exiting propspec, this could be set to awkward values
ply:SpectateEntity(nil)

local alive = util.GetAlivePlayers()

if #alive < 1 then return end

local target = table.Random(alive)
if IsValid(target) then
   ply:SetPos(target:EyePos())
   ply:SetEyeAngles(target:EyeAngles())
end
```
I feel like it is more intuitive for it to spectate the previous player, as right click spectates the next player.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/player.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player.lua
@@ -316,20 +316,13 @@ function GM:KeyPress(ply, key)
       ply:ResetViewRoll()
 
       if key == IN_ATTACK then
-         -- snap to random guy
-         ply:Spectate(OBS_MODE_ROAMING)
-         ply:SetEyeAngles(angle_zero) -- After exiting propspec, this could be set to awkward values
-         ply:SpectateEntity(nil)
+        -- spectate either the previous guy or a random guy in chase
+        local target = util.GetPreviousAlivePlayer(ply:GetObserverTarget())
 
-         local alive = util.GetAlivePlayers()
-
-         if #alive < 1 then return end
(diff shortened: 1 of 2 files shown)


svdm:
Thanks for the PR. I've considered this for a while, it is more consistent/intuitive to mirror left and right clicks but in the end I just prefer the current mechanism.

mcNuggets1:
In which way?
It doesnt have any downsides, only benefits.
Nothing'll really change for players using the "old mechanism".

colemclaren:
> Thanks for the PR. I've considered this for a while, it is more consistent/intuitive to mirror left and right clicks but in the end I just prefer the current mechanism.

The current spectating system is awfully outdated and needs an update. This is a great improvement to the spectating system.

actuallyNothing:
I agree that this should be reconsidered as an improvement to the current system. Not only is it unintuitive to have it work the way it does right now, but it is less convenient and the "feature" of jumping a freecam to a random player does not fulfill any meaningful purpose. The only use it has is going into freecam, which is already done by the CTRL key.

lokeisloke:
Please reopen this for consideration. This is an objective improvement for navigating to a specific player while spectating. If you're cycling through 15+ players and accidently click again and pass by, you have to cyle through the entire playerset again.

mcNuggets1:
+1

PR #1903, opened 2022-08-02
TTT: Remove ASS

colemclaren:
Inappropriate comment. Not necessary.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_voice.lua
@@ -165,7 +165,6 @@ function RADIO:ShowRadioCommands(state)
 
          radioframe:CenterVertical()
 
-         -- ASS
          radioframe.ForceResize = function(s)
                                      local w, label = 0, nil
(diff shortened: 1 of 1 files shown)


Kefta:
Replace it with something that at least conveys its hacky nature.

colemclaren:
> Replace it with something that at least conveys its hacky nature.

I'd rather just replace the hacky nature all together. It's not really hacky if you read it. All it does is extend the width of the panel in case the label is too long. I don't think it's hacky for what it does. What do you propose it be changed to?

colemclaren:
Perhaps a new pull request should be made for fixing the hacky nature? I only want to remove the awful comment. Nothing else.

colemclaren:
> > Replace it with something that at least conveys its hacky nature.
> 
> I'd rather just replace the hacky nature all together. It's not really hacky if you read it. All it does is extend the width of the panel in case the label is too long. I don't think it's hacky for what it does. What do you propose it be changed to?

I take this back. It's in a think hook. I've replaced the comment with a better way of conveying the message across. I appreciate the idea thank you.

PR #1905, opened 2022-08-02
Remove offensive comment

colemclaren:
It's 2022. This is not a word I would like my kids using. I don't want them to start saying this because they've just read the sent_ball.lua file for the first time. Not a joke. I will not hear the end of it.


robotboy655:
See #1906

PR #1906, opened 2022-08-02
Remove Todo

colemclaren:
There's no better way to do this. This works fine. Does not need to be changed. Inappropriate word in comment.


robotboy655:
You do not need to make a separate PR to remove a single comment.

PR #1907, opened 2022-08-07
Fix NPC vs. NPC Deathnotice Icons

Conrad33:
Shows NPC killicons for NPC vs NPC kills that don't have an inflictor instead of just the worldspawn icon, matching the behavior for NPC vs Player deaths


robotboy655:
I have decided to got for a different fix that matches how and where it happens for players.

https://github.com/Facepunch/garrysmod/commit/23d6d962916d82e45d261acc092139b7adf3a461

PR #1911, opened 2022-08-19
Add Server Operator Rules to Legal in options tab

colemclaren:
I've added the modding guidelines, community server and hosting guidelines, as well as the server operator guidelines to the legal tab in options, which was previously not made aware before. It's important to note, the server operator guidelines are not linked anywhere on Facepunch's legal section of the Garry's Mod website.

Screenshot on beta branch: x86-64 - Chromium + 64-bit binaries
S6L0m


colemclaren:
This will not work out of the box, there are engine changes which need to be made. Did I miss anything?

mechdabosh:
👍 I'd like to see this, the wiki exist but this is more accessible for people who don't that exist.

PR #1912, opened 2022-08-27
Fix tooltip height position miscalculation with newlines.

0x098:
1.5 causes the whole tooltip to elevate .5 extra on each newline(\n) line this: 

1.0 or just no multiplier "fixes" it like this:


colemclaren:
I believe the tooltip is supposed to hover a little above the element it's tooltipping for, and the 1.5 is intended to give it this extra space. This change removes this space. Why not add a 15px spacer between the tooltip and the element instead of multiplying using height?

0x098:
10px height gap appears like this

colemclaren:
Yeah, this 10px spacer looks more appropriate than no spacer.

colemclaren:
This would be a neat change. 🤠👍

robotboy655 (approved):

PR #1913, opened 2022-08-30
Add DDialogBox element

jorjic:
DDialogBox is a customizable [dialog box](https://en.wikipedia.org/wiki/Dialog_box) implementation. 

Out of the box, it contains a label, an accept button, and a reject button. Each of these components may be customized or disabled as desired. Arbitrary VGUI elements may be added by passing either an existing panel or a panel class name to the `AddPanel` function. The panel must be docked to TOP or the layout will probably be incorrect. After custom elements have been added, `Invalidate()` must be called to automatically resize the dialog box.

The `OnAccept` and `OnReject` callback functions will be fired once the user clicks one of the available option buttons. If you want to perform some action once the dialog has been acknowledged/confirmed, you should override one or both of these functions.

It is [modeless](https://en.wikipedia.org/wiki/Mode_(user_interface)) by default, but can be made modal with the existing [Panel:DoModal](https://wiki.facepunch.com/gmod/Panel:DoModal) function.

# Examples

## Simple information dialog

```lua
-- Regular dialog example
local dialog = vgui.Create( "DDialogBox" )
dialog:SetIcon( "icon16/information.png" )
dialog:SetTitle( "Example Dialog" )
dialog:SetText( "This is an example dialog.\n\nInform the user about something here." )
dialog:SetRejectButtonEnabled( false )
```

image

---

## Color selection dialog

```lua
-- Color selector example
local dialog = vgui.Create( "DDialogBox" )
dialog:SetIcon( "icon16/color_wheel.png" )
dialog:SetTitle( "Choose Color" )
dialog:SetText( "" )
dialog:SetWide( 280 )  -- Carefully selected for color swatch alignment

local mixer = dialog:Add( "DColorMixer" )

function dialog:OnAccept()
	local color = mixer:GetColor()
	-- Do something with the chosen color
end

function dialog:OnReject()
	-- User cancelled, do something...
end
```

image

---

## Text input dialog

```lua
-- Text input example
local dialog = vgui.Create( "DDialogBox" )
dialog:SetIcon( "icon16/application_form.png" )
dialog:SetTitle( "Choose A Name" )
dialog:SetText( "" )

local firstNameInput = dialog:Add( "DTextEntry" )
firstNameInput:SetPlaceholderText( "First name" )
firstNameInput:DockMargin( 0, 0, 0, 0 )

local lastNameInput = dialog:Add( "DTextEntry" )
lastNameInput:SetPlaceholderText( "Last name" )
lastNameInput:DockMargin( 0, 4, 0, 0 )
lastNameInput.OnEnter = function()
	dialog:OnAccept()
	dialog:Close()
end

function dialog:OnAccept()
	local firstName = firstNameInput:GetText()
	local lastName = lastNameInput:GetText()
	print( "Hello, " .. firstName .. " " .. lastName )
end
```

image

---

## Custom accept/reject button texts

```lua
local dialog = vgui.Create( "DDialogBox" )
dialog:SetIcon( "icon16/eye.png" )
dialog:SetTitle( "Example Dialog" )
dialog:SetText( "This is an example dialog." )
dialog:SetRejectButtonEnabled( false )
dialog:SetAcceptButtonText( "I'm a little teapot" )
```

image


neico:
Just what I was looking for earlier (not that it's hard to do yourself)

The variable naming (as in, the casing) might not be completely consistent with the rest of the codebase, but then again, much of the vgui element code aren't...

Why do you introduce a new `AddPanel` instead of the commonly used `Add` though?
Also, why `SetLabelText` and not `SetMessage` or `SetContent`, and when no message is set it should implicitly disable the label instead of the extra function
Generally, it might need a few more tweaks to be in line with the rest of the vgui elements, because I don't see having to explicitly call `Invalidate` being a good idea

And maybe provide a `GenerateExample` that has a few settings and a button to open the dialog, so it's easy to see and how to use going forward

colemclaren:
This is great addition, amazing job. The above comment sums up pretty much everything. I don't think having to call Invalidate all the time is very noob friendly either. You could add assert checks for the docking.

jorjic:
> Also, why `SetLabelText` and not `SetMessage` or `SetContent`

Because I named it something even worse before `SetLabelText`. 

> when no message is set it should implicitly disable the label instead of the extra function

Good idea, I'll add that in, thanks.

> Why do you introduce a new `AddPanel` instead of the commonly used `Add` though?

I was a little afraid of overriding `Add` in case somebody wanted to do something outside of what `AddPanel` does.

> I don't see having to explicitly call Invalidate being a good idea

I wrote a majority of this many months ago. I think I struggled with getting it to work any other way. I'll try again, but if anyone can point out how to make it work like this, it would be great not to require `Invalidate`.

> And maybe provide a GenerateExample that has a few settings and a button to open the dialog, so it's easy to see and how to use going forward

I tried to do this before opening the PR, but it wasn't showing up in `derma_controls`. I'll try again in case restarting fixes it.

jorjic:
Added GenerateExample:

image

garryspins:
Adding this to the base game seems kind of silly to me for a few reasons.
1. I don't actually know how useful this will be, or how often it will be used
2. Most people will just rewrite this for their own reasons anyway, having a builtin component for it is kind of redundant, its something that will only really be used by beginners.
3. Its very extra to have a whole separate component for this, it is just a DFrame with a few buttons, maybe if it had some more functionality.

neico:
Your second point could be applied to pretty much all existing elements, so that's rather moot

The third point is valid, but at the same time saves a bit of trouble if all you want is a quick (modal) dialog to notify or warn the user of something.
Having a standardized and customizable element in the base allows you to worry less about it and just use it.

Pretty much all gui frameworks/libraries in existence have such an element, so I don't see why this shouldn't.
And yes, even VGui has one, but with the more commonly used name: [MessageBox](https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/public/vgui_controls/MessageBox.h)

Bringing this code more in line with that one would be ideal.
Especially because you can port most existing derma elements (and the stuff you make with it) into native VGui without much adjustment, as most Methods are identical.

garryspins:
My second point can and does apply to most elements, except most of those elements are actually somewhat useful or have some sort of functionality that's relatively complex to emulate, DFrame dragging and resizing, DScrollPanel's... scrolling, DNumSlider, ect. This doesn't stick out as an element that has any real pros to it, there's nothing beginners will gain, nonbeginners will rewrite the element in the first place because there's nothing complex about it.

It saves time for beginners, I don't mean to be insulting but if you have a hard time adding 2 buttons to a DFrame.... you have bigger things to worry about.

Also you seem to be forgetting the quick and dirty popups we have (`Derma_Message`, `Derma_Query` and `Derma_StringRequest`), want to warn a user? great use these, this just adds an extra step that's not all that powerful.

I dont have a problem with helping beginners, I do have a problem with cluttering the codebase with just more useless stuff. 

tl;dr we should remove DPanel.

neico:
Those 3 functions are actually quite useful, but it feels like they aren't really highlighted much. (First time I've heard of them)
Especially because most Derma* functions are mostly for internal use of some sort.
And looking at their code shows quite a bit of duplicated code, which could be simplified with this element present.

A drawback on those however is that they are all modals (so they'd cover my use cases at least)
But this element is non-modal by default, giving the ability for more use cases than simple notifications

"cluttering the codebase with useless stuff" is quite subjective though

garryspins:
You skipped over the main point, which is that this wont add anything really useful. And _not_ being a modal isn't a feature, it really does just become a DFrame with a few buttons at that point.

jorjic:
I reinvented this wheel often enough that I decided it was worth spending time writing a generic implementation of it 18 months before buffing away my special use case functionality to submit it as a pull request. What if somebody decides to continue re-inventing the wheel anyways? Believe it or not, there is very little I can do about that.

garrysmod/lua/vgui/ddialogbox.lua
```diff
@@ -0,0 +1,203 @@
...
+	lastNameInput.OnEnter = function()
+		dialog:OnAccept()
+		dialog:Close()
+	end
+
+	propertySheet:AddSheet( className, dialog, nil, true, true )
+end
+
+function PANEL:OnClose()
+	if not self.handled then
```
robotboy655, line 49: Code formatting is not compliant with [CONTRIBUTING.md](https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md)

Specifically missing brackets around the `if-` block. There are more instances that I have not pointed out.

garrysmod/lua/vgui/ddialogbox.lua
```diff
@@ -0,0 +1,203 @@
...
+-- See: https://en.wikipedia.org/wiki/Dialog_box
+
+function PANEL:Init()
+	self:SetTitle( "Untitled Dialog Box" )
+	self:SetSize( 400, 200 )
+	self:MakePopup()
+	self:Center()
+	self:SetDrawOnTop( true )
+	
+	self.children = {}
```
robotboy655, line 13: Is there even a point of having this table, since you are putting every child you create into it?

You can just use `Panel:GetChildren()`, or the built-in `Panel:Clear()` for what this variable is being used for.

garrysmod/lua/vgui/ddialogbox.lua
```diff
@@ -0,0 +1,203 @@
...
+end
+
+-- The reject button is the choice button that denies the requested action
+function PANEL:GetRejectButtonEnabled()
+	return self.isRejectButtonEnabled
+end
+
+function PANEL:SetRejectButtonEnabled( value )
+	self.isRejectButtonEnabled = value
+	self:RebuildContents()
```
robotboy655, line 173: I think this way of doing this unnecessarily recreates all the panels multiple times during panel creation, when you call multiple functions, like `SetRejectButtonEnabled`, `SetRejectButtonText`, etc. on the same panel.

I think creating your panels once in `PANEL:Init` and then toggling visibility/`SetText`ing in these setters would be a better idea, or doing something similar to how https://wiki.facepunch.com/gmod/PANEL:PerformLayout functions (i.e. using a 'dirty' flag and rebuilding only once per frame).

robotboy655 (changes requested):
I have a few issues with the implementation of this, but I think overall it's a good idea.

PR #1918, opened 2022-10-25
New way to move EyePoser, click and drag panel

DanielSvoboda:
When selecting the eyes of an entity by right clicking it is possible to:
Set the position from where you are looking through a panel; 
Where it is also possible, I left a squint, in or out. 

Did not understand? See the  
Demo video:  

https://user-images.githubusercontent.com/28203966/197897650-18053b35-84e9-4527-974b-8068f7f0bc71.mp4

The "TOOL:RightClick" function has been changed to "TOOL:Reload" 
Because the new function will be in "TOOL:RightClick"

This is the panel:
The_Panel_EyePoser

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/eyeposer.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/eyeposer.lua
@@ -4,8 +4,23 @@ TOOL.Name = "#tool.eyeposer.name"
 
 TOOL.Information = {
 	{ name = "left" },
-	{ name = "right" }
+	{ name = "right" },
+	{ name = "reload" }
 }
+
+if ( CLIENT ) then
+	CreateClientConVar( "eye_x", "0", false, false )
+	CreateClientConVar( "eye_y", "0", false, false )
+	CreateClientConVar( "eye_z", "360", false, false )
+	CreateClientConVar( "eye_strabismus" , "false", false, false )
+else	
+	util.AddNetworkString( "EyeMouseClickSide" )
+end
+
+local eye_x = GetConVar( "eye_x" )
+local eye_y = GetConVar( "eye_y" )
+local eye_z = GetConVar( "eye_z" )
+local eye_strabismus = GetConVar( "eye_strabismus" )
 
 local function SetEyeTarget( Player, Entity, Data )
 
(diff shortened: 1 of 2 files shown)


DanielSvoboda:
On line 188 inside the ```function TOOL:Reload( trace )```  , add this:
```lua 
        SetEyeTarget( self:GetOwner(), ent, { EyeTarget = vector_origin } )
	RunConsoleCommand( "eyeposer_x", 0.5 )
	RunConsoleCommand( "eyeposer_y", 0.5 )
	RunConsoleCommand( "eyeposer_strabismus", 0 )
```
Yes, the middle button on the panel already returns to the default, 
however the `R` button does not work after clicking with RightClick, and this is a little uncomfortable.

robotboy655:
What do you mean `the R button does not work after clicking with RightClick`? Reload deselects the entity, so it should be fine? It's really meant to undo left clicks, especially on NPCs.

DanielSvoboda:
Yes, it works well on NPC, however the ragdoll does not return to default using RightClick:

On an NPC's: If you use the LeftClick or RightClick functions and then use TOOL:Reload , it works normally.
In a ragdol:If you use RightClick and then try to use TOOL:Reload, it is not reset to default.

In my opinion TOOL:Reload should work at any time, even knowing that with RightClick, there is a possibility to reset to default using MOUSE_MIDDLE. Adding the RunConsoleCommand as I mentioned seems to solve this.

https://user-images.githubusercontent.com/28203966/213609520-64e7b206-64c0-4e7d-a610-a306f0f215ec.mp4

robotboy655:
Fixed that issue in https://github.com/Facepunch/garrysmod/commit/313c02fe19621c3d5372f896193ba44bd55ea302 thanks for reporting it.

DanielSvoboda:
Excellent thank you, Just two observations:
1. That the comment about TOOL:Reload on line 176 is in line 153. 
And in place of line 176 a new comment is needed.

2. It is necessary to include the slider limit in order not to leave the 'SliderBackground',
I don't know if these magic numbers are the best way, but it solves it.

Remove this:
EyeSlider:Dock( FILL )

Replace with this:
EyeSlider:SetPos( 7, 7 )
EyeSlider:SetSize( 282 , 211)

image

robotboy655:
https://github.com/Facepunch/garrysmod/commit/a17058cb524e2052ff1dd68c3edd43c699f42f34

It's perfect now.

PR #1922, opened 2022-11-29
Improved net.WriteType and net.ReadType. reduced bits for net.ReadEntity and net.WriteEntity

RaphaelIT7:
## net.WriteType and net.ReadType
Reduced the amount of bits networked for each type from 8 bits to 4 bits.

## net.WriteEntity and net.ReadEntity
Reduced the amount of bits from 16 to 14 because EntIndex doesn't seem to return a value above, 8127

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -1,6 +1,17 @@
 
 TYPE_COLOR = 255
 
+local NET_TYPE_NIL = 0
+local NET_TYPE_STRING = 1
+local NET_TYPE_NUMBER = 2
+local NET_TYPE_TABLE = 3
+local NET_TYPE_BOOLEAN = 4
+local NET_TYPE_ENTITY = 5
+local NET_TYPE_VECTOR = 6
+local NET_TYPE_ANGLE = 7
+local NET_TYPE_VMATRIX = 8
+local NET_TYPE_COLOR = 9
+
 net.Receivers = {}
 
 --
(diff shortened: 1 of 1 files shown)


Riddle1001:
Takes away to much readability

RaphaelIT7:
I forgot to clean it up. Has now been done

> Takes away to much readability

neico:
I'm not sure using "magic numbers" instead of "constants" is such a good idea, at the very least it reduces readability as mentioned.

Nevermind that `TypeID()` also supports custom types created by addons and modules, and the original code allows extending those functions for those, so the entire 255 type ID range should be kept.
GMod effectively has support for 16384 entities, of which 8192 are edicts (and the rest being realm specific entities), so you really shouldn't limit the entity bits.

If you follow that, then this whole PR becomes pointless.

RaphaelIT7:
> I'm not sure using "magic numbers" instead of "constants" is such a good idea, at the very least it reduces readability as mentioned.

If it is wished that Enums should be used, I should add it like this:
```lua
NET_TYPE_NIL = 0
NET_TYPE_STRING = 1
NET_TYPE_NUMBER = 2
NET_TYPE_TABLE = 3
NET_TYPE_BOOLEAN = 4
NET_TYPE_ENTITY = 5
NET_TYPE_VECTOR = 6
NET_TYPE_ANGLE = 7
NET_TYPE_MATRIX = 8
```

<hr>

> Nevermind that TypeID() also supports custom types created by addons and modules, and the original code allows extending those functions for those, so the entire 255 type ID range should be kept.

TypeID doesn't even use the full range of 255 and the TYPE_COLOR won't be reached because it will return [TYPE_TABLE](https://wiki.facepunch.com/gmod/Enums/TYPE#TYPE_TABLE)
before it. 

I will change the Net_Types to net.Types so addons can add their custom types to It. With 4 bits(0 - 15 range) and 9 numbers used by default, they can add 6 custom types and functions to the net.WriteVars, net.ReadVars and net.Types table.
> Colors are implemented in Lua thus [type](https://wiki.facepunch.com/gmod/Global.type)(color) == "table" and [TypeID](https://wiki.facepunch.com/gmod/Global.TypeID)(color) == [TYPE_TABLE](https://wiki.facepunch.com/gmod/Enums/TYPE#table).(https://wiki.facepunch.com/gmod/Color)
<hr>

> GMod effectively has support for 16384 entities, of which 8192 are edicts (and the rest being realm specific entities), so you really shouldn't limit the entity bits.

EntIndex will return -1 for realm specific entities, and it doesn't seem to return a value over 8127 which is in the 13 bit range(0 - 8191). Currently, it uses 14 bits(0 - 16384 range), so if it should ever go over it will be networked correctly.
There is no need to use a 16 bit UInt (0 - 65535 range)

RaphaelIT7:
> I'm sure there is also a compatibility problem, the length is changed.

It works completly fine. Only net.ReadType will read the UInt.

<hr>

> Magic numbers aren't better than globals. Localize instead.

Now added localized Enums below TYPE_COLOR

<hr>

> I don't see the point in NetType indirection using a table

It is a cleaner way than doing an obsessive amount of if checks, and It allows the adding custom Types

robotboy655:
I could agree to `net.Read/WriteEntity` changes, but I do not like the idea of potentially breaking existing code, or trying to optimise a function that realistically should be avoided (`net.WriteTable`).

PR #1732, opened 2020-10-13
Markup Library Improvements

WilliamVenner:
- New argument to `MarkupObject:Draw()` (`textAlign`) allows for text alignment inside the maxWidth bounds (i.e. align text center, align text right)
- Added `MarkupObject:GetMaxWidth()`
- Added `markup.Color` which converts a Color into a string for use in `<color>` tags (this does not modify the original Color object)
- #1614 Added `markup.Escape` which escapes markup-sensitive characters (`& < >`)
- Fixed trailing spaces on new lines
- Fixed issue where new blocks were unnecessarily being split in the middle of a word
- Changed pairs to ipairs
- Improved code readability, cleaned up comments, removed a redundant comment
- Removed some unused variables
- Some small optimizations

#### `TEXT_ALIGN_CENTER`
TEXT_ALIGN_CENTER

UTF-8 Test

#### `TEXT_ALIGN_RIGHT`
TEXT_ALIGN_RIGHT

#### Default

Default

Default

PR #1924, opened 2022-12-06
Fix markup alignment inconsistency

danx91:
## tl;dr
Changed text position calculation, so it is aligned as expected while using center or right text alignment.

## Description
Currently, markup object is aligned according to it's size (`self.totalWidth`) but text is aligned according to maximum size (`self.maxWidth`). This results in misaligned text while using center or right alignment. 
offcentered
_Image of current markup behavior with `halign` and `textAlign` both set to `TEXT_ALIGN_CENTER` (white line - center of frame, green box - size of markup object, red box - max size of markup object)._

As visible on the image, the markup object itself is aligned correctly, but the text is off-centered. The same problem occurs while using right alignment, making the text aligned to the right side of the red box.

correct
_Image of changed markup behavior with `halign` and `textAlign` both set to `TEXT_ALIGN_CENTER` (white line - center of frame, green box - size of markup object, red box - max size of markup object)._

Now the text is also aligned according to actual size, resulting in text aligned perfectly in the center of the frame, and it also works correctly with `TEXT_ALIGN_RIGHT`.

## Possible backward compatibility issues
Some developers may have corrected misaligned markup on their own, so changing this will cause the markup text to be 'corrected twice' and making it off-centered for them. However, in my opinion, current behavior is wrong and should be changed to work as expected - centering both markup object and text should produce a perfectly centered result.

### DIFF
--- a/garrysmod/lua/includes/modules/markup.lua
+++ b/garrysmod/lua/includes/modules/markup.lua
@@ -251,9 +251,9 @@ function MarkupObject:Draw( xOffset, yOffset, halign, valign, alphaoverride, tex
 			local lineWidth = self.lineWidths[ blk.offset.y ]
 			if ( lineWidth ) then
 				if ( textAlign == TEXT_ALIGN_CENTER ) then
-					surface.SetTextPos( x + ( ( self:GetMaxWidth() - lineWidth ) / 2 ), y )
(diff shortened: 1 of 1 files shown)


robotboy655 (approved):
Seems the issue was introduced by https://github.com/Facepunch/garrysmod/pull/1732 when the text align argument was added.

PR #1925, opened 2022-12-20
Update string.lua

SuperCALIENTITO:
Added an alternative to string.Comma to replace the pattern with dots instead of commas


CornerPin:
Could you make it accept a custom separator instead? This will make the function more generic and it would be useful if someone wants to separate digits with `'` or a whitespace character.

SuperCALIENTITO:
Done!

robotboy655 (approved):

PR #1928, opened 2023-01-14
Fix for infinite resizing in `DScrollPanel`

PrikolMen:
I have long played Garry's Mod and often notice that the panels in the Context Menu are doing weird things, it is very confusing, so I fixed it, and it would be nice if it was fixed for everyone.

This bug can be seen in the materials menu on any `sandbox` based gamemode.

### DIFF
--- a/garrysmod/lua/vgui/dscrollpanel.lua
+++ b/garrysmod/lua/vgui/dscrollpanel.lua
@@ -5,20 +5,23 @@ AccessorFunc( PANEL, "Padding", "Padding" )
 AccessorFunc( PANEL, "pnlCanvas", "Canvas" )
 
 function PANEL:Init()
-
 	self.pnlCanvas = vgui.Create( "Panel", self )
-	self.pnlCanvas.OnMousePressed = function( self, code ) self:GetParent():OnMousePressed( code ) end
+	self.pnlCanvas.OnMousePressed = function( pnl, code )
+		pnl:GetParent():OnMousePressed( code )
+	end
+
 	self.pnlCanvas:SetMouseInputEnabled( true )
 	self.pnlCanvas.PerformLayout = function( pnl )
-
 		self:PerformLayoutInternal()
 		self:InvalidateParent()
-
 	end
 
 	-- Create the scroll bar
-	self.VBar = vgui.Create( "DVScrollBar", self )
-	self.VBar:Dock( RIGHT )
+	local scrollBar = vgui.Create( "DVScrollBar", self )
+	if IsValid( scrollBar ) then
+		self.VBar = scrollBar
+		scrollBar:Dock( RIGHT )
+	end
 
 	self:SetPadding( 0 )
 	self:SetMouseInputEnabled( true )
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dscrollpanel.lua
```diff
@@ -5,20 +5,23 @@ AccessorFunc( PANEL, "Padding", "Padding" )
...
 		self:PerformLayoutInternal()
 		self:InvalidateParent()
-
 	end
 
 	-- Create the scroll bar
-	self.VBar = vgui.Create( "DVScrollBar", self )
-	self.VBar:Dock( RIGHT )
+	local scrollBar = vgui.Create( "DVScrollBar", self )
+	if IsValid( scrollBar ) then
```
robotboy655, line 21: Code formatting style is inconsistent with the rest of the repository. (Many more cases of this not commented on that should also be changed)

garrysmod/lua/vgui/dscrollpanel.lua
```diff
@@ -27,119 +30,106 @@ function PANEL:Init()
 	self:SetPaintBackgroundEnabled( false )
 	self:SetPaintBorderEnabled( false )
 	self:SetPaintBackground( false )
-
 end
 
 function PANEL:AddItem( pnl )
-
```
robotboy655, line 34: Pointless whitespace changes (Many other instances not commented on)

garrysmod/lua/vgui/dscrollpanel.lua
```diff
@@ -27,119 +30,106 @@ function PANEL:Init()
...
 	self:AddItem( child )
-
 end
 
 function PANEL:SizeToContents()
-
-	self:SetSize( self.pnlCanvas:GetSize() )
-
+	local canvas = self:GetCanvas()
+	if IsValid( canvas ) then
```
robotboy655, line 45: Could be changed to an early return

garrysmod/lua/vgui/dscrollpanel.lua
```diff
@@ -27,119 +30,106 @@ function PANEL:Init()
...
 
-	local x, y = self.pnlCanvas:GetChildPosition( panel )
-	local w, h = panel:GetSize()
-
-	y = y + h * 0.5
-	y = y - self:GetTall() * 0.5
-
-	self.VBar:AnimateTo( y, 0.5, 0, 0.5 )
+	local canvas = self:GetCanvas()
+	if IsValid( canvas ) then
```
robotboy655, line 80: Again, could be changed to an early return to prevent scope pyramids

garrysmod/lua/vgui/dscrollpanel.lua
```diff
@@ -27,119 +30,106 @@ function PANEL:Init()
...
+		local canvasHeight = canvas:GetTall()
+		local width = self:GetWide()
+
+		self:Rebuild()
+
+		local vbar = self:GetVBar()
+		if IsValid( vbar ) then
+			local height = self:GetTall()
+			vbar:SetUp( height, canvasHeight )
+			if vbar.Enabled and (canvasHeight < height) then
```
robotboy655, line 114: This check will almost never pass as `vbar.Enabled` is set to `canvasHeight > height`, resulting in bugged spacing:

image

(Note the text entry going behind the scrollbar)

garrysmod/lua/vgui/dscrollpanel.lua
```diff
@@ -27,119 +30,106 @@ function PANEL:Init()
...
-	YPos = self.VBar:GetOffset()
-
-	if ( self.VBar.Enabled ) then Wide = Wide - self.VBar:GetWide() end
-
-	self.pnlCanvas:SetPos( 0, YPos )
-	self.pnlCanvas:SetWide( Wide )
-
-	self:Rebuild()
-
-	if ( Tall != self.pnlCanvas:GetTall() ) then
```
robotboy655, line 127: This was removed for seemingly no good reason?

robotboy655 (changes requested):
Please make your changes consistent with the rest of the code in the repository.

garrysmod/lua/vgui/dscrollpanel.lua
PrikolMen: in my logic it makes no sense, it is already rebuilded above

garrysmod/lua/vgui/dscrollpanel.lua
PrikolMen: Hmmm, I must have missed it somewhere, I'll try to fix it on this branch

garrysmod/lua/vgui/dscrollpanel.lua
PrikolMen: Well, if it's so important, it's not hard for me to do a style match, I'll do it in the next commit

garrysmod/lua/vgui/dscrollpanel.lua
robotboy655: The comment is not about the Rebuild call, but regarding that call, if the size of the canvas has changed - you gotta rebuild because the size and position of everything on the canvas may have changed, which is in fact what is causing the "infinite loop" problem you are trying to solve.

It's not as simple as removing that function call.

garrysmod/lua/vgui/dscrollpanel.lua
PrikolMen: I will do it, chance that it will not be there is extremely small...

garrysmod/lua/vgui/dscrollpanel.lua
PrikolMen: I don't see anything wrong with additional checks, but given that they weren't there originally, I'll remove them

garrysmod/lua/vgui/dscrollpanel.lua
robotboy655: You will have to follow code formatting style of the repository if you wish your code to be merged. Simple as that.

The code formatting style is chosen in such a way to be easier to read.

garrysmod/lua/vgui/dscrollpanel.lua
robotboy655: As per my previous message, the problem is not the presence of additional checks, but the fact that they can be implemented in a way that makes the code easier to understand and read.

i.e. Instead of doing this:
```
local canvas = self:GetCanvas()
if IsValid( canvas ) then
	if ( blah ) then
		if ( foo ) then
			-- Code
		end
	end
end
```

You would do this:
```
local canvas = self:GetCanvas()
if ( !IsValid( canvas ) ) then return end

if ( blah ) then
	if ( foo ) then
		-- Code
	end
end
```

Or even this if/when possible
```
local canvas = self:GetCanvas()
if ( !IsValid( canvas ) ) then return end

if ( !blah ) then return end

if ( !foo ) then return end

-- Code
```

PrikolMen: Yes, alright, I'll keep that in mind.

garrysmod/lua/vgui/dscrollpanel.lua
PrikolMen: No, I understand when the style has to match the repository style, but wouldn't it be easier to just remove all those spaces, it's a bunch of extra characters that take up space and make no sense, why don't you just remove them all from the repository, they are not part of the code..

PR #1932, opened 2023-01-19
Changes in ukrainian.lua

ErickMaksimets:
It looks like Garry's mod development continues. Well, it is necessary to correct the translation standards for the release version. Over the pastime, user Aliemann has proposed a number of stylistic updates to the translation which I have supported. Also found several hidden mistakes. Many words have been removed because they can be ignored for brevity.

I hope to be able to contact more translators to make a perfect version of the translation in the future. If they offer only stylistic updates to the text and don't find mistakes, then I will ignore it(spam is bad).


Satton2:
@svdm

UAVXP:
> нотатку
> воно

PR #1937, opened 2023-01-26
Fixed ipairs being nil and MarkupObject no longer caring about "paragraph" alignment

StarLight-Oliver:
You get errors when adding your own effects as ipairs is nil in the module.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/base_gmodentity.lua
@@ -63,6 +63,10 @@ function ENT:GetOverlayText()
 
 	local PlayerName = self:GetPlayerName()
 
+	if ( !PlayerName or PlayerName == "" ) then
+		return txt
+	end
+
 	return txt .. "\n(" .. PlayerName .. ")"
 
 end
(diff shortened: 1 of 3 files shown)


StarLight-Oliver:
Alignment has been broken, someone wanted the textAlign argument of MarkupObject:Draw to do the same thing as halign and valign args do.
This feels like a bug but if it isn't just revert my change.

Before the update
image

After the update
image

StarLight-Oliver:
```
gamemodes/sandbox/entities/entities/base_gmodentity.lua:66: attempt to concatenate local 'PlayerName' (a nil value)
    1. GetOverlayText - gamemodes/sandbox/entities/entities/base_gmodentity.lua:66
        2. unknown - gamemodes/sandbox/entities/entities/base_gmodentity.lua:38
```
Someone got this error so fixed it as well. Though I don't think the update is the cause for this

robotboy655:
Regarding markup alignments, you were using the bugged behavior to center your text, which I would argue is the incorrect approach. No, it doesn't "gives us fewer ways we can align our text.", if you use the functionality properly, namely set the `xAlign` argument to `TEXT_ALIGN_CENTER`.

It does not do the same thing as `textAlign` argument. The example of the wiki shows the difference and the difference is still there.

https://wiki.facepunch.com/gmod/MarkupObject:Draw

robotboy655 (approved):

PR #1938, opened 2023-01-26
Add math.Factorial

wasied:
Nothing too complicated. I needed a function to calculate the factorial value of a number and I couldn't find any existing function to calculate it. So I added it in the math library, I think it could be useful.

```lua
print( math.Factorial(-5) ) -- nil (factorial must be >= 0)
print( math.Factorial(0) ) -- 1
print( math.Factorial(1) ) -- 1
print( math.Factorial(5) ) -- 120
```

I decided to return nil if the value is negative. I don't know if it's consistent enough to you.


TomDotBat:
Your usage of recursion in this solution makes it extremely slow, a for loop is much faster:
```lua
function math.Factorial( num )

	if ( num < 0 ) then
		return nil
	elseif ( num < 2 ) then
		return 1
	end

	local res = 1

	for i = 2, num do
		res = res * i
	end

	return res

end
```

MaaxIT:
Yeah you're right. After benchmarking, the difference is very small but it is still interesting if you call it a lot. Edited in changes, thank you!

Kefta:
This is kinda misleading unless ints are exposed from ffi; I'd expect this to return the decimal factorial which would/should require implementing a basic gamma function

CornerPin:
You forgot to change the comment after you converted the function to use a `for` loop.

smokingplaya:
loop would have been better, because with a loop you did not have a stack overflow

MaaxIT:
I'm actually using a loop, check out 551d4240943ea6c9f373d1c0e4b0d97f980852fb

robotboy655 (approved):

PR #1941, opened 2023-02-03
concommand.Exists( name ) & little improvements

PrikolMen:
Added a new function, and corrected the localization.


Kefta:
It also had an error - ``commandList`` instead of ``CommandList``.

PR #1942, opened 2023-02-10
Minor Fix: Prevent medkit timer spam

brandonsturgeon:
### The problem
Medkits currently create a new timer every time they're spawned.
In the `OnRemove` ent hook, the timers are `.Stop`'d, but not `.Remove`'d.
On servers where people die and respawn with a medkit frequently, this can lead to situations like this:
image

They're _probably_ fine, but.. why have 'em?

### This PR
I changed `timer.Stop` to `timer.Remove` in `OnRemove`, which will properly clean up medkit timers when they're removed.

I also returned early for `CLIENT` in `OnRemove` because the timers only exist on `SERVER`.

### DIFF
--- a/garrysmod/lua/weapons/weapon_medkit.lua
+++ b/garrysmod/lua/weapons/weapon_medkit.lua
@@ -126,8 +126,10 @@ end
 
 function SWEP:OnRemove()
 
-	timer.Stop( "medkit_ammo" .. self:EntIndex() )
-	timer.Stop( "weapon_idle" .. self:EntIndex() )
+	if ( CLIENT ) then return end
+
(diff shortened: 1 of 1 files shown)


Kefta:
I believe these timers can be removed altogether for using SWEP:Think instead. I will make a separate PR for this next week if you don't get to it first.

brandonsturgeon:
That sounds like the right way to do it - good idea.

Kefta:
Knocked it out but it still needs some testing https://github.com/Facepunch/garrysmod/pull/1943

robotboy655:
Merging this for now.

PR #1943, opened 2023-02-10
Refactor weapon_medkit

Kefta:
- Converted the ammo/idle timers to predicted netvars like weapon_fists
- Divided up the Primary/SecondaryAttack functions into SWEP:CanHeal(entity), SWEP:DoHeal(entity), SWEP:HealEntity(entity), and SWEP:HealFail(entity)
- Changed many constants to class variables to promote addon SWEP modification instead of copying and modifying the code
- Heal success/failures now punishes both primary and secondary fire no matter which heal type it is to prevent being able to interrupt the medpack animation with the opposite heal type attack
- Regen time is now tied to when the last heal was performed or when the medkit was deployed instead of being arbitrarily tied to a timer's creation time
- Removed SWEP.MaxAmmo as SWEP.Primary.ClipSize was already serving its purpose
- General optimisations/cleanup

Supersedes https://github.com/Facepunch/garrysmod/pull/1942.

### DIFF
--- a/garrysmod/lua/weapons/weapon_medkit.lua
+++ b/garrysmod/lua/weapons/weapon_medkit.lua
@@ -108,7 +108,7 @@ end
 function SWEP:HealEntity( ent, amount )
 
 	-- Heal ent
-	self:SetClip1( math.min( self:GetClip1() - amount, 0 ) )
+	self:SetClip1( math.max( self:Clip1() - amount, 0 ) )
 	ent:SetHealth( ent:Health() + amount )
 
 	-- Do effects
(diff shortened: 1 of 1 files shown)


Kefta:
Tested in local multiplayer and works without regressions as far as I can tell other than the changed behaviour noted.

Kefta:
Can this please be reviewed? There's really not that many built-in weapons and it would be good to have at least one example of SWEP design that takes runtime modification into account

robotboy655 (approved):

garrysmod/lua/weapons/weapon_medkit.lua
```diff
@@ -18,135 +18,178 @@ SWEP.UseHands = true
...
+	if ( isplayer ) then
 		owner:LagCompensation( true )
 	end
 
+	local startpos = owner:GetShootPos()
 	local tr = util.TraceLine( {
-		start = owner:GetShootPos(),
-		endpos = owner:GetShootPos() + owner:GetAimVector() * 64,
+		start = startpos,
+		endpos = startpos + owner:GetAimVector() * 64,
```
2048khz-gachi-rmx, line 74: I believe `self.HealRange` was intended to be used here

Kefta: You're right

PR #1944, opened 2023-02-11
Allow suppressing or overriding halo rendering

Python1320:
May be used to reimplement halos without having to override the library or for situations where halos may be clashing with aesthetics or rendering (better to hide halos temporarily than have them glow through areas or force people to hook the halo library).

### DIFF
--- a/garrysmod/lua/includes/modules/halo.lua
+++ b/garrysmod/lua/includes/modules/halo.lua
@@ -144,9 +144,13 @@ function Render( entry )
 	render.SetStencilReferenceValue( 0 )
 end
 
+function GetTable()
+	return List	
+end
+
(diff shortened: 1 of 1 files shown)


Kefta:
Add parentheses around the if condition.

Kefta (approved):

storm37000:
`List` will memory leak really badly if you stop the hook at that position.  Halos would still be added with `Add` function but not cleared from the list.

garryspins:
Should probably use a new hook, `CanDrawHalo(entry)` for verbosity and backwards compatibility sake as some random addon might return false making issues hard to diagnose in the event of this being merged.

robotboy655:
Closing as stale, since it's been a while and @storm37000 's comment has not been addressed.

PR #1939, opened 2023-01-29
TTT: more translatability improvements and fixes

r47orr:
- Implement actual translations for TTT start new game menu

- Implement actual translations for mute states notifications on chat, thus removing its hardcodes

- Remove hardcodes and refactor hl2 ents replacements for maps with no spawnscripts

- Suggest new effect and sound for discombulator and remove its pulling function related to non-players (only push stuff away)

- This is a complement for c833b5892875a7c7c0a1fdf112d1d34653a4669d by @svdm, trying to fix Facepunch/garrysmod-issues#3544 (one of the pendencies from it)

PS: I'd also recommend transfering all of the TTT lang files to Crowdin, to improve the translatability for the translator's community. I've tested it the english lang file, and it works. Not as straightforward as the rest of the translation, it will involve respecting the syntax and stuff like that, but it could work pretty well.

PR #1946, opened 2023-02-15
TTT: more translatability improvements and fixes²

r47orr:
This is a remake of the #1939 PR. As per request by @svdm, only the translations part has been sent in the PR, to avoid changes to the base game mode.

Just need a maintainer to add the TTT label to it. Thanks in advance.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua
@@ -235,6 +235,18 @@ local function TraitorGlobalVoice(ply, cmd, args)
 end
 concommand.Add("tvog", TraitorGlobalVoice)
 
+-- from shared.lua
+-- MUTE_NONE = 0
+-- MUTE_TERROR = 1
+-- MUTE_ALL = 2
+-- MUTE_SPEC = 1002
+local MuteModes = {
+   [0] = "mute_off",
+   [1] = "mute_living",
+   [2] = "mute_all",
+   [1002] = "mute_specs"
+}
+
 local function MuteTeam(ply, cmd, args)
    if not IsValid(ply) then return end
    if not (#args == 1 and tonumber(args[1])) then return end
@@ -246,13 +258,9 @@ local function MuteTeam(ply, cmd, args)
    local t = tonumber(args[1])
    ply.mute_team = t
 
-   if t == MUTE_ALL then
-      ply:ChatPrint("All muted.")
-   elseif t == MUTE_NONE or t == TEAM_UNASSIGNED or not team.Valid(t) then
-      ply:ChatPrint("None muted.")
-   else
-      ply:ChatPrint(team.GetName(t) .. " muted.")
-   end
+   -- remove all ifs
+   LANG.Msg(ply, MuteModes[t])
+   
 end
 concommand.Add("ttt_mute_team", MuteTeam)
 
--- a/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua
@@ -235,16 +235,11 @@ local function TraitorGlobalVoice(ply, cmd, args)
 end
 concommand.Add("tvog", TraitorGlobalVoice)
 
--- from shared.lua
--- MUTE_NONE = 0
--- MUTE_TERROR = 1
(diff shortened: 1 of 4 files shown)


garrysmod/gamemodes/terrortown/gamemode/gamemsg.lua
```diff
@@ -235,6 +235,18 @@ local function TraitorGlobalVoice(ply, cmd, args)
...
+-- MUTE_NONE = 0
+-- MUTE_TERROR = 1
+-- MUTE_ALL = 2
+-- MUTE_SPEC = 1002
+local MuteModes = {
+   [0] = "mute_off",
+   [1] = "mute_living",
+   [2] = "mute_all",
+   [1002] = "mute_specs"
+}
```
svdm, line 248: Wouldn't it be possible to use MUTE_NONE etc as the indexes here directly? `[MUTE_NONE] = "mute_off"` and so on.

r47orr: Oh, yes, I've made a mistake when testing. I didn't use `[]` around the indexes, and using those indexes was returning an error. But I just tested it and it works fine. Will update it

r47orr: I also thought about adding translations to the last words. But it ended up being a bit more complicated, since the translation functions only work client-side. Although I could "fetch" a translation by storing `LANG.Msg()` in a variable, which felt pretty weird and stupid for me lmao

What do you think about it?

svdm: I think the way to do that would be to add the player's death type to the `TTT_LastWordsMsg`, and then do the interpretation (with `LastWordContext[ply.death_type]` etc) in the clientside handling (in cl_voice). Then retrieving a translation is trivial.

Of course, last words is off by default and probably quite rare in practice (voicechat is more popular, it's rare to die while typing, etc) so maybe it's not worth the effort.

r47orr: It doesn't sound like a bad idea for me. Actually could be a good project to practice some stuff.

By the way, could you do something about the TTT translation to move it to Crowdin? Since you don't see problems with that change, maybe you could ask the folks at Facepunch to implement it?

Satton2:
WOW! @robotboy655 does this mean we will finally be able to localize the TTT's text in the main menu (server creation) via Crowdin? If so, I can't wait to do it into Russian language and approve all the stuff!

r47orr:
> WOW! @robotboy655 does this mean we will finally be able to localize the TTT's text in the main menu (server creation) via Crowdin? If so, I can't wait to do it into Russian language and approve all the stuff!

Yes, if **everything** inside the file main_menu.properties goes to Crowdin, then yes; TTT's main menu text will now be translatable via Crowdin. But still, I hope we can have all of TTT's strings there for translation. Let's hope they implement it.

Satton2:
@Tiagoquix 
> as Crowdin's proofreaders are people hand-selected by the GMod development team.

I can't handle myself from laughing, since it's not actually true. 😅 You know, there was a time when project proofreaders were able to assign other proofreaders by themselves and who knows who assigned who due to that period of time since there were not any logs from our side.

When I become a project proofreader and when it was possible to assign others, me and another project proofreader took kinda long trip to search for any inactive or language-specifed proofreaders. Eventually, we changed some of their access to proofread a specific language only and not the project in general, or demoted some **very** inactive ones (more than 5-6 years) or abusive ones (there were just 2 as I remember), or assigned new ones (or just one? I believe that only one was assigned since others didn't get any trust from our side).

So, as I said in the mentioned PR, unless FP team assigns someone as a manager (currently, only managers can assign proofreaders on Crowdin) or will decide to handle assigning proofreaders on their own (which is probably not gonna happen any time soon due to only one dev that works on the game and has **a lot** of other tasks), I believe that GMod's localization will have some complicated problems like:
- old and outdated approves (I can actually unapprove any, but it would be complicated since I am **not** actually **allowed** to do so, Reports tab is now missing for me, and currently not often look in langauge-specific issues, since there is almost no activity).
- "localization war of votes" (or comments, eventually?).
- constant inconsistency in terminology and the translation.

But once again, I will say that I also would like to see TTT's localization on Crowdin. 😋

PR #1947, opened 2023-02-18
Check "CanProperty" for serverside editvariable

giinuu:
Currently, "CanProperty" is not checked server-side when using the "Edit Properties.." option in the context menu, meaning it cannot be blanket disabled by a sandbox derived gamemode.

Made change locally and tested on a dedicated server, now allows blocking "Edit Properties..."


robotboy655:
https://wiki.facepunch.com/gmod/GM:CanEditVariable ?

giinuu:
> https://wiki.facepunch.com/gmod/GM:CanEditVariable ?

You're absolutely correct, how I didn't notice this.. apologies!

PR #1952, opened 2023-03-10
remove math.clamp in think

rezyaw:


Kefta (approved):

robotboy655:
This doesn't handle duplicator, i.e. saving a dupe without the limits and then spawning the dupe on the server.

Kefta:
This should be reopened. At worst, an additional set of clamps can be added to the dupe loading function, and at best, the TOOL.ClientConVar loader can support table values that specify min/maxes which allow the clamping to be done in the engine.

robotboy655:
They can create a new pull request once they address the issue, can they not?

PR #1954, opened 2023-03-19
Crop DButton's image size if it's too big

doxzter:
It will allow higher resolution icons to be used with lower sized DButtons. For example, on first pic 72x72 icons used with DButtons before changes and after

pic1 pic2


robotboy655:
It should try to preserve aspect ratio when the image is not a square.

robotboy655 (approved):

PR #1955, opened 2023-03-28
Fix "bug" with serverlist panel

smokingplaya:
Before: https://prnt.sc/gzpGGMiVLBf3
After: https://prnt.sc/R8w5DfUr-aNK

Of course, it would be better to write a gmod menu in lua, it's very easy, but no one wants to rewrite it...


robotboy655:
What is the point of this exactly? The "bug" you are referring to is it being intentionally centered?

smokingplaya:
It looks like a bug. It's not just for me, a lot of people don't like it.

PR #1957, opened 2023-04-17
add util.ScreenToWorld

homonovus:
util.ScreenToWorld, a simplification and wrapper around util.AimVector

supply it a position on screen and it returns an accurate direction regardless of your FOV


robotboy655:
https://wiki.facepunch.com/gmod/gui.ScreenToVector

Recent changes to that function (FOV correction) should make this obsolete.

PR #1958, opened 2023-04-19
Fix error due to derma_icon_browser

callumok2004:
derma_icon_browser which was added in the latest patch was not added to send.txt, resulting in a file not found error

image


robotboy655:
Its already fixed in the game, I will update the code in the repo later today.

PR #1961, opened 2023-04-25
Fix weapon_medkit holstered healing regression

Kefta:
Calculate the health to regen on Deploy/Think.

### DIFF
--- a/garrysmod/lua/weapons/weapon_medkit.lua
+++ b/garrysmod/lua/weapons/weapon_medkit.lua
@@ -42,7 +42,7 @@ SWEP.AmmoRegenAmount = 2 -- Amount of ammo refilled every AmmoRegenRate seconds
 function SWEP:Initialize()
 
 	self:SetHoldType( self.HoldType )
-	
+
 	-- Prevent large ammo jumps on-creation
 	-- if DefaultClip < ClipSize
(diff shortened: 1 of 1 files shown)


Kefta:
A Should/CanTakeDamage binding would be useful here to avoid using GetInternalVariable in the codebase

robotboy655:
I think it would be better to save holder time, and calculate and add how much should've regen'd between that time and the deploy function. Should result in better looking code as well.

Kefta:
I thought of that, but if any addons wanted to display the current ammo of the medkit on the HUD while holstered, like for a dedicated healer class, then this would keep it accurately updated. This could be left for SWEP override territory instead though.

Kefta:
New solution to be tested tomorrow.

Kefta:
Good to go from my testing. Only prediction error is from the SetHealth call but there seems to be something funky happening with that function in particular during prediction.

brandonsturgeon:
LGTM 👍
I'll apply this PR to our Build/Kill server and let you know if we see any problems.

brandonsturgeon:
Running it for about half a day, everything seems good 👍

T-Rizzle12:
Any ETA when this would be released?

robotboy655 (approved):

PR #1962, opened 2023-04-26
Make hint text more useful if the command key is unbound

Kefta:
Additionally
- Add ``number length`` and ``string sound`` arguments to ``GM:AddHint`` that default to ``20`` and ``"ambient/water/drip" .. math.random( 1, 4 ) .. ".wav"`` respectively
- Add ``GM:UnprocessHint( string name )``
- Make ``ProcessedHints`` a gamemode table var rather than a local
- Add helptext to ``cl_showhints``
- Adjust ``Annoy#`` hint times to account for the full-load time discrepancy
- Change ``HintTimer_`` hook prefix to ``GMOD_HintTimer_``
- Various optimizations

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
@@ -73,8 +73,8 @@ function GM:UnprocessHint( name )
 end
 
 -- Show opening menu hint if they haven't opened the menu within 30 seconds
-GM:AddHint( "OpeningMenu", 10 )
+GM:AddHint( "OpeningMenu", 30 )
 
 -- Tell them that they can turn off hints as they fully load in
-GM:AddHint( "Annoy1", 5 )
-GM:AddHint( "Annoy2", 7 )
+GM:AddHint( "Annoy1", 10 )
+GM:AddHint( "Annoy2", 13 )
--- a/garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
@@ -1,6 +1,5 @@
 
-
-CreateClientConVar( "cl_showhints", "1", true, false )
+local cl_showhints = CreateClientConVar( "cl_showhints", "1", true, false )
 
 -- A list of hints we've already done so we don't repeat ourselves`
 local ProcessedHints = {}
@@ -8,25 +7,26 @@ local ProcessedHints = {}
 --
 -- Throw's a Hint to the screen
 --
-local function ThrowHint( name )
+local function ThrowHint( GM, name, force, length )
 
-	local show = GetConVarNumber( "cl_showhints" )
-	if ( show == 0 ) then return end
-
-	if ( engine.IsPlayingDemo() ) then return end
+	if ( !cl_showhints:GetBool() ) then return end
 
 	local text = language.GetPhrase( "Hint_" .. name )
 
 	local s, e, group = string.find( text, "%%([^%%]+)%%" )
 	while ( s ) do
 		local key = input.LookupBinding( group )
-		if ( !key ) then key = "<NOT BOUND>" end
+		if ( !key ) then
+			if ( !force ) then return end
+			key = "<NOT BOUND>"
+		end
 
 		text = string.gsub( text, "%%([^%%]+)%%", "'" .. key:upper() .. "'" )
 		s, e, group = string.find( text, "%%([^%%]+)%%" )
 	end
 
-	GAMEMODE:AddNotify( text, NOTIFY_HINT, 20 )
+	if ( length == nil ) then length = 20 end
+	GM:AddNotify( text, NOTIFY_HINT, length )
 
 	surface.PlaySound( "ambient/water/drip" .. math.random( 1, 4 ) .. ".wav" )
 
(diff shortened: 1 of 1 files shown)


robotboy655:
I disagree. Displaying the hints is still useful to let the player know that a feature exists at all, and that they may have some actions unbound.

Kefta:
How about putting the command they have unbinded in-place of <NOT BOUND> with an additional note like "(command not bound)"?

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
```diff
@@ -1,62 +1,80 @@
 
+local cl_showhints = CreateConVar( "cl_showhints", "1",
+	bit.bor( FCVAR_ARCHIVE, FCVAR_USERINFO ), "Show hints called thru SANDBOX:AddHint", 0, 1 )
 
-CreateClientConVar( "cl_showhints", "1", true, false )
+-- A list of hints we've already done so we don't repeat ourselves
+GM.ProcessedHints = {}
```
robotboy655, line 6: What is the point of exposing an internal table to public?

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
```diff
@@ -1,62 +1,80 @@
...
 
---
--- Adds a hint to the queue
---
-function GM:AddHint( name, delay )
+	if ( self.ProcessedHints[ name ] ) then return end
 
-	if ( ProcessedHints[ name ] ) then return end
+	if ( !engine.IsPlayingDemo() ) then
+		timer.Create( "GMOD_HintSystem_" .. name, delay, 1, function()
```
robotboy655, line 54: What is the point of adding `GMOD_` prefix to the timer's ID? Do we not know what game we are?

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
```diff
@@ -1,62 +1,80 @@
...
-		text = string.gsub( text, "%%([^%%]+)%%", "'" .. key:upper() .. "'" )
-		s, e, group = string.find( text, "%%([^%%]+)%%" )
+	local startpos, endpos, bind = text:find( "%%([^%%]+)%%" )
+
+	if ( startpos ~= nil ) then
+		local buffer = {}
+		local bufferlen = 1
+		local nextstartpos = 1
+		
+		repeat
```
robotboy655, line 20: Why was the `while` loop replaced with a `repeat` loop? This new code looks infinitely less readable, while seemingly providing absolutely no benefit.

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
```diff
@@ -1,62 +1,80 @@
...
+	if ( length == nil ) then length = 20 end
+	GM:AddNotify( text, NOTIFY_HINT, length )
 
-	surface.PlaySound( "ambient/water/drip" .. math.random( 1, 4 ) .. ".wav" )
+	if ( sound == nil ) then sound = "ambient/water/drip" .. math.random( 1, 4 ) .. ".wav" end
+	surface.PlaySound( sound )
 
 end
 
+function GM:AddHint( name, delay, length, sound )
```
robotboy655, line 49: I am not sure adding `sound` parameter is a good idea. What's next, custom icons? Custom render function?

This system should be kept self contained.

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
```diff
@@ -1,62 +1,80 @@
...
--- Removes a hint from the queue
---
 function GM:SuppressHint( name )
 
-	timer.Remove( "HintSystem_" .. name )
+	timer.Remove( "GMOD_HintSystem_" .. name )
+
+end
+
+function GM:UnprocessHint( name )
```
robotboy655, line 69: Is this really needed? At what point would you want a hint to be displayed again?

robotboy655 (changes requested):

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
Kefta: This was to introduce the auto-refresh table idiom to the file but I forgot to write the full statement.

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
Kefta: Timer identifiers are global. Hook identifiers are global. ConCommands/ConVars are global. Many other built-in systems (Lua or engine based) share common identifiers - adding the prefix is just a formality to avoid conflict.

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
Kefta: There are no future gateways to customizability in this library because the vanilla hint system is already firmly limited; the additional arguments introduced are just extensions of the default impl to allow scripts to use this system without reimplementing/copying all the Hint code to change a hardcoded time/sound.

The hint system here is a barebones notification feature to introduce people to Sandbox, but since it's in the Sandbox gamemode table, why not allow the standard impl to be configured without expanding its scope or breaking anything?

garrysmod/gamemodes/sandbox/gamemode/cl_hints.lua
Kefta: If a server wants to allow specific classes/roles/teams/whatever division hints to be displayed every time they switch, then they should be able to use this system freely since it's available in the Sandbox gamemode table. This entire sandbox system sucks but it should be fleshed out into the standard Add + Remove template so it can be utilised as the official "gmod hint system"

Kefta:
I will update the code to be more concise and readable this week

PR #1965, opened 2023-05-03
Update dlistview_column fixed width update wide

dvdvideo1234:


dvdvideo1234:
Hi, it is very simple actually. You see when you create `ListView` column with a fixed width it i called:
```lua
pnListView:AddColumn(languageGetPhrase("tool."..gsToolNameL..".pn_routine_lb1")):SetFixedWidth(25)
local w = pnListView:ColumnWidth(1) -- Returns the value of column:GetWide()
```
Probably it is my expectation, but since the column width is being updated with a fixed value, [I would rather expect `GetWide`](https://github.com/Facepunch/garrysmod/blob/f7feb921ee2f02a4654ca97b60804687db40e4d9/garrysmod/lua/vgui/dlistview.lua#L125) to return that same fixed value instead of zero or another value not relevant to the list view  column width.

robotboy655 (approved):

PR #1966, opened 2023-05-14
Send achievements through net. instead of :SendLua

wrefgtzweve:
I've noticed that pretty much all achievements are currently being send through SendLua which is suboptimal, my proposal is to send it through the net library instead so there's less bandwith being used.

I'm open for suggestions and improvements as this might not be the best way to do it. I've considered sending the int of achievements instead of using the achievement string but those aren't easily available on serverside so i went with this.

It's hard to test how much SendLua really sends over so i am not 100% sure if this can be considered an improvement, although in the best case sending (for SendLua) `achievements.SpawnedProp()` is 26 bytes while with this method it'd be 14 bytes of data.


robotboy655:
I feel like this is change for the sake of change. Not only this will break the achievements (the functions check where they are ran from), it's a half measure if the goal is to save some bytes over the network.

PR #1967, opened 2023-05-14
Add AddLegacyType & GetLegacyTypes

Zaurzo:
The notice material types are hard-coded and we have such a small arsenal of these materials, it would be nice to have an easy way to add our own. This might be a useless addition, I'd like to see feedback on this.


Zaurzo:
Update I did 2 weeks ago:

On success, it will now return the IMaterial object that was created.
On fail, it will use the generic notify material and return false.

PR #1968, opened 2023-05-15
Expose RealFrameTime() to the menu state

Zaurzo:
This one is simple and more of a fix than anything. notification.AddLegacy() is available to both client and menu state and utilizes this function, but errors on the menu state because RealFrameTime() isn't exposed to that state.


Kefta:
Please move this to an engine binding Rubat 🙏

robotboy655:
Menu state is internal and is not meant to be modified by players.

PR #1969, opened 2023-05-16
Added http.HandleCode

vicentefelipechile:
This function returns the name of the request (e.g., 200 -> "OK") and the description of the request (e.g., 200 -> "The resource has been obtained").
If the request with that value is not found or does not exist, it will return nil.

For example:
```lua
local msg, desc = http.HandleCode(200)
print( msg, desc )
print( http.HandleCode(499) )
```

Output:
```
OK      The resource has been obtained
nil
```

### DIFF
--- a/garrysmod/lua/includes/modules/http.lua
+++ b/garrysmod/lua/includes/modules/http.lua
@@ -103,3 +103,60 @@ local request = {
 HTTP( request )
 
 --]]
+
+
+
+
+
+-- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status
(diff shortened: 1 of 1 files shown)


WilliamVenner:
Maybe `http.GetStatusDescription` would be a better name? `http.HandleCode` implies that some processing takes place when called.

vicentefelipechile:
Done!

Zaurzo (changes requested):
All of the code does not follow the preferred code format btw, see  https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md

PR #1971, opened 2023-05-28
Math for Color class

Be1zebub:
new metaevents support `__concat`, `__unm`, `__add`, `__sub`, `__mul`, `__div`, `__lt`, `__le`
new methods: `:Invert()`, `:Copy()`, `:Normalize()`, `:GetNormalized()`

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -47,6 +47,86 @@ function COLOR:__tostring()
 end
 
 --[[---------------------------------------------------------
+	Returns unary color
+-----------------------------------------------------------]]
+function COLOR:__unm()
+
+	local col = self:Copy()
+	col:Invert()
+
+	return col
+
+end
+
+--[[---------------------------------------------------------
+	Returns colors summ
+-----------------------------------------------------------]]
+
+function COLOR:__add( other )
+
+	return Color( self.r + other.r, self.g + other.g, self.b + other.b, self.a + other.a )
+
+end
+
+--[[---------------------------------------------------------
+	Returns colors subtraction
+-----------------------------------------------------------]]
+
+function COLOR:__sub(other)
+
+	return Color( self.r - other.r, self.g - other.g, self.b - other.b, self.a - other.a )
+
+end
+
+--[[---------------------------------------------------------
+	Returns color multiplication with another color or number
+-----------------------------------------------------------]]
+
+function COLOR:__mul(other)
+
+	if ( type( other ) == "number" ) then
+		return Color( self.r * other, self.g * other, self.b * other, self.a * other )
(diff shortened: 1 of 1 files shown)


Kefta:
Could you give some examples/documentation that makes these operations defined and sensible for rgb colours?

garrysmod/lua/includes/util/color.lua
```diff
@@ -105,3 +185,35 @@ function COLOR:ToTable()
...
+function COLOR:Copy()
+
+	return Color( self.r, self.g, self.b, self.a )
+
+end
+
+function COLOR:Normalize()
+
+	self.r = self.r / 255
+	self.g = self.r / 255
```
Malivil, line 206: Shouldn't these lines be `g`, `b`, and `a` rather than `r` three more times?

Be1zebub:
> Could you give some examples/documentation that makes these operations defined and sensible for rgb colours?

arithmetic operators allows to write cleaner code, this is useful in complex code - eg for pseudo shaders (we dont have shaders api, so cope with rt & render.ReadPixel)
I actively use pseudo shaders in my projects, for example bloom
here are some examples how color math can be used:
https://gist.github.com/Be1zebub/66a16af468e2f2d7e2a9fc0dc9ee9311
image

garrysmod/lua/includes/util/color.lua
Be1zebub: woops, typo

garrysmod/lua/includes/util/color.lua
```diff
@@ -46,6 +46,86 @@ function COLOR:__tostring()
 
 end
 
+--[[---------------------------------------------------------
+	Returns unary color
+-----------------------------------------------------------]]
+function COLOR:__unm()
```
CornerPin, line 52: I would expect this function to just negate all color components so that these two lines produce the same result
```lua
local invert1 = color_white - color
local invert2 = color_white + (-color)
````
The current implementation should rather be named `GetInverted` to be consistent with `Normalize`/`GetNormalized`.

CornerPin (commented):
Can you add functions like `Add`, `Sub`, etc. that modify the original color rather than create a new one? Vectors, angles and matrices already have these, so it would make sense for colors to have them too.

stepa2:
If color is not in linear RGB, you can't just multiply or add it by-component.
You'll need to convert it from RGB to linear RGB first, then do math, then convert it back.
http://www.ericbrasseur.org/gamma.html?i=1#explanation

> This all, to explain that gray number 127 does not imply a luminosity half that of number 255. Gray number 127 implies that the power of light emitted by the screen must be 21% of the power of bright white. Conversely, the number of the gray that asks for 50% power is 187.

PR #1972, opened 2023-05-29
Updated `cvars` library

PrikolMen:
I updated the outdated `cvars` library code, corrected the comments, replaced depricated functions with actual ones, and improved it by removing unnecessary checks.

I also formatted the entire code according to [CONTRIBUTING.md](https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md)
All tabs were replaced by 4 spaces, function argument names were fixed, and a few small changes were made.

### DIFF
--- a/garrysmod/lua/includes/modules/cvars.lua
+++ b/garrysmod/lua/includes/modules/cvars.lua
@@ -4,131 +4,131 @@ local type				= type
 local GetConVar			= GetConVar
 
 --[[---------------------------------------------------------
-    Name: cvar
-    Desc: Callbacks when cvars change
+	Name: cvar
+	Desc: Callbacks when cvars change
 -----------------------------------------------------------]]
 module( "cvars" )
 
 local ConVars = {}
 
 --[[---------------------------------------------------------
-    Name: GetConVarCallbacks
-    Desc: Returns a table of the given ConVars callbacks
+	Name: GetConVarCallbacks
+	Desc: Returns a table of the given ConVars callbacks
 -----------------------------------------------------------]]
 function GetConVarCallbacks( name, createIfNotFound )
 
-    local tab = ConVars[ name ]
-    if ( createIfNotFound and !tab ) then
-        tab = {}; ConVars[ name ] = tab
-    end
+	local tab = ConVars[ name ]
+	if ( createIfNotFound and !tab ) then
+		tab = {}; ConVars[ name ] = tab
+	end
 
-    return tab
+	return tab
 
 end
 
 --[[---------------------------------------------------------
-    Name: OnConVarChanged
-    Desc: Called by the engine
+	Name: OnConVarChanged
+	Desc: Called by the engine
 -----------------------------------------------------------]]
 function OnConVarChanged( name, old, new )
 
-    local tab = GetConVarCallbacks( name )
-    if ( !tab ) then return end
+	local tab = GetConVarCallbacks( name )
+	if ( !tab ) then return end
 
-    for i = 1, #tab do
-        local callback = tab[ i ]
-        if ( type( callback ) == "table" ) then
-            callback[ 1 ]( name, old, new )
-        else
-            callback( name, old, new )
-        end
-    end
+	for i = 1, #tab do
+		local callback = tab[ i ]
+		if ( type( callback ) == "table" ) then
+			callback[ 1 ]( name, old, new )
+		else
+			callback( name, old, new )
+		end
+	end
 
 end
 
 --[[---------------------------------------------------------
-    Name: AddChangeCallback
-    Desc: Adds a callback to be called when convar changes
+	Name: AddChangeCallback
+	Desc: Adds a callback to be called when convar changes
 -----------------------------------------------------------]]
 function AddChangeCallback( name, func, identifier )
 
-    local tab = GetConVarCallbacks( name, true )
+	local tab = GetConVarCallbacks( name, true )
 
-    if ( !identifier ) then
-        table.insert( tab, func )
-        return
-    end
+	if ( !identifier ) then
+		table.insert( tab, func )
+		return
+	end
 
-    for i = 1, #tab do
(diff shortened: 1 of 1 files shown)


Kefta:
I do not see this as an improvement. In fact it introduces regressions, such as as no longer checking that the identifier to cvars.Add/RemoveChangeCallback is a string and cvars.Bool can now return nil.

robotboy655:
[CONTRIBUTING.md](https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md) specifically states to use `Tab` characters, not spaces.

PrikolMen:
what do you mean, then just write `4 tabs` instead of `tab = 4 spaces`

image

PrikolMen:
> I do not see this as an improvement. In fact it introduces regressions, such as as no longer checking that the identifier to cvars.Add/RemoveChangeCallback is a string and cvars.Bool can now return nil.

Have you used cvars at all? String checking is not needed there at all because it gives nothing and stupidly limits developers and wastes CPU time on checking, there are no operations with strings. And return `nil` is much easier to check than `false` plus it's much more universal.

PrikolMen:
> [CONTRIBUTING.md](https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md) specifically states to use `Tab` characters, not spaces.

fixed

Kefta:
These are ancient functions with defined behaviours that addons rely on - you can't just change them because you think it's better or more efficient.

PrikolMen:
> These are ancient functions with defined behaviours that addons rely on - you can't just change them because you think it's better or more efficient.

For example for me the behavior of `cvars.Bool` is not at all obvious and not predictable, and I have not seen any addon that checks `cvars.Bool( "cvar" ) == false `, I say more if it causes problems it is extremely easy to correct the developers, so that your statement is silly to say the least.

Personally, I'll tell you an amazing fact, you can't update the game without breaking something, you either don't update it and it dies, or you update it and the developers adapt, you shouldn't look at the "old addons" you should look at how simple, clear and productive it is, unless you want to turn the game into a graveyard of good ideas.

garrysmod/lua/includes/modules/cvars.lua
```diff
@@ -18,111 +11,124 @@ module( "cvars" )
...
+	Desc: Returns a table of the given ConVars callbacks
+-----------------------------------------------------------]]
+function GetConVarCallbacks( name, createIfNotFound )
 
-	if ( CreateIfNotFound && !Tab ) then
-		Tab = {}
-		ConVars[ name ] = Tab
+	local tab = ConVars[ name ]
+	if ( createIfNotFound and !tab ) then
+		tab = {}; ConVars[ name ] = tab
```
robotboy655, line 22: What's the point of putting these on the same line?

garrysmod/lua/includes/modules/cvars.lua
```diff
@@ -18,111 +11,124 @@ module( "cvars" )
...
+	local tab = GetConVarCallbacks( name )
+	if ( !tab ) then return end
 
-	for k, v in pairs( Callbacks ) do
-
-		if ( istable( v ) ) then
-			v[ 1 ]( name, oldvalue, newvalue )
+	for i = 1, #tab do
+		local callback = tab[ i ]
+		if ( type( callback ) == "table" ) then
```
robotboy655, line 40: What's the point of changing `istable( v )` calls  to `type( callback ) == "table"`?

garrysmod/lua/includes/modules/cvars.lua
```diff
@@ -18,111 +11,124 @@ module( "cvars" )
...
-	Name: OnConvarChanged
-	Desc: Called by the engine
+	Name: AddChangeCallback
+	Desc: Adds a callback to be called when convar changes
 -----------------------------------------------------------]]
-function AddChangeCallback( name, func, sIdentifier )
-
-	if ( sIdentifier ) then
-		assert( isstring( sIdentifier ), format( "bad argument #%i (string expected, got %s)", 3, type( sIdentifier ) ) )
-	end
```
robotboy655, line 62: What is the point of removing these? This is a regression that will cause more confusion for the people incorrectly using these functions during development.

robotboy655 (changes requested):
I am seeing a bunch of unnecessary changes here,

garrysmod/lua/includes/modules/cvars.lua
PrikolMen: just looks shorter and better, i can split it into two lines if you think it's unacceptable

garrysmod/lua/includes/modules/cvars.lua
PrikolMen: purpose is to make the code shorter, why do we need to localize `istable` if we already have `type`, why do we need both functions?

garrysmod/lua/includes/modules/cvars.lua
PrikolMen: what is the point of this restriction, what confusion are we talking about if it doesn't affect anything, in fact the identifier can be really anything, there are no operations with string

PrikolMen: This value can be `nil` altogether isn't that confusing? Why a value that should be a string can be `nil`, how we can find and remove this value?

garrysmod/lua/includes/modules/cvars.lua
PrikolMen: Corrected

garrysmod/lua/includes/modules/cvars.lua
PrikolMen: Okay

garrysmod/lua/includes/modules/cvars.lua
PrikolMen: Returned

robotboy655 (approved):

PR #1973, opened 2023-05-29
`util.IsLuaModuleInstalled( name )`

PrikolMen:
New function to check if there is a lua module in the gmod file system.

```lua
local name = "promise"
if util.IsLuaModuleInstalled( name ) then
    require( name )
end

local asyncFunc = promise.Async( function()
    -- something
end )
```

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -395,3 +395,11 @@ function util.IsBinaryModuleInstalled( name )
 
 	return false
 end
+
+--[[---------------------------------------------------------
+	Name: IsLuaModuleInstalled( name )
+	Desc: Returns whether a lua module with the given name is present on disk
+-----------------------------------------------------------]]
+function util.IsLuaModuleInstalled( name )
+	return file.Exists( "includes/modules/" .. name .. ".lua", "GAME" )
+end
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -401,5 +401,5 @@ end
 	Desc: Returns whether a lua module with the given name is present on disk
 -----------------------------------------------------------]]
 function util.IsLuaModuleInstalled( name )
-	return file.Exists( "lua/includes/modules/" .. name .. ".lua", "GAME" )
+	return file.Exists( "includes/modules/" .. name .. ".lua", "LUA" )
 end
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -401,5 +401,5 @@ end
 	Desc: Returns whether a lua module with the given name is present on disk
 -----------------------------------------------------------]]
 function util.IsLuaModuleInstalled( name )
-	return file.Exists( "includes/modules/" .. name .. ".lua", "GAME" )
+	return file.Exists( "lua/includes/modules/" .. name .. ".lua", "GAME" )
 end


robotboy655:
Aren't we going a bit too far with this? I can understand this for binary modules - you can't ship them on workshop yourself, but an alias to check if a Lua file exists?

PrikolMen:
I wouldn't add this if I didn't need it, people are now more active in creating lua modules, and it would be much more convenient to check this using one function instead of `file.Exists("lua/includes/modules/" ... name ... ".lua", "GAME" )` every time.

FlorianLeChat:
The reason behind adding `util.IsBinaryModuleInstalled` was to have a robust function which could check whether a binary module was installed in different environments and operating systems. Your suggestion only serves to shorten the code, which in my opinion is more than pointless.

PrikolMen:
> The reason behind adding `util.IsBinaryModuleInstalled` was to have a robust function which could check whether a binary module was installed in different environments and operating systems. Your suggestion only serves to shorten the code, which in my opinion is more than pointless.

What's wrong with code shortening? In addition, there are many more pointless functions that exist just for code reducing. Not all simple solutions are bad.

PrikolMen:
It also looks like a great addition to `require` and `util.IsBinaryModuleInstalled` to fully check the ability to require.

FlorianLeChat:
> What's wrong with code shortening? In addition, there are many more pointless functions that exist just for code reducing.

I'm not saying your solution is bad, I'm just saying your solution doesn't need to be included in the codebase. And even if some functions already exist to "shorten" or "simplify" code, I doubt Rubat would support adding new ones (like #1769).

PrikolMen:
> > What's wrong with code shortening? In addition, there are many more pointless functions that exist just for code reducing.
> 
> I'm not saying your solution is bad, I'm just saying your solution doesn't need to be included in the codebase. And even if some functions already exist to "shorten" or "simplify" code, I doubt Rubat would support adding new ones (like #1769).

In your example is really a very questionable solution for inclusion in the codebase...

ExtReMLapin:
Gmod repo isn't what it used to be

Kefta:
The point is that require (what you probably meant) will error from the file not existing.

garryspins:
Ah yea i didnt look at the implementation mb, either way it doesnt seem useful at all to be in the codebase to me, i cant imagine this would get much usage plus lua modules are their own things separate from just files so it could add another level of confusion.

Plus this only takes into account whats in lua/modules/ which is a bit silly, if this allowed relative include checking itd be vastly more useful but as it is i really dont see the point, either way a large majority of people use their own loaders that completely go around the established subfolders which would make this completely redundant.

PrikolMen:
> ```lua
> if not promise then include("promise.lua") end
> ```
> 
> seems fine to me

There's no example of checking for an installed lua module, just "protection" against a call, which is what `require` does by default, you said nonsense

PrikolMen:
> Ah yea i didnt look at the implementation mb, either way it doesnt seem useful at all to be in the codebase to me, i cant imagine this would get much usage plus lua modules are their own things separate from just files so it could add another level of confusion.
> 
> Plus this only takes into account whats in lua/modules/ which is a bit silly, if this allowed relative include checking itd be vastly more useful but as it is i really dont see the point, either way a large majority of people use their own loaders that completely go around the established subfolders which would make this completely redundant.

Sorry but you're talking nonsense, according to your logic `lua` modules as well as in general the folder `modules` should not exist in principle because all the developers use loaders.

Kefta:
I don't think that's an absurd take considering its static nature and greater rate of naming conflicts.

FredyH:
> Your suggestion only serves to shorten the code, which in my opinion is more than pointless.

Using your logic we should not have any function in the repository that can be implemented in pure LUA.

Clearly, adding a maintainable function that _correctly_ checks if a lua module can be required is a good idea to avoid code duplication and loads of people doing it incorrectly.

garryspins:
There's no such thing as doing it "incorrectly", however in my opinion the incorrect way to do it would be this. Too many downsides for no upsides, a unified module folder with the risk of conflicts vs managing your own files, managing your own versioning of those files and making it impossible for things to conflict. This encourages bad practices in a way.

lifestorm (changes requested):
Module can be placed in "modules/" subfolder. I guess it should search for specified module in every "modules/" subfolder.

lifestorm (changes requested):
Since module can be placed in "modules/" subfolder, we should search for it through all the subfolders

dvdvideo1234:
In TA dedicated meta addons I check if the library is included properly as @garryspins suggested [here](https://github.com/Facepunch/garrysmod/pull/1973#issuecomment-1598383157). Seems simple enough to me as well.
https://github.com/dvdvideo1234/TrackAssemblyTool/blob/master/lua/autorun/z_autorun_%5Bshinji85_s_rails%5D.lua#L62

PrikolMen:
> There's no such thing as doing it "incorrectly", however in my opinion the incorrect way to do it would be this. Too many downsides for no upsides, a unified module folder with the risk of conflicts vs managing your own files, managing your own versioning of those files and making it impossible for things to conflict. This encourages bad practices in a way.

I'm sorry, but you're talking pure nonsense, what versions? Where have you seen in the game even a semblance of version control, I do not understand why you talk about "conflicts" when the game otrutututet any mechanism to protect against them, and their presence is not regulated by anything, for example, my library breaks "Advanced Player Selector" because the original author decided that it would be a great idea to overwrite `PLAYER.SetModel` every time it loads or someone switches "protection against model changes", so do not talk nonsense.

PrikolMen:
> In TA dedicated meta addons I check if the library is included properly as @garryspins suggested [here](https://github.com/Facepunch/garrysmod/pull/1973#issuecomment-1598383157). Seems simple enough to me as well. https://github.com/dvdvideo1234/TrackAssemblyTool/blob/master/lua/autorun/z_autorun_%5Bshinji85_s_rails%5D.lua#L62

Great solution to check for library loading `if (myLib) then` but you are missing one small detail, how are you going to load this library without `require` if it is a game module. It's not about a universal solution for libraries, it's about more convenient interaction with gLua modules. Your example has no relation to the topic.

PrikolMen:
> Since module can be placed in "modules/" subfolder, we should search for it through all the subfolders

I'm not sure that gLua modules work like this, but if you want:
```lua
util.IsLuaModuleInstalled( "my_complex_library/init" )
```

garryspins:
> I'm sorry, but you're talking pure nonsense, what versions? Where have you seen in the game even a semblance of version control, I do not understand why you talk about "conflicts" when the game otrutututet any mechanism to protect against them, and their presence is not regulated by anything, for example, my library breaks "Advanced Player Selector" because the original author decided that it would be a great idea to overwrite `PLAYER.SetModel` every time it loads or someone switches "protection against model changes", so do not talk nonsense.

Nowhere did i talk about overriding functions, the discussion is about _file conflicts_, if you have _one folder_ with _every module_ you WILL encounter conflicts, it is inevitable, if you have 5 different versions of CPPI, all under `modules/cppi.lua`, only one will take priority, potentially breaking your addon. if you can have your own version, that you use in your own way, you can guarantee it will be valid, this adds ambiguity and another window for addon conflict. This "version control" I speak of, is addons not universally using the same version of a file, as in, my addon has `my_addon/libs/sh_lib.lua`, your addon has `your_addon/libs/sh_lib.lua`. If these libraries are written properly, they will never conflict, atleast the file names will not conflict. This would not solve code conflicts regardless, so your point is null.
Also refrain from insulting me when its clear you didn't read my message, atleast in the proper context.

PrikolMen:
> > I'm sorry, but you're talking pure nonsense, what versions? Where have you seen in the game even a semblance of version control, I do not understand why you talk about "conflicts" when the game otrutututet any mechanism to protect against them, and their presence is not regulated by anything, for example, my library breaks "Advanced Player Selector" because the original author decided that it would be a great idea to overwrite `PLAYER.SetModel` every time it loads or someone switches "protection against model changes", so do not talk nonsense.
> 
> Nowhere did i talk about overriding functions, the discussion is about _file conflicts_, if you have _one folder_ with _every module_ you WILL encounter conflicts, it is inevitable, if you have 5 different versions of CPPI, all under `modules/cppi.lua`, only one will take priority, potentially breaking your addon. if you can have your own version, that you use in your own way, you can guarantee it will be valid, this adds ambiguity and another window for addon conflict. This "version control" I speak of, is addons not universally using the same version of a file, as in, my addon has `my_addon/libs/sh_lib.lua`, your addon has `your_addon/libs/sh_lib.lua`. If these libraries are written properly, they will never conflict, atleast the file names will not conflict. This would not solve code conflicts regardless, so your point is null. Also refrain from insulting me when its clear you didn't read my message, atleast in the proper context.

I say again, what does this have to do with versions, my function just allows you to check glua module file existence. How does that stop you from using your copies of the libraries? If it is already installed as a module, then the fact that you have this function or not will not change anything, you will still have your "scary" conflicts. Your example with CPPI is maximally incorrect, because CPPI and CAMI are interfaces and they physically cannot have different versions because it is guaranteed to cause a conflict. In fact, even your "genius" way of copying a library will not save you from a CPPI/CAMI copy conflicts. A name conflict in this situation is the most harmless consequence.

P.S. By the way, the file limit of 8192 files doesn't scare you in any way, you literally tell addon authors to duplicate all used libraries in one addon, don't you think it's a bit too much?

robotboy655:
Closing this due to lack of support from other people.

PR #1974, opened 2023-05-30
Update hook.lua

smokingplaya:
Added a new function hook.GetData()


robotboy655:
I feel like this goes against how hooks are expected to work. You can have this in your addon if you really need it using `hook.GetTable`.

PR #1977, opened 2023-06-10
Do not set ClassName in vgui.CreateFromTable

FredyH:
Panels created by this functions do not have their own class and are anonymous. The reference classname in the existing function does not exist and is nil.


neico:
It does have a classname: `metatable.Base`, but it indeed isn't correctly set.
If you look around in the neighboring functions, you'll notice what `classname` is supposed to be

FredyH:
No, metatable.Base is the classname of the baseclass. Panels created via vgui.CreateFromTable do not really have their own classnames because they are essentially anonymous classes.
Of course it could be argued that if we have to give it a classname, maybe we should give it the name of the base, but that would alter the current behaviour.

robotboy655 (approved):

PR #1979, opened 2023-06-22
Removed the old SetDeploySpeed function.

RaphaelIT7:
This removes the old SetDeploySpeed function. This function is already removed on every branch except the main branch, but it was never removed here.


robotboy655:
I have updated the repo to reflect current changes on the `dev` beta.

PR #1981, opened 2023-06-24
Clamp DListView_Column resize to width restraints

lionbryce:
w/o this it'll resize neighboring columns when you've already hit the limit


robotboy655:
An code example where this has an effect would be appreciated.

PR #1983, opened 2023-06-26
TTT: Fix broken sound cue and make it toggleable.

figardo:
The "Hud.Hint" notification sound cue has remained untouched in the gamemode since the very first public release, and was working until the script file containing the sound's definition was removed with GMod 12.

[Here's an old video showing the sound cue working.](https://youtu.be/C9YRWZ2iXog?t=3)

I've also added a convar to enable the sound, and an entry in the settings menu to toggle this convar.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_help.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_help.lua
@@ -103,8 +103,6 @@ function HELPSCRN:Show()
 
    cb = dgui:CheckBox(GetTranslation("set_cues"), "ttt_cl_soundcues")
 
-   cb = dgui:CheckBox(GetTranslation("set_msg_cue"), "ttt_cl_msg_soundcue")
-
    dsettings:AddItem(dgui)
 
    --- Gameplay area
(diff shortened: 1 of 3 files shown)


svdm:
- The name of the sound should changed to be TTT-specific given that we're adding it, so we don't clash with any addons etc. Something like `TTT.MessageCue` or whatever.
- I appreciate the convar etc, but given that this is restoring something that should always have been there I don't think it needs a setting in the menu. I doubt it will see much use so it just adds clutter there.

Malivil:
Since this has been missing for so long, I'm sure there will be people who want to keep it disabled so I think the setting is a good idea.

wgaylord:
> * The name of the sound should changed to be TTT-specific given that we're adding it, so we don't clash with any addons etc. Something like `TTT.MessageCue` or whatever.
> * I appreciate the convar etc, but given that this is restoring something that should always have been there I don't think it needs a setting in the menu. I doubt it will see much use so it just adds clutter there.

Since this has not existed in such a long time (over 13 years based on them saying it disappeared in Gmod 12) I wish there was a setting in F1 with it defaulted off. Had players coming to staff complaining about the messages now having sound thinking it was a change we made.

TimGoll (commented):
ah, bummer, wrong tab :D

PR #1984, opened 2023-06-26
Send the entity to send fewer bits than sending the class

MoskalykA:
Hi
I suggest sending an entity to send fewer bits than sending the class, which is the same as sending a string.
An entity corresponds to 13 bits whereas a string with two characters corresponds to a higher number of bits and we know that the class of an entity is not two characters long.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/npc.lua
+++ b/garrysmod/gamemodes/base/gamemode/npc.lua
@@ -3,6 +3,8 @@
 util.AddNetworkString( "PlayerKilledNPC" )
 util.AddNetworkString( "NPCKilledNPC" )
 
+local worldspawn = Entity(0)
+
 --[[---------------------------------------------------------
    Name: gamemode:OnNPCKilled( entity, attacker, inflictor )
    Desc: The NPC has died
@@ -30,8 +32,8 @@ function GM:OnNPCKilled( ent, attacker, inflictor )
 	
 	end
 	
-	local InflictorClass = "worldspawn"
-	local AttackerClass = "worldspawn"
+	local InflictorClass = worldspawn
+	local AttackerClass = worldspawn
 	
 	if ( IsValid( inflictor ) ) then InflictorClass = inflictor:GetClass() end
 	if ( IsValid( attacker ) ) then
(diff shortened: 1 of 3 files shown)


robotboy655:
Have you actually tested this change on an actual server?

Kefta:
I believe net messages will always reach the client before its knowledge of entity deletion when dispatched at the same time, but this should be tested extensively. I'll do some this weekend.

garrysmod/gamemodes/base/gamemode/npc.lua
```diff
@@ -3,6 +3,8 @@
 util.AddNetworkString( "PlayerKilledNPC" )
 util.AddNetworkString( "NPCKilledNPC" )
 
+local worldspawn = Entity(0)
```
Kefta, line 6: Just call game.GetWorld() at the time the hook is ran - this optimisation is not needed. Also the variable names should be changed to reflect the fact that they're an entity now, not a class string.

Kefta (changes requested):

MoskalykA:
> Have you actually tested this change on an actual server? The reason for sending the classname is that the entity may be deleted by the time the message gets to the client.

I've actually tested my modifications but I hadn't thought of that, I'll test it tomorrow.

MoskalykA:
> I believe net messages will always reach the client before its knowledge of entity deletion when dispatched at the same time, but this should be tested extensively. I'll do some this weekend.

Interesting, I'll let you check it out for yourself.

FlorianLeChat:
Have you tested this in an environment where a server is under heavy network load due to a large number of players and network requests can take a **longer** time to be broadcasted to players? Your suggestion seems to be pretty weak in such cases where bad timing is enough to cause an error.

MoskalykA:
> Have you tested this in an environment where a server is under heavy network load due to a large number of players and network requests can take a **longer** time to be broadcasted to players? Your suggestion seems to be pretty weak in such cases where bad timing is enough to cause an error.

I've just tried to get as many requests as possible and it's still working, but I have no way of testing this with a server full of players and sending more requests than I've been able to get.

MoskalykA:
Hi, I added the modification to a friend's server, you were right

PR #1985, opened 2023-06-27
Add more robust error checking to weapons/scripted_ents.Register

Kefta:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/5495 - unlike SWEPs, SENTs do not set a ClassName variable. While SWEPs really shouldn't either to prevent the same multi-registration issue, addons and [even base code](https://github.com/Facepunch/garrysmod/blob/3d0d965168a9e58ad669f90667f02bee13048cb0/garrysmod/gamemodes/sandbox/gamemode/commands.lua#L860-L863) relies on this key to be set on the data table. While the entire initial SWEP table could be copied, I don't want to break addons that may rely on that ClassName being set on their passed table (yet).

Also added a bunch of type checks to provide more premature errors in case the SENT/SWEP sets something invalid before its creation.

A third optional bool argument (defaulting to false) "forcename" has been added to both Register functions which allows the caller to ignore ClassName overrides.

Additionally, the ai and nextbot SENT bases were added to the default base table.

Lastly, X.Register now returns true if the registration was successful and false if the PreRegisterX hook blocked it.

### DIFF
--- a/garrysmod/lua/includes/modules/scripted_ents.lua
+++ b/garrysmod/lua/includes/modules/scripted_ents.lua
@@ -131,29 +131,29 @@ function Register( t, name, forcename )
 
 	end
 
-	if ( t.Spawnable != true ) then return true end
+	if ( t.Spawnable == true ) then
+		local printname = t.PrintName
+		if ( !isstring( printname ) ) then printname = name end
 
-	local printname = t.PrintName
-	if ( !isstring( printname ) ) then printname = name end
+		local category = t.Category
+		if ( !isstring( category ) ) then category = "Other" end
 
-	local category = t.Category
-	if ( !isstring( category ) ) then category = "Other" end
+		list.Set( "SpawnableEntities", name, {
+			-- Required information
+			ClassName		= name,
+			PrintName		= printname,
+			Category		= category,
 
-	list.Set( "SpawnableEntities", name, {
-		-- Required information
-		ClassName		= name,
-		PrintName		= printname,
-		Category		= category,
-
-		-- Optional information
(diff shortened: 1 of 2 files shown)


garrysmod/lua/includes/modules/scripted_ents.lua
```diff
@@ -9,11 +9,14 @@ module( "scripted_ents", package.seeall )
...
-local BaseClasses = {}
-BaseClasses[ "anim" ] = "base_anim"
-BaseClasses[ "point" ] = "base_point"
-BaseClasses[ "brush" ] = "base_brush"
-BaseClasses[ "filter" ] = "base_filter"
+local BaseClasses = {
+	[ "ai" ]		= "base_ai",
+	[ "anim" ]		= "base_anim",
+	[ "brush" ]		= "base_brush",
+	[ "filter" ]	= "base_filter",
```
Kefta, line 16: This tabbing looks correct on tab = 4 spaces.

Kefta:
I'll keep this one open and just put the minimal classname setting code in a new PR.

robotboy655:
I had to revert your previous PR (https://github.com/Facepunch/garrysmod/commit/7a137541dbf86290a72805c4f8d38cac1f816ba1), I think moving forward if we want this functionality, perhaps it would be better if we use a name like `SWEP/ENT.ClassNameOverride` or something, to avoid any issues like this again.

Kefta:
Fair point, I'll use that.

Kefta:
I'll rebase and use the new class override field once https://github.com/Facepunch/garrysmod/pull/1987 is merged.

PR #1986, opened 2023-06-27
Update weapon_ttt_phammer.lua

Flapchik:


Kefta:
Better commit name would be appreciated.

PR #1987, opened 2023-06-27
Add SWEP/SENT.ClassNameOverride

Kefta:
Done after the PreRegisteredX hooks in-case those want to modify the class name override.


robotboy655:
This implementation sort of defeats the purpose of the hooks tho, doesn't it?

Kefta:
That's for the hook to check and modify if they want. The unoverrided name would be more useful in most cases anyway; take a CS:S weapon pack where all the weapons have a custom unoverrided name (weapon_mycustomcssbase_ak47) then specifies a name override (weapon_ak47) if a config option/convar is set - you'd want the hook to have access to the custom name to affect only the custom base instead of all weapon_ak47s.

robotboy655 (approved):

PR #1988, opened 2023-06-28
Prevent indexing IsValid from a number or boolean value

MaaxIT:
Here's an example of some hook code I recently had to use for a public addon for a DarkRP compatibility:
```lua
hook.Add("canDropPocketItem", "RRB:Home:canDropPocketItem", function(ply, item)
    if not IsValid(ply) or not IsValid(item) then return end
    -- ...
end)
```

However, this code seems to generate an error when "item" is a number (which can happen with DarkRP). Since there are no types in GLua, I can never be sure which value I'm going to check, and it would be a pain to constantly have to add checks to make sure it's an indexable type variable.

After analyzing the code, it soon becomes clear that this is due to the function trying to retrieve the IsValid function from the passed object. My patch therefore proposes to return false if a number or boolean is passed, rather than generating an error that stops execution.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -224,6 +224,7 @@ end
 function IsValid( object )
 
 	if ( !object ) then return false end
+	if ( isnumber(object) || isbool(object) ) then return false end
 
 	local isvalid = object.IsValid
 	if ( !isvalid ) then return false end


Kefta:
This has been suggested ad nauseum in the past and it's never been added because IsValid is a very commonly called function, thus adding two additional C calls would have major performance detriments.

MaaxIT:
The problem of numbers for the most part, and booleans a little less so, which are used a lot, seemed to me more suitable in terms of code and two calls to the C API didn't seem all that crazy to me, but it's true that it depends on how it's used.

It's also possible to integrate other types, but I thought this was relatively rare and wouldn't really be of any use in terms of code (and it would add even more calls to the C API).

But you're right, since it's been working like this for years, it wouldn't really be worth changing all that. It's just a bit annoying to have to do it manually every time, but I'll get used to it, never mind! Thanks for the feedback.

PR #1990, opened 2023-07-03
Be able to hide the tool menu in the context menu

MoskalykA:
Hi there,
I suggest not displaying the tool menu in the context menu.
This functionality was already planned but I need it.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/contextmenu.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/contextmenu.lua
@@ -38,9 +38,7 @@ function PANEL:Open()
 
 	RestoreCursorPosition()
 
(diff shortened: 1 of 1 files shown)


Zaurzo:
This is going to make it not show up at all if nothing is returned in the hook. Default it to true, and set it to false if ContextMenuShowTool returns false.

```
local bShouldShow = true

if ( hook.Run( "ContextMenuShowTool" ) == false then )
    bShouldShow = false
end
```

Kefta:
You can condense the above code into a single line by moving the hook call to the variable declaration

MoskalykA:
👍🏻

robotboy655 (approved):

Be1zebub (commented):
all you need is just `local bShouldShow = hook.Run( "ContextMenuShowTool" ) ~= false`

PR #1997, opened 2023-07-22
Fix Malformed CSS Comments

Noahbg:
The correct syntax for CSS comments is `/* */`, as specified by the [W3C CSS2.1 specification](https://www.w3.org/TR/CSS21/syndata.html#comments). This is consistent with the commenting style used in other CSS files.


Aws0mee:
hi noah

PR #2000, opened 2023-07-27
Add more NPCs categorized as hostile and friendly towards players

FlorianLeChat:
It's all in the title. These NPCs are hostile and friendly towards players by default in Garry's Mod but also in Half-Life 1/2.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -331,6 +331,14 @@ end
 	IsEnemyEntityName
 -----------------------------------------------------------]]
 local EnemyNames = {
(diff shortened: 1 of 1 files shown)


robotboy655:
Should probably also have the HL1 npc classnames here as well

FlorianLeChat:
Let me know if it's good for you!

robotboy655:
You should do the same for the friendly types as well.

robotboy655 (changes requested):
You should do the same for the friendly types as well.

FlorianLeChat:
It wasn't the original purpose of this pull request, but obviously it made more sense in this way.

PR #2001, opened 2023-07-30
Add custom icons for tree nodes in some spawnmenu categories

Dead0Duck:


robotboy655:
I wonder, is there any particular reason to have 4 separate lists for this, when one might be just fine? I mean, if you are giving `Half-Life 2` entities category an icon, you are probably wanting the exact same icon for the other categories anyway?

robotboy655 (approved):

PR #2002, opened 2023-08-01
Add math.BezierLerp & math.BezierSpline

Be1zebub:
Cubic/Quadratic/Linear bezier curve lerp & spline constructor

tests:
https://gist.github.com/Be1zebub/a2d1383a6ba09a2c5d71888dbeedf284
bezier.lua

docs:

Vector `math.BezierLerp( number frac, table points )`
Lerp point between points with bezier algorithms
example: [math.BezierLerp.lua](https://gist.github.com/Be1zebub/1e0bbb8d227c7df6803431447d686971)

table `math.BezierSpline( table points, number steps, table spline )`
Calc spline between points with bezier algorithms
example: [math.BezierSpline.lua](https://gist.github.com/Be1zebub/d9154bbf10668802bb89708532791b6c)

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -250,3 +250,59 @@ end
 function math.SnapTo( num, multiple )
 	return math.floor( num / multiple + 0.5 ) * multiple
 end
+
+--[[---------------------------------------------------------
+	Name: BezierLerp( frac, points )
+	Desc: Lerp point between points with bezier algorithms
+-----------------------------------------------------------]]
+
+function math.BezierLerp( frac, points )
+	local mu = frac * frac
+	local mum = 1 - frac
+	local mum2 = mum * mum
+
+	if ( points[4] ) then -- Cubic
+		return mum2 * mum * points[1] + 3 * mum2 * frac * points[2] + 3 * mum * mu * points[3] + mu * frac * points[4]
+	elseif ( points[3] ) then -- Quadratic
+		return mum2 * points[1] + 2 * mum * frac * points[2] + mu * points[3]
+	else -- Linear
+		return mum * points[1] + frac * points[2]
+	end
+end
+
+--[[---------------------------------------------------------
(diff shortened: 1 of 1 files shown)


robotboy655:
How about we change this to `math.CubicBezier( frac, p1, t1, p2, t2 )` + `math.QuadraticBezier( frac, p1, p2, p3 )` and drop `math.BezierSpline`?

I take issue with `math.BezierLerp` accepting a table of points, but only using 3 or 4 of them. I also see no good reason to have another linear lerp.

As for `math.BezierSpline`, I do not see it being necessary to be included in the base game, as what it does is equivalent of a for-loop. It could be added as example on the wiki page though.

Be1zebub:
what `t1`, `t2` means?
why not `p1, p2, p3, p4`?

robotboy655:
p1 = point1
t1 = tangent1

At least that's how I understand cubic beizer splines. It's not really that important.

Be1zebub:
i think tangents are confusing.
I also changed naming to `p0, p1, p2, p3` to make it match the formulas I used
Quadratic
Cubic

robotboy655 (approved):

robotboy655:
```
local points = {
	Vector( 128, 128 ),
	Vector( 384, 128 ),
	Vector( 384, 384 ),
	Vector( 128, 384 )
}

hook.Add( "HUDPaint", "math.BezierLerp", function()
	local frac = RealTime() % 1
	local point = math.CubicBezier( frac, unpack( points ) )

	surface.SetDrawColor( 255, 255, 0 )
	surface.DrawRect( point.x - 4, point.y - 4, 8, 8 )
	
	surface.DrawRect( points[1].x - 4, points[1].y - 4, 8, 8 )
	surface.DrawRect( points[2].x - 4, points[2].y - 4, 8, 8 )
	surface.DrawRect( points[3].x - 4, points[3].y - 4, 8, 8 )
	surface.DrawRect( points[4].x - 4, points[4].y - 4, 8, 8 )
end )
```
Example for the wiki

PR #2005, opened 2023-08-09
Small optimisation

MoskalykA:
Hi, there,
I suggest using `Player:GetWeapon` directly to get the toolgun


MoskalykA:
Sorry for the commit message

robotboy655 (approved):

robotboy655:
I changed it a bit, the function was basically duplicated for client and server separately, so I made it be one shared function for both.

PR #2007, opened 2023-08-10
Being able to configure the fog of the world and the sky at the same time

MoskalykA:
Hi, there,
I suggest applying what the todo suggested


robotboy655 (changes requested):
I think if we are to add this, it should have a toggle (bool networkvar) that is disabled by default to preserve old behavior (for dupes and stuff), as well as ease of use incase you do NOT want this behavior, so you do not need to manually copy over values in the UI or wherever else.

PR #2011, opened 2023-09-11
terrortown: use `IsValid()` to check `wep.AmmoEnt` in `DropActiveAmmo()`

RalphORama:
While most of the TTT code uses the global `IsValid()` to check validity of entities, a less accurate `if not wep.AmmoEnt` check is currently used in `weaponry.lua`.

Changing this check to use `IsValid()` produces more accurate results and prevents issues when porting other SWEP bases to TTT.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/weaponry.lua
@@ -248,7 +248,7 @@ local function DropActiveAmmo(ply)
    local wep = ply:GetActiveWeapon()
    if not IsValid(wep) then return end
 
-   if not wep.AmmoEnt then return end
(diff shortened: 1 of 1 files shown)


svdm:
`AmmoEnt` is a string, the name of the ammo entity, rather than an entity itself. For example: `item_ammo_pistol_ttt`. Worth mentioning that this is also [documented as such in the custom weapon docs](https://ttt.badking.net/development/sweps/#required-fields), which is interesting material for anyone porting a SWEP base.

So, this check is correct as it is and IsValid would break it.

Kefta:
Should probably throw some message/error if the ammo ent couldn't be created so SWEP authors aren't scratching their heads from a minor spelling error.

PR #2012, opened 2023-09-14
Allow clamp to catch NaNs

thegrb93:
math.min and math.max will return NaN if the input is NaN so this fixes that.


thegrb93:
Probably better to catch nan where its possible to find them actually

Kefta:
Agreed.

PR #2014, opened 2023-09-28
Digital Paintball maps

BreakinBenny:
From this discontinued mod: https://www.moddb.com/mods/digital-paintball-source

Sucks that Digital Paintball Redux (which is a Half-Life mod) was quickly abandoned.


robotboy655:
I don't think it would make sense to add integrations with obscure mods that are not mountable.

BreakinBenny:
That's a good point, and just because the map rp_evocity_v2d made use of some textures from that dead mod it doesn't get special treatment.

PR #2016, opened 2023-10-08
Fix error when using sent_ball on non-singleplayer server

callumok2004:
Fixes `Trying to remove server entity sent_ball#335 on client!`

### DIFF
--- a/garrysmod/lua/entities/sent_ball.lua
+++ b/garrysmod/lua/entities/sent_ball.lua
@@ -141,7 +141,7 @@ end
 
 function ENT:Use( activator, caller )
 
-	if SERVER and activator:IsPlayer() then
+	if ( SERVER && activator:IsPlayer() ) then
 
 		self:Remove()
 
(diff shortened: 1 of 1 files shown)


garrysmod/lua/entities/sent_ball.lua
```diff
@@ -141,9 +141,9 @@ end
 
 function ENT:Use( activator, caller )
 
-	self:Remove()
+	if ( activator:IsPlayer() ) and ( ( game.SinglePlayer() and CLIENT ) or SERVER ) then
```
Kefta, line 144: Just change this check to SERVER, entities should never be removed on the client.

Kefta (changes requested):

Kefta:
Add the parentheses back around the if-condition and this is good to go.

robotboy655:
I cannot reproduce such an issue.

https://wiki.facepunch.com/gmod/ENTITY:Use
`ENTITY:Use` is not even called on the client by the base game as far as I am aware.

robotboy655:
Closing, as this tries to fix a non-existent issue as far as I can tell, and I see no replies to my comment above.

PR #2019, opened 2023-10-15
Add OnPlayerRequestRespawn hook

oninoni:
If a player presses a button, to respawn the base gamemode currently just calls spawn.
This way there was no way of actually detecting this kind of spawn.

I added a simple hook, that is called before Spawning the player, to allow programmers, to have behavior depending on a "normal respawn"

I feel like people could use such a hook in general, so I'm writing a pull request.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/player.lua
+++ b/garrysmod/gamemodes/base/gamemode/player.lua
@@ -98,6 +98,8 @@ function GM:PlayerDeathThink( pl )
 
 	if ( pl:IsBot() || pl:KeyPressed( IN_ATTACK ) || pl:KeyPressed( IN_ATTACK2 ) || pl:KeyPressed( IN_JUMP ) ) then
 
+		hook.Run( "OnPlayerRequestRespawn ", pl )
+
 		pl:Spawn()
 
 	end


robotboy655:
https://wiki.facepunch.com/gmod/GM:PlayerSpawn

This hook will detect such spawns.

oninoni:
Yes, but not exclusively. 

If ply:Spawn() is called from somewhere else the function is also called.
This would give a nice exclusive way of detecting "normal spawns"

robotboy655:
But this IS the "normal respawn" hook, as opposed to the initial respawn one.

I do not think we need a new hook for every possible scenario where a certain function might be called.
What exactly do you want to do differently when the player respawns from that hook anyway? 

Not to mention, there's also no way to guarantee that in non Sandbox gamemodes this new hook will be called at all.

oninoni:
I want to decide between respawning through the normal means and custom means from inside the spawn hook.
For this i need to mark it from somewhere.

Its fine if you dont want it, i can just override the functionality in my gamemode.

robotboy655:
That would make most sense in my head.

What I'd do is set a variable on a player just before calling `ply:Spawn()` and then check for it in `GM:PlayerSpawn`, do my custom logic, and unset it.

That way you can have your custom logic for your special respawns, and it will work across any gamemode.

PR #2020, opened 2023-10-18
Clear DOF_Ents on DOF_Kill

brandonsturgeon:
### Summary
`DOF_Kill` doesn't actually clean up the `DOF_Ents` table, meaning the table could expand indefinitely.

With this change, the `DOF_Ents` table is efficiently emptied whenever `DOF_Kill` is called.

### Demo
```lua
local ed = EffectData()

print( "Initial count:", #DOF_Ents )

util.Effect( "dof_node", ed )
print( "First Created:", #DOF_Ents )

DOF_Kill()
print( "After normal DOF_Kill(): ", #DOF_Ents )

util.Effect( "dof_node", ed )
print( "Second Created: ", #DOF_Ents )

_DOF_Kill()
print( "After modified DOF_Kill(): ", #DOF_Ents )
```

Outputs:
```
Initial count:	0
First Created:	1
After normal DOF_Kill(): 	1
Second Created: 	2
After modified DOF_Kill(): 	0
```

### DIFF
--- a/garrysmod/lua/postprocess/dof.lua
+++ b/garrysmod/lua/postprocess/dof.lua
@@ -13,11 +13,15 @@ local NUM_DOF_NODES = 16
 
 function DOF_Kill()
 
-	for k, v in pairs( DOF_Ents ) do
+	for i = #DOF_Ents, 1, -1 do
+
+		local v = DOF_Ents[i]
 
 		if ( IsValid( v ) ) then
 			v:Remove()
 		end
+
+		DOF_Ents[i] = nil
 
 	end
 


garrysmod/lua/postprocess/dof.lua
```diff
@@ -13,12 +13,16 @@ local NUM_DOF_NODES = 16
 
 function DOF_Kill()
 
-	for k, v in pairs( DOF_Ents ) do
+	for i = #DOF_Ents, 1, -1 do
 
-		if ( IsValid( v ) ) then
+		local v = DOF_Ents[i]
+
+		if ( v:IsValid() ) then
```
brandonsturgeon, line 20: I believe this change is safe - it's only entities in here, and even `NULL`s have `:IsValid`

garrysmod/lua/postprocess/dof.lua
```diff
@@ -13,12 +13,16 @@ local NUM_DOF_NODES = 16
 
 function DOF_Kill()
 
-	for k, v in pairs( DOF_Ents ) do
+	for i = #DOF_Ents, 1, -1 do
```
brandonsturgeon, line 16: Looping through in reverse means we can safely set `DOF_Ents[i] = nil` without the cost of reordering any elements, and without needing to do another loop or call `table.Empty`

robotboy655: Is there any reason to bother with this over just doing `DOF_Ents = {}` at the end of the function?

garrysmod/lua/postprocess/dof.lua
robotboy655: Unfortunately DOF_Ents is a global table, so addons could be adding things into it.

I also see no reason to change this in the first place as it is not performance intensive code.

robotboy655 (approved):

PR #2021, opened 2023-10-27
Polyfill table.move for 32-bit Gmod

Denneisk:
Implements [`table.move`](https://wiki.facepunch.com/gmod/table.move) for 32-bit Gmod as accurately as possible, including errors, to preserve forward-compatibility. This solution is, naturally, not completely optimal.

Comment out the if statement and rename the function to compare outputs.

```lua
PrintTable( table.move( { 1, 2, 3, 4 }, 1, 4, -4 ) )
--[[ Outputs:
-4	=	1
-3	=	2
-2	=	3
-1	=	4
1	=	1
2	=	2
3	=	3
4	=	4
]]
```

Additionally, the documentation on the wiki is incorrect. The 4th argument, `dest`, is required (you can test this yourself).

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -752,3 +752,30 @@ function table.Flip( tab )
 	return res
 
 end
+
+-- Polyfill for table.move on 32-bit
+-- Don't forget to remove this when it's no longer necessary
+if ( !table.move ) then
+	function table.move( sourceTbl, from, to, dest, destTbl )
+
+		if ( !istable( sourceTbl ) ) then error( "bad argument #1 to 'move' (table expected, got " .. type( sourceTbl ) .. ")" ) end
+		if ( !isnumber( from ) ) then error( "bad argument #2 to 'move' (number expected, got " .. type( from ) .. ")" ) end
+		if ( !isnumber( to ) ) then error( "bad argument #3 to 'move' (number expected, got " .. type( to ) .. ")" ) end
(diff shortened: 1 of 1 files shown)


Kefta:
LuaJIT will be upgraded on main when it's branch merge time. I don't see the need to implement this in Lua right now.

Grocel:
> LuaJIT will be upgraded on main when it's branch merge time. I don't see the need to implement this in Lua right now. Servers can do it themselves for main branch clients.

I don't think this will happen any time soon. We have been in limbo state on several awaited features for YEARS. So better account for things taking a while, then speculate on "soon™". Things taking time is how it.

So add polyfills as long they are needed, especially then they were already done for you via a PR. If you want cross branch compatibility than do it properly. Otherwise it will just create additional unnecessary community pressure on the dev.

I have seen a lot of tip toeing around over potential braking compatibility changes over the years.

robotboy655 (approved):

bjcscat (commented):
Could be cleaner if assert was used.
`
        assert(istable( sourceTbl ), "bad argument #1 to 'move' (table expected, got " .. type( sourceTbl ) .. ")" )

        assert(isnumber( from ), "bad argument #2 to 'move' (number expected, got " .. type( from ) .. ")" )

        assert(isnumber( to ), "bad argument #3 to 'move' (number expected, got " .. type( to ) .. ")" )

        assert(isnumber( dest ), "bad argument #4 to 'move' (number expected, got " .. type( dest ) .. ")" )

        if destTbl ~= nil then

            assert(istable( destTbl ), "bad argument #5 to 'move' (table expected, got " .. type( destTbl ) .. ")")

        else

            destTbl = sourceTbl

        end
`

PR #2023, opened 2023-10-31
TTT: Prevent tttbase Lua errors with NPCs

BuzzHaddaBig:
NPCs create Lua errors when shooting tttbase weapons by calling self:GetOwner():Alive().

Return early if an NPC to prevent calling Alive().


Malivil:
This change puts the `IsNPC` check before the owner is checked for validity.

You can fix the original error by rearranging the if to this: `if (not IsValid(self:GetOwner())) or self:GetOwner():IsNPC() or (not self:GetOwner():Alive()) then return end`

PR #2024, opened 2023-11-05
Refactor list.lua

Kefta:
- Refactor for optimisation, better var naming, and new code style standards
- Remove doc comments
- Add GetStored which returns a list without creating one if it doesn't exist unlike GetForEdit

### DIFF
--- a/garrysmod/lua/includes/modules/list.lua
+++ b/garrysmod/lua/includes/modules/list.lua
@@ -1,90 +1,62 @@
---=============================================================================--							 
+--=============================================================================--
 --
 --  A really simple module to allow easy additions to lists of items
 --
 --=============================================================================--
 
-local table 	= table
 local pairs		= pairs
+local table 	= table
+
 
 module( "list" )
 
 
-local g_Lists = {}
+local Lists = {}
 
---
---	Get a list
---
-function Get( list )
-
-	g_Lists[ list ] = g_Lists[ list ] or {}
-	return table.Copy( g_Lists[ list ] )
-	
+function GetStored( listid )
+	return Lists[ listid ]
 end
 
---
---	Get a list
---
-function GetForEdit( list )
+function GetForEdit( listid )
+	local list = Lists[ listid ]
 
-	g_Lists[ list ] = g_Lists[ list ] or {}
-	return g_Lists[ list ]
-	
+	if ( list == nil ) then
+		list = {}
+		Lists[ listid ] = list
+	end
+
+	return list
 end
 
---
---	Get all list names
---
+function Get( listid )
+	return table.Copy( GetForEdit( listid ) )
+end
+
 function GetTable()
-
-	return table.GetKeys( g_Lists )
-
+	return table.GetKeys( Lists )
 end
 
---
---	Set a key value
---
-function Set( list, key, value )
-
-	local list = GetForEdit( list )
-	list[ key ] = value
-
+function Set( listid, key, val )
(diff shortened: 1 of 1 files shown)


garrysmod/lua/includes/modules/list.lua
```diff
@@ -1,90 +1,56 @@
...
---
---	Get a list
---
-function GetForEdit( list )
+local Lists = {}
 
-	g_Lists[ list ] = g_Lists[ list ] or {}
-	return g_Lists[ list ]
-	
+function GetStored( listid )
```
robotboy655, line 10: This could probably be just an argument for `GetForEdit`.

garrysmod/lua/includes/modules/list.lua
```diff
@@ -1,90 +1,56 @@
...
+	return table.GetKeys( Lists )
 end
 
---
---	Returns true if the list contains the value (as a value - not a key)
---
-function Contains( list, value )
-
-	if ( !g_Lists[ list ] ) then return false end
+function Set( listid, key, val )
```
robotboy655, line 33: How is `val` is "better var naming" than `value`? There's no good reason for this change as far as I can tell

It's shorter? Sure. Why does it need to be 2 characters shorter?

robotboy655 (changes requested):
I really don't understand why so many unnecessary changes are included.

Just look how much nicer the diffs can be without them:
image

I can understand the function body changes, but what does rearranging functions around and removing all the comments and whitespace achieve exactly?

garrysmod/lua/includes/modules/list.lua
Kefta: For naming consistency. ``key, val`` is the standard arg naming for most Lua codebases I've come across (and ``k, v`` for loops). ``val`` ( https://github.com/search?q=repo%3AFacepunch%2Fgarrysmod+val+language%3ALua&type=code&l=Lua ) is used more than ``value`` ( https://github.com/search?q=repo%3AFacepunch%2Fgarrysmod+value+language%3ALua&type=code ) in the gmod codebase currently (rough search but val seems to be the most used for function args). This doesn't really "matter" but the goal of these file-level refactor PRs is to set a consistent and optimal standard of practices for the entire codebases that devs can learn from, and common arg naming is just an extension of that.

Kefta:
I have restored the function ordering but I am leaving the doc comments removed as per our discussion in Discord relating to duplicate/outdated documentation.

garrysmod/lua/includes/modules/list.lua
robotboy655: This is an unbelievably silly argument. Not only that, your example doesn't really work if you do it more properly:

image
image

(This is across the GMod repo, granted the word `value` appears in some comments as well)

This feels more of a you forcing your own preferences into the repo.

robotboy655: I'd also would like to invite you to read https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md

Kefta: I highly disagree that's it's "silly" since it's introducing consistency though I'll concede that it's pedantic. If it will get the review thru, I'll revert it.

Kefta: Also, value is inflated in searches because of member variables which (correctly) use the full name, but alas.

robotboy655 (approved):

PR #2025, opened 2023-11-06
Bug Fix: Infinite loop on string.Comma when second parameter is numerical.

Krelsis:
# Description
When string.Comma is called with a number as the second parameter, the function enters an infinite loop.

**Current behaviour:** Gets stuck in an infinite loop. Effectively crashing the game.
**Suggested Behaviour:** Default behaviour ("," is inserted)

## Changes
- Default behaviour is exhibited when the second parameter is a number.

# Testing
Testable on https://www.lua.org/cgi-bin/demo , paste the code, hit run.
- isnumber is stubbed to return true for the purpose of the test on the demo.

## Original
```lua
-- Stub isnumber to return true.
function isnumber(val)
	return true
end

function string.Comma( number, str )

	local replace = str == nil and "%1,%2" or "%1" .. str .. "%2"

	if ( isnumber( number ) ) then
		number = string.format( "%f", number )
		number = string.match( number, "^(.-)%.?0*$" ) -- Remove trailing zeros
	end

	local index = -1
	
	-- keep track of iteration count
	local loopCount = 1
	
	while index ~= 0 do 
		number, index = string.gsub( number, "^(-?%d+)(%d%d%d)", replace ) 

		-- break out of loop after 30 iterations, no point infinitely looping.
		 loopCount = loopCount + 1
		 if loopCount > 30 then
			 break
		 end
	end

	return number

end

print(string.Comma(23432))     -- Output: 23,432
print(string.Comma(23432, "2"))  -- Output: 23222222222222222222222222222222432
```

## New
```lua
-- Stub isstring and isnumber to return true.
function isnumber(val)
        return true
end

function isstring(val)
        return true
end
 
function string.Comma( number, str )

	if ( str ~= nil and not isstring( str ) ) then
		error( "bad argument #2 to 'string.Comma' (string expected, got " .. type( str ) .. ")" )
	elseif ( str ~= nil and string.match(str, "%d") ~= nil ) then
		error( "bad argument #2 to 'string.Comma' (Non-numerical values expected, got " .. str .. " )" )
	end

	local replace = str == nil and "%1,%2" or "%1" .. str .. "%2"

	if ( isnumber( number ) ) then
		number = string.format( "%f", number )
		number = string.match( number, "^(.-)%.?0*$" ) -- Remove trailing zeros
	end

	local index = -1
	while index ~= 0 do number, index = string.gsub( number, "^(-?%d+)(%d%d%d)", replace ) end

	return number

end

print(string.Comma(23432))     -- Output: 23,432
print(string.Comma(23432, "2"))  -- bad argument #2 to 'string.Comma' (Non-numerical values expected, got 22
```

### DIFF
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -358,8 +358,8 @@ function string.Comma( number, str )
 
 	if ( str ~= nil and !isstring( str ) ) then
 		error( "bad argument #2 to 'string.Comma' (string expected, got " .. type( str ) .. ")" )
-	elseif ( str ~= nil and isnumber( str ) ) then
-		error( "bad argument #2 to 'string.Comma' (Non-numerical value expected, got " .. str .. " )" )
+	elseif ( string.match(str, "%d") ~= nil ) then
+		error( "bad argument #2 to 'string.Comma' (Non-numerical values expected, got " .. str .. " )" )
 	end
 
 	local replace = str == nil and "%1,%2" or "%1" .. str .. "%2"
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -355,8 +355,7 @@ function string.ToColor( str )
 end
 
 function string.Comma( number, str )
-
-	local replace = str == nil and "%1,%2" or "%1" .. str .. "%2"
+	local replace = ( str == nil or isnumber( str ) ) and "%1,%2" or "%1" .. str .. "%2"
 
 	if ( isnumber( number ) ) then
 		number = string.format( "%f", number )
--- a/garrysmod/lua/includes/extensions/string.lua
+++ b/garrysmod/lua/includes/extensions/string.lua
@@ -356,7 +356,11 @@ end
 
 function string.Comma( number, str )
 
-	local replace = ( str == nil or isnumber( str ) ) and "%1,%2" or "%1" .. str .. "%2"
+	if ( isnumber( str ) ) then
+		error( "bad argument #2 to 'string.Comma' (Non-numerical value expected, got " .. str .. " )" )
+	end
+
+	local replace = str == nil and "%1,%2" or "%1" .. str .. "%2"
 
 	if ( isnumber( number ) ) then
 		number = string.format( "%f", number )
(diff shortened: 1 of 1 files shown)


Kefta:
Instead of ignoring the number, cast it to a string to match built-in string lib functions.

Krelsis:
> Instead of ignoring the number, cast it to a string to match built-in string lib functions.

Even if the second parameter was cast to a string the pattern used in gsub will still acknowledge it as a number, thus still encountering the infinite loop. 

It's likely better to ignore numbers that are given as a second parameter as it will only make the output string ambiguous and remain illegible (232432 would be output rather than 23432 with some sort of legible separator; and suddenly a whole different number is represented). I don't see a situation where this behaviour would be desired.

Kefta:
Ah you're right, I would throw an explicit arg error then instead of silently ignoring the second argument for a single type.

garrysmod/lua/includes/extensions/string.lua
```diff
@@ -356,7 +356,7 @@ end
 
 function string.Comma( number, str )
 
-	local replace = str == nil and "%1,%2" or "%1" .. str .. "%2"
+	local replace = ( str == nil or isnumber( str ) ) and "%1,%2" or "%1" .. str .. "%2"
```
robotboy655, line 359: Yeah I think it'd be better to throw an explicit error rather than fail silently, similarly like it is done here:
https://github.com/Facepunch/garrysmod/blob/9bbd7c8af0dda5bed88e3f09fbdf5d4be7e012f2/garrysmod/lua/includes/extensions/util.lua#L234

robotboy655 (changes requested):

garrysmod/lua/includes/extensions/string.lua
Krelsis: Made a change, now throwing an explicit error, let me know if the thrown message is acceptable.

robotboy655:
```
	if ( str ~= nil and !isstring( str ) ) then
		error( "bad argument #2 to 'string.Comma' (string expected, got " .. type( str ) .. ")" )
	end
```
This would be more consistent with other functions.

Krelsis: > ```
> 	if ( str ~= nil and !isstring( str ) ) then
> 		error( "bad argument #2 to 'string.Comma' (string expected, got " .. type( str ) .. ")" )
> 	end
> ```
> 
> This would be more consistent with other functions.

Whilst this will stop a true `2` from being passed, my only concern with this if statement, is that it will still allow `"2"` to be passed as `"2"` is a valid string where `isstring` is concerned, however the pattern matching of `gsub` (` ^(-?%d+)(%d%d%d) `) cares not for the data type and will still match on `"2"` and "2" indiscriminately; the infinite loop would still be encountered. We need to explicitly check that it is not a number here to prevent the infinite loop.

robotboy655 (approved):

PR #2026, opened 2023-11-09
Fix GetHighestKey not choosing anything if values are all 0

WinterPhoenix:
Fixes totally empty gamemodes not using *any* description, instead falling back to internal folder name.

Before:
image

After:
image

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -535,7 +535,7 @@ function GetHighestKey( obj )
 
 	for ( k in obj )
 	{
-		if ( obj[k] > h )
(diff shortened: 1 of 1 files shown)


garrysmod/html/js/menu/control.Servers.js
```diff
@@ -535,7 +535,7 @@ function GetHighestKey( obj )
 
 	for ( k in obj )
 	{
-		if ( obj[k] > h )
+		if ( h == 0 || obj[k] > h )
```
robotboy655, line 538: Probably better to just set initial value of `h` to `Number.MIN_VALUE`, no?

robotboy655 (changes requested):

robotboy655 (approved):

PR #2027, opened 2023-11-17
Create turkish.lua

NovaDiablox:
Directly copied from my localization from TTT2 "https://github.com/TTT-2/TTT2/tree/master" project. Since I don't know any sorting formula in excel, I leave devs to sort the different lines that doesn't exist in the original TTT.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/turkish.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/turkish.lua
@@ -1,0 +1,1977 @@
+-- Turkish language strings
+
+local L = LANG.CreateLanguage("Turkish")
+
+-- General text used in various places
+L.traitor = "Hain"
+L.detective = "Dedektif"
+L.innocent = "Masum"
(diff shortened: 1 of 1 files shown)


robotboy655:
> Since I don't know any sorting formula in excel, I leave devs to sort the different lines that doesn't exist in the original TTT.

You should definitely remove the extraneous stuff. Your file has almost twice as many lines as the english translation one does.  

Have you even tested that your translation file provides all the necessary lines for TTT? Just random spot checking shows lines such as `L.ev_goomba` being untranslated. I bet all of the tips are also mismatched, and some of them I bet are irrelevant for original TTT.

NovaDiablox:
The translation was made with a help of a CAT tool and it might be the reason why some code was translated or broken in the process. I didn't really had time to test it in game.I will try to find errors and fix it soon. Can probably deliver it in less than a week. Thanks for your feedback.

NovaDiablox:
Currently translating it from scratch you can close or delete the request sir.

PR #2029, opened 2023-11-19
TTT: Replace most SteamID usage with SteamID64

T-Rizzle12:
#1182 made changes in TTT that replaced UniqueID usage with SteamID. There was an issue with it breaking usage with bots. A recent Gmod update made bots return proper SteamID64 values in the client. This pull request changes the same things #1182 changed, but using SteamID64 instead. I did my best to maintain backward compatibility.


svdm:
For compatibility I suspect it would be better to not rename all the various fields/vars from `sid` to `sid64` (and `killer_sid` -> `killer_sid64` etc). This rename will definitely break anyone interacting with those things, whereas the value was a string and will be a string, so unless the content is being inspected things could just continue to work without a rename.

T-Rizzle12:
> For compatibility I suspect it would be better to not rename all the various fields/vars from `sid` to `sid64` (and `killer_sid` -> `killer_sid64` etc). This rename will definitely break anyone interacting with those things, whereas the value was a string and will be a string, so unless the content is being inspected things could just continue to work without a rename.

Ok, when I get the time, I will change sid64 back to sid and killer_sid64 to killer_sid. I will also test to make sure the changes don't break anything internally.

T-Rizzle12:
Done, `sid64` is back to `sid` and `killer_sid64` is back to `killer_sid`. I still have to test and make sure that nothings breaks first!

mgetJane:
i don't think this is a good idea, ttt has been using steamid over steamid64 for 7 yrs and we don't know how many addons created within that time expect this behaviour

i really doubt bot support is a real concern for ttt, people only ever use bots in ttt for debugging/testing purposes and when i need bots to have unique steamids i just override Player:SteamID for the session

Malivil:
I agree with Jane. I think the original implementation with sid64 and killer_sid64 being added and used while leaving sid for backwards compatibility is the better direction.

T-Rizzle12:
I doubt that many addons will be expecting a full SteamID instead of SteamID64, but I rather leave this decision to @svdm

T-Rizzle12:
Feel free to test this with whatever TTT addons you use and let me know if any errors occur.

Malivil:
I know for a fact that there are addons that use a player's SteamID to find `prop_ragdoll` whose `sid` matches in order to delete that ragdoll. Are there better ways that don't require doing that? Of course, but that doesn't stop the fact that it is being done and this will break things like that.

T-Rizzle12:
> I agree with Jane. I think the original implementation with sid64 and killer_sid64 being added and used while leaving sid for backwards compatibility is the better direction.
I just saw that Robotboy agrees with the backward compatibility approach, so I will change it back when I get the time. Unless @svdm has any objections to this.

svdm:
I can get behind an implementation where the old SteamID values are left in place for compatibility, while TTT uses the 64 values (ie. a `rag.sid` next to a `rag.sid64`). However, while those ragdoll fields are easy to duplicate, this is not feasible for other cases:

- Player and ragdoll `kills` list: we'd have to maintain two lists here which is nasty. I don't think that's worth it, and from a Github search I don't see any obvious reliance on old SteamIDs here.
- `GAMEMODE.LastRole[ply:SteamID()]` and a similar case in KARMA. I can find a few addons looking at this on Github, but pretty old stuff and nothing that would explode, so this could just go to 64.
- The scoring system, it would be quite bloating to have double IDs here. Probably not common for addons to do anything here (it's harder to search for), so 64 IDs should again be acceptable.

So unless I'm very wrong on any of these, I think the only change would be to turn it into `sid` + `sid64` and `killer_sid` + `killer_sid64`, only for those two cases.

T-Rizzle12:
> I can get behind an implementation where the old SteamID values are left in place for compatibility, while TTT uses the 64 values (ie. a `rag.sid` next to a `rag.sid64`). However, while those ragdoll fields are easy to duplicate, this is not feasible for other cases:
> 
> * Player and ragdoll `kills` list: we'd have to maintain two lists here which is nasty. I don't think that's worth it, and from a Github search I don't see any obvious reliance on old SteamIDs here.
> * `GAMEMODE.LastRole[ply:SteamID()]` and a similar case in KARMA. I can find a few addons looking at this on Github, but pretty old stuff and nothing that would explode, so this could just go to 64.
> * The scoring system, it would be quite bloating to have double IDs here. Probably not common for addons to do anything here (it's harder to search for), so 64 IDs should again be acceptable.
> 
> So unless I'm very wrong on any of these, I think the only change would be to turn it into `sid` + `sid64` and `killer_sid` + `killer_sid64`, only for those two cases.

Alright, I will work on this when I get the time.

Malivil:
> I can get behind an implementation where the old SteamID values are left in place for compatibility, while TTT uses the 64 values (ie. a `rag.sid` next to a `rag.sid64`). However, while those ragdoll fields are easy to duplicate, this is not feasible for other cases:
> 
>     * Player and ragdoll `kills` list: we'd have to maintain two lists here which is nasty. I don't think that's worth it, and from a Github search I don't see any obvious reliance on old SteamIDs here.
> 
>     * `GAMEMODE.LastRole[ply:SteamID()]` and a similar case in KARMA. I can find a few addons looking at this on Github, but pretty old stuff and nothing that would explode, so this could just go to 64.
> 
>     * The scoring system, it would be quite bloating to have double IDs here. Probably not common for addons to do anything here (it's harder to search for), so 64 IDs should again be acceptable.
> 
> 
> So unless I'm very wrong on any of these, I think the only change would be to turn it into `sid` + `sid64` and `killer_sid` + `killer_sid64`, only for those two cases.

I think that's fair. The ragdoll was my main point on concern and i agree that doing double values in other places is more bloat with less benefit.

mgetJane:
imo just do the change only for bots since the goal here is bot compatibility

T-Rizzle12:
> imo just do the change only for bots since the goal here is bot compatibility

I'm afraid, that is not feasible. It's pointless to have a bunch of checks for bots since like svdm said before it's not worth the hassle. When I get the time here is what I will do:
- [ ] Readd rag,sid64 while keeping rag.sid for backwards compatibility
- [ ] LastRole will use SteamID64 instead of SteamID as keys
- [ ] KARMA will use SteamID64 instead of SteamID as keys
- [ ] The scoring system will use sid64 and killer_sid64
- [ ] Readd killer_sid64 while keeping killer_sid for backwards compatibility

Let me know if I missed anything.....

T-Rizzle12:
Done, I added back `sid64` and `killer_sid64`, but left `sid` and `killer_sid` for backward compatibility.

TimGoll:
FYI, we switched to steamID64 in TTT2 years ago and we never noticed any addon breaking because of this

PR #2032, opened 2023-11-30
Improve SortedPairs even more

Be1zebub:
PR #2015 is good, but we can make it much better.

Improvements:

- `table.insert( tbl, val )` > `tbl[ #tbl + 1 ] = val`
- `next( keys, i )` > `keys [ i ]`

benchmarks:

- brandonsturgeon's version is 11.23% better than default
- my version is 36% better than default
- or 27.92% better than brandonsturgeon's

bench src: https://gist.github.com/Be1zebub/4b90791fc7ab2d89a18c7a93da143095
<details>
	<summary>bench results:</summary>
	
</details>

benchmarks have been tested on a default branch, in p2p server.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -536,10 +536,11 @@ end
 
 local function getKeys( tbl )
 
-	local keys = {}
+	local keys, i = {}, 0
 
 	for k in pairs( tbl ) do
-		keys[ #keys + 1 ] = k
+		i = i + 1
+		keys[ i ] = k
 	end
 
 	return keys
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -539,7 +539,7 @@ local function getKeys( tbl )
 	local keys = {}
 
 	for k in pairs( tbl ) do
-		table.insert( keys, k )
+		keys[ #keys + 1 ] = k
 	end
 
 	return keys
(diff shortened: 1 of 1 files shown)


Kefta:
Does caching #keys and incrementing a local improve perf?

Be1zebub:
@Kefta do you mean replace
```lua
local keys = {}

for k in pairs( tbl ) do
	keys[ #keys + 1 ] = k
end
```
with
```lua
local keys, i = {}, 0

for k in pairs( tbl ) do
	i = i + 1
	keys[ i ] = k
end
```

`#` doesnt actually calculate the amount, lua just get the length from the cache.
||at least jit store len in cache, idk how it works in lua puc - but I has read about how jit store and process tables||
but because its C-call, its 30% slower then use manual calculated number.
we can get 2-5% more perf with this microoptimization.

<details>
	<summary>bench results:</summary>
	
</details>

Heyter:
:)

```lua
-- SortedPairs:
	-- sum = 0.168
	-- avg = 0.00168
	-- median = 0.0010000000000012
-- SortedPairs new:
	-- sum = 0.082000000000001
	-- avg = 0.00082000000000001
	-- median = 0.00099999999999945
do
	local pairs_aux = pairs({})
	local index, keys = 0, {}
	local sort = table.sort
	local function fn1(a, b) return a > b end
	local function fn2(a, b) return a < b end
	function SortedPairs(tbl, desc)
		keys, index = {}, 0

		for k in pairs_aux, tbl do
			index = index + 1
			keys[index] = k
		end

		sort(keys, desc and fn1 or fn2)
		return pairs_aux, tbl
	end
end
```

robotboy655 (approved):

PR #2033, opened 2023-11-30
table.Print - a better version of PrintTable

Be1zebub:
A better version of `PrintTable` with lua syntax based output.

there is also serializer `local str = table.ToPlain( tbl, nopretty )`
this is useful for things like configs

<details>
	<summary>How it looks:</summary>
	
	
</details>

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -801,3 +801,153 @@ if ( !table.move ) then
 
 	end
 end
+
+--[[---------------------------------------------------------
+	table.Print
+		A better version of PrintTable
+-----------------------------------------------------------]]
+
+local tostringGmodTypes
+
+do
+	local format1 = "%s(%s)"
+	local format2 = "%s(%s, %s)"
+	local format3 = "%s(%s, %s, %s)"
+	local format4 = "%s(%s, %s, %s, %s)"
+
+	local function Round( num )
+		return math.Round( num, 3 )
+	end
+
+	local formaters = {}
+
+	function formaters.Vector( vec )
+		if ( vec.z == 0 ) then
+			if ( vec.y == 0 ) then
+				return format1:format( "Vector", Round( vec.x ) )
+			end
+
+			return format2:format( "Vector", Round( vec.x ), Round( vec.y ) )
+		end
+
+		return format3:format( "Vector", Round( vec.x ), Round( vec.y ), Round( vec.z ) )
+	end
+
+	function formaters.Angle( ang )
+		if ( ang.r == 0 ) then
+			if ( ang.y == 0 ) then
+				return format1:format( "Angle", Round( ang.p ) )
+			end
+
+			return format2:format( "Angle", Round( ang.p ), Round( ang.y ) )
+		end
+
+		return format3:format( "Angle", Round( ang.p ), Round( ang.y ), Round( ang.r ) )
+	end
+
+	function formaters.Color( col )
+		if ( col.a == 255 ) then
+			return format3:format( "Color", col.r, col.g, col.b )
+		end
+
+		return format4:format( "Color", col.r, col.g, col.b, col.a )
+	end
+
+	function tostringGmodTypes( val )
+		if ( formaters[ type( val ) ] ) then
+			return formaters[ type( val ) ](val)
+		end
+
+		if ( IsColor( val ) ) then
+			return formaters.Color( val )
+		end
+
+		return tostring( val )
+	end
+end
+
+function table.ToPlain( tbl, nopretty, lvl, already )
+	already = already or {
+		[ tbl ] = true
+	}
+	lvl = lvl or 1
+
+	local out = {}
+
+	local len = 0
+	for _ in pairs( tbl ) do
+		len = len + 1
+	end
+
+	if ( len == 0 ) then
+		return "{}"
+	end
+
+	local isSeq = len == #tbl
+	local not_isSeq = not isSeq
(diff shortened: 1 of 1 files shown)


Aws0mee:
I think this is a really nice improvement! I've been confused before by the current PrintTable in some scenarios, specifically when dealing with tables in tables. I think you could improve this even further by adding different colors for keys/values and maybe different colors for different types if you want.

Be1zebub:
it would be fun to make the output with syntax highlighting, but i don't think its a good idea.

Aws0mee:
> it would be fun to make the output with syntax highlighting, but i don't think its a good idea.

Why don't you think it's a good idea? I've seen a good amount of people make their own table printing functions similar to yours, but most of them with color. I think it just makes it a little easier to read.

Aws0mee:
> I think its cool, but not necessary feature. it would also inflate the code, turning it into a separate module with 300+ lines - I think we need cleaner output, not a separate overloaded module.
> 
> making syntax highlighting is not so difficult, I know some techniques todo it effectively. I can implement it if the community will support this feature.

I think the community will support it. You can look at Luvit's p() function for example. I think the colors on it are quite useful for quickly determining the type just by glancing at different values in the table.

Zaurzo:
I really like this, 1000 times better than PrintTable. Also, I suggest localizing the result of `formaters[ type( val ) ]` in the tostringGmodTypes function:

```
local format = formaters[ type( val ) ] )

if ( format ) then
	return format(val)
end
```

Be1zebub:
Syntax highlighting has been added!
`table.ToPlain` was deleted because it conflict with new mechanics of work.
`nopretty` arg is removed, it made sense for serialization but not for print.
added `Player` & `Entity` classes support.

image
<details>
	<summary>Source table:</summary>
	
</details>

Zaurzo:
I think the syntax highlighting should be an optional argument; it could get confusing on what realm these values are defined in

Aws0mee:
I think it looks really nice. This will finally remove the need for external libraries for this. Hopefully this can get merged.

> I think the syntax highlighting should be an optional argument; it could get confusing on what realm the values are defined in

I don't see how this can get confusing. If you call the function serverside then it will print serverside values, if you call it clientside then it will print clientside values. However, I think PrintTable should be kept as well to avoid breaking old code or for people who just want more basic functionality.

Zaurzo:
> I think it looks really nice. This will finally remove the need for external libraries for this. Hopefully this can get merged.
> 
> > I think the syntax highlighting should be an optional argument; it could get confusing on what realm the values are defined in
> 
> I don't see how this can get confusing. If you call the function serverside then it will print serverside values, if you call it clientside then it will print clientside values. However, I think PrintTable should be kept as well to avoid breaking old code or for people who just want more basic functionality.

Fair enough, I looked at it from a "calling it shared" view. Just so used to the blue/yellow colors in console lol

Be1zebub:
I was currently developing a grid inventory system, and I realized that I made a few mistakes while making rework that adds colors
image

the current version has few issues, rn im working on a fix and more thorough testing.

Be1zebub:
solved
image
hl2_ihuqyxzbwT
I tested this on 10 large tables from `_G` from large addons like sam admin mod, it works well rn.

Be1zebub:
Now the multiline strings output has a syntax correct.
image
image

CornerPin:
You can use `string.format`'s `q` specifier to make syntactically correct strings. Currently, this doesn't handle cases when a string contains `]]`, `"`, `\`, or any non-printable characters.

Be1zebub:
> You can use `string.format`'s `q` specifier to make syntactically correct strings. Currently, this doesn't handle cases when a string contains `]]`, `"`, `\`, or any non-printable characters.

nah, it does not handle newlines properly.
I think I should add `[[`, `"`, `\12345` escaping manually.

image
image

CornerPin:
```lua
"Hello\
world"
```
is a valid Lua syntax, though I think `\n` would look prettier. You could fix this with `string.gsub`.
```lua
string.gsub(str, "\\(.)", function(char)
    if char == "\n" then return "\\n" end
end)
```

Be1zebub:
> ```lua
> "Hello\
> world"
> ```
> 
> is valid Lua syntax, though I think `\n` would look prettier. You could fix this with `string.gsub`.
no, it doesnt.
> 
> ```lua
> string.gsub(str, "\\\n", "\\n")
> ```
nope, it doesnt works.

Zaurzo:
Small suggestion: Localize `type` at the top to avoid errors from bad addons overriding the function

Be1zebub:
why should anybody care about shitty addons? just uninstall it, if its broken.

Zaurzo:
> why should anybody care about shitty addons? just uninstall it, if its broken.

Fair enough.

Kefta:
It's a library file so all global funcs should be localised but that's for another PR.

garryspins:
love this, wrote this myself many years ago and has been a massive help, even added the function src file on my own version, massive help.

Aspect12:
This is great, I hope it gets merged soon, either as a direct replacement to PrintTable() or as a separate function.

Be1zebub:
PTAL

Be1zebub:
added `compact` & `noPretty` flags
`table.Print( tbl, compact, noPretty )`

image



default is facepunch code style, no behaivor changed.
compact is like `[key] = Color(255, 255, 255)` eg
noPretty is minified, like no indent, no newlines, no spaces.

This commit is aimed at making it easier to read large logs when your debugging is huge.

PR #2035, opened 2023-12-09
made ammo types reflect game.GetAmmoTypes

EntranceJew:
to the benefit of anyone reading the ammo type and trying to pair it up to things without running string.lower/upper across the board.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_stungun.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_stungun.lua
@@ -33,7 +33,7 @@ SWEP.Primary.ClipSize      = 30
 SWEP.Primary.ClipMax       = 60
 SWEP.Primary.DefaultClip   = 30
 SWEP.Primary.Automatic     = true
-SWEP.Primary.Ammo          = "smg1"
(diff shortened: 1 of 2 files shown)


robotboy655:
Ammo types are case insensitive, what is this meant to achieve?

EntranceJew:
In TTT2 I added a [HUD ammo type indicator](https://github.com/TTT-2/TTT2/commit/7485034ea7abacb32443234a3cd27a1a1198f3bf) and SMG1 was the only odd-one-out for code that wasn't already cased to match the lookup. If everything else was right, I figure, what's the harm in this also matching?

robotboy655:
I personally find this to be the wrong approach. What happens when someone installs a weapon that doesn't follow the casing conventions your code expects? Does it just break? Ammo types are case insensitive. They should be treated as such at all times. Yes that means you will have to call `string.lower` on some strings.

Kefta:
I agree that this is useless semantics, but any approach that seeks to match base Lua code/strings with base engine code/strings is fine. At least no one will PR trying to revert this - it should just be merged and we all move on with our lives.

robotboy655 (approved):
This is definitely pointless, but I guess it doesn't hurt either.

PR #2036, opened 2023-12-09
Enhance table.Copy

Zaurzo:
### Changes
**Added Arguments**
`fullCopy`- Will make the function copy Vector and Angle objects.
`noMeta`- Will not set the metatable of the given table to the copy result, if there is one.

**Additional Changes**
The function will yield an error if the first argument isn't a table.
The function now only sets the metatable of the given table to the copy result after everything has been copied.
Updated the command description in the comment above it.

### There is some changes I want to discuss in this PR, however:
1. Is making the function error a bad idea? Will it break some addons?

2. Will making the function set the metatable to the copy after it's done copying cause issues?
- I made this change because a table I was trying to copy had its own __newindex behavior, and since the original function set the metatable before it added all of the values, it caused some problems.

3. Should the `fullCopy` argument even exist, or should I make it always copy Vector and Angle objects?

4. Is the backwards compatibility for the `lookup_table` argument even necessary?

I am looking forward to reading your thoughts on this. Thanks

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -20,30 +20,56 @@ function table.Inherit( t, base )
 end
 
 --[[---------------------------------------------------------
-	Name: Copy(t, lookup_table)
-	Desc: Taken straight from http://lua-users.org/wiki/PitLibTablestuff
-		and modified to the new Lua 5.1 code by me.
-		Original function by PeterPrade!
------------------------------------------------------------]]
-function table.Copy( t, lookup_table )
-	if ( t == nil ) then return nil end
+	Name: Copy( tbl, fullCopy, noMeta, lookupTable )
+	Desc: Creates a deep copy of the given table. 
+	      The fullCopy argument will make it copy Vector and Angles.
+	      The noMeta argument will make it ignore the metatable of the given table. 
+	      The lookupTable argument is optional and is for internal use.
 
-	local copy = {}
-	setmetatable( copy, debug.getmetatable( t ) )
-	for i, v in pairs( t ) do
-		if ( !istable( v ) ) then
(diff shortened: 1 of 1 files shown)


Denneisk:
Would suggest changing `fullCopy` to `copyUserdata` or something more precise to indicate that it's specifically for vectors/angles; might be confused with shallow/deep copying otherwise.

I believe sub-tables shouldn't be effected by `noMeta`, since they might be used for objects.

Zaurzo:
> Would suggest changing `fullCopy` to `copyUserdata` or something more precise to indicate that it's specifically for vectors/angles; might be confused with shallow/deep copying otherwise.
> 
> I believe sub-tables shouldn't be effected by `noMeta`, since they might be used for objects.

Changing `fullCopy` to `copyUserdata` might better, yeah, however this implies is copies all userdata which it doesn't. 

Yeah, changing `noMeta` to not affect sub-tables seems better. I was already aware of this problem of it not setting the metatables for custom objects like util.Timer()

Denneisk:
> Changing `fullCopy` to `copyUserdata` might better, yeah, however this implies is copies all userdata which it doesn't.

Right, then `copyVecAng` or something.

Zaurzo:
> > Changing `fullCopy` to `copyUserdata` might better, yeah, however this implies is copies all userdata which it doesn't.
> 
> Right, then `copyVecAng` or something.

Gotcha.

Kefta:
Why copy vectors and angles but not matrices and colors?

Zaurzo:
> Why copy vectors and angles but not matrices and colors?

Colors: Colors are tables, it already copies them.
Matrices: I forgot about them

Zaurzo:
Added VMatrix support.

PR #2037, opened 2023-12-10
Add ENT:SetXYZ() and ENT:SetAxisIndex()

SeafouryTheBird:
- Add ENT:SetXYZ()
- Add ENT:SetAxisIndex()
Feel free to change my pull request! (:

### DIFF
--- a/garrysmod/lua/entities/widget_axis.lua
+++ b/garrysmod/lua/entities/widget_axis.lua
@@ -12,12 +12,17 @@ local widget_axis_arrow = { Base = "widget_arrow" }
 function widget_axis_arrow:Initialize()
 
 	BaseClass.Initialize( self )
-
+	--///added by seafoury
+	BaseClass.Initialize( base )
+	--///
 end
 
 function widget_axis_arrow:SetupDataTables()
 
 	BaseClass.SetupDataTables( self )
+	--///added by seafoury
+	BaseClass.SetupDataTables( base )
+	--///
 	self:NetworkVar( "Int", 0, "AxisIndex" )
 
 end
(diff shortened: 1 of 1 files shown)


garrysmod/lua/entities/widget_axis.lua
```diff
@@ -12,12 +12,17 @@ local widget_axis_arrow = { Base = "widget_arrow" }
 function widget_axis_arrow:Initialize()
 
 	BaseClass.Initialize( self )
-
+	--///added by seafoury
+	BaseClass.Initialize( base )
```
robotboy655, line 16: What is this meant to do? `base` in this context is `nil`, isn't it?

garrysmod/lua/entities/widget_axis.lua
```diff
@@ -115,6 +122,24 @@ function ENT:Setup( ent, boneid, rotate )
 
 end
 
+--
+-- This was added by Seafoury
```
robotboy655, line 126: It is not necessary to include your name in every line you modify.

garrysmod/lua/entities/widget_axis.lua
```diff
@@ -115,6 +122,24 @@ function ENT:Setup( ent, boneid, rotate )
...
 end
 
+--
+-- This was added by Seafoury
+-- widget_axis.lua
+--
+-- if ENT:SetXYZ() and ENT:SetAxisIndex() don't work
+-- then please tell me and I will try to fix it! (:
+function ENT:SetXYZ( x, y, z, d )
+	base.ArrowZ:SetColor( Color( x, y, z, d ) )
```
robotboy655, line 132: What is this meant to do/achieve?
Why is it only applied to `ArrowZ`?
Again, `base` in this context is `nil`, so this would just error?

robotboy655 (changes requested):
You will need to provide more information on what this is meant to do, why these changes should be accepted.

As it stands right now, exactly none of the proposed changes actually work from what I can see.

PR #2038, opened 2023-12-13
player_manager.GetPlayerClassTable(Player ply)

MerekiDor:
This is a method to access **player class definition tables** more quickly.

Instead of doing this:
```lua
player_manager.GetPlayerClasses()[ player_manager.GetPlayerClass( ply ) ]
```
We can do this:
```lua
player_manager.GetPlayerClassTable( ply )
```

When I access the class table, **I still create a deep copy of the table before it is returned**, since we don't want players to mess with the internal tables! However, it should be more performant, because I avoid copying the _entire_ `Type` table.

It's a very simple change that I came up with while working with player classes for my gamemode, but I'd be glad to contribute even something as small as this.

### DIFF
--- a/garrysmod/lua/includes/modules/player_manager.lua
+++ b/garrysmod/lua/includes/modules/player_manager.lua
@@ -404,7 +404,7 @@ function GetPlayerClassTable( ply )
 	local ct = Type[ util.NetworkIDToString( id ) ]
 	if ( !ct ) then return end
 	
-	return table.Copy(t)
+	return table.Copy(ct)
 
 end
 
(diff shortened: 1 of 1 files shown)


Denneisk:
Wouldn't it be more efficient to copy the single element instead of the entire `Type` table?

MerekiDor:
Oh my. I thought this was exactly what I did, but now that I double-checked, I still copy the entire Type table, even though my description suggests otherwise.

MerekiDor:
Updated & fixed, hopefully. I should keep my eyes peeled next time... Thanks for noticing.

robotboy655 (approved):

PR #2040, opened 2023-12-25
Make DeathNoticeEvent verify if the entities are players

Zaurzo:
Small fix


robotboy655 (approved):

robotboy655:
I don't know when exactly it's possible to hit this, but I guess it couldn't hurt to check if the entity given is valid and is a player.

PR #2041, opened 2024-01-07
Color MetaTable is not accessible. replace debug.getregistry() with setmetatable

KennyHarrer:
debug.getregistry() is deprecated, and returns an empty table. The Color metatable is currently not accessible.

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -12,12 +12,6 @@ function Color( r, g, b, a )
 end
 
 --[[---------------------------------------------------------
-	Register our metatable to make it accessible using FindMetaTable
------------------------------------------------------------]]
-
-setmetatable(Color, COLOR)
-
---[[---------------------------------------------------------
 	Change the alpha of a color
 -----------------------------------------------------------]]
 function ColorAlpha( c, a )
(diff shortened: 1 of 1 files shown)


KennyHarrer:
shoot. this breaks. hold on.

Kefta:
The repo is currently not updated with what has been released, wait for Rubat to push the latest changes.

Quphoria:
The following code was added to lua/include/init.lua
Which overwrites the debug.getregistry function with a version that uses the new MetaTable system
(Change in https://steamdb.info/patchnotes/13095403/)

You may need to update your server/client so it has the latest version of init.lua
In my case the server was pushing a custom init.lua which didn't include this fix

```lua
-- Temporary hack
local meta = {}
function meta.__index(self, key)
    return FindMetaTable(key)
end

local metas = {}
function meta.__newindex(self, key, value)
    metas[ key ] = value
end

debug.getregistry = function()
    local tbl = {}
    setmetatable(tbl, meta)

    return tbl
end

local oldFindMetaTable = FindMetaTable
FindMetaTable = function( name )
    local f = oldFindMetaTable( name )
    if ( f ) then return f end

    return metas[ name ]
end
```

Aws0mee:
This is a little unrelated, but does anyone know if debug.registry will ever be fixed or re-implemented? I noticed a bunch of my code randomly stopped working, and it took me a while to figure out it was because of debug.getregistry being removed. Does anyone happen to have any info whatsoever on why this change happened or why the entire debug library is deprecated now?

PR #2042, opened 2024-01-15
Add vgui.FindByClass() to fix PANEL:Pre/PostAutoRefresh()

Zaurzo:
Due to `debug.getregistry()` being removed, it broke the `PANEL:Pre/PostAutoRefresh()` hooks. This PR adds `vgui.FindByClass()` for `derma` to use to fix this problem.

### DIFF
--- a/garrysmod/lua/derma/derma.lua
+++ b/garrysmod/lua/derma/derma.lua
@@ -25,30 +25,6 @@ SkinMetaTable.__index = function ( self, key )
 
 end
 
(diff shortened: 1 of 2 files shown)


Kefta:
This should be binded in the engine instead.

robotboy655:
I appreciate the idea and effort, but I have implemented a different fix, via a newly added `vgui.GetAll()`.

Zaurzo:
Sounds good to me.

PR #2043, opened 2024-01-19
Fortress Forever spawn support

BreakinBenny:
Fortress Forever may be very niche (niche enough to not even be a mountable game by default), but I'd still like to add this entity... and this is also for futureproofing in the event that there's an engine upgrade. (It won't happen now, but I believe it will. It won't be a modlov, *winks & smiles at Nikolay Aulov*)


robotboy655:
I would maybe agree to add the spawnpoint entity, but not the rest. It's addon territory.

BreakinBenny:
> I would maybe agree to add the spawnpoint entity, but not the rest. It's addon territory.

Oh dear… I made the mistake of having the unrelated extra entities in the master branch, hence the added extra nonsense.

BreakinBenny:
I had to reboot my fork, here's the new PR without the above needless bits:
https://github.com/Facepunch/garrysmod/pull/2044

PR #2045, opened 2024-01-27
Small optimizations to GM:DrawDeathNotice()

Zaurzo:
- Localize the cl_drawhud convar and replace the use of the deprecated GetConVarNumber function with it
- Don't reset the Deaths table if there's nothing in it
- Don't iterate through the Deaths table twice


CoolDude034:
add sex

Zaurzo:
@CoolDude034 No.

PR #2049, opened 2024-02-02
Set cl_smooth to 0 so there is no client smoothing applied to the local player

CoolDude034:
This should fix some network related bugs and maybe improve optimization


robotboy655:
This will not achieve what you claim it will achieve, but feel free to set it to 0 in your own game.

PR #2052, opened 2024-02-02
Clamp damage in TTT

CoolDude034:
This fixes a godmode bug in TTT gamemode


CoolDude034:
pls add pls add pls add pls add

PR #2055, opened 2024-02-10
Fix table.Add infinite loop

zohart75:
Calling table.Add with the same tables will freeze the game forever.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -114,6 +114,8 @@ end
 	Desc: Unlike merge this adds the two tables together and discards keys.
 -----------------------------------------------------------]]
 function table.Add( dest, source )
+	-- The tables should be different otherwise this will just freeze the whole game
+	if ( dest == source ) then return dest end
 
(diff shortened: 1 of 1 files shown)


Aws0mee:
I'm not sure what specific case you would want this for, but I think it would be best to just implement it in your code.
For example:

```lua
if table1 ~= table 2 then
    table.Add(table1, table2)
end
```

I don't think it makes sense to add this check to the function since this won't be a problem in most circumstances.

Kefta:
The function should have protection from stalling the game no matter what arguments are passed. It's a library function. I would even support an error being thrown so developers can catch their logic mistake more easily.

garryspins:
Its a free check, +1

robotboy655 (approved):

robotboy655:
I couldn't reproduce an infinite loop, but it does hang the game for a bit.

zohart75:
Idk why, but my srcds just stopped responding lol

zohart75:
UPD: Lua throws an error when trying to table.insert for more than max table key limit.

PR #2056, opened 2024-02-21
Add util.In( any thing, vararg values )

bydlocodd:
Allows you to check if `thing` is among `values`.
Using the function, we can do this:

```lua
local myUsergroup = 'user'
if util.In( curState, 'user', 'vip' ) then
    -- ...
end
```

... instead of:
```lua
local allowedUsergroups = {
    user = true,
    vip = true,
}

local myUsergroup = 'user'
if allowedUsergroups[ myUsergroup ] then
    -- ...
end
```

Of course, the second code is faster because it performs a key lookup (which is O(1) vs. my O(n) implementation), but `util.In` is useful in contexts where performance is not so important

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -408,3 +408,15 @@ function util.IsBinaryModuleInstalled( name )
 
 	return false
 end
+
+function util.In( thing, ... )
+    for i = 1, select( '#', ... ) do
+        local cur = select( i, ... )
+
+        if thing == cur then
+            return true
+        end
+    end
+
+    return false
+end


robotboy655:
Bad naming aside, 
```
local myUsergroup = 'user'
if table.HasValue( { 'user', 'vip' }, myUsergroup ) then
    -- ...
end
```

I understand this is less efficient than the proposed method, but by your own admission that is not the focus here.

Kefta:
Not creating the table would indeed be much more efficient when JIT compiles the select calls. Not as efficient as creating a hashtable with true vals, but for dynamic vararg values, this seems like the best solution.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -408,3 +408,15 @@ function util.IsBinaryModuleInstalled( name )
 
 	return false
 end
+
+function util.In( thing, ... )
+    for i = 1, select( '#', ... ) do
```
brandonsturgeon, line 413: nit: I think it was Spar who demonstrated that localizing your for limit makes it marginally faster.

Something like:
```suggestion
    local limit = select( "#", ... )

    for i = 1, limit do
```

_might_ be a nice microoptimization

bydlocodd:
Not sure how I can improve on my solution ( except for brandonsturgeon's suggestion). Of course, pretty much the same thing can be done using `table.HasValue` (honestly, I just forgot this function existed), but my method is a bit faster.

garrysmod/lua/includes/extensions/util.lua
Kefta: I don't see how this is an improvement since the loop limit is only evaluated once.
image

bydlocodd:
> Bad naming aside,
> 
> ```
> local myUsergroup = 'user'
> if table.HasValue( { 'user', 'vip' }, myUsergroup ) then
>     -- ...
> end
> ```
> 
> I understand this is less efficient than the proposed method, but by your own admission that is not the focus here.

Yes, performance is not a key factor, because in that case I will use a key lookup. However, the developed solution, although similar to `table.HasValue`, is slightly more optimized. In this case, I see my function as a compromise between readability and speed - we sacrifice speed for readability, but not as much as in the case of using `table.HasValue`

garrysmod/lua/includes/extensions/util.lua
Grocel: > I don't see how this is an improvement since the loop limit is only evaluated once.

Does it be have like that in GLua too?

Kefta: Yes

TankNut:
With regards to the name, `util.InValues`?

PR #2058, opened 2024-02-26
Prevent infinite loop from `panel:SetDragParent`

Timmy-the-nobody:
Calling `panel:SetDragParent` with the panel itself as `parent` argument freezes the game

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel/dragdrop.lua
+++ b/garrysmod/lua/includes/extensions/client/panel/dragdrop.lua
@@ -329,7 +329,6 @@ end
 -- dragging the defined parent
 --
 function meta:SetDragParent( parent )
-	if ( parent == self ) then return end
 	self.m_pDragParent = parent
 end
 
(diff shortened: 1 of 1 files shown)


robotboy655:
I think it's best to add a check in `meta:DragMousePress( mcode )` instead, in case some megamind directly sets the internal variable `m_pDragParent`.

https://github.com/Facepunch/garrysmod/blob/0d8135682e7353d75688c71f57a3cb170d2f3a40/garrysmod/lua/includes/extensions/client/panel/dragdrop.lua#L418

robotboy655 (changes requested):
I think it's best to add a check in `meta:DragMousePress( mcode )` instead, in case some megamind directly sets the internal variable `m_pDragParent`.

https://github.com/Facepunch/garrysmod/blob/0d8135682e7353d75688c71f57a3cb170d2f3a40/garrysmod/lua/includes/extensions/client/panel/dragdrop.lua#L418

Timmy-the-nobody:
That's right, I just made the change

robotboy655:
I was thinking more in the way of 
```
if ( IsValid( self.m_pDragParent ) && self.m_pDragParent != self ) then
```

Which would make most sense and not prevent code below the IF block from running.

Timmy-the-nobody:
> I was thinking more in the way of
> 
> ```
> if ( IsValid( self.m_pDragParent ) && self.m_pDragParent != self ) then
> ```
> 
> Which would make most sense and not prevent code below the IF block from running.

Done 🙌

PR #2060, opened 2024-02-28
TTT: Replace self.Weapon/self.Entity with self

figardo:


robotboy655 (approved):
Looks safe to merge at a quick glance.

PR #2062, opened 2024-02-29
TTT: Use player.Iterator

figardo:
I replaced instances of ipairs(player.GetAll()) within the gamemode on my server a while back and haven't run into any issues with it since, so there shouldn't be anything that breaks with these changes (hopefully).

Left out 3 instances of player.GetAll where it probably wouldn't be worth replacing them. Let me know if they should be replaced anyway (or on the contrary, if the radar change also isn't worth it).

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
@@ -97,7 +97,7 @@ local function DrawPropSpecLabels(client)
    local scrpos = nil
    local text = nil
    local w = 0
-   for _, ply in ipairs(player.GetAll()) do
+   for _, ply in player.Iterator() do
       if ply:IsSpec() then
          surface.SetTextColor(220,200,0,120)
 
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
@@ -34,9 +34,7 @@ end
 local indicator_mat = Material("vgui/ttt/sprite_traitor")
 local indicator_col = Color(255, 255, 255, 130)
 
-local client, inext, plys, pos, dir, tgt
-local num = 0
-local GetPlayers = player.Iterator
+local client, pos, dir, tgt
 
 local propspec_outline = Material("models/props_combine/portalball001_sheet")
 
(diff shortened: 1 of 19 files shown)


garrysmod/gamemodes/terrortown/gamemode/cl_targetid.lua
```diff
@@ -34,25 +34,25 @@ end
...
+local GetPlayers = player.Iterator
 
 local propspec_outline = Material("models/props_combine/portalball001_sheet")
 
 -- using this hook instead of pre/postplayerdraw because playerdraw seems to
 -- happen before certain entities are drawn, which then clip over the sprite
 function GM:PostDrawTranslucentRenderables()
    client = LocalPlayer()
-   plys = GetPlayers()
+   inext, plys = GetPlayers()
```
robotboy655, line 47: This looks like a mess. IMO it should either be left as is, or the loops be converted to use `player.Iterator`. I can't imagine caching the 3 returns is going to save a lot of performance, the most important performance savings are already done by `player.Iterator`, by caching the player table.

figardo: Does this apply to the ones in [init.lua](https://github.com/Facepunch/garrysmod/pull/2062/files#diff-7a8b22d2b219d4f4dd9b2d78f1ef16e700592dc75bd8d9ecf42353db31ab8758R530) and [traitor_state.lua](https://github.com/Facepunch/garrysmod/pull/2062/files#diff-5da3e5f509939356952f4e4a75a4f1def4798c2439bc831223d9b446e8fff24eR95) as well?

robotboy655: Yeah

garrysmod/gamemodes/terrortown/gamemode/radar.lua
```diff
@@ -17,7 +17,11 @@ local function RadarScan(ply, cmd, args)
 
          ply.radar_charge =  CurTime() + chargetime
 
-         local scan_ents = player.GetAll()
+         local scan_ents = {}
+         for i, pl in player.Iterator() do
+            scan_ents[i] = pl
+         end
```
svdm, line 23: If this was faster than calling `player.GetAll`, then wouldn't `player.GetAll` just be reimplemented this way?

Regardless, this function does not get called very frequently so the simpler implementation of just calling `player.GetAll()` is preferable here.

Kefta: player.GetAll returns a table, not an iterator, so it can't be reimplemented.

PR #2064, opened 2024-03-02
TTT: Reduce radar ping bit count

figardo:
I know there's been some talk about increasing the map size limits, but to be perfectly honest with you 4 billion units cubed seems a tad bit excessive.

[This used to use umsg.Short](https://github.com/Facepunch/garrysmod/blob/329e7979ea0cbb9dfbf9e89f27510baba791fd12/garrysmod/gamemodes/terrortown/gamemode/radar.lua#L60), but another 16 bits were added to each axis along with the conversion to the net library for some reason.

Testing with 31 other scan targets, the message length goes from 3046 bits to 1465 bits.

Using WriteVector creates a 1744 bit message, and the coordinates are rounded to integers anyway so there's no need for the extra precision.

Thanks @mgetJane for pointing it out.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_radar.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_radar.lua
@@ -240,9 +240,9 @@ local function ReceiveRadarScan()
       local r = net.ReadUInt(2)
 
       local pos = Vector()
-      pos.x = net.ReadInt(32)
-      pos.y = net.ReadInt(32)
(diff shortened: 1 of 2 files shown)


mgetJane:
> I know there's been some talk about increasing the map size limits, but to be perfectly honest with you 4 billion units cubed seems a tad bit excessive.

btw this pr would actually make it potentially easier for such an increase, since net.WriteVector and net.ReadVector would need to be changed
(this is hypothetical of course)

PR #2065, opened 2024-03-02
Allow custom error message for CanArmDupe

brandonsturgeon:
Adds a second return for the `CanArmDupe` hook that, if provided, will override the default failure message.


robotboy655 (changes requested):
This new behavior should also be mimicked in `arming.lua`.

brandonsturgeon:
Thanks! Done ✔️

robotboy655 (changes requested):
You still missed one instance of the hook call here:
https://github.com/Facepunch/garrysmod/blob/a3453cb1e3e9090d8c77b8c739ae5c54766158f4/garrysmod/gamemodes/sandbox/entities/weapons/gmod_tool/stools/duplicator/arming.lua#L66-L67

brandonsturgeon:
Oops! I added it there too ✔️

PR #2067, opened 2024-03-03
Make toolgun reload also use util.TraceHull, to make hit detection consistent with primary/secondary

NO-LOAFING:
In the current dev branch, the toolgun's primary and secondary fire use util.TraceHull, while reload uses util.TraceLine. This makes hit detection inconsistent when using the toolgun on ents like NPCs - if you aim somewhere like the space between an NPC's head and shoulder, or between an NPC's legs, the primary and secondary fire will still hit, but the reload will not. This is because TraceHull traces against the NPC's entire collision hull, while TraceLine traces against the NPC's hitboxes instead.

This pull request changes gmod_tool's Reload function to also use TraceHull, to fix this problem. It looks like the original TraceHull pull request also did this (https://github.com/Facepunch/garrysmod/pull/1890/commits/01e22b972833f1b7e7f64e1c6362fab7921479a9#diff-ce3c1fb60d7d926168d4c8160817ec2bd6b843cf3fed5398b54d6f995a37ba9aR258), but then a later commit removed it for some reason? (https://github.com/Facepunch/garrysmod/commit/24eaca30e0cab3f0c003c65e22b953ead3497924#diff-ce3c1fb60d7d926168d4c8160817ec2bd6b843cf3fed5398b54d6f995a37ba9aR272) I couldn't tell if this was on purpose, since it was part of a huge "Minor code adjustments" commit that didn't mention it or have any other changes like this, but if there was a good reason for it, then feel free to disregard this.

https://github.com/Facepunch/garrysmod/assets/12566919/194a4e5a-2121-47c3-aaca-6399a7c88c73


NO-LOAFING:
Here's a video of the issue being fixed:
https://github.com/Facepunch/garrysmod/assets/12566919/ea554482-9479-4fe7-a6b3-17640f85ce70

robotboy655 (approved):

robotboy655:
Looks like it was undone by mistake.

PR #2069, opened 2024-03-10
Add killicon overriding logic

StrawWagen:
Allow for the overriding of killicons.
Fluff change, for adding custom killicons, without creating/spawning a custom inflictor.

Adds
entity:Set/GetKilliconOverride( string )
entity:SetDeathKillicon( string )

+Gamemode hook,
GM:GetDeathNoticeInflictorClass( killed, inflictor )

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/npc.lua
+++ b/garrysmod/gamemodes/base/gamemode/npc.lua
@@ -87,11 +87,11 @@ AccessorFunc( entMeta, "m_strKillIcon", "KilliconOverride", FORCE_STRING )
 -----------------------------------------------------------]]
 function entMeta:SetDeathKillicon( classOverride )
 	if ( !classOverride ) then return end
+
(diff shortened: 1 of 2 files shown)


EntranceJew:
> I am not a fan of this implementation as opposed to sending the killicon index with the dmginfo.

inclined to agree here, an entity deciding what its kill icon is or needing it set / a handler assigned prior to their death seems like it'd really be wack for replication/demos/etc

however, the server has no concept of killicon so that would be even more complicated to utilize
image
(does the client know about this icon? can they be iterated so something could pragmatically decide it?)

Kefta:
I mentioned this in the gmod discord, but the best solution if this idea is necessary at all would be to pool killicon names in a shared network table so that the dmginfo struct could include an extra uint to decide what the killicon would be on the client.

StrawWagen:
All of these sound like good ideas to me, this is as far i can take the concept though, not skilled enough in this area haha

PR #2073, opened 2024-04-05
Use ents/player.Iterator instead of GetAll

aStonePenguin:
Self explanatory.

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_init.lua
@@ -196,7 +196,7 @@ function GM:OnChatTab( str )
 
 	if ( LastWord == nil ) then return str end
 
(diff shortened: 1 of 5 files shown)


robotboy655:
This depends on https://github.com/Facepunch/garrysmod-issues/issues/5800 getting solved first.

mgetJane:
> This depends on [Facepunch/garrysmod-issues#5800](https://github.com/Facepunch/garrysmod-issues/issues/5800) getting solved first.

player.Iterator should be unaffected by this

@aStonedPenguin imo you should make a new pr only for the player.Iterator() changes

storm37000:
Tested player.GetBySteamID and player.GetBySteamID64 in singleplayer and both are 200-300% faster.  Both had similar gains so im assuming the other functions that have the player.Iterator replacement will be similar.

PR #2076, opened 2024-04-20
TTT: Replace GetConVarNumber/String

figardo:


garrysmod/gamemodes/terrortown/gamemode/shared.lua
```diff
@@ -139,9 +139,9 @@ local ttt_playercolors = {
    }
 };
 
-CreateConVar("ttt_playercolor_mode", "1")
+local playercolor_mode = CreateConVar("ttt_playercolor_mode", "1")
 function GM:TTTPlayerColor(model)
-   local mode = GetConVarNumber("ttt_playercolor_mode") or 0
+   local mode = playercolor_mode:GetInt() or 0
```
Malivil, line 144: I don't think the `or 0` is needed anymore since `GetInt` already returns `0` on failure

garrysmod/gamemodes/terrortown/gamemode/player_ext.lua
```diff
@@ -52,13 +52,13 @@ function plymeta:SubtractCredits(amt) self:AddCredits(-amt) end
 
 function plymeta:SetDefaultCredits()
    if self:GetTraitor() then
-      local c = GetConVarNumber("ttt_credits_starting")
+      local c = GetConVar("ttt_credits_starting"):GetInt()
```
robotboy655, line 55: FYI, this does not achieve anything. This is exactly what `GetConVarNumber` does behind the scenes.

figardo: Just means there's less deprecated shit for glualint to blind me with is all.
Also, I wanted to cache the result of GetConVar outside of each function, but it caused errors in the area you've highlighted so I left it as is in case it caused anything else to break.

PR #2081, opened 2024-04-28
Synchronize width of the gamemodes tab and the serverlist for visual consistency

rubiksw:
This PR makes it so that the width of the gamemodes tab and the serverlist are the same for more visual consistency. 

ezgif-5-0f9f7cfb4b

### DIFF
--- a/garrysmod/resource/serverevents.res
+++ b/garrysmod/resource/serverevents.res
@@ -113,6 +113,7 @@
 	{
 		"userid"	"short"		// user ID on server
 		"text"		"string"	// the say text
+		"teamonly"	"bool"		// is it teamonly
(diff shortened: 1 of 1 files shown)


robotboy655:
The reason they are different is because the gamemodes list doesn't benefit from being stretched (it will just be a bunch of wasted space), but the server list has a few columns with varying text lengths, so the width has to be wider. This is especially an issue on lower resolutions.

RubikSW:
This change really only effects ultra-wide resolutions where you get some funky behavior with the gamemodes title bar not lining up with the body. See below:

image

On more standard lower resolutions and it remains unchanged.

image

PR #2083, opened 2024-05-03
Don't draw effect rings w/gravgun equipped.

StrawWagen:
I get why this was a feature, since the gravgun _can_ pick up, move effects. 

But I'm gonna be straight, nobody places, moves effects with the gravgun, it's a weapon you use to "play the game".
If I'm wrong, and someone does use the gravgun to place effects, I feel very confident saying they're vastly outnumbered by the players who use it solely as a weapon.

"cl_draweffectrings 0" does exist, but it's not the "default player experience", it's another, albeit very small, "barrier to entry".

Thanks for considering this very small PR, for what I feel is an equally small oversight.

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/prop_effect.lua
+++ b/garrysmod/gamemodes/base/entities/entities/prop_effect.lua
@@ -79,7 +79,7 @@ function ENT:Draw()
 
 	local weapon_name = wep:GetClass()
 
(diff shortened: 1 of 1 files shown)


StrawWagen:
Yeah, but my opinion is there's more people fighting with the gravity gun, than people using it to interact with effects.

This change is really just for the small save/course building community.
It sucks to make an elaborate course with cool details everywhere, only for a bunch of yellow circles to show up when you give players the best half life 2 weapon haha.

robotboy655 (approved):

PR #2084, opened 2024-05-04
Fix DMenuOptionCVar tripping over itself

Tenrys:
RunConsoleCommand does set the CVar a frame later, but the panel's Think runs right after OnChecked, and by that point the CVar hasn't been updated yet, so it sets itself back to its previous value.

This works but I feel like it's a bit too dirty?

Also remove deprecated use of GetConVarString

### DIFF
--- a/garrysmod/lua/vgui/dmenuoptioncvar.lua
+++ b/garrysmod/lua/vgui/dmenuoptioncvar.lua
@@ -3,6 +3,7 @@ local PANEL = {}
 
 DEFINE_BASECLASS( "DMenuOption" )
 
+AccessorFunc( PANEL, "m_strConVar", "ConVar" )
 AccessorFunc( PANEL, "m_strValueOn", "ValueOn" )
 AccessorFunc( PANEL, "m_strValueOff", "ValueOff" )
 
@@ -20,12 +21,6 @@ function PANEL:SetConVar( strConVar )
 
 	self.m_strConVar = strConVar
 	self.m_ConVar = GetConVar( strConVar )
-
-end
-
-function PANEL:GetConVar()
-
-	return self.m_strConVar
 
 end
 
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dmenuoptioncvar.lua
```diff
@@ -17,12 +16,34 @@ function PANEL:Init()
...
+	self.m_strConVar = strConVar
+	self.m_ConVar = GetConVar( strConVar )
+
+end
+
+function PANEL:GetConVar()
+
+	return self.m_strConVar
+
+end
```
robotboy655, line 30: This can definitely be removed in favor of restoring line 6. `AccessorFunc( PANEL, "m_strConVar", "ConVar" )`

Tenrys: While keeping the custom SetCVar right? so it gets overridden?

robotboy655: Ye

robotboy655 (changes requested):
Are you able to present a case where this issue actually happens?

I have tried adding some `print()`s when the convar value is being set and read, and I am not noticing any issues with the spawnmenu menubar setting the cvar, either by clicking the option to set cvar, or by setting the cvar from the console.

image

garrysmod/lua/vgui/dmenuoptioncvar.lua
```diff
@@ -17,12 +17,28 @@ function PANEL:Init()
...
 end
 
+function PANEL:SetConVar( strConVar )
+
+	self.m_strConVar = strConVar
+	self.m_ConVar = GetConVar( strConVar )
+
+end
+
+local justChecked = false
```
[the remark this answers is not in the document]
Tenrys: Right, just to be sure, would `PANEL.JustChecked` work out better then? I do feel like the solution I presented is a bit dirty, but I can't think of anything better on the spot.

PR #2086, opened 2024-05-20
util.IntervalTimer

T-Rizzle12:
Currently, we only have util.Timer which creates a countdown timer. What this pull request adds is the Interval Timer as found in the Source 2013 SDK. This is useful for addons that want certain things to happen after a certain amount of time elapsed
Interval Timer according to [Source 2013 SDK](https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/shared/util_shared.h#L487):
This is a simple class for tracking intervals of game time.
Upon creation, the timer is invalidated.  To measure time intervals, start the timer via Start().


robotboy655:
I fail to see the necessity for a separate function and class for this purpose. 

This appears to be a simple timer which is normally done manually by just storing the start time in a variable, and then doing whatever math you need where you need it.

If you really absolutely must have an object-oriented timer like this, wouldn't it make more sense to just store the start time in the original timer object, and add a single new function to return the time since timer was started?

robotboy655 (changes requested):
I fail to see the necessity for a separate function and class for this purpose.

This appears to be a simple timer which is normally done manually by just storing the start time in a variable, and then doing whatever math you need where you need it.

If you really absolutely must have an object-oriented timer like this, wouldn't it make more sense to just store the start time in the original timer object (util.Timer), and add a single new function to return the time since timer was started?

T-Rizzle12:
> I fail to see the necessity for a separate function and class for this purpose.
> 
> This appears to be a simple timer which is normally done manually by just storing the start time in a variable, and then doing whatever math you need where you need it.
> 
> If you really absolutely must have an object-oriented timer like this, wouldn't it make more sense to just store the start time in the original timer object (util.Timer), and add a single new function to return the time since timer was started?

Sure, I have nothing against merging this with the default util.Timer class. I will update the pull request to add the start time to util.Timer. Along with a few functions that will interact with it.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -195,7 +198,7 @@ local T =
 	--
 	Started = function( self )
 
-		return self.endtime != nil
+		return self.endtime != nil and self.m_timestamp > 0
```
robotboy655, line 201: I don't think this change is necessary at all?

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -207,6 +210,33 @@ local T =
...
+	IsLessThan = function( self, duration )
+
+		return CurTime() - self.m_timestamp < duration
+
+	end
+
+	--
+	-- Returns true if the amount of time that has elapsed is greater than the entered duration
+	--
+	IsGreaterThan = function( self, duration )
```
robotboy655, line 235: Why do you think these 2 are necessary?

Surely we don't need a binding to `GetElaspedTime() > 3` in the form of `IsGreaterThan( 3 )`?

robotboy655 (changes requested):

garrysmod/lua/includes/extensions/util.lua
T-Rizzle12: It's not necessary but helps with making the code look cleaner.
For example:
```
local timer = util.Timer()
timer:Start()

hook.Add( "Think", "TimerExample", function()

    if timer:IsGreaterThan( 3.0 ) then
    
        local ply = Entity(1)
        if IsValid( ply ) then

            ply:Kill()
            hook.Remove( "Think", "TimerExample" )
        
        end

    end

end)
```

Compared to:
```
local timer = util.Timer()
timer:Start()

hook.Add( "Think", "TimerExample", function()

    if timer:GetElaspedTime() > 3 then
    
        local ply = Entity(1)
        if IsValid( ply ) then

            ply:Kill()
            hook.Remove( "Think", "TimerExample" )
        
        end

    end

end)
```
This might not be the strongest example, but if you want the two functions removed, I will remove them.

garrysmod/lua/includes/extensions/util.lua
T-Rizzle12: Ok, I will undo that change.

garrysmod/lua/includes/extensions/util.lua
robotboy655: Yeah, I don't think this warrants adding 2 extra function aliases for a built-in operator.

I think you should remove them from this PR.

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -220,12 +232,12 @@ function util.Timer( startdelay )
 
 	local t = {}
 	setmetatable( t, T )
+	t.m_timestamp = -1.0
```
robotboy655, line 235: Since the timer starts immediately when the function is called, this should be set to `CurTime()`. Perhaps even simply call `Start( startdelay )` here

garrysmod/lua/includes/extensions/util.lua
```diff
@@ -206,6 +209,15 @@ local T =
...
 		return self.endtime == nil or self.endtime <= CurTime()
 
+	end,
+
+	--
+	-- Returns the amount of time that has passed since the Timer was started
+	--
+	GetElaspedTime = function( self )
+
+		return self:Started() and CurTime() - self.m_timestamp or 99999.9
```
robotboy655, line 219: I am not sure it makes sense to return 99999.9 for a stopped timer. Perhaps `-1` will be a more logical return being an "invalid time" rather than returning a large value that may or may not be the real elapsed time.

robotboy655 (changes requested):

garrysmod/lua/includes/extensions/util.lua
robotboy655: Come to think of it, shouldn't it just return the elapsed time at the moment the timer was stopped anyway?

T-Rizzle12: I'm basing this PR on the IntervalTimer object in the Source 2013 SDK. The GetElaspedTime in there internally returns 99999.9. I don't know why Valve did that, but I did change it to -1.0 as you suggested. When the timer is reset, Valve sets m_timestamp to -1.0. If you want to make any other changes, you should have access to edit the PR.

garrysmod/lua/includes/extensions/util.lua
T-Rizzle12: Changed the util.Timer function to call the Start function upon creation.

robotboy655 (approved):
I have changed `GetElaspedTime` to work like expected after timer is stopped. If you see no issues and have no objections, I will merge this.

T-Rizzle12 (commented):
Looks good, feel free to merge when ready!

PR #2091, opened 2024-07-02
Use strict check for HL2/EP1/EP2 maps

TetraTheta:
GMod categorizes maps with their name prefix, like `d1_`, `d2_`, `ep1_`, etc.

Since GMod doesn't allow addons to override this behavior, any map that has these prefixes will always be categorized as 'Half Life 2', 'Half Life 2: Episode 1', etc., not the addon itself.

```plaintext
"my_addon"
{
  "title" "My HL2 Mod Port"
  "maps" "d1_trainstation_reversed_01|d1_trainstation_reversed_02"
  "menusystem" "0"
}
```
In this example gamemode txt, the maps `d1_trainstation_reversed_01` and `d1_trainstation_reversed_02` will be categorized as 'Half-Life 2' instead of 'My HL2 Mod Port' because they have the `d1_` prefix. You can't change the map's name because that will break the map transition.

The only way to work around this is to edit `getmaps.lua` manually, because `getmaps.lua` will take priority over addon's gamemode txt.

So I'm suggesting setting strict rules for categorizing HL2/EP1/EP2 maps. With this, only maps from the actual HL2/EP1/EP2 will be categorized as HL2/EP1/EP2.

---

**PROS**

- HL2 mod port addons can now set their map's category safely without worrying about them being miscategorized as HL2/EP1/EP2.

**CONS**

- `getmaps.lua`'s length has been increased. This may increase the loading time of the 'New Game' screen.


robotboy655:
I have applied this in a different way. Also this PR doesn't really work, because patterns are not accepted in `MapNames` table.

TetraTheta:
Oops, sorry. I should've test it before creating PR first. I completely forgot about that...


Your implementation looks far more superior than mine! Thank you!

PR #2092, opened 2024-07-03
Implemented hook.Exists function into the hook module

The-Lord-of-Owls:
Summary:
Implemented a simple hook.Exists() function to check if a hook exists or not. I've verified that this function works as intended and that the code is consistent with what is already in the hook module. Args are validated just the same as with other functions already in the module.

Why this was implemented:
This would be a much cleaner and straightforward way to check if a hook exists instead of having to do a work around such as hook.GetTable()[ event_name ][ name ]

Name: Exists
Args: string event_name, string name
Desc: Returns if the hook exists or not

Usage:
```lua
//Check if hook doesn't exist yet
if ( not hook.Exists( "InitPostEntity", "test" ) ) then
	//Add the hook if it doesn't exist
	hook.Add( "InitPostEntity", "test", function()
		print( "howdy!" )
	end )
end

//Check if the hook exists now
if ( hook.Exists( "InitPostEntity", "test" ) ) then
	print( "hook exists" )	//This will print because the hook exists now
end
```

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -62,6 +62,25 @@ end
 
 
 --[[---------------------------------------------------------
+    Name: Exists
+    Args: string hookName, identifier
+    Desc: Returns if the hook exists or not
+-----------------------------------------------------------]]
+function Exists( event_name, name )
+
+	if ( !isstring( event_name ) ) then ErrorNoHaltWithStack( "bad argument #1 to 'Exists' (string expected, got " .. type( event_name ) .. ")" ) return end
+
+	local notValid = isnumber( name ) or isbool( name ) or isfunction( name ) or !name.IsValid or !IsValid( name )
+	if ( !isstring( name ) and notValid ) then ErrorNoHaltWithStack( "bad argument #2 to 'Exists' (string expected, got " .. type( name ) .. ")" ) return end
+
+	if ( Hooks[ event_name ] and Hooks[ event_name ][ name ] ) then return true end	--Return true if the hook exists
+
+	return false	--Otherwise return false
+
+end
+
+
+--[[---------------------------------------------------------
     Name: Run
     Args: string hookName, vararg args
     Desc: Calls hooks associated with the hook name.


Aws0mee:
I don't understand how
```lua
if ( hook.Exists( "InitPostEntity", "test" ) ) then
	print( "hook exists" )	//This will print because the hook exists now
end
```
is better than
```lua
if hook.GetTable().InitPostEntity.test then
	print( "hook exists" )	//This will print because the hook exists now
end
```

It might be a little cleaner, but all of those validation checks will likely make it a little slower too. I don't think this addition is 100% needed, but I guess it wouldn't really hurt to add.

Zaurzo:
> I don't understand how
> 
> ```lua
> if ( hook.Exists( "InitPostEntity", "test" ) ) then
> 	print( "hook exists" )	//This will print because the hook exists now
> end
> ```
> 
> is better than
> 
> ```lua
> if hook.GetTable().InitPostEntity.test then
> 	print( "hook exists" )	//This will print because the hook exists now
> end
> ```
> 
> It might be a little cleaner, but all of those validation checks will likely make it a little slower too. I don't think this addition is 100% needed, but I guess it wouldn't really hurt to add.

That will error if no InitPostEntity hooks were added

garryspins:
Whats one use for this outside of debugging

Aws0mee:
> > I don't understand how
> > ```lua
> > if ( hook.Exists( "InitPostEntity", "test" ) ) then
> > 	print( "hook exists" )	//This will print because the hook exists now
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is better than
> > ```lua
> > if hook.GetTable().InitPostEntity.test then
> > 	print( "hook exists" )	//This will print because the hook exists now
> > end
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It might be a little cleaner, but all of those validation checks will likely make it a little slower too. I don't think this addition is 100% needed, but I guess it wouldn't really hurt to add.
> 
> That will error if no InitPostEntity hooks were added

Pretty sure every server will have at least 1 InitPostEntity hook, but you can just check for that too if you want.

```lua
local hooks = hook.GetTable()
if hooks.InitPostEntity and hooks.InitPostEntity.test then
	print( "hook exists" )	//This will print because the hook exists now
end
```

The-Lord-of-Owls:
The use case I've primarily been using this for is when dealing with temporary hooks that are created on an as needed basis that may depend on other hooks existing alongside them, or may need to skip parts of their logic while one of these temporary hooks are in use. This has been really beneficial to dealing with hooks that run very often such as HUDPaint or Think/Tick. Namely with server wide events where something temporarily changes but immediately reverts back the moment one of these temporary hooks is removed.

It becomes especially beneficial when having to check if multiple hooks exist at a time

The-Lord-of-Owls:
UPDATE TO PULL REQUEST:
I have also added my hook.Temporary() function to hook module. Perhaps also providing that for others may help see some reasons for the other function I added to exist aswell.

PURPOSE:
With the hook.Temporary() function you can now specify a predetermined amount of times the hook should run before removing itself. when the hook runs it imposes very minimal overhead as it is only increasing a local variable by 1, and performing a >= check on two variables to determine when the hook should be removed

NOTE: If only event_name, name and max_runs are provided, it will assume that the max_runs is the hook function and that this hook should only be ran once before removing itself. This makes it rather simple to to write onetime use hooks or limited use hooks depending on needs of the addon developer using them.


EXAMPLE: This example below demonstrates the use of both addons as a way of determining how many rounds a gamemode should run a match for before doing a map change for the next game. I'm sure these two functions together will see alot of use. I've found temporary hooks to be very useful in my projects and I think having these two functions added to the game would help make alot of things easier for addon creators by having the logic for it behind a simple built in function to do it.

```lua
--This hook will run 25 times before removing itself
--It will return the winner of the round to where it can be displayed for all players
--If the hook has removed itself then our code knows that it is time to do a map transition to the next map in the list
Temporary( "EndRound", "CheckWinner", 25, function( redTeamScore, blueTeamScore )
	if redTeamScore > blueTeamScore then
		return redTeamScore
	elseif redTeamScore < blueTeamScore then
		return blueTeamScore
	else
		return redTeamScore == blueTeamScore
	end
end )

--Determine if our gamemode should run another round
hook.Add( "ShouldStartRound", "CheckRounds", function()
	return hook.Exists( "EndRound", "CheckWinner" ) or false
end)

local nextMatchMap = "gm_flatgrass_soccer"
function StartRound()
	local shouldStart = hook.Call( "ShouldStartRound" )

	if shouldStart then
		hook.Call( "StartRound" )
	else
		RunConsoleCommand( nextMatchMap )
	end
end
```

EXAMPLE 2: The hook.Temporary() function is also useful during server startup and handling things such as loading config files and spawning things in. This example shows how to make a one time running hook on InitPostEntity

```lua
hook.Temporary( "InitPostEntity", "SpawnEntities", function()
	--Some code that spawns entities at server startup
	--Maybe also load a config file from the data folder
	--Perhaps even perform an sql query to a database on startup
	--What ever this hook does, will run once and then remove itself
end )
```

The-Lord-of-Owls:
In regards to what you said about the isfunction validation happening multiple times. I've made a slight change and moved the same logic from hook.Add and hook.Remove into the hook.Temporary function. So now validation on event_name, name, max_runs and func should only occur once.

So the overhead when adding the hook and the overhead when removing it should no longer be an issue.

garryspins:
I actually really like hook.Temporary

StarLight-Oliver:
Another issue with this, is that the second argument to hook.Add isn't a string, its any. so your hook.Exists doesn't work for all possible hooks

The-Lord-of-Owls:
> Another issue with this, is that the second argument to hook.Add isn't a string, its any. so your hook.Exists doesn't work for all possible hooks

The wiki says that, tho when looking at the actual hook.lua file it doesn't seem to be the case unless I'm misreading this. It looks like if the id of the hook is anything other than a string it would call ErrorNoHaltWithStack on line 34. So I don't see why checking for a string would be an issue if that the only one it sees as a valid type to add to the hook table.

It would probably be useful to have it instead convert those to strings rather than throwing an error. It might also be useful to return the string if that happens too just so the dev is aware of it happening but I've not thought of that till now so I'm unsure what unforeseen issues or overhead that would cause. But given its just for hook.Add it probably wouldn't cause too much issues or break things.

I'll play around with the idea of converting to string from other types and returning the converted ID from hook.Add. I'm not sure if I should include that in this pull request or not however. I'll let somebody else tell me first if I should or make a separate pull

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/modules/hook.lua#L23-L42

StarLight-Oliver:
if its a table or userdata with isvalid true, then it will pass it into the function. useful for adding lifecycle too hooks etc

Denneisk:
> hook.Temporary()

We reinventing timers again with this one

The-Lord-of-Owls:
> > hook.Temporary()
> 
> We reinventing timers again with this one

Not quite, timers would still be delay based when it comes to removing it. hook.Temporary would be based on how many times the hook event was called which could vary between calls

PR #2095, opened 2024-07-09
Improve util.StringToType

brandonsturgeon:
This PR aims to solve these inconsistencies/issues:

- `type( true )` returns `"boolean"` but `StringToType` takes `"bool"`
- `type( 5 )` returns `"number"` but `StringToType` takes `"int"`
- `util.StringToType` does not handle failed `tonumber` conversions and would error on `math.Round`
  ```
  > print( util.StringToType( "blah", "int" ) )...
  L 07/09/2024 - 16:36:33: Lua Error:
  [ERROR] lua/includes/extensions/math.lua:187: attempt to perform arithmetic on local 'num' (a nil value)
  1. StringToType - lua/includes/extensions/math.lua:187
  2. unknown - lua_run:1
  ```

This PR aims to make it more consistent.

I don't like how I did the formatting but I can't decide on a better way. Suggestions appreciated 👍 

Thanks to superjared42 on Discord

### DIFF
--- a/garrysmod/lua/includes/extensions/util.lua
+++ b/garrysmod/lua/includes/extensions/util.lua
@@ -119,7 +119,8 @@ function util.StringToType( str, typename )
 	if ( typename == "string" )	then return tostring( str ) end
 	if ( typename == "entity" )	then return Entity( str ) end
 	if ( typename == "float" )	then return tonumber( str ) end
-	if ( typename == "int" )	then
+	if ( typename == "int" )	or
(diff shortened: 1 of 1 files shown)


robotboy655:
https://github.com/Facepunch/garrysmod/commit/b056fc812c480072b66a8a103eb1b6d8312cedeb

brandonsturgeon:
Handling the `tonumber` failure would be a cherry on top, but I'll let you decide if/how to handle that 👍

PR #2098, opened 2024-07-31
Add Texture Flag constants

A1steaksa:
Adds constants for Texture Flags with the prefix `TEXTUREFLAGS_` to make it eaasier to work with Render Targets and to make [the wiki page that already exists for them](https://wiki.facepunch.com/gmod/Enums/TEXTUREFLAGS) less confusing.


robotboy655:
This should not be done in Lua for non-Lua enums. I am also weary of spam-adding new globals to the global table due to performance concerns.

PR #2100, opened 2024-08-12
Allow return true in ENT:PreEntityCopy() to prevent table.Merge in Duplicator Copy

SpaxscE:
This will allow ENT:PreEntityCopy() to prevent the table.Merge in Duplicator

Useful for more complex entities that rely on a specific way of spawning or have alot of cached things that break when duped/made persistant

Fixes issue in one of my addons here https://github.com/SpaxscE/lvs_base/issues/72


SpaxscE:
ENT:OnEntityCopyTableFinish( data ) exists, isnt needed

PR #2106, opened 2024-08-25
Add SAM support to karma banning

wgaylord:
Add support for the SAM admin addon into the karma ban system of TTT 

https://www.gmodstore.com/market/view/sam


robotboy655:
I feel like a better solution here would be something like adding a "GM:TTTBanPlayer" hook for admin mods to implement themselves (with the hook falling back to GMod's [Player:Ban](https://wiki.facepunch.com/gmod/Player:Ban)), instead of constantly adding special cases for each possible admin mod.

Or perhaps just call [Player:Ban](https://wiki.facepunch.com/gmod/Player:Ban), since wouldn't those admin mods overwrite it anyway?

PR #2107, opened 2024-08-28
Improvement skypaint matproxy

friztailo:
## sky_paint
`Entity:GetNetworkVars()` replaced with `Entity:GetDT...()` functions.
Does not create unnecessary tables and does not perform unnecessary iterations over the table of network values in each frame.
```
benchmark (SysTime):
Entity:GetDT...()	1.6824031256988e-06	1
Entity:Get...()	4.7368874997673e-06	2.8155484422319
Entity:GetNetworkVars()	3.2047187503892e-06	1.9048459322483
```

### DIFF
--- a/garrysmod/lua/matproxy/sky_paint.lua
+++ b/garrysmod/lua/matproxy/sky_paint.lua
@@ -8,35 +8,32 @@ matproxy.Add( {
 	bind = function( self, mat, ent )
 
 		local skyPaint = g_SkyPaint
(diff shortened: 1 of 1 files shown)


friztailo:
Maybe we should stop creating the table every frame?

robotboy655 (approved):
I don't like how it goes back to magic numbers, but given this is a performance intensive area, I will allow it.

PR #2112, opened 2024-09-08
TTT: Update Turkish

NovaDiablox:
I can't believe I missed this one.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/turkish.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/turkish.lua
@@ -213,7 +213,7 @@ L.sb_mia = "Eylem Eksik"
 L.sb_confirmed = "Onaylanmış Ölü"
 
 L.sb_ping = "Gecikme"
-L.sb_deaths = "Ölümler"
+L.sb_deaths = "Ölüm"
 L.sb_score = "Puan"
 L.sb_karma = "Karma"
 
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/terrortown/gamemode/lang/turkish.lua
```diff
@@ -213,7 +213,7 @@ L.sb_mia = "Eylem Eksik"
 L.sb_confirmed = "Onaylanmış Ölü"
 
 L.sb_ping = "Gecikme"
-L.sb_deaths = "Ölümler"
+L.sb_deaths = "Ölüm"
```
robotboy655, line 216: I don't think this change is correct? The English string for this is plural, not singular.

https://github.com/Facepunch/garrysmod/blob/33740f86eea326dbc7f84104e993f7c769c65329/garrysmod/gamemodes/terrortown/gamemode/lang/english.lua#L216

NovaDiablox: Its better this way. The change mimics the CS2 scoreboard and its generally used singular if its in a scoreboard.

NovaDiablox: image
Here is an example.

NovaDiablox:
indir
Another example.

robotboy655:
I am not sure sample size of 1 other game is a good argument, but I guess I don't speak Turkish, so you probably know better.

NovaDiablox:
> I am not sure sample size of 1 other game is a good argument, but I guess I don't speak Turkish, so you probably know better.

image
Here is an another example from CS:S. Thanks for merging.

PR #2113, opened 2024-09-09
Allow for properties to be removed

DeveloperGhost:


robotboy655:
I don't hate this, but what does this do that https://wiki.facepunch.com/gmod/GM:CanProperty doesn't?

DeveloperGhost:
> I don't hate this, but what does this do that https://wiki.facepunch.com/gmod/GM:CanProperty doesn't?

Well, the [CanProperty](https://wiki.facepunch.com/gmod/GM:CanProperty) hook could be used for never calling a property, but imo, if you want various native properties to be removed from a sandbox-derived gamemode, that's a waste of engine hook calling.

If you really want to remove some of the properties permanently, why should the property module treat them as if they exist?

robotboy655 (approved):

PR #2116, opened 2024-09-13
Optimised ``draw.RoundedBoxEx``

Jaffies:
Remakes ``draw.RoundedBoxEx`` to draw only 4 quads, instead of 7.
By bare logic, this decreases draw calls almost for 2 times, meaning that this will work faster 2 times per draw.

Remade logic uses UV stretching and ``CLAMP S/T`` texture flags, instead of filling insides with ``surface.DrawRect``


robotboy655:
Are you serious?

PR #2122, opened 2024-09-24
Localization Support For NPCs & Vehicles

NovaDiablox:
Adds localization support for in game NPCs & vehicles. Please provide feedback as I change the localization of NPCs.


Satton2:
@NovaDiablox Hi!

I don't really see the reason why NPC with proper nouns are commented (while I don't really see them to be actually proper noun instead of just noun).

Like, if some language requires the "Headcrab" part to be localized in "Poison Headcrab", you are limiting a regular "Headcrab" to be inlocalizable, which is not okay. It's up to translators and language localization rules to decide, instead of total limiting in the code for every language.

NovaDiablox:
> @NovaDiablox Hi!
> 
> I don't really see the reason why NPC with proper nouns are commented (while I don't really see them to be actually proper noun instead of just noun).
> 
> Like, if some language requires the "Headcrab" part to be localized in "Poison Headcrab", you are limiting a regular "Headcrab" to be inlocalizable, which is not okay. It's up to translators and language localization rules to decide, instead of total limiting in the code for every language.

Thanks for the feedback! I believe people or the fan base wouldn't like to change or localize "Headcrab" part due to everyone is calling it as it is. Also I commented without changing it as much as possible. In the end of the day it is still up to the player base. I think about adding every NPC since they will roll back to English anyways.

NovaDiablox:
Oh and just to make things clear the comments are temporary, it won't be in the final finished PR.

Satton2:
> Thanks for the feedback! I believe people or the fan base wouldn't like to change or localize "Headcrab" part due to everyone is calling it as it is. Also I commented without changing it as much as possible. In the end of the day it is still up to the player base. I think about adding every NPC since they will roll back to English anyways.

Just for your interest. Most that you commented before the last change are already translated in Russian localization of HL2. Currently, it's not about us (translators) to translate NPCs names from scratch, it's about moving existing translations from games into GMod (if, of course, games were translated into our languages).

For example, all (or at least most) of these names are already translated into Russian in HL2 (Headcrab, Zombine, Combine Gunship, etc. are already there), so as a Russian translator my job will be just to check what the current translation is and put it into GMod, unless, something is not translated, of course.

So, my initial point about was uncommenting and using these strings as they were supposed to.

> I hope the community likes the changes. I will add other NPCs when I'm free.

I hope we will eventually be able to localize these strings, and the weapon ones too. 😅

NovaDiablox:
> > Thanks for the feedback! I believe people or the fan base wouldn't like to change or localize "Headcrab" part due to everyone is calling it as it is. Also I commented without changing it as much as possible. In the end of the day it is still up to the player base. I think about adding every NPC since they will roll back to English anyways.
> 
> Just for your interest. Most that you commented before the last change are already translated in Russian localization of HL2. Currently, it's not about us (translators) to translate NPCs names from scratch, it's about moving existing translations from games into GMod (if, of course, games were translated into our languages).
> 
> For example, all (or at least most) of these names are already translated into Russian in HL2 (Headcrab, Zombine, Combine Gunship, etc. are already there), so as a Russian translator my job will be just to check what the current translation is and put it into GMod, unless, something is not translated, of course.
> 
> So, my initial point about was uncommenting and using these strings as they were supposed to.
> 
> > I hope the community likes the changes. I will add other NPCs when I'm free.
> 
> I hope we will eventually be able to localize these strings, and the weapon ones too. 😅

I agree with you.

NovaDiablox:
@Satton2 I just added localization support for vehicles. Only thing left is portal npcs I believe and I don't have it installed currently. Could you provide me Portal cores names in Garry's Mod spawnmenu?

NovaDiablox:
Thanks for providing me the names :)

PR #2126, opened 2024-09-30
Localization Support for key bindings.

IBRS-4Ever:
image
image


NovaDiablox:
Nice! I always wondered if the missing lines were localizable. Thank you IBRS.

robotboy655 (changes requested):
Usually newly added strings are appended to the end of the file, not at the top.

You will notice there are already GMod related strings there. You should also specify in the text file that the strings are for the keyboard settings specifically.

IBRS-4Ever:
You mean like this?
image

robotboy655 (approved):

PR #2128, opened 2024-10-02
Set gmod_spawnnpc npcs lagcompensated

wrefgtzweve:
This PR feels a bit too simple however i don't see any downsides/previous PR's.


robotboy655:
While I don't know what benefit this brings, the downside I can see is that this will blowup lag compensation usage, possibly resulting in performance issues, especially for those "NPC battle" players.

This feels like something a mod would do, not be in the base game to me.

wrefgtzweve:
A good example would be trying to hit a manhack or headcrab on anything about ~20ms ping, the more ping the worst it gets
at higher ping it's incredibly hard to even hit a npc_zombie in the head while they're walking after you due to their head shaking.

PR #2134, opened 2024-10-11
Create .gitattributes

BreakinBenny:
This'll probably void having to do anything to the files manually when Git does stuff in the repo.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,13 @@
+* text=auto eol=lf
+
+*.cfg text
+*.fgd text
+*.rad text
+*.md text
+*.html text
+*.res text
+*.lst text
+*.scr text
+*.rc text
+*.vbsp text
+*.lua text


robotboy655:
Huh?

BreakinBenny:
> Huh?

Yeah, its purpose is to automatically make EOLs of files with specified extensions use Line-Feed only... At worst it'll only be helpful with Git; I use Github Desktop more.

robotboy655:
I don't think this is necessary, at least not now. I am afraid this will cause negative side-effects to existing open PRs while bringing no benefit.

BreakinBenny:
Got it.

BreakinBenny:
@robotboy655 Can a .gitattributes still be supplied, even if its contents are commented out or it is empty to begin with?

robotboy655:
Why?

BreakinBenny:
Because... then it can be freely changed afterwards when you have fewer PRs that'll require alteration before approval...?

Or maybe not, the amount of PRs that'll also require LF conversion treatment means that has to wait.

PR #2136, opened 2024-10-12
Color manipulation methods

Be1zebub:
Features:
HWB support.
Color manipulation methods with HSV, HSL, HSB, HWB.

The main way to use this methods is to create colors for the UI.
For example, the Discord color scheme consists of the same hue and saturation, they just change lightness to create a beautiful color palette.

All you need to make pallete is background color, text color & accent color.
So addons theme config can be just 3-5 colors, instead of:
image

Demo:
<video src="https://github.com/user-attachments/assets/328d0943-14dc-476c-810a-375741a8b728"/>

demo src: https://gist.github.com/Be1zebub/a015958b71eb8806f0ce5db3e55ed4c0


garrysmod/lua/includes/util/color.lua
```diff
@@ -110,3 +110,178 @@ function COLOR:ToTable()
...
+function COLOR:AddBrightness( brightnessAdd )
+
+	local hue, saturation, brightness = self:ToHSV()
+	brightness = math.Clamp( brightness + brightnessAdd, 0, 1 )
+	ColorCopy( self, HSVToColor( hue, saturation, brightness ) )
+
+end
+
+-- hsv value == hsb brightness
+COLOR.GetValue = COLOR.GetBrightness
```
robotboy655, line 195: I don't like the naming here, `GetValue` is too generic. it is also an alias.

robotboy655:
Do you think it's also worth porting `HSVToColor` and `HSLToColor` to Lua? Not sure how the performance would differ, but it would solve https://github.com/Facepunch/garrysmod-issues/issues/2407

Be1zebub:
is this legit? lua implementation 146% faster
image
https://gist.github.com/Be1zebub/e069b63ca4b96d16ed99fd77af2ae9aa

Be1zebub:
ready 2 merge
image

Be1zebub:
mod leads to unexpected results.
eg hsv `80, 1, 1` gives rgb `170, 0, 0` instead of rgb `170, 255, 0`
`255 % 255 == 0`
better to use `math.min`/`math.max` in this case

it looks like c++ also uses mod, which is incorrect
eg C++ `HSVToColor(80, 0, 1.1)` gives `24, 24, 24`
while lua `HSVToColor(80, 0, 1.1)` that using formulas from wikipedia gives `255, 255, 255` which corresponds to
```lua
if saturation == 0 then -- grayscale, depends only on value
    rgb = math.floor(value * 255) -- 280.5 (255 floor)
end
```
c++ `24` is like `280.5 % 255` which is `25.5` with weird rounding that you are noticed.

robotboy655 (approved):

PR #2138, opened 2024-10-17
Adds Entity:AddSpawnFlags, Entity:RemoveSpawnFlags and Entity:SetSpawnFlags

bloodycop7:
Adds NPC:SetSpawnFlags, NPC:GiveSpawnFlag and NPC:RemoveSpawnFlag
In my opinion, it's easier to use than the bit. library, for those who don't understand it.


Zaurzo:
I feel like this should be an addition for all entities, not just npcs

Zaurzo:
There's already an extension file for entity (includes/extensions/entity.lua) so just add it there, no need for a new file. Also, extensions are included manually within `includes/init.lua`, so this npc.lua file is never going to be included unless you add it to init.lua

Zaurzo (approved):

garrysmod/lua/includes/extensions/player.lua
```diff
@@ -241,6 +241,17 @@ end
 
 end
 
+--[[---------------------------------------------------------
+	OnLadder
+	Returns true if the player is climbing a ladder
+-----------------------------------------------------------]]
+
+function meta:OnLadder()
```
neico, line 249:
```suggestion
function meta:IsOnLadder()
```
`On` prefixes should be reserved for Outputs, besides, we already have a few `IsOn` functions.
There's also to consider that NextBot can also use ladders, making it not bound specifically to the player.

If you want to pollute this PR with unrelated changes...

robotboy655:
Please do not include unrelated changes to your Pull Requests.

If you want to suggest further additions, open a second/new Pull Request.

neico:
You could inline `newSpawnFlags` into the `SetKeyValue` call if you care about micro-optimization, as it's not used anywhere else and therefore saves a variable.

Also, I don't think an explicit check if flags already exist is needed, that way you can add multiple flags in a single call, or just one at a time, leaving it up to the individual.

If you want to complete the set though, add `SetSpawnFlags` and `ClearSpawnFlags`, the latter being a C++ function along those other 2.
Example:
```lua
function meta:SetSpawnFlags( Flags )
	return self:SetKeyValue( "spawnflags", Flags )
end

function meta:ClearSpawnFlags()
	return self:SetSpawnFlags( 0 )
end

function meta:AddSpawnFlags( Flags )
	return self:SetSpawnFlags( bit.bor( self:GetSpawnFlags(), Flags ) )
end

function meta:RemoveSpawnFlags( Flags )
	return self:SetSpawnFlags( bit.band( self:GetSpawnFlags(), bit.bnot( Flags ) ) )
end
```
Just a suggestion, and I'm not sure what robotboy655 thinks of such trivial functions.

garrysmod/lua/includes/extensions/entity.lua
```diff
@@ -4,6 +4,22 @@ local meta = FindMetaTable( "Entity" )
 -- Return if there's nothing to add on to
 if ( !meta ) then return end
 
+function meta:SetSpawnFlags( Flags )
+	self:SetKeyValue( "spawnflags", Flags )
+end
+
+function meta:ClearSpawnFlags()
+	self:SetSpawnFlags( 0 )
+end
```
robotboy655, line 13: This is not necessary, and there's no precedent for it in the codebase.

garrysmod/lua/includes/extensions/entity.lua
```diff
@@ -4,6 +4,22 @@ local meta = FindMetaTable( "Entity" )
 -- Return if there's nothing to add on to
 if ( !meta ) then return end
 
+function meta:SetSpawnFlags( Flags )
```
robotboy655, line 7: Nitpick: the argument name is capitalized here, and not capitalized in your other changes. It should all be lowercase.

robotboy655 (changes requested):

neico:
If you're going to nitpick then also mention that all instances of `flag` should be `flags`, including the function name, as you can add and remove multiple with each function.

robotboy655 (approved):

robotboy655:
I don't think your recent changes are a good idea.

Zaurzo:
> > I don't think your recent changes are a good idea.
> 
> How come?

- It is inconsistent with the other Flag entity methods. None of them utilize varargs. 
- I'm pretty sure varargs are slower. I haven't confirmed that though, I'm not certain

garrysmod/lua/includes/extensions/entity.lua
```diff
@@ -4,6 +4,18 @@ local meta = FindMetaTable( "Entity" )
 -- Return if there's nothing to add on to
 if ( !meta ) then return end
 
+function meta:SetSpawnFlags( flags )
+    self:SetKeyValue( "spawnflags", bit.bor( flags ) )
```
Zaurzo, line 8: Redundant use of `bit.bor` here

garrysmod/lua/includes/extensions/entity.lua
```diff
@@ -4,6 +4,18 @@ local meta = FindMetaTable( "Entity" )
 -- Return if there's nothing to add on to
 if ( !meta ) then return end
 
+function meta:SetSpawnFlags( flags )
+    self:SetKeyValue( "spawnflags", flags)
```
neico, line 8: nitpick: the space after `flags`, or rather, the lack thereof

Zaurzo (approved):

robotboy655 (approved):

PR #2141, opened 2024-10-30
Add 2 convars: cl_show_shiftesc_hint and cl_show_error_duration

osiris-plus:
`cl_show_shiftesc_hint`- 0/1 - Show notification about Shift+ESC after pressing ESC too many times
`cl_show_error_duration` - 1-30, default 10 - For how many seconds the error banner will be displayed

Convars should be added to the blocklist https://wiki.facepunch.com/gmod/Blocked_ConCommands


robotboy655:
No, there will not be an option to disable this warning.

robotboy655:
I don't think that should cause a problem though, unless you have 3 menus nested at the same time or something.

The notification only appears if you spam the escape key AND it gets actually blocked.

PR #2143, opened 2024-11-01
Fix "Bad SetLocalOrigin(x,y,z) on gmod_hands" warnings

thecraftianman:
Closes Facepunch/garrysmod-issues#1184

This PR should fix all instances of this mildly annoying warning that can occur under normal gameplay circumstances. Hands will be (re)parented to the viewmodel only when they actually need to be, and their position will now always be set to prevent the engine from complaining when they are parented or unparented.

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/gmod_hands.lua
+++ b/garrysmod/gamemodes/base/entities/entities/gmod_hands.lua
@@ -34,8 +34,10 @@ end
 
 function ENT:GetPlayerColor()
 
+	--
 	-- Make sure there's an owner and they have this function
 	-- before trying to call it!
+	--
 	local owner = self:GetOwner()
 	if ( !IsValid( owner ) ) then return end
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/base/entities/entities/gmod_hands.lua
```diff
@@ -46,22 +44,33 @@ function ENT:GetPlayerColor()
 
 end
 
-function ENT:ViewModelChanged( vm, old, new )
```
robotboy655, line 49: Please do not include unnecessary changes in your Pull Requests.

thecraftianman:
Should be all good now

robotboy655 (approved):

robotboy655:
Appears to be working, at least when teleporting on `gm_mobenix_v3`.

a-straight-white-male:
Please revert, fixing an "annoying warning" is NOT worth killing half of all weapon addons!!!

PR #2145, opened 2024-11-04
Add text justification to markup

danx91:
# Features
This pull request adds 2 new functionalities to the markup object:
* Passing `true` as `textAlign` in `MarkupObject:Draw` will cause the text to be justified (aligned both to the left and the right)
* Added new tag `<par>` that will tell markup to create new paragraph

<sub>If you think that passing `true` as an argument that is expected to be a number is bad, a new argument can be added, but I simply wanted to avoid bloating this function with new parameters and it won't cause any problem with compatibility.</sub>
# Details
Long texts aligned to the left are ugly and annoying to read, but long portions of text that are justified are much more pleasing to the eye. This pull request adds such functionality to the markup library.

Passing `true` as `textAlign` will cause the markup object to render justified text while preserving all other features of markup. A `<par>` tag can be used to signal markup object to create a new paragraph (`\n` is considered a "soft" line break), breaking the line (with 1.5x line height gap) and disabling justification on the last line of the paragraph (in non-justified mode the new line with 1.5x gap is still added). `\t` is ignored when it's not adding the full 50px spacing while rendering justified text and `halign` argument is ignored as text is always taking up the full space.

Everything is made without altering the original markup functionalities so no need to worry about setting the whole world on fire by this commit. Despite rendering justified text is quite expensive compared to non-justified text because each word has to be rendered separately to maintain consistent gaps between words (it makes little-to-no impact while rendering non-justified text because it's getting rendered in the old way), I think it's still a cool functionality to have.

I tired my best to test everything and I think it's working just fine. If there is something that have to be changed or there is some kind issue with text justification/my code, please let me know.

# Showcase
Lorem Ipsum
<sub>Top image - text is left-aligned, bottom image - text is justified</sub>

Text text
<sub>Top image - text is left-aligned, bottom image - text is justified</sub>

### DIFF
--- a/garrysmod/lua/includes/modules/markup.lua
+++ b/garrysmod/lua/includes/modules/markup.lua
@@ -4,6 +4,7 @@ local table = table
 local surface = surface
 local tostring = tostring
 local ipairs = ipairs
+local pairs = pairs
 local setmetatable = setmetatable
 local tonumber = tonumber
 local math = math
@@ -53,6 +54,7 @@ local Color = _Color
 -----------------------------------------------------------]]
 local colour_stack = { Color( 255, 255, 255 ) }
 local font_stack = { "DermaDefault" }
+local paragraph = 1
 local blocks = {}
 
 local colourmap = {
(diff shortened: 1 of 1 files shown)


danx91:
Just noticed that I made a little mistake with `halign` argument. fixed in 9f85409

PR #2147, opened 2024-11-07
Add 2 Toolgun Hooks for easier access to Disabling Tracer and Sound Play

bloodycop6385:
Adds **ShouldToolgunEmitEffects** and **ShouldToolgunEmitSound**, both with `Player owner, Entity toolgun, Entity entity` arguments


Zaurzo:
For disabling sound play, you can just make an EntityEmitSound hook. Also, what is the purpose of either of these?

PR #2149, opened 2024-11-10
ScreenScale and ScreenScaleH with custom resolution

bailmast:


garryspins:
Adjacent functions have been common place in UI libraries for the better part of a decade now, I see this change as nothing but beneficial.

PR #2155, opened 2024-11-20
Add string.IsNullOrEmpty and string.IsNullOrWhitespace

marchc1:


garryspins:
Feel like a pattern is a little unneeded here

garryspins:
I guess if this is intended to be used on a string and not just a char that's valid!

PR #2124, opened 2024-09-29
Micro optimizations to `player.GetBy*` functions

joeyjumper94:
speed up player.GetBy____ID(ID) functions by relying primarily on indexing a table

PR #2161, opened 2024-11-24
Move `player finders` to a proper place & optimize them

noaccessl:
I thought it would be better to move these functions to a specialized place.
There was even a comment about it there.
```lua
-- These are totally in the wrong place.
```

~Also, optimized them by using bidirectional mapping between IDs and players for easier fetching and cleanup.~

~UPD 25.04.06: [commit "Let the Lua GC be responsible for maps cleanup"](https://github.com/Facepunch/garrysmod/pull/2161/commits/d7ac3da0fe92f70bbe0fea73aed5ccec101e3c25). Now there's no need in utilizing `EntityRemoved` hook, meaning less potential troubles.~
UPD2 25.04.06: Brought back the initial changes; Lua GC is unreliable regarding GMod's NULL objects.
UPD3 25.04.06: [commit "Rework"](https://github.com/Facepunch/garrysmod/pull/2161/commits/0074bebd7d51690fc797f6efee5e764152c00b76)
* OnEntityCreated hook for update -> Validation-in-place
* EntityRemoved hook for GC -> Timed GC


noaccessl:
> Are you sure this cannot have a client side side effect just like ents.Iterator, invalid entities that even IsValid check doesn't fix

Most likely it should never happen as this just clears the certain player from stored data. `ents.Iterator` invalidates cache entirely, setting recaching up for the next call.

As far as I know, these two hooks shouldn't provide invalid players and only called once.

Cheatoid:
Let's have another one of those again, yay! #2124 
I can't think of why would you do this, like this is not optimizing anything, it is just introducing bugs and reliability issues.
Are you also not aware that UniqueID has hash collisions because it is CRC32? Also SteamID == `BOT` for all bots while their SteamID64 is different, you see where I am going here(no?), add 1 bot, `player.GetBySteamID("BOT")` would return it, add 1 more bot, it returns the different (latest bot), kick the second bot, and do `player.GetBySteamID("BOT")`, what do you think is the result? nil... whereas in the original code, it would always return the first bot it would find under the entity list. On top of that, the lengths of your lookup/map tables would forever differ from that point onwards.
This could cause some spectacular headache when a player's entity is briefly removed (e.g. after being killed), or some ID being empty or nil during `OnEntityCreated` due to its nature... when this happens, player object might become NULL entity, at which point something is going to mess up the lookup table... Why would you not invalidate whole cache on full update?

...

..

---

You want to optimize "player finders"? My advice is to stop doing it.

Astralcircle:
Weak tables don't clear NULL keys as far as I remember

artemking4:
I am against the merge here, because this can simply return `false` in case a player has not been referenced anywhere else in the state for *long enough*.

Vitroze:
Hello @noaccessl, 
I don't understand the point of having a timer that executes every 30 seconds to know if the player is valid or not. Why not make a hook with PlayerDisconnected, which is server-side, or player_disconnect, which is server and client-side (https://wiki.facepunch.com/gmod/gameevent/player_disconnect). This will save performance and avoid two loops every 30 seconds. 


If I'm wrong, don't hesitate to tell me ;)

noaccessl:
> Hello @noaccessl, I don't understand the point of having a timer that executes every 30 seconds to know if the player is valid or not. Why not make a hook with PlayerDisconnected, which is server-side, or player_disconnect, which is server and client-side (https://wiki.facepunch.com/gmod/gameevent/player_disconnect). This will save performance and avoid two loops every 30 seconds.
> 
> As for the validate function, why not use this player_spawn (https://wiki.facepunch.com/gmod/gameevent/player_spawn) hook to set it? It's a client- and server-side hook, so to simulate a PlayerInitialSpawn, you can set a variable that Init only once.
> 
> If I'm wrong, don't hesitate to tell me ;)

> This will save performance and avoid two loops every 30 seconds.
The performance cost of two loops in a single tick every 30 seconds is extremely low.

Hooks might be unreliable (some addon returns `false` or errors and everything is stopped). Gameevent utilizes network.

PR #2163, opened 2024-11-26
Fixed EFL_KEEP_ON_RECREATE_ENTITIES not being respected on Admin Cleanup

StarLight-Oliver:
I was debating adding it to cleanup.Add and Client Cleanup.
I think it might be worth while adding it to them and to the specific forms of clean up as well.
But I have left this as minimal, but enough to work for my use case.

I also moved all the checks next to each other inside the Add function, so its clearer what is being checked.

### DIFF
--- a/garrysmod/lua/includes/modules/cleanup.lua
+++ b/garrysmod/lua/includes/modules/cleanup.lua
@@ -68,7 +68,7 @@ if ( SERVER ) then
 		cleanup_list[ id ][ type ] = cleanup_list[ id ][ type ] or {}
 
 		table.insert( cleanup_list[ id ][ type ], ent )
-	
+
 	end
 
 	function ReplaceEntity( from, to )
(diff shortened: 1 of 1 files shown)


robotboy655:
I am not sure this is such a good idea. This would leave entities essentially "orphaned", not tied to any game system. Do you have a specific example where absence of this change causes problems?

StarLight-Oliver:
I was working on an efficient perma propped system. A huge amount of lag on map clean up, comes from having permanent entities respawned. And this made map refreshes alot nicer to do.

Another example is a vehicle store that uses a server side entity distance check to validate and place your requested vehicle. If a client has the menu open when a map cleanup happens, the vendor will get removed and replaced. And now your server side validation fails, which will degrade the user experience. This can be fixed by checking in pre and then resetting in post, but this adds added complexity and the more complex the entity and its interactions are, it becomes more painful to achieve.

robotboy655:
In both of these cases, why is https://wiki.facepunch.com/gmod/cleanup.Add being explicitly called for those entities that need to survive a map clean up?

You are explicitly tying server-wide entities to a specific player. That sounds wrong.

robotboy655:
In that case using this would be the proper solution:
https://wiki.facepunch.com/gmod/cleanup.ReplaceEntity

Call it and the `undo` library one when a spawned entity gets moved from player spawned to your own system.

PR #2164, opened 2024-11-28
Add utf8.lower & utf8.upper

Be1zebub:
Implementation based on mapping (50kb for 2 tables).  
This is a compromise for performance and simplicity, in exchange for 50kb mem.


garrysmod/lua/includes/modules/utf8_case.lua
```diff
@@ -0,0 +1,15 @@
...
+
+local uc_lc = {}
+for lc, uc in pairs( lc_uc ) do uc_lc[ uc ] = lc end
```
WilliamVenner, line 4: You could use [table.Flip](https://wiki.facepunch.com/gmod/table.Flip) here

Be1zebub:
@CornerPin 
I dont think it plays a big role.
Just looked at the godot code and tested few other projects, it seems no one cares about special casing.
It also goes beyond a simple and effective implementation, better to make os bindings in this case.

JarosLucky:
needed

xPizu (commented):
Good job !

Malivil:
I just found a situation where this would be very useful -- please merge it in =)

PR #2165, opened 2024-11-30
player/ents.GetCached

Astralcircle:
An addition to ents.Iterator and player.Iterator that allows you to modify the returned table. It is still much faster than ents/player.GetAll, but slower than ents/player.Iterator. Below is a comparison

Benchmark code:
```lua
local SysTime = SysTime
local ipairs = ipairs
local MsgN = MsgN

function GMODBenchmark(repeats, ...)
    local times = {}

    for k, benchmark_function in ipairs({...}) do
        local start_time = SysTime()

        for i = 1, repeats do
            benchmark_function()
        end

        local end_time = SysTime()
        times[k] = end_time - start_time
    end

    MsgN(string.format("\nBenchmark result %s:", SERVER and "(SERVER)" or "(CLIENT)"))
    for k, v in ipairs(times) do
        MsgN(k, ": " .. v)
    end
end
```
Benchmark:
```lua
--gm_construct with 128 bots
GMODBenchmark(10000, 
function()
    local players = player.GetAll()
end,
function()
    local players = player.GetCached()
end)

GMODBenchmark(10000, 
function()
    local players = ents.GetAll()
end,
function()
    local players = ents.GetCached()
end)
```

Result:
```
Benchmark result (SERVER):
1: 0.056322099999988
2: 0.012729499999978

Benchmark result (SERVER):
1: 1.1994984999999
2: 0.10440289999997
```

### DIFF
--- a/garrysmod/lua/includes/extensions/ents.lua
+++ b/garrysmod/lua/includes/extensions/ents.lua
@@ -38,6 +38,19 @@ function ents.Iterator()
 
 end
 
+function ents.IteratorForEdit()
+
+	if ( EntityCache == nil ) then EntityCache = ents.GetAll() end
+
+	local entities = {}
+	for k, v in ipairs( EntityCache ) do
+		entities[k] = v
+	end
+
+	return entities
+
+end
+
 local function InvalidateEntityCache( ent )
 
 	EntityCache = nil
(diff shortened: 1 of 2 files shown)


Astralcircle:
I don't think this should be added

PR #2166, opened 2024-12-02
Localization: added more missing strings/moved to .properties

Satton2:
- Moved ammo strings from "garrysmod_english.txt" to "entities.properties" (to make pick up notification localizable).
- - Some characters' capitalization in those strings was updated (for some reason after moving to .properites they were trying to find different strings).
- - Some ammo names were updated to avoid confusion when picking up HL1 ammo similar to HL2, which are actually different.
- Moved strings from "\language\npc-ents-english.txt" to "entities.properties".
- - for example, this will allow the kill feed to get their localized alternatives.
- - Removed \language\npc-ents-english.txt since now it's empty.
- Added localization support for "Author:" in the spawn menu tooltips (for entities, weapons, vehicles).
- Added localization support for "Force:" in the world tooltip for balloons.
- ~~Garry's Mod~~ (already done by Rubat here: https://github.com/Facepunch/garrysmod/commit/a413b41ddfcc77aac57656073ab9b63417a2fe1f), HL2 and HL1 weapons now use localization from entites.properties in the weapon selector and the pickup notifications.
Added "npc_tf2_ghost" to "entities.properties": turns out we didn't have it in the past.

It works fine with GMod's localization system:

https://github.com/user-attachments/assets/cfb8f767-3bc4-4a49-b040-a1b8bd53c7ac

https://github.com/user-attachments/assets/ffbaaf7a-e1b3-4bf3-8e66-9fea26452ecb

Partly solves this: https://github.com/Facepunch/garrysmod-issues/issues/6096


NovaDiablox (changes requested):

Satton2:
> A lot of this I have already done separately without realizing it, but I do not agree with localizing every entity name. A lot of these entities are very rarely encountered. I really don't think we need to localize things like "Base Entity" or `basehlcombatweapon`

Yeah, I saw that. I will wait for the next update before revisiting this and _maybe_ creating a new PR, if there will be something worth adding. I guess, next time I should make smaller PR's.

> but I do not agree with localizing every entity name. A lot of these entities are very rarely encountered.

Fair enough. 👍

PR #2167, opened 2024-12-02
Optimize ```ents.FindByClassAndParent```

Astralcircle:
Optimizes this function by about 25%

Benchmark code:
```lua
local SysTime = SysTime
local ipairs = ipairs
local MsgN = MsgN

function GMODBenchmark(repeats, ...)
    local times = {}

    for k, benchmark_function in ipairs({...}) do
        local start_time = SysTime()

        for i = 1, repeats do
            benchmark_function()
        end

        local end_time = SysTime()
        times[k] = end_time - start_time
    end

    MsgN(string.format("\nBenchmark result %s:", SERVER and "(SERVER)" or "(CLIENT)"))
    for k, v in ipairs(times) do
        MsgN(k, ": " .. v)
    end
end
```
Result:
```
Benchmark result (SERVER):
1: 0.26664929999993
2: 0.35286960000008
```

### DIFF
--- a/garrysmod/lua/includes/extensions/ents.lua
+++ b/garrysmod/lua/includes/extensions/ents.lua
@@ -1,7 +1,7 @@
 
 function ents.FindByClassAndParent( classname, entity )
 
-	if !IsValid( entity ) then return end
+	if ( !IsValid( entity ) ) then return end
 
 	local out = {}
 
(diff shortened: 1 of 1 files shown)


aStonedPenguin:
Just use `ipairs`. With jit it should be the same speed, and cleaner.

PR #2168, opened 2024-12-07
Make MAX_EDICT_BITS globally available

wrefgtzweve:
Having `MAX_EDICT_BITS` globally available would help a lot for other addons that write entity indexes, currently they hardcode `13` but this can eventually be broken if gmod ever changes the entity limit.
Feel free to do it in another way, i figured having it on the net library table would make the most sense.


robotboy655:
I think it's best to do it in engine so its value is automatically synched with the real in-engine value.

wrefgtzweve:
Yeah i agree

PR #2170, opened 2024-12-10
Add option to view C Arms on viewmodel effects

A1steaksa:
This pull request adds an option to show C Arms on `prop_effect`s with a ViewModel that uses them.

It validates that the model on the `prop_effect` has all of the bones required by the C Arms bonemerge and only shows the option when they are all present.

By default, it will show a list of all the C Arms models that ship with Garry's Mod by default (Citizen, Chell, Combine, Counter-Strike, Day of Defeat, HEV, and Refugee) but the table holding those models is global so that more can be added by Addons if they want to.

C Arm choices are networked to other Players and should appear appropriately when new Players join or enter the `prop_effect`'s PVS.

**You can see it in action here:**

https://github.com/user-attachments/assets/f34d3c6b-0132-447d-9f7f-7d1a606efde8

### DIFF
--- a/garrysmod/lua/autorun/properties.lua
+++ b/garrysmod/lua/autorun/properties.lua
@@ -13,4 +13,4 @@ include( "properties/editentity.lua" )
 include( "properties/kinect_controller.lua" )
 include( "properties/bodygroups.lua" )
 include( "properties/skin.lua" )
-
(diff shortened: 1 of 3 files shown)


robotboy655:
I don't think this belongs in the base game. This is too much of a "just because" and "for testing" thing.

PR #2171, opened 2024-12-10
Bring Sandbox context menu icons to spawn menu

github-is-garbage:
Passes the panel back and forth as the menus open


garryspins:
This and the recent change regarding playermodel selection are bandaid fixes to the real issue (which arguably doesn't exist) that the contextmenu isn't obvious enough. How about we make it more obvious instead of changing established iconic visual layouts for the sake of 5 or 6 rarely used icons.

If I'm wrong as to why this would exist, please let me know!

A1steaksa:
Please include screenshots/video if you're going to make a UI change

github-is-garbage:
> Please include screenshots/video if you're going to make a UI change

discord dot gg slash gmod

I think there are better UI concepts than this anyway, this was not thought out 100% before creation. Death to #2171!

PR #2172, opened 2024-12-12
Add math.IsNearlyZero() and math.IsNearlyEqual()

xCynDev:
These simple functions should be used when comparing numbers instead of `==`, since floating point precision issues easily arise.
Felt like making them since someone on the Garry's Mod Discord server was confused as to why the following didn't work:

image

So now, you can use `math.IsNearlyEqual()` or `math.IsNearlyZero()` which solves the issue:

image

The functions allow you to give them a tolerance of your choice, but they'll default to `math.exp( -8 )` otherwise.

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -255,3 +255,29 @@ function math.Factorial( num )
 
 	return res
 end
+
+local smallNumber = math.exp( -8 )
+
+--[[---------------------------------------------------------
+	Name: IsNearlyZero( num )
+	Desc: Checks if a floating point number is nearly zero
+-----------------------------------------------------------]]
+function math.IsNearlyZero( num, tolerance )
+	if tolerance == nil then
+		tolerance = smallNumber
+	end
+
+	return math.abs( num ) <= tolerance
+end
+
+--[[---------------------------------------------------------
+	Name: IsNearlyEqual( a, b )
+	Desc: Checks if two floating point numbers are nearly equal
+-----------------------------------------------------------]]
+function math.IsNearlyEqual( a, b, tolerance )
+	if tolerance == nil then
+		tolerance = smallNumber
+	end
+
+	return math.abs( a - b ) <= tolerance
+end


xCynDev:
> - I think `math.IsNearlyEqual` should be renamed to `math.IsEqualTol` for consistency because we already have methods like [`Angle:IsEqualTol`](https://wiki.facepunch.com/gmod/Angle:IsEqualTol) and [`Vector:IsEqualTol`](https://wiki.facepunch.com/gmod/Vector:IsEqualTol). Similarly, `math.IsNearlyZero` can be renamed to `math.IsEqualZeroTol`.

Wouldn't be against it. In my opinion, IsNearlyEqual is more readable than IsEqualTol, I would keep it that way if it isn't going to prevent it from being merged as it won't affect new code.

> - An epsilon of $e^{-8}$ seems arbitrary to me, did you mean to set it to `1e-8` instead?

Yep! TIL Lua supports 1e-8 as a constant, I'll fix that in a bit.

robotboy655:
Fixes https://github.com/Facepunch/garrysmod-requests/issues/1991

garrysmod/lua/includes/extensions/math.lua
```diff
@@ -255,3 +255,29 @@ function math.Factorial( num )
 
 	return res
 end
+
+local epsilon = 1e-8
```
robotboy655, line 259: Does this need to be a separate variable?

garrysmod/lua/includes/extensions/math.lua
```diff
@@ -255,3 +255,29 @@ function math.Factorial( num )
...
 end
+
+local epsilon = 1e-8
+
+--[[---------------------------------------------------------
+	Name: IsNearlyZero( num, tolerance )
+	Desc: Checks if a floating point number is nearly zero
+-----------------------------------------------------------]]
+function math.IsNearlyZero( num, tolerance )
+	if tolerance == nil then
```
robotboy655, line 266: This does not adhere to CONTRIBUTING.md

robotboy655 (approved):
Regarding the names, I think it's fine as is, you are searching for this type of function, you'd type either `zero` or `equal` into the search bar and find both the existing vector ones, and these names.

robotboy655 (changes requested):
I didn't mean to approve the changes with my previous comment.

garrysmod/lua/includes/extensions/math.lua
xCynDev: Fixed

garrysmod/lua/includes/extensions/math.lua
xCynDev: Fixed.

xCynDev:
Everything has been fixed & pushed, let me know if you require any other changes.

robotboy655 (approved):

PR #2173, opened 2024-12-12
Make all post-processing conditional

Denneisk:
Optimizes post-processing by not having their hooks run when they're never used. Also removes some unnecessary checks or moves them to run only once.

Now post-processing hooks run only when needed.

I tested it and it all works fine, but I would appreciate anyone looking over this to make sure I didn't do any copying errors.

Was waiting for someone else to do this before I realized it'd be faster to do it myself.

### DIFF
--- a/garrysmod/lua/postprocess/bloom.lua
+++ b/garrysmod/lua/postprocess/bloom.lua
@@ -58,21 +58,25 @@ function DrawBloom( darken, multiply, sizex, sizey, passes, color, colr, colg, c
 
 end
 
---[[---------------------------------------------------------
-	The function to draw the bloom (called from the hook)
------------------------------------------------------------]]
-hook.Add( "RenderScreenspaceEffects", "RenderBloom", function()
+cvars.AddChangeCallback( "pp_bloom", function( _, _, newValue )
 
 	-- No bloom for crappy gpus
 
 	if ( !render.SupportsPixelShaders_2_0() ) then return end
-	if ( !pp_bloom:GetBool() ) then return end
 	if ( !GAMEMODE:PostProcessPermitted( "bloom" ) ) then return end
 
-	DrawBloom( pp_bloom_darken:GetFloat(), pp_bloom_multiply:GetFloat(),
+	if ( newValue != "0" ) then
+		hook.Add( "RenderScreenspaceEffects", "RenderBloom", function()
+
+			DrawBloom( pp_bloom_darken:GetFloat(), pp_bloom_multiply:GetFloat(),
 				pp_bloom_sizex:GetFloat(), pp_bloom_sizey:GetFloat(),
 				pp_bloom_passes:GetFloat(), pp_bloom_color:GetFloat(),
 				pp_bloom_color_r:GetFloat() / 255, pp_bloom_color_g:GetFloat() / 255, pp_bloom_color_b:GetFloat() / 255 )
+
+		end)
+	else
+		hook.Remove( "RenderScreenspaceEffects", "RenderBloom" )
+	end
 
 end )
 
--- a/garrysmod/lua/postprocess/dof.lua
+++ b/garrysmod/lua/postprocess/dof.lua
@@ -27,6 +27,8 @@ function DOF_Kill()
 
 	DOFModeHack( false )
 
+	hook.Remove( "Think", "DOFThink" )
+
 end
 
 function DOF_Start()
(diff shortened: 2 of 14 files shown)


Cheatoid:
I like the idea, but:
- Change this logic: `newValue != "0"` to check particular convar's boolean value instead (`ConVar:GetBool()`), this is because any non-zero value is considered true in engine, even `-1` and `123.45`. Consistency.
- Have you considered what would happen if any of those Lua files were (hot)reloaded (or loaded twice for whatever reason), no? Well, it would add another instance of those change callbacks into the list, and there be double shenanigans.

Denneisk:
Thanks for the response

> Change this logic: `newValue != "0"`

I thought all nonzero values were true for convars. I thought it would be simpler this way.

> (hot)reloaded

Thank you for considering this, that's a very important oversight.

> `render.SupportsPixelShaders_2_0()`

Good idea. Thanks.

garrysmod/lua/postprocess/bloom.lua
```diff
@@ -58,22 +58,26 @@ function DrawBloom( darken, multiply, sizex, sizey, passes, color, colr, colg, c
...
+			DrawBloom( pp_bloom_darken:GetFloat(), pp_bloom_multiply:GetFloat(),
 				pp_bloom_sizex:GetFloat(), pp_bloom_sizey:GetFloat(),
 				pp_bloom_passes:GetFloat(), pp_bloom_color:GetFloat(),
 				pp_bloom_color_r:GetFloat() / 255, pp_bloom_color_g:GetFloat() / 255, pp_bloom_color_b:GetFloat() / 255 )
 
+		end)
+	else
+		hook.Remove( "RenderScreenspaceEffects", "RenderBloom" )
+	end
+
```
robotboy655, line 80: Doing it this way introduces a regression where the result of `GAMEMODE:PostProcessPermitted` is not updated live.

I can understand `SupportsPixelShaders_2_0` being a static thing and can be moved outside of the hook.

I also think it's a better idea to make a the hook a local function and reference that in the `cvars.AddChangeCallback` callback, rather than nesting functions like this.

garrysmod/lua/postprocess/bloom.lua
```diff
@@ -113,4 +117,4 @@ list.Set( "PostProcess", "#bloom_pp", {
 
 	end
 
-} )
+} )
```
robotboy655, line 120: Unnecessary change.

garrysmod/lua/postprocess/dof.lua
```diff
@@ -43,14 +45,14 @@ function DOF_Start()
...
-end
+	hook.Add( "Think", "DOFThink", function()
 
-hook.Add( "Think", "DOFThink", function()
+		DOF_SPACING = pp_dof_spacing:GetFloat()
+		DOF_OFFSET = pp_dof_initlength:GetFloat()
 
-	DOF_SPACING = pp_dof_spacing:GetFloat()
-	DOF_OFFSET = pp_dof_initlength:GetFloat()
+	end)
```
robotboy655, line 53: Unnecessary change that does not conform to the style guidelines:

garrysmod/lua/postprocess/frame_blend.lua
```diff
@@ -186,6 +186,18 @@ hook.Add( "PostRender", "RenderFrameBlend", function()
 	frame_blend.AddFrame()
 	frame_blend.DrawPreview()
 
+end
+
+cvars.AddChangeCallback( "pp_fb", function( _, _, newValue )
+
+	if ( !GAMEMODE:PostProcessPermitted( "fb" ) ) then return end
```
robotboy655, line 193: Should be `"frame_blend"`, not `"fb"`

garrysmod/lua/postprocess/frame_blend.lua
```diff
@@ -174,9 +174,9 @@ end
 --
 if ( engine.IsPlayingDemo() ) then return end
 
-hook.Add( "PostRender", "RenderFrameBlend", function()
+local function RenderFrameBlend()
 
-	if ( !frame_blend.IsActive() ) then return end
+	if ( !frame_blend.IsActive() ) then hook.Remove( "RenderFrameBlend", "RenderFrameBlend" ) return end
```
robotboy655, line 179: Why is this necessary?

garrysmod/lua/postprocess/super_dof.lua
```diff
@@ -354,23 +342,35 @@ concommand.Add( "pp_superdof", function()
...
-hook.Add( "GUIMouseReleased", "SuperDOFMouseUp", function( mouse )
+		if ( !IsValid( SuperDOFWindow ) ) then hook.Remove( "GUIMouseReleased", "SuperDOFMouseUp" ) return end
 
-	if ( !IsValid( SuperDOFWindow ) ) then return end
+		vgui.GetWorldPanel():MouseCapture( false )
+		FocusGrabber = false
 
-	vgui.GetWorldPanel():MouseCapture( false )
-	FocusGrabber = false
+	end )
```
robotboy655, line 373: This entire file is a mess. See previous comments. 

I also think this can be all done better, such as removing the hooks from `Panel.OnRemove` of the `SuperDOFWindow` for example.

robotboy655 (changes requested):

robotboy655:
I think 2 months is enough time to react to requested changes.

PR #2175, opened 2024-12-18
Add hook.ProtectedRun and hook.ProtectedCall functions

Srlion:
`hook`.`ProtectedCall`/`ProtectedRun` will execute hooks without interruptions. If a hook returns a value or encounters an error, it will not halt the execution of the remaining hooks.
These functions are particularly useful for hooks that do not expect a return value, such as Initialize hooks.
Many poorly designed addons in the workshop return values within critical hooks, making it difficult to track down what is preventing a hook from working correctly or causing issues.

Here is a simple list of hooks that could benefit from it:
[GM:PreGamemodeLoaded](https://gmodwiki.com/GM:PreGamemodeLoaded)
[GM:OnGamemodeLoaded](https://gmodwiki.com/GM:OnGamemodeLoaded)
[GM:PostGamemodeLoaded](https://gmodwiki.com/GM:PostGamemodeLoaded)
[GM:Initialize](https://gmodwiki.com/GM:Initialize)
[GM:OnEntityCreated](https://gmodwiki.com/GM:OnEntityCreated)
[GM:EntityRemoved](https://gmodwiki.com/GM:EntityRemoved)
[GM:InitPostEntity](https://gmodwiki.com/GM:InitPostEntity)

### DIFF
--- a/garrysmod/lua/includes/modules/hook.lua
+++ b/garrysmod/lua/includes/modules/hook.lua
@@ -7,6 +7,7 @@ local isbool = isbool
 local IsValid = IsValid
 local type = type
 local ErrorNoHaltWithStack = ErrorNoHaltWithStack
+local GProtectedCall = ProtectedCall
 
 module( "hook" )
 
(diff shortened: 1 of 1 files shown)


Aws0mee:
Would be very nice to avoid some random workshop script breaking servers when it causes a minor error in an initialization hook.

PR #2176, opened 2024-12-20
Allow passing varargs from vgui.Create to Panel:Init

luca1197:
This PR makes it possible to pass varargs from [`vgui.Create`](https://wiki.facepunch.com/gmod/vgui.Create) to [`Panel:Init`](https://wiki.facepunch.com/gmod/PANEL:Init) for custom (user-created) panels.

Passing arguments to Panel:Init can be quite useful in a few cases, generally when the goal is to allow changing some behavior in the Init function without it being really necessary to separate these different behaviors into individual functions.

It would be possible to define a custom init function for the panel and call that after `vgui.Create`, but I don't see a reason why a cleaner solution like this shouldn't exist.

Thanks :)

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel/scriptedpanels.lua
+++ b/garrysmod/lua/includes/extensions/client/panel/scriptedpanels.lua
@@ -20,7 +20,7 @@ function vgui.Exists( classname )
 	return PanelFactory[ classname ] != nil
 end
 
-function vgui.Create( classname, parent, name )
+function vgui.Create( classname, parent, name, ... )
 
 	-- Is this a user-created panel?
 	if ( PanelFactory[ classname ] ) then
@@ -38,7 +38,7 @@ function vgui.Create( classname, parent, name )
 
 		-- Call the Init function if we have it
 		if ( panel.Init ) then
-			panel:Init()
+			panel:Init( ... )
 		end
 
 		panel:Prepare()


darkfated:
A conditional constructor that allows arguments to be passed during initialization... In the context of GLua, this can cause confusion, because it is not known which of the ui elements has support for this, and the elements for which things we transmit. It is better to adhere to encapsulation, and formalize everything through methods. A useless thing

luca1197:
I see what you're saying, but in my opinion the goal should be to allow different implementation "styles" depending on the many possible use-cases. If you prefer not to make use of it, that's totally fine, but the option remains available for those who do. Especially since this is just a minor change that applies only to panels that explicitly choose to make use of it.

garryspins:
I think this is better suited for your own wrapper/create function. Its valid to omit Init for your own constructor!

robotboy655:
I see a few problems with this:
1) Inheritance handling. If you want to make it happen, you'd probably have to use tables, making the vararg part obsolete immediately
2) Usage of varargs locks us out of ever modifying the arguments of vgui.Create (can't add any new ones), not to mention, it already has quite a few.
3) Does it really add any useful functionality?

Is the above actually cleaner than the below?
```
local editor = vgui.Create( "IconEditor", nil, nil, "test1", "test2" )

local editor = vgui.Create( "IconEditor" )
editor:MySetup( "test1", "test2 )
````

I'd argue that this make the code less readable and therefore less clean, as you don't know what the varargs in the top example do without opening the source code for the `Init` of the panel, whereas with a separate function you could likely infer what they do from the function name.

PR #2178, opened 2024-12-31
Improve file.Write function to return success status

Srlion:


darkfated:
Why not, good

robotboy655:
Might as well do this for `file.Append` too?

robotboy655 (approved):

PR #2181, opened 2025-01-04
Only update model if it has been changed

Cheatoid:
(untested)


robotboy655:
Please actually test your changes before submitting them, and explain their purpose.

It also does not conform to the contributor's guidelines.

As it stands, I see no good reason to merge this.

PR #2182, opened 2025-01-08
Remove metatable deletion ToolObj

animerem:
If removed, this makes it possible to create an instrument almost anywhere, useful for modular gamemodes.

If this cannot be done, please explain why.


robotboy655:
Don't bother submitting a PR blanket replacing `GetConVarNumber`, I am not going to accept it.

robotboy655:
Because it already caches the GetConVar internally, so the changes would infact be worse for performance.

PR #2189, opened 2025-01-17
net.Write65533String because net.WriteString's limit of 65532 characters is not enough!

blobles-dev:
net.WriteString has a limit of 65532 characters and sometimes, that's just not enough, and you need just an extra byte. If we're going to be filling the net message to it's maximum size, why do we need to waste a byte for a null terminator?

Introducing: `net.Write65533String` & `net.Read65533String`

An alternative approach would be not wasting a byte on a null terminator with net.WriteString when the size is 65533

Usage example (SV&CL)
```
util.AddNetworkString("please_sir_may_I_have_some_more")

concommand.Add("test_long_string", function()
    local strbuildtbl = {}
    for i = 1, 65533 do
        local randomChar = string.char(math.random(32, 126))
        strbuildtbl[i] = randomChar
    end
    local longStr = table.concat(strbuildtbl)

    net.Start("please_sir_may_I_have_some_more")
    net.Write65533String(longStr) -- net.WriteString would NEVER dream of writing a string this long!
    net.Broadcast()
end)

net.Receive("please_sir_may_I_have_some_more", function(len)
	print(net.Read65533String())
end)
```

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -183,6 +183,62 @@ function net.ReadTable( seq )
 
 end
 
+local function bitsToChar( b1, b2, b3, b4, b5, b6, b7, b8 )
+
+	local asciiValue = bit.bor( bit.lshift(b1, 7), bit.lshift(b2, 6), bit.lshift(b3, 5), bit.lshift(b4, 4),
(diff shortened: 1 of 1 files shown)


robotboy655:
Good laugh, but this is just ridiculous.

PR #2190, opened 2025-01-21
Update AngularJS from 1.1.2 to 1.8.2

DarthTealc:
I have updated AngularJS from 1.1.2 to the final version, 1.8.2

It works in both CEF (x86-64 branch) and Awesomium (main branch) on Windows 10. Have not tested Linux but I don't expect any problems since it's all browser-level.

There are no changes to functionality or design, but would be good to know if users on lower performing devices can notice performance improvements. 

Implementation notes and explanations:
- angular.js and angular-route.js are sourced from their .min.js versions here https://code.angularjs.org/1.8.2 - for security, you can add these yourself rather than trusting the versions I submitted. Remove `.min` from the filename. 
- Some changes to menuApp.js have been explained via comments, to assist other developers.
- the 'old' angular.js (1.1.2) has been renamed angular.legacy.js, and is used by Sandbox's spawn menu saves/dupes. These sandbox spawn menu panels continue to use 1.1.2 as I was not able to resolve bugs with them. They are mostly isolated from the rest of the menu system so this isn't a problem. But would welcome anyone updating them to use 1.8.2.
- Developers: you can edit menuApp.js and comment out the `$compileProvider.debugInfoEnabled( false );` line, and use the non-minified angular script, for more detailed console errors during development. 

Other contributors are very welcome to add improvements.


robotboy655 (changes requested):
Alright, I have tested this, and it seems to work well, but I'd like to get rid of `angular.legacy.js`. 

What are the issue you were having with the spawnmenu tabs?

After some messing about I got them to work with these changes:
image


I tested main and `x86-64`, both seem to work.

robotboy655 (approved):

PR #2191, opened 2025-01-22
Avatar quality improvement

darkfated:
The avatar has a size of 32x32, and with an empty argument it pumps up a 32x32 image of standart, which has a blur, hence the quality is lost. If you put a 64px picture on a 32px avatar, then sharpness is added, and the blur effect disappears.

**Before**
gmod_e1455uviSs

**After**
gmod_sGWXlfppqe


darkfated:
image
image

robotboy655:
It's fine as is.

PR #2193, opened 2025-01-23
Balloon Crash Fix

anatolidiatlov:
Prevent player from crashing servers with balloon


robotboy655:
This disables ability for balloons to have negative buoyancy, surely the limit can be simply be made closer to zero, without compromising functionality.

anatolidiatlov:
some player abuse it by crashing servers with and yeah it can be better to just made it closer to zero

Astralcircle:
I don't know what crash you're talking about, but I couldn't reproduce this problem in singleplayer.
image

PR #2195, opened 2025-01-24
Manual Unix LF conversion

BreakinBenny:
Because the repository is not ready for a **.gitattributes**, why not start by manually removing the Carriage Returns using NPP?... Maybe files in particular folders with particular naming schemes should be targeted, rather than all instances of them?

I made sure nothing conflicts with other PRs, not even #2166. Testing shows nothing bad has happened after files were edited to only use LF line-endings.


robotboy655:
image

BreakinBenny:
…Do make sure to decide when you people will opt for the Unix-style line-endings. For your sakes, not mine nor anyone else's.

robotboy655:
You are the only one who cares it would seem though. It really doesn't make any difference to me, or most other people it would seem. I personally do not see any good reasons to change this.

PR #2196, opened 2025-01-28
Fixing ragdoll spawning with wrong faces in Sandbox.

RubisetCie:
Hello everyone.

A couple of weeks ago, I raised an issue about [ragdolls spawning with wrong face in Sandbox](https://github.com/Facepunch/garrysmod-issues/issues/6146) (it may have flown under the radar, I'm still curious about if you too experience this problem).

This pull request fixes the problem, although **I have no clue on what causes it in the first place**; so this is more of patch than a fix...

If someone with more knowledge and access to the closed-source GMod is willing to take the time to investigate, I encourage you. :)

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/commands.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/commands.lua
@@ -113,6 +113,11 @@ function GMODSpawnRagdoll( ply, model, iSkin, strBody )
 		gamemode.Call( "PlayerSpawnedRagdoll", ply, model, e )
 	end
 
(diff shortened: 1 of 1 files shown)


robotboy655:
This is not a good way to do this. `Entity:SetFlexWeight` creates a separate entity for every ragdoll that gets face posed, which gets quite expensive fast if we do it for every ragdoll spawned.

RubisetCie:
Hi, finally, I thought nobody would pick up on this issue. :)

What do you suggest to fix this bug ?

PR #2198, opened 2025-01-29
Algorithmic optimizations on math.ease functions

RubisetCie:
Good day.

Because my addons rely heavily on `math.ease` functions (which is an awesome addition for the glua, btw), I took the responsibility to optimize some of its expressions.

Mainly, it is about converting the `^` operator to a succession of multiplications `*`; it is faster because [exponentiation in Lua relies](https://www.lua.org/pil/3.1.html) on the `pow` C function, which is too much overhead for simple squaring, cubing, etc.

So, `x ^ 2` becomes `x * x`, and so on...

- I've made some benchmarks on my end, and **the performance gain is compelling** (at best, 200% faster, never slower).
- I've made sure the output value for a same input stayed the same for each modified function, it's a 100% match, so **no compatibility break**.

### DIFF
--- a/garrysmod/lua/includes/extensions/math/ease.lua
+++ b/garrysmod/lua/includes/extensions/math/ease.lua
@@ -73,7 +73,7 @@ function OutCubic( x )
 end
 
 function InOutCubic( x )
-	if x < 0.5 then
+	if ( x < 0.5 ) then
 		return 4 * ( x * x * x )
 	else
 		local w = -2 * x + 2
@@ -91,7 +91,7 @@ function OutQuart( x )
 end
 
 function InOutQuart( x )
-	if x < 0.5 then
+	if ( x < 0.5 ) then
 		return 8 * ( x * x * x * x )
 	else
 		local w = -2 * x + 2
@@ -109,7 +109,7 @@ function OutQuint( x )
 end
 
 function InOutQuint( x )
-	if x < 0.5 then
+	if ( x < 0.5 ) then
 		return 16 * ( x * x * x * x * x )
 	else
 		local w = -2 * x + 2
@@ -144,7 +144,7 @@ function OutCirc( x )
 end
 
 function InOutCirc( x )
-	if x < 0.5 then
+	if ( x < 0.5 ) then
 		local w = 2 * x
 		return ( 1 - sqrt( 1 - ( w * w ) ) ) / 2
 	else
@@ -165,7 +165,7 @@ function OutBack( x )
 end
 
 function InOutBack( x )
-	if x < 0.5 then
+	if ( x < 0.5 ) then
 		local w = 2 * x
 		return ( ( w * w ) * ( ( c2 + 1 ) * 2 * x - c2 ) ) / 2
 	else
--- a/garrysmod/lua/includes/extensions/math/ease.lua
+++ b/garrysmod/lua/includes/extensions/math/ease.lua
@@ -47,7 +47,7 @@ function InOutSine( x )
 end
 
 function InQuad( x )
-	return x ^ 2
+	return x * x
 end
 
 function OutQuad( x )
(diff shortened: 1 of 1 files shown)


RubisetCie:
Sure, I will post the benchmarking code tomorrow.

robotboy655:
GMod uses LuaJIT, which is Lua 5.1. 

I am testing on a fairly good machine with a Ryzen 5900x.

goodusername123:
Runtime performance between PUC Lua 5.1 (or really any version of PUC Lua) and LuaJIT is **wildly** different, especially on the math side of things since in LuaJIT most of that stuff (including pow) gets optimized into machine instructions and even when the relevant code that's being run doesn't make it through the JIT most of the math functions themselves are implemented in assembly.

Any benchmarking that's done **NEEDS** to be done with/in LuaJIT or the results won't matter, better yet benchmarking should probably be done inside GMod itself as done in a prior comment on this PR if possible. @RubisetCie

RubisetCie:
OK, here are the scripts I made to do my benchmarking:

[easing-benchmark.zip](https://github.com/user-attachments/files/18602959/easing-benchmark.zip)

- `bench1.lua`: Measures the time of each easing function before my change.
- `bench2.lua`: Measures the time of each easing function after my change.
- `results-original.txt`: Reports the results of bench1 (before the change).
- `results-new.txt`: Reports the results of bench2 (after the change).

It may be naive, but I did another test this morning using LuaJit, and the results were like even more drastic... especially for _InCubic,_ where I get an improvement so big I suspect there may be an error... so I'm curious about your analysis. :)

RubisetCie:
> in LuaJIT most of that stuff (including pow) gets optimized into machine instructions and even when the relevant code that's being run doesn't make it through the JIT most of the math operations themselves are implemented in assembly rather then C.

This is true. That could explain the lack of improvements if the benchmarks are refuted.

> Any benchmarking that's done **NEEDS** to be done with/in LuaJIT or the results won't matter, better yet benchmarking should probably be done inside GMod itself as done in a prior comment on this PR if possible. @RubisetCie

LuaJit is done, and shows improvements. I'll test in GMod right now.

goodusername123:
After doing a bunch of benchmarking and testing and digging (I ended up testing 12 different compiles of LuaJIT) I have found out and concluded that this behavior is LuaJIT version dependent.
more specifically LuaJIT performing worse with the current `^` operator math is due to `pow` related changes that happened to LuaJIT in 2022 (I originally thought it was due to changes in 2020 but testing revealed I was wrong)
however the reason this doesn't happen in GMod (both branches) is because it's on an older version of LuaJIT (`main` is on 2.0.4 and `x86-64` is on 2.1.0beta3 from 2017)

If GMod ever updates it's copy of LuaJIT to a modern version (hopefully it will one day) then this pull request would actually be useful if the performance of `pow` in modern LuaJIT stays as the slow behavior introduced in 2022, However current day Gmod with it's old copy won't really see much of an improvement or regression with this pull request.

[LuaJIT POW benchmark results.rar.zip](https://github.com/user-attachments/files/18615586/LuaJIT.POW.benchmark.results.rar.zip)

related LuaJIT commits/issues:
https://github.com/LuaJIT/LuaJIT/issues/684#issuecomment-1020116837 (https://github.com/LuaJIT/LuaJIT/commit/9512d5c1aced61e13e7be2d3208ec7ae3516b458)
https://github.com/LuaJIT/LuaJIT/issues/817 (https://github.com/LuaJIT/LuaJIT/commit/96d6d5032098ea9f0002165394a8774dcaa0c0ce https://github.com/LuaJIT/LuaJIT/commit/8b8304f16e03aa652be728f0dc957eadf42e4353 https://github.com/LuaJIT/LuaJIT/commit/43ebb949a249a16c49d232ad24dc9127cc4302bb)

RubisetCie:
Thank you for this deep analysis, @goodusername123 :)
It answers all of my questions about benchmark differences.

So, I guess it's better to merge my PR, but I'm bothered by the current Garry's Mod benchmarks, which showed the new version being sometimes slower than the old one... even if there's a confounding factor, I'm troubled. :|

HEV-Scientist:
I've noticed You've increased MAX_MAP_BRUSHES from 16384 to 65535, does this mean really big maps is possible?

PR #2201, opened 2025-02-03
type function in lua

Srlion:
This PR solves two issues:

1. Small/Simple functions that make use of `type` don't get jitted because `type` is reintroduced as a normal C function so luajit just blacklists them from getting jitted.
2. https://github.com/Facepunch/garrysmod-issues/issues/4100

I understand that it may be annoying to be implemented, [but from looking at luajit source code](https://github.com/LuaJIT/LuaJIT/blob/62e362afbb1d100c892d2782c5862ad18bc464f2/src/vm_x86.dasc#L1379), luajit already uses assembly for `type` function that is pushed to lua, so there won't be any need to modify `lua_type` and `lua_typename` C functions (modified by gmod) for this change to work.
All that needs to be done is just [pushing the assembly version of `type` just like how luajit does it](https://github.com/LuaJIT/LuaJIT/blob/62e362afbb1d100c892d2782c5862ad18bc464f2/src/lib_base.c#L68).

This `type` does it just like how `lua_shared.dll` does it:
image

image


artemking4:
im sorry but why the actual fuck does `type` return `"UserData"` instead of `"userdata"`???

Bagellll:
What is the performance comparison for this and the other `is<type>()` functions?

Srlion:
> What is the performance comparison for this and the other `is<type>()` functions?

There is no comparing here, default `type` function is implemented by luajit in asm, which allows small or jittable functions to be jitted, so it's extremely fast.

Also once this gets in, we can just reimplement all `is` functions to use the new type OR directly use `C_type` to have them be as fast too.

JarosLucky (changes requested):
"UserData" -> "userdata"
Because breakable change behavior

Srlion:
> "UserData" -> "userdata" Because breakable change behavior

Did you even see the screenshot I have added? gmod returns `UserData` not `userdata`, all I did here was to make it 1:1 to gmod implementation, if rubat wants to make it just like Lua then he can, but I can't.

Srlion:
@robotboy655 I'm curious on why was this slapped with "Requires engine changes"? Isn't all modifications by gmod done in Lua only?

artemking4:
As it says in the tag description, this requires internal source code changes to be merged

JarosLucky:
> As it says in the tag description, this requires internal source code changes to be merged

But it isn't. So I think Rubarth probably wants to fix this problem at the root.

JarosLucky:
> As it says in the tag description, this requires internal source code changes to be merged

But it isn't. So I think Rubarth probably wants to fix this problem at the root.

artemking4:
> All that needs to be done is just [pushing the assembly version of `type` just like how luajit does it](https://github.com/LuaJIT/LuaJIT/blob/62e362afbb1d100c892d2782c5862ad18bc464f2/src/lib_base.c#L68).
This requires an internal source code change.

Srlion:
Ah, I misunderstood it. I thought it requires engine changes to push default `type` function lol

Srlion:
> I did some basic testing, it appears to be ~15-30% faster.
> 
> It will have to be done over 2 updates, we need to ensure every server has the Lua part before making changes to the C++ part for the clients.

The point is not to make it itself faster, on main branch it will allow code around it to be jitted, for example
```lua
local function test()
    type("safsa")
    local s = {}
end
```
This won't be jitted because in gmod `type` is a normal C function, so jit aborts the trace one that call happens.
You can test something like
```lua
local function hey()
    os.time()
end

for i = 1, 60 do
    hey()
end
```
on https://luajit.me and select 2.0.4 from right side, you will see it shows red because "NYI", but if you select 2.1 it will be jitted just fine because it allows some NYI, but I don't know how it works internally.

And on the x64 branch, it should solve the issue with code calling type/TypeID (we will have to reimplement the latter in lua to fix it's issue too, but I don't know how it works internally but shouldn't be hard to do)

PR #2202, opened 2025-02-06
Fix middle-clicking links breaking menu

DarthTealc:
In CEF (x86-64), middle-clicking links tries to open them in a new tab, which overrides the current tab, which breaks the menu. This bug is described in https://github.com/Facepunch/garrysmod-issues/issues/6104

This pull request fixes it by preventing middle-click's event from firing in the menu pages.


robotboy655:
This is a hack. I fixed it in-engine. Apparently CEF internals changed that one more callback needed to be watched for opening new tabs.

PR #2203, opened 2025-02-06
TTT: Fix C4/Radio sounds not playing outside of PAS

figardo:
Same issue as #957, but I've made a video to better demonstrate the problem here: https://www.youtube.com/watch?v=6nweYCxRAQ4

If you'd like to reproduce this yourself:
- The map can be downloaded [here](https://steamcommunity.com/sharedfiles/filedetails/?id=3422213891), and the vmf can be downloaded [here](https://drive.google.com/file/d/1Enipyc7vp78z7WyvNjv1RDlnmoR6LLx-/view?usp=sharing).
- Make sure you launch the map c4test in a **multiplayer** game.
- Any sound played within one trollface zone will not be audible from the other trollface zone.
- This is done with hint/skip and areaportals, and as such there are plenty of cases in real maps where this also happens.

Luckily the fix for this has become much simpler since #957 was made, thanks to the _filter_ option that was added to Entity:EmitSound.

This PR fixes 3 things that I feel would benefit most from this change:
- The C4 beep, because dying to a C4 you couldn't hear sucks immensely. (Definitely counts as a C4 nerf unfortunately, maybe tweak C4_MINIMUM_TIME to compensate?)
- Radio sounds, which should make the radio a bit less useless.
- The C4 explosion, to fix cases where people's screens are shaken but no explosion sound is played. (The timer for this sound had to be removed so that the C4 entity can be used before it's removed. As far as I can tell it wasn't there to fix anything, but if it's absolutely necessary to keep it then I can try and find a workaround.)

This _could_ be done with gunshots and death screams as well (maybe dependent on SWEP.IsSilent for gunshots?), but that's a much more radical change to gameplay that I'd rather let @svdm decide on.

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
@@ -249,7 +249,7 @@ function ENT:Explode(tr)
       util.Effect("Explosion", effect, true, true)
       util.Effect("HelicopterMegaBomb", effect, true, true)
 
-      timer.Simple(0.1, function() sound.Play(c4boom, pos, 100, 100) end)
+      self:BroadcastSound(c4boom, 100)
 
       -- extra push
       local phexp = ents.Create("env_physexplosion")
@@ -351,7 +351,7 @@ function ENT:Think()
       end
 
       if SERVER then
-         sound.Play(beep, self:GetPos(), amp, 100)
+         self:BroadcastSound(beep, amp)
       end
 
       local btime = (etime - CurTime()) / 30
@@ -360,7 +360,7 @@ function ENT:Think()
 end
 
 function ENT:Defusable()
-	return self:GetArmed()
+   return self:GetArmed()
 end
 
 -- Timer configuration handlign
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_radio.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_radio.lua
@@ -192,7 +192,8 @@ function ENT:PlayDelayedSound(snd, ampl, last)
          snd = table.Random(snd)
       end
 
-      sound.Play(snd, self:GetPos(), ampl)
+      self:BroadcastSound(snd, ampl)
+
       self.Playing = not last
 
       --print("Playing", snd, last)
@@ -200,10 +201,9 @@ function ENT:PlayDelayedSound(snd, ampl, last)
 end
 
 function ENT:PlaySound(snd)
-   local pos = self:GetPos()
    local this = self
    if simplesounds[snd] then
-      sound.Play(table.Random(simplesounds[snd]), pos)
+      self:BroadcastSound(table.Random(simplesounds[snd]))
    elseif gunsounds[snd] then
       local gunsound = gunsounds[snd]
       local times = math.random(gunsound.times[1], gunsound.times[2])
--- a/garrysmod/gamemodes/terrortown/gamemode/entity.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/entity.lua
@@ -16,3 +16,11 @@ function meta:IsExplosive()
    local kv = self:GetKeyValues()["ExplodeDamage"]
    return self:Health() > 0 and kv and kv > 0
 end
+
+-- Some sounds are important enough that they shouldn't be affected by CPASAttenuationFilter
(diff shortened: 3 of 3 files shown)


mgetJane:
can you check if this fix still works if the c4 wasn't networked to the player?

1. observer stays in the left trollface zone
2. planter plants c4 within the right trollface zone
3. planter walks over to the left trollface zone

if all goes well, then both the observer and planter should hear the c4 beeps, if not then only the planter can hear it

figardo:
> can you check if this fix still works if the c4 wasn't networked to the player?
> 
> 1. observer stays in the left trollface zone
> 2. planter plants c4 within the right trollface zone
> 3. planter walks over to the left trollface zone
> 
> if all goes well, then both the observer and planter should hear the c4 beeps, if not then only the planter can hear it

Seems to work fine.

https://github.com/user-attachments/assets/2c8323ac-8a8c-4494-a096-1e5fec3a4b50

svdm:
I'm not sure it's super relevant, but it's worth noting we do lose the other non-PAS part of CPASAttenuationFilter here, where it [removes recipients outside the audible radius](https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/server/recipientfilter.cpp#L367). With this change every player's client will receive the C4 beep sound messages, even if they subsequently can't hear them. Strictly speaking that's information their client shouldn't have and could be cheated with. Now of course there's many more interesting bits of info Source networks with no way for TTT to stop it, so this is perhaps not worth worrying about. The alternative is re-implementing the attenuation filtering in Lua, because there's no recipient filter that does attenuation filtering without involving PAS.

robotboy655:
There's this:
https://wiki.facepunch.com/gmod/CRecipientFilter:AddPAS

mgetJane:
> There's this: https://wiki.facepunch.com/gmod/CRecipientFilter:AddPAS

afaik this is what emitsound does by default already, which is what causes the issue in the first place (the way source calculates pas is just odd)

figardo:
> I'm not sure it's super relevant, but it's worth noting we do lose the other non-PAS part of CPASAttenuationFilter here, where it [removes recipients outside the audible radius](https://github.com/ValveSoftware/source-sdk-2013/blob/0d8dceea4310fde5706b3ce1c70609d72a38efdf/mp/src/game/server/recipientfilter.cpp#L367). With this change every player's client will receive the C4 beep sound messages, even if they subsequently can't hear them. Strictly speaking that's information their client shouldn't have and could be cheated with. Now of course there's many more interesting bits of info Source networks with no way for TTT to stop it, so this is perhaps not worth worrying about. The alternative is re-implementing the attenuation filtering in Lua, because there's no recipient filter that does attenuation filtering without involving PAS.

Had a go at recreating that attenuation check in lua, but it seems to fall short of the actual audible distance of the sound.
(e.g. the c4 with a timer set to 45 seconds makes the beep sndlvl 58.65, which can normally be heard from ~1110u assuming direct LOS but the attenuation check cuts it off at 865u)

I'm definitely not a source sdk expert so let me know if I've missed something here, but worst case I can just add an extra argument to cap the max audible range manually.

	function meta:BroadcastSound(snd, lvl, pitch, vol, channel, flags, dsp)
	   local rf = RecipientFilter()
	
	   local pos = self:GetPos()
	   local attenuation = lvl > 50 and 20.0 / (lvl - 50) or 4.0 -- https://github.com/ValveSoftware/source-sdk-2013/tree/master/mp/src/public/soundflags.h#L106
	   local maxAudible = 2000 / attenuation -- https://github.com/ValveSoftware/source-sdk-2013/tree/master/mp/src/game/server/recipientfilter.cpp#L409 and https://github.com/ValveSoftware/source-sdk-2013/blob/master/mp/src/public/const.h#L427
	
	   for _, ply in player.Iterator() do
	      if (ply:EyePos() - pos):Length() > maxAudible then continue end
	
	      rf:AddPlayer(ply)
	   end
	
	   self:EmitSound(snd, lvl, pitch, vol, channel, flags, dsp, rf)
	end

mgetJane:
@figardo from my tests, that should be about right

though for some reason, EmitSound always cuts off at 2500 units max regardless of soundlvl, and sound.Play at 2000 units max

(that's only for serverside sounds, there is no cutoff when the sound is played clientside)

so imo do this:

```lua
local maxAudible = math.min(2500, 2000 / attenuation)
```

PR #2205, opened 2025-02-08
Add list.GetEntry

Lexicality:
Dramatically more efficient if you've got a big list and the thing you want might not even be on it

### DIFF
--- a/garrysmod/lua/includes/modules/list.lua
+++ b/garrysmod/lua/includes/modules/list.lua
@@ -13,6 +13,19 @@ function Get( listid )
 
 end
 
+function GetEntry( listid, key )
+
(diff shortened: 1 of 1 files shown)


Zaurzo:
Suggestion; perhaps add an argument that decides whether or not it should return a copy?

Zaurzo:
> > Suggestion; perhaps add an argument that decides whether or not it should return a copy?
> 
> I figured the entire point of the `Get` / `GetForEdit` distinction was that "normal" gets shouldn't be able to mutate the list itself. I suppose I could add `GetEntryForEdit` but at that point you might as well call `list.GetForEdit(list)[key]`

Yeah, that's true

robotboy655 (approved):

PR #2206, opened 2025-02-08
Add "Assert" Global Function

Zaurzo:
A custom assert function that has more features. Unlike the default `assert` function, this function allows you to provide a level to which the error is thrown at. It also has a `noHalt` argument to make the error not halt the code. The varargs are now used to format the error message.

Allowing to provide an error level will allow us to make it actually point to the faulting file, whereas the default `assert` only throws the error at stack level 1, making it seem like we caused the error.

For those with cases that need the error to not halt, or need to format the error message, this function has it built in for convenience, keeping it looking nice and clean.

On a successful assertion, it behaves the same as `assert`, returning all of the arguments.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -438,7 +438,7 @@ function Assert( expression, errorMessage, errorLevel, noHalt, ... )
 
     if ( errorMessage ) then
         if ( ... ) then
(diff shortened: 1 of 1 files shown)


Aws0mee:
Seems like a nice addition, but it might belong better in a 3rd party library imo.

PR #2207, opened 2025-02-09
New Localisation Phrases

DarthTealc:
I've added some new localisation phrases that I believe are general enough to benefit multiple addons/servers.

Explaining some:
- Added plural/singular of some existing phrases, where the plural/singular isn't available elsewhere.
- Added `saves.singular=Save` - There is a `saves=Saves` in main_menu.properties but no singular. I clarified it as a singular of Saves, as this is a singular save file, not the concept of saving "save your game" "save your text" etc.
- Added `dialog.ok=OK` - This one isn't essential, there is already a `preset.okay=Okay` but it's common for dialogs to also use "OK". Some languages may be identical, some may also have a standard of using a shorter form in some dialogs.
- ~Added `ownership.mine=Mine` and `ownership.my=My` - the "My" variant would be useful as a prefix for stuff like "My Props", "My Top Scores", etc.~
- Added `paginate.previous=Previous` `paginate.next=Next` `paginate.first=First` `paginate.last=Last` `browser.back=Back` `browser.forward=Forward` - They're grouped as pagination or browser to avoid ambiguity, and to avoid confusion with the existing player directional values `right=right` `left=left` `forward=forward` `backward=backward`

Some contain #comments. I'm unsure if comments work inline so please remove if they don't.

You can change the keys used if wanted. I think some of the existing keys are vague as to purpose, so having them be a bit more specific can help with the translation.

As pointed out by @Satton2 below, Crowdin's [duplicate setting](https://support.crowdin.com/project-settings/import/#duplicate-strings-management) needs to be changed, otherwise words that mean different things in different languages are impossible to use.


garrysmod/resource/localization/en/community.properties
```diff
@@ -973,3 +976,24 @@ combat=combat
...
+search.none=No Results
+sort.random=Random
+sort.newest=Newest
+sort.oldest=Oldest
+category.featured=Featured
+dialog.ok=OK
+ownership.mine=Mine
+ownership.my=My #eg "My Props" "My Top Scores"
+saves.singular=Save #Singular of main_menu's "saves=Saves". Not the concept of saving, but the concept of a singluar Save.
+fav.addto=Add to Favorites
```
NovaDiablox, line 990: You may want to change it to "favourites". GMod prefers British English.

garrysmod/resource/localization/en/community.properties
```diff
@@ -973,3 +976,24 @@ combat=combat
...
+fav.removefrom=Remove from Favorites
```
NovaDiablox, line 991: Same here, British English.

NovaDiablox (changes requested):

garrysmod/resource/localization/en/community.properties
```diff
@@ -973,3 +976,24 @@ combat=combat
 technology=technology
 cancel=cancel
 direct=direct
+search.results=Results
+search.result=Result
+search.none=No Results
+sort.random=Random
```
Satton2, line 982: Hint: look at the big comment below first.

The same "master string" issue.

Satton2 (changes requested):

NovaDiablox:
I remember changing "Favorites" to "Favourites" as Rubat requested. It seems that the changes I made were reverted. I agree on the cases of mixed British & American English strings.

Satton2:
> For Result/Results: I wasn't intending to use it with a number, but I expect it would get used that way. My use would've been in the non-specific number case, eg a "Results" heading, or listing with the search term eg "'rollermine' results". In a sentence it could be used like "some results" "view results" "a handful of results" etc. 

Sadly, it still depends on the context. Let's take "результат<ins>ы</ins>" as a base translation once again:

- Rollermine's results → Шаровая мина результаты ❌Результаты шаровая мина ✅ (_note: rollermine's affixes will also need to be changed in order to look right, but it usually okay in case of player name if it's possibly or intended to be a player name here_) — the main part here is that you have to move the "results" word before the target's name to make it right in Russian.
- some results → какие-то результаты ✅
- view results → просмотреть результаты ✅ (BUT: it depends on how the "view" word will be used and if it be localized as "просмотр" then the proper localization will look like: просмотр результат<ins>ов</ins>)
- a handful of results → несколько результат<ins>ы</ins> ❌→ несколько результат<ins>ов</ins> ✅

> I assume Russian has a way of presenting "results" when the specific number is unknown? And given Gmod's use of "entities" "maps" "players" etc, I'm guessing Russian translation is already using the "non-exact number" version in those cases?

Maybe not in the way you might think. If we talk about tabs, sections, menus, titles and there is no presented context, you are fine with "результаты", but if the context presents and if there meant to be some number it may (and probably will have to) change its affix. Even if we talk about counters or some summary it may change it affix, like in "Results:" it can be either "Результат<ins>ов</ins>:" (counter) or "Результат<ins>ы</ins>:" (e.g. end of the round results).

"Entities" is one of the exceptions, since it was transliterated to "энтити" (for GMod specifially) in a way that it doesn't get affected by any affixes and its only form is "энтити" even if it's singular, while in English it would be "entity".

---

Regarding `community.properties` localization file, I am still really worried about its quality since I believe it needs a big review and every single term should have its strict context otherwise 60% (or more) of its content will be messed up in one way or another. Since there is no context for many of them, translators will and do interpret them in their own way, and when addon developers come with the same approach, it results in a total mess. 😔 (and the duplicate strings setting makes it even worse)

robotboy655 (approved):

Satton2:
@robotboy655 
Hello!

Is there any chance you will look into Crowdin's [duplicate setting](https://support.crowdin.com/project-settings/import/#duplicate-strings-management)? 👀 Context to that can be found [here](https://github.com/Facepunch/garrysmod/pull/2207#discussion_r1948109775).

robotboy655:
It was already disabled once. I think it was re-enabled by someone else at the company, probably for money savings purposes. If there's a particular string you think should be deduplicated, I think it can be done. You can email me a list if you wish or something.

PR #2210, opened 2025-02-14
Cleaner server list layout code & Better display on small screens

DarthTealc:
I've updated the server list to use a 2012-era webkit box flow (precursor to today's flexbox) instead of using position:absolute on hundreds of elements. This should be beneficial for performance, and it makes it much easier to add/remove/size columns. 
I'm using the outdated webkit box model as it works in Awesomium and CEF, though once Awesomium is gone it should be updated with the actual modern flexbox. Both are still better than using position:absolute everywhere with hardcoded positions at different breakpoints.

For users on regular sized displays, they shouldn't notice much visual difference.

For users on very small displays such as 800x600, I've improved the layout to make servers readable.

Example of the small display improvement. Current game (left) vs this update (right)
image

In-game screenshot of this update, 800x600 [note: see below, this now makes use of more space]
image

~I didn't extend the server list into the player list area but think it would be a good idea for a future update.~
I've made use of the server info space (player list) by allowing users to close it. See more details: https://github.com/Facepunch/garrysmod/pull/2210#issuecomment-2668289496
[Example](https://github.com/user-attachments/assets/62e5f3bf-29db-431b-aa48-288421fdf463) at 800x600

Please ask if you're unsure why I made a particular change, or if you want me to make any adjustments to it.


darkfated:
Good job, buddy. I checked it myself and found no problems.

robotboy655 (commented):
I am not a fan of the close button styling.

Also do you think we can solve the vertical alignment issues? I understand it's not a regression in this PR, but since we are touching the styles, might as well try to solve this.

Here are some examples:
image
image

PR #2211, opened 2025-02-14
[TTT] Update the Magneto stick to use the HL2DM viewmodel used in sandbox that has c_hands support

SvveetMavis:
Changes the magneto stick to use the hl2dm stunstick model instead of the hl2beta model so that it has proper c_hands support 

before 
https://github.com/user-attachments/assets/abf71291-c7fa-440a-8051-454b944ce64d

after
https://github.com/user-attachments/assets/64694194-4db8-44d3-9837-55d4e676c538

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_zm_carry.lua
@@ -12,13 +12,15 @@ if CLIENT then
 
    SWEP.DrawCrosshair       = false
    SWEP.ViewModelFlip       = false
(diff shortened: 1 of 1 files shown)


mgetJane:
i personally quite prefer the current viewmodel, that other one looks too different from it imo and its animations dont match the magneto stick's delays, it's way too fast

though, you could probably hack the hl2dm viewmodel to look very close to the current viewmodel with the cycle speed and viewmodel offset functions, would require a bunch of trial and error

SvveetMavis:
my main thing with it is seeing the hev suit hand just like really bugs me lol, but yeah somewhere in between would be much better

PR #2212, opened 2025-02-15
[TTT] Change holdtypes for consistency with viewmodels

SvveetMavis:
changes the holdtypes of the pistol weapons to from "pistol" to "revolver" to better match what is shown in the view model

before 
Desktop Screenshot 2025 02 14 - 16 12 01 60

after
Desktop Screenshot 2025 02 14 - 17 22 56 61

### DIFF
--- a/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_flaregun.lua
+++ b/garrysmod/gamemodes/terrortown/entities/weapons/weapon_ttt_flaregun.lua
@@ -1,7 +1,7 @@
 
 AddCSLuaFile()
 
-SWEP.HoldType              = "pistol"
+SWEP.HoldType              = "revolver"
 
 if CLIENT then
    SWEP.PrintName          = "flare_name"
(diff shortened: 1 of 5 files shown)


SvveetMavis:
the normal "pistol" viewmodel it uses was made for hl2 usp which has a one handed viewmodel so it just feels weird to use that with the 2 handed guns

mgetJane:
personally seems fine imo, though here's some unforeseen consequences of this:

- playermodel viewed from the front is a little narrower, so you're a bit less likely to get hit
- your arms also now cover more of your torso when viewed from the front, which acts a bit like armour because arms take about half as much damage than the torso
- there's some items on the workshop that use the deagle model for the purpose of being deliberately indistinguishable from the vanilla deagle, which imo is a little silly but regardless this pr would make those distinguishable from the vanilla deagle
- i think many ppl aesthetically prefer the one-handed anims

Malivil:
The workshop items thing will probably be the biggest issue, though it's easy enough to write a quick addon that fixes that using the `PreRegisterSWEP` hook if authors don't update their weapons.

svdm:
I'm fairly sure the one-handed version did not even exist when I originally set the holdtypes, so the two-handed version is more accurate to the intention there.

mgetJane:
btw something to note is that the revolver holdtype doesn't look exactly like the old pistol holdtype (but close enough i guess)

for the revolver holdtype, the player stands up straight, torso facing forward, and both arms stretched far forward

for the old pistol holdtype, the player is a bit hunched, torso facing to the right, and left arm bent a lot more than the right arm

screenshots from some old ttt vid:

image

image

image

image

PR #2214, opened 2025-02-16
Sort favourite servers at the top of the server list

DarthTealc:
Favourited servers will now show at the top of the server list. They will still follow the sort options, but within that favourites grouping. 
A small divider is shown between the last favourite server and the first non-favourite server to assist visually, however the servers are still all part of the same listing so there isn't an issue with multiple scrollbars or anything like that.

Works in Awesomium and CEF86.

image

### DIFF
--- a/garrysmod/html/css/menu/Servers.css
+++ b/garrysmod/html/css/menu/Servers.css
@@ -245,6 +245,11 @@ H1 small
 	background-color: #F1F6FF;
 }
 
+.serverlist .server.fav + .server:not(.fav)
+{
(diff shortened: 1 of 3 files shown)


robotboy655:
Why do you think this is necessary? What's wrong with having them in the favorites tab?

robotboy655:
I don't think this belongs in the base game. If you wish to see your favourite servers - you go to the favorites tab. If you want to find a new server you go to the Internet tab, and listing favourite servers at the top feels counterproductive to that goal.

PR #2220, opened 2025-02-19
TTT: Fix pistols using revolver reload gesture

figardo:
Since #2212 changed the pistol holdtypes to be "revolver", the pistols now perform a revolver reload gesture complete with audible revolver-y sounds.

I've added an option to decouple the reload gesture from the weapon's holdtype and used it on the pistols that shouldn't be reloaded like a revolver.


SvveetMavis:
cool as hell

PR #2221, opened 2025-02-19
TTT: Close player volume slider along with scoreboard

figardo:
If the player volume slider is open when the scoreboard closes, it will remain stuck on the screen until you either mess with the slider or wait 10 seconds for it to autoclose.

This adds a hook to make it close when letting go of tab.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
@@ -173,7 +173,7 @@ function PANEL:SetPlayer(ply)
 
    self.voice.DoRightClick = function()
       if IsValid(ply) and ply != LocalPlayer() then
-         self:ShowMicVolumeSlider()
+         self.voice.volume = self:ShowMicVolumeSlider()
       end
    end
 
@@ -320,14 +320,6 @@ function PANEL:DoRightClick()
    menu:Open()
 end
 
-local frame
-local function HideVolume()
-   if IsValid(frame) then
-      frame:Close()
-   end
-end
-hook.Add("ScoreboardHide", "TTT_HideVolume", HideVolume)
-
 function PANEL:ShowMicVolumeSlider()
    local width = 300
    local height = 50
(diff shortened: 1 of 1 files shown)


garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -408,9 +408,16 @@ function PANEL:ShowMicVolumeSlider()
          )
          draw.DrawText(textValue, "cool_small", sliderHeight / 2, -20, Color(255, 255, 255, 255), TEXT_ALIGN_CENTER)
       end
-      
+
       draw.RoundedBox(100, 0, 0, sliderHeight, sliderHeight, Color(255, 255, 255, 255))
    end
 end
 
+local function HideVolume()
```
Malivil, line 416: If you move this above `PANEL:ShowMicVolumeSlider`, you can use it inside the timer as well to reduce code duplication.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -320,6 +320,13 @@ function PANEL:DoRightClick()
    menu:Open()
 end
 
+local frame
```
robotboy655, line 323: This approach with the global variable does not handle situation where multiple voice volume panels are open.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
```diff
@@ -350,7 +357,7 @@ function PANEL:ShowMicVolumeSlider()
    end
 
    -- Automatically close after 10 seconds (something may have gone wrong)
-   timer.Simple(10, function() if IsValid(frame) then frame:Close() end end)
+   timer.Create("TTT_CloseVolumeSlider", 10, 1, HideVolume)
```
robotboy655, line 360: This is a pointless change that also is a regression, because with this only 1 panel will close in 10 seconds if multiple are opened, since only 1 timer with given name can exist.

robotboy655 (changes requested):
This doesn't really address the issue adequately, and causes regressions.

garrysmod/gamemodes/terrortown/gamemode/vgui/sb_row.lua
figardo: For the record, this was done so that a timer created for a previous volume panel would not close any future volume panels too early. Regardless, this can be reverted now that each volume panel is localized again.

robotboy655 (approved):

PR #2222, opened 2025-02-23
`math`: Optimize `Distance`, `Clamp`, `Approach`

noaccessl:
`Distance`: `math.sqrt( ... )` —> `( ... ) ^ 0.5`

image
<details> <summary>Benchmark Code</summary>

```lua

times = 64

print( Format( 'branch => %s\niterations => %i', BRANCH, times ) )

function main()

	print( Format( '\njit status => %s', jit.status() ) )

	sum = 0
	avg1 = 0

	local sqrt = math.sqrt

	for i = 1, times do

		util.TimerCycle()

			local x = sqrt( 5 )

		sum = sum + util.TimerCycle()

	end

	avg1 = sum / times
	print( Format( 'math.sqrt\n\t%.8f ms on average (100%%)', sum / times ) )

	sum = 0
	avg2 = 0

	for i = 1, times do

		util.TimerCycle()

			local x = 5 ^ 0.5

		sum = sum + util.TimerCycle()

	end

	avg2 = sum / times
	print( Format( 'pow operator\n\t%.8f ms on average (%.1f%%)', sum / times, ( avg2 * 100 ) / avg1 ) )

end

jit.off()
main()

jit.on()
main()

```
</details>

`Clamp` & `Approach`: Replaced functions with operators, which are faster in action.

image
image

Note: Approach becomes a bit slower with enabled JIT if used more than once at the moment (applies to the default branches, but not to `x86-64`)

### DIFF
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -54,7 +54,8 @@ function math.Clamp( num, low, high )
 	if ( num < low ) then return low end
 	if ( num > high ) then return high end
 
-	return num
+	-- If the passed number is NaN, return the lowest limit
+	return num ~= num and low or high
 end
 
 --[[---------------------------------------------------------
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -51,11 +51,13 @@ end
 	Desc: Clamp value between 2 values
 ------------------------------------------------------------]]
 function math.Clamp( num, low, high )
-	if ( num < low ) then return low end
-	if ( num > high ) then return high end
+	if ( num < low or num ~= num --[[NaN-check]] ) then
+		return low
+	elseif ( num > high ) then
+		return high
+	end
 
-	-- If the passed number is NaN, return the lowest limit
-	return num ~= num and low or num
+	return num
 end
 
 --[[---------------------------------------------------------
--- a/garrysmod/lua/includes/extensions/math.lua
+++ b/garrysmod/lua/includes/extensions/math.lua
@@ -20,7 +20,7 @@ end
 function math.Distance( x1, y1, x2, y2 )
 	local xd = x2 - x1
 	local yd = y2 - y1
-	return math.sqrt( xd * xd + yd * yd )
+	return ( xd * xd + yd * yd ) ^ 0.5
 end
 math.Dist = math.Distance -- Backwards compatibility
 
@@ -47,11 +47,14 @@ function math.IntToBin( int )
 end
 
 --[[---------------------------------------------------------
-	Name: Clamp( in, low, high )
+	Name: Clamp( num, low, high )
 	Desc: Clamp value between 2 values
 ------------------------------------------------------------]]
-function math.Clamp( _in, low, high )
-	return math.min( math.max( _in, low ), high )
+function math.Clamp( num, low, high )
+	if ( num < low ) then return low end
+	if ( num > high ) then return high end
+
+	return num
 end
 
 --[[---------------------------------------------------------
@@ -172,11 +175,13 @@ end
 
 function math.Approach( cur, target, inc )
 	if ( cur < target ) then
-		return math.min( cur + math.abs( inc ), target )
+		cur = cur + ( inc < 0 and -inc or inc )
+		return cur < target and cur or target
 	end
 
 	if ( cur > target ) then
-		return math.max( cur - math.abs( inc ), target )
+		cur = cur - ( inc < 0 and -inc or inc )
+		return cur > target and cur or target
 	end
 
 	return target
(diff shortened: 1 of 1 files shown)


Astralcircle:
New math.Clamp function does not handle nan properly:
```lua
print(math.Clamp(0 / 0, 1, 5), math.Clamp2(0 / 0, 1, 5))
```
```
1	nan
```

noaccessl:
Now it should

JarosLucky:
Update benchmarks then)

goodusername123:
replacing `math.sqrt` with `^ 0.5` seems like a really really bad idea future proofing wise considering that in modern versions of LuaJIT the performance characteristics of these are somewhat wildly different then they are in the versions of LuaJIT GMod currently uses.
https://www.freelists.org/post/luajit/Bad-performance-of-x05-versus-mathsqrtx
https://github.com/LuaJIT/LuaJIT/issues/684#issuecomment-1020116837
https://github.com/Facepunch/garrysmod/pull/2198#issuecomment-2626726754

noaccessl:
Ok, I have figured out some balanced optimization for `math.Clamp`. Decided to replace two conditions with one with `elseif`-element. Also, moved NaN-check to the first condition. 

# Previous
```lua
function math.Clamp( num, low, high )
	if ( num < low ) then return low end
	if ( num > high ) then return high end

	-- If the passed number is NaN, return the lowest limit
	return num ~= num and low or num
end
```
luac.nl
image
luajit.me
image

# New
```lua
function math.Clamp( num, low, high )
	if ( num < low or num ~= num --[[NaN-check]] ) then
		return low
	elseif ( num > high ) then
		return high
	end

	return num
end
```
luac.nl
image
luajit.me
image

# Benchmark
[Benchmark I use](https://gist.github.com/noaccessl/2336bb390aca15bc7ba50d2c9d966daa)

<details> <summary>Benchmark code</summary>

```lua
local min, max = math.min, math.max

local function Clamp_mathlib( num, low, high )

	return min( max( num, low ), high )

end

local function Clamp_prev( num, low, high )

	if ( num < low ) then return low end
	if ( num > high ) then return high end

	-- If the passed number is NaN, return the lowest limit
	return num ~= num and low or num

end

local function Clamp_new( num, low, high )

	if ( num < low or num ~= num --[[NaN-check]] ) then
		return low
	elseif ( num > high ) then
		return high
	end

	return num

end

local Tests = {

	'Clamp_mathlib', function()

		Clamp_mathlib( -1, 0, 5 )
		Clamp_mathlib( 0 / 0, 0, 5 )
		Clamp_mathlib( 10, 0, 5 )
		Clamp_mathlib( 3, 0, 5 )

	end;

	'Clamp_prev', function()

		Clamp_prev( -1, 0, 5 )
		Clamp_prev( 0 / 0, 0, 5 )
		Clamp_prev( 10, 0, 5 )
		Clamp_prev( 3, 0, 5 )

	end;

	'Clamp_new', function()

		Clamp_new( -1, 0, 5 )
		Clamp_new( 0 / 0, 0, 5 )
		Clamp_new( 10, 0, 5 )
		Clamp_new( 3, 0, 5 )

	end;

}

local Frames		= 1000
local Digit 		= 4
local Iterations	= 30

SetupBenchmark( Frames, Digit, Iterations, 'us', 'average' )

local function RunTests()

	for i = 1, #Tests, 2 do
		Benchmark( Tests[i], Tests[i + 1], i == 1 --[[The first test will be considered as standard]] )
	end

end

jit.off()
RunTests()

jit.on()
RunTests()
```
</details>

# Results
## Pre-Release Branch
image

## x86-64 Branch
image

---

Waiting for feedback.

IVogel:
Used same benchmark as you, and your parameters and got these results :/

_Windows 10 x86_64, Ryzen 5 3600_
 Dev branch  |  x86-64 branch, 64bit  |  x86-64 branch, 32bit 
:--------------:|:-------------------------:|:-------------------------:
benchmark-dev | benchmark-x86-64-64 | benchmark-x86-64-32

JarosLucky:
Windows 10 x86_64, Intel I5 3550
x86-64 branch, 64bit
image
Since the x86-64 branch uses the updated LuaJIT with Stitching, I don't care about the results with JIT disabled as a measure of efficiency. After all, nobody uses JIT disabling besides special cases, besides functions that could not compile with JIT

JarosLucky:
If you look at your own benchmark result, you can see that the original feature is already working quite well. It makes no sense to optimize the version without jit when regressing performance with jit. **My opinion this change should definitely be brought back. And the others are measured again, as the benchmark attached originally has questionable validity**

JarosLucky:
> # Benchmark
> [Benchmark I use](https://gist.github.com/noaccessl/2336bb390aca15bc7ba50d2c9d966daa)

Also, it seems you need to rewrite your benchmark to measure performance without JIT because the change in execution speed between with and without JIT differs by 76.67 times

noaccessl:
Ok, I've reworked my benchmark. 

`math.sqrt( x )` is more stable in speed than `x ^ 0.5` (especially on the `x86-64`). And the previous `math.Clamp` & `math.Approach` was handled by JIT well enough.

Overall, the initial functions are well-optimized for JIT; and the majority doesn't turn it off globally. Was a bit presumptuous with my optimizations as I was prioritizing the results with disabled JIT.

PR #2223, opened 2025-02-24
Fix DButton:SetConsoleCommand() not accepting vararg

Kari-Jukka:
Makes variable arguments in DForm:Button() function properly.

### DIFF
--- a/garrysmod/lua/vgui/dbutton.lua
+++ b/garrysmod/lua/vgui/dbutton.lua
@@ -126,10 +126,18 @@ function PANEL:PerformLayout( w, h )
 
 end
 
-function PANEL:SetConsoleCommand( strName, strArgs )
+function PANEL:SetConsoleCommand( strName, strArg, ... )
+
+	if ( select( "#", ... ) > 0 ) then
+		local extraArgs = { ... }
+		self.DoClick = function( slf, val )
+			RunConsoleCommand( strName, strArg, unpack( extraArgs ) )
+		end
+		return
+	end
 
 	self.DoClick = function( slf, val )
-		RunConsoleCommand( strName, strArgs )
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dbutton.lua
```diff
@@ -126,10 +126,18 @@ function PANEL:PerformLayout( w, h )
...
+		local extraArgs = { ... }
+		self.DoClick = function( slf, val )
+			RunConsoleCommand( strName, strArg, unpack( extraArgs ) )
+		end
+		return
+	end
 
 	self.DoClick = function( slf, val )
-		RunConsoleCommand( strName, strArgs )
+		RunConsoleCommand( strName, strArg )
```
robotboy655, line 140: Um, `RunConsoleCommand( strName, ...)` and `PANEL:SetConsoleCommand( strName, ... )` Instead?

Kari-Jukka: No; the vararg would be outside the scope, and it would error. That's why you need to wrap it inside an array and unpack it again. You can thank the compiler for that.

Anyways, sorry for typo!

robotboy655: Surely its achievable without the `select` call at all though?

Kari-Jukka: It's a memory optimization. 99% of the buttons won't be utilizing varargs anyways. You may also avoid vararg completely by having fixed extra arguments `a, b, c, d, e, f, g` and so on. But that's not what the documentation says.
See [DForm:Button](https://wiki.facepunch.com/gmod/DForm:Button). Currently the varargs here do nothing at all.

robotboy655 (approved):

PR #2224, opened 2025-03-01
Loading screen cleanup/tweaks

DarthTealc:
I've made a few tweaks to the loading screen html/css. 

Mostly to clean up some code, remove redundant bits, etc. Developed and tested in Awesomium to ensure compatibility.

For visual changes:
- If the server name is too long to fit the user's screen width, it will now use "..." ellipsis at the end of the screen instead of a hard cut
- ~The gamemode name will have the gamemode icon next to it. If the icon does not exist it will just have no icon, which is how it is currently. Having this in the loading screen shows developers how they can easily do it on their own loading screen, and having it on the right means no change to text alignment.~

### DIFF
--- a/garrysmod/html/loading.css
+++ b/garrysmod/html/loading.css
@@ -150,17 +150,5 @@ body.dark
 	height: 128px;
 	background: #666;
 	margin-right: 16px;
-}
-.server_info #serverGamemode img
-{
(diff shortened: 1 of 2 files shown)


DarthTealc:
Fixed the padding and removed the gamemode icon

PR #2230, opened 2025-03-05
`env_skypaint` I/O (incl. `color255` inputs)

piqey:
A hacky implementation of `ENT:AcceptInput( ... )` to proxy incoming I/O traffic to existing `ENT:KeyValue( ... )` data table logic.

Historically, `AddOutput` would be used by mappers to change keyvalues on `env_skypaint`, but any `AddOutput` setup requires static string parameter overrides, thus barring them from utilizing entities such as `math_remap`, `math_colorblend` or any other logic entities that dynamically output their own data (`float`, `vector`, `color255`).

These changes keep the old `AddOutput` behavior, but also add `SetVariableName` inputs for all fields, including `color255` inputs for all traditionally `color1` fields (in the form of `SetColorField255` vs. `SetColorField`, both existing and functional).

Given the hacky nature of the implementation, any of `env_skypaint`'s KV pairs accessible via `ENT:SetNetworkKeyValue( ... )` can be set this way, but only the variables that make sense to modify in this manner have been added to the FGD as inputs.

### DIFF
--- a/bin/garrysmod.fgd
+++ b/bin/garrysmod.fgd
@@ -30,7 +30,7 @@
 [
 	topcolor(color1)		: "Sky Top Color" : "0.2 0.5 1.0" : "The colour of the top of the sky."
 	bottomcolor(color1)		: "Sky Bottom Color" : "0.8 1.0 1.0" : "The colour of the bottom of the sky."
-	fadebias(float)			: "Sky Fade Bias" : "1.0" : "Controls the bias of the fade between top/bottom. (1.0 is even)"
+	fadebias(float)			: "Sky Fade Bias" : "1.0" : "Controls the bias of the fade between top/bottom (1.0 is even)."
 
 	sunsize(float)			: "Sun Size" : "2.0" : "Controls the size of the sun glow."
 	sunnormal(vector)		: "Sun Normal" : "0.4 0.0 0.01" : "The position of the sun, expressed as a normal from the center of the world."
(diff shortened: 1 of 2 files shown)


piqey:
> OK, I like this, but let's not have this hacky implementation on the Lua side. Please make it be explicit, which could also be a good example for modders.
> 
> An example can be found here: https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/base/entities/entities/lua_run.lua#L56-L63

I'll look into this.

> Can you also please elaborate on the necessity of `color1` and `color255` inputs?

I'm not sure if you mean the necessity of including them both, or the necessity of having an additional `color255` input. I'll include both explanations.

For `color1`, I don't necessarily even like `color1`, but figuring as how it's the way the system's been implemented it seems right to have it.

Concerning `color255`, though, it's hard for me to know where to begin—there's a lot of examples I could think of. Let's say, for example, I want to create some form of transition between sky colors over time in my map. The `ENT:KeyValue`/`AddOutput` implementation is already there and functional, but all that does is let me run an output from some arbitrary entity to the `env_skypaint` entity where I have to manually type in the value I want as part of the parameter override:

image

If I want to create a smooth transition from one color to another, I have no way to do this other than:
  1. A Lua script
  2. An elaborate system of precomputed numeric transitions generated by external tools that add up to somewhere over 200-300 `AddOutput` calls triggered over 9 seconds (looks great by the way, and it still can't change the color)
  image
  3. Add a Lua API for `color255`, as `math_colorblend` does exactly what I want (but there isn't a Source engine entity on the planet that uses `color1` for the I/O system)

`AddOutput` is great, I use it all the time, but it's functionally impossible to pass anything to it because it necessitates you to include the key you want to set within the parameter override. This problem, combined with the fact that `color1` is esoteric to Source engine entity I/O, is my impetus behind making the pull request.

robotboy655:
I have implemented this in a better way here:

https://github.com/Facepunch/garrysmod/commit/409266b0c43a44c2b1967aa8b9ba3f3f091b65f8

This should cover all the inputs you wanted to add, with additional benefit of making it much easier to add Map Input support for other entities, including for addons.

PR #2232, opened 2025-03-11
Localized "Menu failed to load"

IBRS-4Ever:
Screenshot 2025-03-11 121357


IBRS-4Ever:
How about add a button to open Lua Menu by Rubat?
https://github.com/robotboy655/gmod-lua-menu

robotboy655:
I don't think this needs localization, ideally you'd never see this menu, it is only provided because of Linux 64bit builds, so you can at least quit the game. It should be removed at some point in the future.

Maybe options/quit could reuse the already existing localizations though.

PR #2235, opened 2025-03-12
TTT: Crosshair Customisation

figardo:
Based on #1376, but with additional changes:

# Improve crosshair alignment

- ttt_crosshair_thickness 1 - 10
- mat_antialias 4

top-left of red dot = centre of screen

By comparison, the old PR implementation had a tendency to leave a larger gap in the bottom right than the top left when using odd thickness levels (3 here):

ttt_misaligned

# Ensure there's no overlap if thickness is set too high

- ttt_crosshair_thickness 1 (mat_antialias 1) vs 5 vs 10

When using the H.U.G.E-249, the distance between the crosshair prongs remains 18x18 no matter what thickness level is set.

Worst case scenario from the old PR (thickness 20):

ttt_overlap

All fixed:

ttt_nooverlap

# Use surface.DrawLine if thickness is 1 to preserve its behaviour when antialiasing is enabled

- mat_antialias 4 vs mat_antialias 1
- ttt_crosshair_thickness 1

ttt_crosshair_outlinethickness adjusts for the extra pixel caused by antialiasing

# Remove any settings that can tighten the crosshair gap/scale

New settings section:

image


figardo:
> I think it would be nice if you could integrate this into the default GMOD crosshair customization feature and replace instances of the TTT crosshair in the TTT game mode to use only the GMOD crosshair customization options. You could leave a separate checkbox for using role-based colors (Inno/T/DT).

If we used the GMod crosshair customisation settings while preserving the "no tightening the crosshair gap" rule, we'd have to ignore the settings on the Classic option for gap, dot, and maybe the T-style, as well as outright ignoring the Dot Image and HL2 options. This arrangement would just cause confusion over why the custom crosshair looks different in TTT from that of the preview in the customisation menu.

I also feel like most people would prefer to keep their sandbox crosshair settings separate from their TTT crosshair settings in general, since they are very different gamemodes.

figardo:
> One bug I noticed testing this PR in my local server is that the Newton Launcher does not follow the crosshair customization.

Thank you, fixed now.

> One suggestion: I believe you could take this opportunity and enable outline thickness by default. Helps visibility.

While I also prefer crosshair outlines, I think it's best to keep the crosshair behaviour the same as it has been by default.

> Also, could you add an option to disable the crosshair gap/size changing when shooting?

This was one of the settings removed for tightening the crosshair gap, but I'm open to readding those settings if @svdm thinks they wouldn't mess with weapon balancing too much.

> Not sure if this is possible, but could you make it so the spectator crosshair (when not spectating anyone) matches the TTT-defined crosshair instead of using the game's crosshair?

It's possible, but it requires moving the crosshair code out of the weapon base into the gamemode files somewhere. I'll see if there's an elegant solution.

figardo:
> In other Source games that support crosshair outlines, it is enabled by default. Most notably Counter-Strike games if I am not mistaken.

We're adding to a gamemode that already has an established crosshair, so keeping that existing crosshair as the default would make the most sense over forcing change upon everyone after 15 years.

> Do you mean that what I described (not changing the gap while shooting) is now the default behavior and not toggle-able? If so, does this also stop the crosshair changing sizes depending on the held weapon?

If the gap isn't changing when you shoot, try testing in a multiplayer server instead of a singleplayer server, since WEAPON:LastShootTime doesn't work in singleplayer. Both gap changes still work as usual on my end.

figardo:
Ah I get you now, to be clear I'm refraining from adding anything that can tighten the crosshair gap, including options to remove the gap increase when shooting or the gap increase with higher bullet spread weapons, because a tighter crosshair gap could potentially have an impact on weapon balance.

If this is determined to be a non-issue then I can re-add all of the omitted settings, including a toggle for a static crosshair.

figardo:
> If Rubat agrees, the crosshair settings needs its own tab for ease of access next to the current "Settings" tab or somewhere else.

I did consider that the new section makes the settings tab a bit long, but it feels weird to relegate new settings to a different tab just because there's a lot of them.

> I think the word "crosshair" seems repetitive. I advise something more in line with CS2 crosshair settings.

Sure thing, here's what it looks like without the repeat mention of "Crosshair" (except for disable, cause it feels weird to just say "Disable completely" imo):

image

garrysmod/gamemodes/terrortown/gamemode/lang/english.lua
```diff
@@ -244,7 +244,7 @@ L.set_tips      = "Show gameplay tips at the bottom of the screen while spectati
 L.set_startpopup = "Start of round info popup duration"
 L.set_startpopup_tip = "When the round starts, a small popup appears at the bottom of your screen for a few seconds. Change the time it displays for here."
 
-L.set_cross_opacity   = "Ironsight crosshair opacity"
+L.set_cross_opacity   = "Ironsight opacity"
```
NovaDiablox, line 247: "Iron sight opacity"

garrysmod/gamemodes/terrortown/gamemode/lang/english.lua
```diff
@@ -244,7 +244,7 @@ L.set_tips      = "Show gameplay tips at the bottom of the screen while spectati
 L.set_startpopup = "Start of round info popup duration"
 L.set_startpopup_tip = "When the round starts, a small popup appears at the bottom of your screen for a few seconds. Change the time it displays for here."
 
-L.set_cross_opacity   = "Iron sight crosshair opacity"
+L.set_cross_opacity   = "Iron sight opacity"
```
robotboy655, line 247: I feel like this is misleading now, it makes no sense for "iron sight opacity" is a nonsensical statement. Maybe "ADS opacity"  would be better, but it could be more difficult to localize. Maybe just revert this string to what it was.

figardo: It only makes sense in the context of being located underneath crosshair settings along with the other settings (see the last image i posted), but I agree that it feels awkward when compared with the other opacity option.

Maybe swapping the order to "Opacity (iron sights)" would fit better with the other option?

NovaDiablox: > It only makes sense in the context of being located underneath crosshair settings along with the other settings (see the last image i posted), but I agree that it feels awkward when compared with the other opacity option.
> 
> Maybe swapping the order to "Opacity (iron sights)" would fit better with the other option?

I also agree to revert the changes in this one. Translators can use the mentioned examples when translating but it won't be consistent.

svdm:
Seems good as-is, not allowing closing of the gap is the right choice.

PR #2236, opened 2025-03-13
hook instead of gamemode function

bailmast:
1. `PlayerCanJoinTeam` probably should be called as a hook instead of being a gamemode function
2. Fix errors when client provides a nil argument in changeteam command
3. Get rid of funny comments duplicating code logic


bailmast:
And probably functions like [OnPlayerChangedTeam](https://wiki.facepunch.com/gmod/GM:OnPlayerChangedTeam) should have tag "internal" on wiki

robotboy655:
Do you have a use case for it to be called as a hook though?

I am struggling to think of a scenario where an addon should have control over what team(s) can be joined, and that's provided the gamemode would use the base gamemode hook and not their own system.

Any sort of team player limits or things like that would surely need to enforced by the gamemode?

Do you actually need this to be added, or is this a "just because" idea?

bailmast:
So, I'm currently working on something, and I think it's important for new gamemodes to have that standard to stick with. For example, if someone is creating a new admin system, they will think, 'Ah, it's not an internal hook, so I can use it for punishments,' or they might create mini games unrelated to the gamemode's purpose.

PR #2237, opened 2025-03-13
PlayerEditedVariable

bailmast:
Call a hook after a variable was actually changed


robotboy655:
What is this useful for?

private-effort:
I don't think anyone except me will ever use it, so I guess it's not useful at all, or just easy to override.

robotboy655:
Well what do you want to use it for?

private-effort:
I've been using similar addition to store history of such changes to undo them using spawnmenu.

PR #2239, opened 2025-03-15
Optimize widgets (but better)

thecraftianman:
Based on (and supersedes) #915
Closes [#2902](https://github.com/Facepunch/garrysmod-issues/issues/2902)

In addition to the changes made in the original PR, the following improvements have been made:
- The filter function now uses a whitelist table instead of a function
- The hooks handling widget processing are only added when widgets are in use
- `widgets.RenderMe` is no longer necessary (its logic has been optimized and moved into the rendering hook directly) and has been stubbed out. I'm not sure if any addons that call this function even exist, so any suggestions on whether it should be kept functional, removed entirely, or kept stubbed out would be appreciated
- All widget entities have had some amount of extra optimization done (mostly just reusing variables more often)

The widget system still won't be particularly performant, but it runs at least decently faster now, and manually removing the system entirely should no longer be necessary for a performance boost.

### DIFF
--- a/garrysmod/lua/entities/widget_bones.lua
+++ b/garrysmod/lua/entities/widget_bones.lua
@@ -3,6 +3,7 @@ AddCSLuaFile()
 
 local matBone = Material( "widgets/bone.png", "smooth" )
 local matBoneSmall = Material( "widgets/bone_small.png", "smooth" )
+local c = Color( 255, 255, 255, 255 )
 
 local widget_bone = {
 	Base = "widget_base",
@@ -25,7 +26,6 @@ local widget_bone = {
 		local pp = self:GetParentPos()
 		if ( !pp ) then return end
 
-		--local fwd = self:GetAngles():Forward()
 		local len = self:GetSize() / 2
 		local w = len * 0.2
 
@@ -35,21 +35,23 @@ local widget_bone = {
 			render.SetMaterial( matBoneSmall )
 		end
 
-		local c = Color( 255, 255, 255, 255 )
+		local hovered = LocalPlayer():GetHoveredWidget()
 
-		if ( self:IsHovered() ) then
-			self.a = math.Approach( self.a, 255, FrameTime() * 255 * 10 )
-		elseif ( self:SomethingHovered() ) then
+		if ( IsValid( hovered ) && hovered != self ) then
 			self.a = math.Approach( self.a, 20, FrameTime() * 255 * 10 )
 		else
 			self.a = math.Approach( self.a, 255, FrameTime() * 255 * 10 )
 		end
 
+		local pos = self:GetPos()
+
 		cam.IgnoreZ( true )
-			render.DrawBeam( self:GetPos(), pp, w, 0, 1, Color( c.r, c.g, c.b, self.a * 0.5 ) )
+			c.a = self.a * 0.5
+			render.DrawBeam( pos, pp, w, 0, 1, c )
 		cam.IgnoreZ( false )
 
-		render.DrawBeam( self:GetPos(), pp, w, 0, 1, Color( c.r, c.g, c.b, self.a ) )
+		c.a = self.a
+		render.DrawBeam( pos, pp, w, 0, 1, c )
 
 	end,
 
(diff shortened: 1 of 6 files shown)


garrysmod/lua/entities/widget_bones.lua
```diff
@@ -35,21 +35,23 @@ local widget_bone = {
...
 		end
 
-		local c = Color( 255, 255, 255, 255 )
+		local hovered = LocalPlayer():GetHoveredWidget()
 
-		if ( self:IsHovered() ) then
-			self.a = math.Approach( self.a, 255, FrameTime() * 255 * 10 )
-		elseif ( self:SomethingHovered() ) then
+		if ( IsValid( hovered ) && hovered != self ) then
 			self.a = math.Approach( self.a, 20, FrameTime() * 255 * 10 )
```
robotboy655, line 41: Unexpected change in functionality

garrysmod/lua/entities/widget_disc.lua
```diff
@@ -34,14 +35,15 @@ function ENT:OverlayRender()
...
+	local ang = angles.roll - 90
 
 	render.DepthRange( 0, 0.01 )
 	render.SetMaterial( matDiscAlpha )
-	render.DrawQuadEasy( self:GetPos(), fwd, size, size, Color( c.r, c.g, c.b, c.a * 0.2 ), ang )
+	render.DrawQuadEasy( pos, fwd, size, size, Color( c.r, c.g, c.b, c.a * 0.2 ), ang )
 	render.DepthRange( 0, 1 )
 	render.SetMaterial( matDisc )
-	render.DrawQuadEasy( self:GetPos(), fwd, size, size, Color( c.r, c.g, c.b, c.a ), ang )
+	render.DrawQuadEasy( pos, fwd, size, size, Color( c.r, c.g, c.b, c.a ), ang )
```
robotboy655, line 46: `Color( c.r, c.g, c.b, c.a )` can be just `c`

garrysmod/lua/includes/modules/widget.lua
```diff
@@ -26,54 +34,38 @@ local function UpdateHovered( pl, mv )
...
 	widgets.Tracing = true
-	local tr = util.TraceLine( trace )
+	util.TraceLine( trace )
 	widgets.Tracing = false
 
-	if ( !IsValid( tr.Entity ) ) then return end
-	if ( tr.Entity:IsWorld() ) then return end
-	if ( !tr.Entity:IsWidget() ) then return end
+	if ( !IsValid( tr.Entity ) || tr.Entity:IsWorld() ) then
+		pl:SetHoveredWidget( NULL )
```
robotboy655, line 52: Was this tested?

robotboy655 (changes requested):

garrysmod/lua/entities/widget_bones.lua
thecraftianman: I don't believe so. Correct me if I'm wrong here, but this new check is equivalent to the `not self:IsHovered() and self:SomethingHovered()` one that would occur before with fewer C calls, and the previous result of `self:IsHovered()` now falls under the `else` because it produces exactly the same behavior.

garrysmod/lua/includes/modules/widget.lua
thecraftianman: Yes? This causes the hovered widget to only be set to NULL when there is no valid target widget instead of doing so every single frame (and potentially twice in one frame if there is a valid widget). I don't believe that there should be any regressions in this change (I couldn't find any myself in my testing), but if this is an issue I can revert it back to how it's done in the current implementation.

robotboy655 (changes requested):
image

Found this regression when testing for 1 second. (Just the `widget.lua` file)

The blue box is not supposed to appear.

thecraftianman:
There's a corresponding change in `widgets_axis.lua` (if I remember right) that fixes the issue with this particular widget. This should only affect this one widget because of the way that it attempted to prevent itself from being drawn before when `widgets.RenderMe` was still functional.

robotboy655:
Do you not see the potential issues this can create with 3rd party addon content?

PR #2245, opened 2025-03-19
Net anti-spam protection.

Vitroze:
**_Hello,_**
After browsing through some addons on steam or online shops/downloads I could see that some developers don't protect their net against spam. 

I'm well aware that the **majority of developers** have protected their nets against spam, which is why I'm proposing to deactivate this protection vis-à-vis a convar (activated as standard). 

The algorithm I've set up only works on the SERVER side, since the CLIENT doesn't need any real protection against spam. 

# Function changed
### net.Incoming
In this function I set up a cooldown to display a message if a lot of nets have been sent in a short time, i.e. 0.5 seconds (Customizable by a convar).

I've created two new tables, net.ReceiversConfigTime (for more information, see net.Receive) and net.ReceiversProtectionSpam, and we're going to take a look at the net.ReceiversProtectionSpam table. In this table we'll have this structure: 
```
[“76561198389123776”]:
                [“testNet”] = 1538.0303955078 (CurTime() + Float Value)
```           

Example (Print) : 
```
Player Vitroze Gaming (STEAM_0:0:214429024) tried to spam the net message testNet too much.
```

We're going to emit a (simple) timer that deletes the table value if the cooldown is exceeded. I'm aware that if the player disconnects, his table will still be there, but **_I don't know if we're allowed to create hooks (PlayerDisconnected) in this file, so I'd rather not do it and let others have their say._**

### net.Receive
In the net.Receive modification I've just added an argument (iTime: which can be a float/int) that just changes the cooldown of the specific net.

The net.ReceiversConfigTime table will be used to modify this value.

# ConVar
`sv_net_time` (Float ; Default : 0.5) : The time in seconds between each net message a player can send.
`sv_net_protect` (Int --> Boolean ; Default : 1 (true)) : Whether or not to protect the net messages from spam.

# Documentation (Wiki)

## net.Receive 
```html 
<arg name="TimeCooldown" type="number">Modify net cooldown. Cooldown uses CurTime()</arg>
```
Preview : 
<arg name="TimeCooldown" type="number">Modify net cooldown. Cooldown uses CurTime()</arg>

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -1,24 +1,33 @@
 
--- TODO: Hack. Move to where color is defined?
-TYPE_COLOR = 255
-
--- TODO: Temp hack, remove meta
+-- This is just enough for the entity index. This however is not perfect
+-- as the entity at given index may have changed during transport.
+-- If this becomes a problem, inclusion of entity's serial will also be necessary
 local MAX_EDICT_BITS = 13
 
+TYPE_COLOR = 255
+
 net.Receivers = {}
+net.ReceiversConfigTime = {}
+
+CreateConVar("sv_net_time", 0.5, {FCVAR_ARCHIVE, FCVAR_NOTIFY, FCVAR_REPLICATED}, "The time in seconds between each net message a player can send.")
+CreateConVar("sv_net_protect", 1, {FCVAR_ARCHIVE, FCVAR_NOTIFY, FCVAR_REPLICATED}, "Whether or not to protect the net messages from spam.")
 
 --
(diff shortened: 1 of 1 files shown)


yogwoggf:
I think you're conflating the net library's responsibility as a mechanism for data transfer across the network and as a firewall against potentially malicious attacks. In any case, why would it be built-in? This could be an addon. (I could also see the defaults breaking lots of addons...)

Vitroze:
I understand your comment, I made this pull request to improve net security for developers who don't think or know how to make a “cooldown”. I totally agree that it's the developer's responsibility to secure his net, but unfortunately I see a lot of addons with no real protection against this “threat”.

PR #2247, opened 2025-03-19
Add team.getTeamIDByName(String sName) 

Vitroze:
**_Hello,_**
One thing that might be useful but that I don't necessarily see (and that avoids looping through the team.GetAllTeams()) is the ability to get the team number by the team name. 

# Function added
### team.getTeamIDByName(String sName)
Description : This function will return the team ID of the team that contains the string sName in its name.
Args : sName (String) 
Return : boolean

Example : 
```lua
local sNameJob = "Spectator"
print(team.getTeamIDByName(sNameJob))
-- Output : 1002
```

### DIFF
--- a/garrysmod/lua/includes/modules/team.lua
+++ b/garrysmod/lua/includes/modules/team.lua
@@ -1,11 +1,16 @@
 module( "team", package.seeall )
 
 local TeamInfo = {}
+local TeamInfoByName = {}
 local DefaultColor = Color( 255, 255, 100, 255 )
 
 TeamInfo[TEAM_CONNECTING] 	= { Name = "Joining/Connecting", 	Color = DefaultColor, 	Score = 0, 	Joinable = false }
 TeamInfo[TEAM_UNASSIGNED] 	= { Name = "Unassigned", 			Color = DefaultColor, 	Score = 0, 	Joinable = false }
 TeamInfo[TEAM_SPECTATOR] 	= { Name = "Spectator", 			Color = DefaultColor, 	Score = 0, 	Joinable = true }
+
+TeamInfoByName["Joining/Connecting"] = TEAM_CONNECTING
+TeamInfoByName["Unassigned"] = TEAM_UNASSIGNED
+TeamInfoByName["Spectator"] = TEAM_SPECTATOR
 
 --[[------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


robotboy655 (changes requested):
What is the use case for this? 

Also the function name does not follow capitalization of other functions in the game.

Vitroze:
The interest of this function is more for the understanding of the user who will configure an addon, for example if I need to be in a certain team instead of making a loop and finding IDs that interest (to retrieve information) a function with just a table will suffice and will do optimization. 

For the user, it's a better understanding of the configuration, and for the developer, it avoids having to loop through all the teams in order to retrieve ID

I'm going to change the name of the function, thank you for reporting it.

robotboy655:
`team.GetName(  Player:Team() ) == "Spectator"` will get you there though, without any looping.

Vitroze:
Sure, but it's not necessarily about the player. For example, if you need to retrieve data via the team name, such as its color, he'll have to make a loop to retrieve this information.

For example : 
```lua
local cColor 
local sNameTeam = "Spectator"
for _, tData in pairs(team.GetAllTeams()) do
    if tData.Name == sNameTeam then
        cColor = tData.Color
        break
    end
end

print("Color for spectator team: ", cColor)
```

Vitroze (commented):
I've changed the function name to GetIDByName

robotboy655:
My point is, what is a situation where you'd know the name but not the ID of a team?

Vitroze:
In fact, it depends on how the system works. For example, if we were to use a team-based game mode and wanted to know the number of people in a team by name, we'd currently have to loop back and forth. With this function, we wouldn't have to.
Some addons are configured using team names rather than team IDs. This is simply another configuration method, easier for the user and simple to set up, without having to create a loop.

robotboy655:
In a team based gamemode, the gamemode would store the team IDs and use those. It wouldn't store teams by name.

Can you provide an addon that targets teams by name? Why is looping such a big issue in your mind?

My concern here is just the reliability of the implementation as well as the miniscule increased memory usage.

Because `team.GetAllTeams()` exists, it can be used to change the name of a team, and `team.GetIDByName` would no longer function correctly.

Vitroze:
> Can you provide an addon that targets teams by name? 

I don't have a specific addon in mind but I know that in the darkrp gamemode on the workshop or on other sites this was done a lot. 


> Why is looping such a big issue in your mind?

It's only micro-optimization, because if you have to loop over 100 teams to find just one ID, it's a useless loop, but if this function exists, it won't need this famous loop.

PR #2250, opened 2025-03-25
TTT: Russian localization update

Satton2:
- Localized strings that were added/updated in: https://github.com/Facepunch/garrysmod/pull/2235


Satton2:
@svdm

PR #2253, opened 2025-03-25
DProperty_Float SetMin/SetMax passthrough

blobles-dev:
I can't think of a good reason why these weren't added, you've got to do some panel child-hunting to locate it otherwise, thx

### DIFF
--- a/garrysmod/lua/vgui/prop_float.lua
+++ b/garrysmod/lua/vgui/prop_float.lua
@@ -58,6 +58,16 @@ function PANEL:Setup( vars )
 		ctrl:SetValue( val )
 	end
 
+	-- Set the min
+	self.SetMin = function( slf, val )
+		ctrl:SetMin( val )
+	end
+
+	-- Set the max
+	self.SetMax = function( slf, val )
+		ctrl:SetMax( val )
+	end
+	
 	-- Alert row that value changed
 	ctrl.OnValueChanged = function( slf, newval )
 


blobles-dev:
I realised the potential issues with NetworkVar/editable entities:

perhaps either:
add warning to wiki (this wont effect editable entity panels), you can edit the slider anyway by playing child-fu
   or
ignore editable entity setmax/setmin calls

robotboy655:
I am not quite sure what you mean by the second comment? Do you think this is a good idea to merge or no? Is your intention to use these in conjunction with NetworkVars, which I think already set the min/max for you?

blobles-dev:
I think we can disregard my previous stupor. I haven't used editable entities so I assumed they were manually created panels - which led me to think that if you set a value on a network var DProperty there would be mis-alignment. I think the only way to achieve this misalignment with editable entities if this pull request is merged would be through intentional fuckery so yeah by all means boss let's merge

ToxiWatch:
Hi @blobles-dev, we're researchers from Stockholm University.

Your recent [comment](https://github.com/Facepunch/garrysmod/pull/2253#issuecomment-2791171346) was flagged by [Detoxify](https://github.com/unitaryai/detoxify) (an automated tool) as potentially toxic.

> i think we can disregard my previous stupor. i haven't used editable entities so i assumed they...

While detection isn't perfect, [research](https://ieeexplore.ieee.org/document/9397528) shows toxic interactions on GitHub can discourage developers, leading to stress and burnout. If you'd like, here are a few ways to improve clarity or tone:
- Rephrase with a more constructive style
- Add context or soften strong language
- Try tools like ChatGPT for a neutral rewrite

You can share feedback in our [1-minute survey](http://otree-extra-url-d0bb646a8437.herokuapp.com/?hash_id=166bd35a2b0456be183a5206c517e944443789df7e838db777f76347b399a9d1) (opt-out available at the end).

Thanks for helping make open source a more welcoming place — click our profile to learn more about us!

React to this comment to let us know you saw it!

robotboy655:
I think if we are to merge this, it should follow established way to do this by DProperty_Combo, by putting the methods on the Row panel, not the sub panel.

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/prop_combo.lua#L54-L60

robotboy655 (changes requested):
I think if we are to merge this, it should follow established way to do this by DProperty_Combo, by putting the methods on the Row panel, not the sub panel.

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/prop_combo.lua#L54-L60

PR #2259, opened 2025-03-29
Fix an error with taking ammo from NPCs

github-is-garbage:
If an NPC is holding a weapon where Clip1 (or 2) <= 0 then TakePrimaryAmmo (or TakeSecondaryAmmo) will cause an error because the Ammo1 and Ammo2 functions call GetAmmoCount which doesn't exist for NPCs.

If GetAmmoCount was added for NPCs it would still error because of the RemoveAmmo call, since that also does not exist for NPCs.

Reproduce:
```lua
local Guy = ents.Create("npc_combine_s")
Guy:Spawn()
Guy:Give("weapon_base")

timer.Simple(1, function()
    Guy:GetActiveWeapon():SetClip1(0)
    Guy:GetActiveWeapon():TakePrimaryAmmo(1) -- Should error here
end)
```

EDIT: Minor spelling mistake

### DIFF
--- a/garrysmod/gamemodes/base/entities/weapons/weapon_base/shared.lua
+++ b/garrysmod/gamemodes/base/entities/weapons/weapon_base/shared.lua
@@ -162,6 +162,8 @@ function SWEP:TakePrimaryAmmo( num )
 	-- Doesn't use clips
 	if ( self:Clip1() <= 0 ) then
 
+		if ( self:GetOwner():IsNPC() ) then return end
+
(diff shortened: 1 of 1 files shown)


robotboy655:
Fixed it here: https://github.com/Facepunch/garrysmod/commit/6af7c201a09430f9c5715a84292ab8c391aec3b1

The end result is same, but also fixes potential Lua errors when simply calling Ammo1() on the SWEP.

PR #2260, opened 2025-04-01
Add net.WriteSteamID64, net.ReadSteamID64

Heyter:
- Added functions:
  - `net.WriteSteamID64( string steamid64 )` - writes SteamID64 in compact binary form (40 bits vs 144 bits for net.WriteString)
  - `net.ReadSteamID64()` - reads SteamID64 from compact binary form

### DIFF
--- a/garrysmod/lua/includes/extensions/net.lua
+++ b/garrysmod/lua/includes/extensions/net.lua
@@ -238,16 +238,36 @@ end
 do
 	local mask = 0x100000000
 
-	function net.WriteSteamID64( steamid64 )
-		local unique_part = tonumber( string.sub(steamid64, 6) )
+	function net.WriteSteamID64(steamid64)
+		local prefix = tonumber(string.sub(steamid64, 1, 5))
+		local unique_part = tonumber(string.sub(steamid64, 6))
+
 		local high = unique_part % mask
 		local low = math.floor(unique_part / mask)
 
-		net.WriteUInt(high, 32) net.WriteUInt(low, 8)
+		net.WriteUInt(high, 32)
+
+		if prefix == 76561 then
+			net.WriteBool(false)
+			net.WriteUInt(low, 8)
+		else
+			net.WriteBool(true)
+			net.WriteUInt(low, 10) -- 0x0170000000000000
+			net.WriteUInt(prefix, 17)
+		end
 	end
 
 	function net.ReadSteamID64()
-		local high, low = net.ReadUInt(32), net.ReadUInt(8)
-		return "76561" .. tostring(low * mask + high)
+		local high = net.ReadUInt(32)
+		local is_special = net.ReadBool()
+		local low = net.ReadUInt(is_special and 10 or 8)
+		local part = tostring(low * mask + high)
+
+		if is_special then
+			local prefix = net.ReadUInt(17)
+			return tostring(prefix) .. part
+		end
+
+		return "76561" .. part
 	end
 end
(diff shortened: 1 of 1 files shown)


robotboy655:
You are losing a lot of data the way you are doing it. Perhaps this is best done in C++. SteamIDs are 64 bit numbers, you are throwing 24 bits of that data away.

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -234,3 +234,20 @@ function net.ReadType( typeid )
...
+		local unique_part = tonumber( string.sub(steamid64, 6) )
+		local high = unique_part % mask
+		local low = math.floor(unique_part / mask)
+
+		net.WriteUInt(high, 32) net.WriteUInt(low, 8)
+	end
+
+	function net.ReadSteamID64()
+		local high, low = net.ReadUInt(32), net.ReadUInt(8)
+		return "76561" .. tostring(low * mask + high)
```
WilliamVenner, line 251: Perhaps on the Write side you can check if the SteamID64 _doesn't_ start with 76561 and if so, `net.WriteBool(true)` then write the missing 24 bits Rubat mentioned. On the read side, you read that bool, if it's true you read the missing 24 bits, otherwise you assume `76561`.

Srlion:
[net.WriteUint64](https://gmodwiki.com/net.WriteUInt64)?

WilliamVenner:
This function saves a few bits in comparison

Heyter:
> [net.WriteUint64](https://gmodwiki.com/net.WriteUInt64)? I don't feel like this would be useful to add because you can directly use `net.WriteUint64` and it will just be a wrapper around it?

> WriteUInt64 bits        64

> WriteSteamID64 bits     41

WilliamVenner:
Yeah, I can see this saving a significant amount of bits if used to send a lot of SteamID64s at once.

Heyter:
> Yeah, I can see this saving a significant amount of bits if used to send a lot of SteamID64s at once.

Why did Rubat close the PR?

robotboy655:
Just use [net.WriteUint64](https://gmodwiki.com/net.WriteUInt64) as mentioned above?

WilliamVenner:
As mentioned above in 99.9% of cases you don't need to write all 64 bits though, which is what this function optimises for.

Srlion:
This actually sounds stupid. Who is going to send lots of steamid64s at once? You doing it? Then you are probably doing something wrong and should do paging and not send them all at once.

robotboy655:
Is SteamID64 sent over network so often that it can be noticed at all though? It's not like players/entities, which are sent over in most net messages. 

Also I am not sure how reliable it is to just omit a bunch of data, and to do it with a string check.

WilliamVenner:
> This actually sounds stupid. Who is going to send lots of steamid64s at once? You doing it? Then you are probably doing something wrong and should do paging and not send them all at once.

What are you talking about? By "all at once" that includes the pagination use case. This would also save bits for pagination...

> Is SteamID64 sent over network so often that it can be noticed at all though? It's not like players/entities, which are sent over in most net messages.

Would probably be mostly useful for administrative-type menus as an example. E.g. a big list of SteamID64s in an admin mod menu. Regardless, any bit savings in these net functions coalesce into the simple fact that we can fit more data into a single packet/reliable data buffer frame, ergo achieving overall more bandwidth.

> and to do it with a string check

This is a valid concern TBH and would justify implementing this on the C++ side.

PR #2261, opened 2025-04-02
Improving the problems menu

darkfated:
- Combined duplicate problem and error list refresh functions into a single shared RefreshList function to improve code maintainability and reduce code duplication
- The changes improve sorting performance in `ReceivedError()` function by implementing a single-pass array sorting with numeric indices instead of multiple iterations. This optimization reduces memory allocations and improves efficiency when handling large numbers of error panels
- Added menu closing functionality using the ESCAPE key. (Implementation uses `input.IsKeyDown()` in the `Think()` method since the standard OnKeyCodePressed handler doesn't capture the ESCAPE key in Garry's Mod menu system). This improves UX by allowing users to close the menu using the standard game key

### DIFF
--- a/garrysmod/lua/menu/problems/problem_lua.lua
+++ b/garrysmod/lua/menu/problems/problem_lua.lua
@@ -1,3 +1,4 @@
+
 local realmColors = {}
 realmColors[ "menu" ] = Color( 121, 221, 100 )
 realmColors[ "client" ] = Color( 255, 222, 102 )
@@ -74,10 +75,7 @@ function PANEL:Paint( w, h )
 
 	-- The error count
 	if ( count > 0 ) then
-		local txt = "x" .. count
-		surface.SetFont( "DermaMedium" )
-
-		draw.SimpleText( txt, "DermaMedium", w - 16 - 16, h / 2, clr, draw.TEXT_ALIGN_RIGHT, draw.TEXT_ALIGN_CENTER )
+		draw.SimpleText( "x" .. count, "DermaMedium", w - 16 - 16, h / 2, clr, draw.TEXT_ALIGN_RIGHT, draw.TEXT_ALIGN_CENTER )
 	end
 
 	-- The error
--- a/garrysmod/lua/menu/problems/problem_lua.lua
+++ b/garrysmod/lua/menu/problems/problem_lua.lua
@@ -1,4 +1,3 @@
-
 local realmColors = {}
 realmColors[ "menu" ] = Color( 121, 221, 100 )
 realmColors[ "client" ] = Color( 255, 222, 102 )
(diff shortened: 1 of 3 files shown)


darkfated:
Let me briefly explain my optimization 🙃

Both algorithms (the existing one and mine) have the same O(n log n) complexity, but my version has fewer operations and a better constant factor. At low error counts, the difference is negligible, but at scale, mine will be faster. I also use an array instead of a hash table, which is more efficient, and perform fewer sorting passes

garrysmod/lua/menu/problems/problems.lua
```diff
@@ -9,26 +8,33 @@ local MenuUpdated = false
...
-			ProblemsPanel:ReceivedError( id, err )
-		end
-		ProblemsPanel:InvalidateLayout()
-	end
+	RefreshList(
+		ProblemsPanel.LuaErrorList,
+		ErrorLog,
+		ProblemsPanel.ErrorPanels,
+		ProblemsPanel.ReceivedError
+	)
```
robotboy655, line 37: image

Sigh.

garrysmod/lua/menu/problems/problem_lua.lua
```diff
@@ -1,4 +1,3 @@
-
```
robotboy655, line 1: Pointless change.

garrysmod/lua/menu/problems/problem_lua.lua
```diff
@@ -77,8 +76,6 @@ function PANEL:Paint( w, h )
 	if ( count > 0 ) then
 		local txt = "x" .. count
 		surface.SetFont( "DermaMedium" )
-		local tW, tH = surface.GetTextSize( txt )
-		tW = tW
```
robotboy655, line 81: The `surface.SetFont` call can be removed as well, the the `txt` variable too can just be moved into `draw.SimpleText` call.

garrysmod/lua/menu/problems/problems_pnl.lua
```diff
@@ -48,6 +48,14 @@ function PANEL:Init()
...
 end
 
+function PANEL:Think()
+
+    if ( input.IsKeyDown( KEY_ESCAPE ) ) then
+        self:Remove()
+    end
+
+end
+
```
robotboy655, line 58: I am not sure how much I like this when you are in-game. Pressing ESC would close both the main menu, AND the problems panel.

robotboy655 (changes requested):
Please actually test your changes before making a Pull Request. How many times do I have to encounter PRs that do not work?

darkfated:
I made the problem menu close on ESP only in the main menu. I also rolled back the changes related to the list refresh processing function. The mistake was that I carelessly passed the table that I made a copy of in the argument and cleaned it up locally. I decided to return the old version

darkfated:
I would suggest the idea of closing the menu also by clicking in a free area of the screen

{2D1C0D69-CF46-47DB-A812-C19A829E06CD}

darkfated:
https://github.com/user-attachments/assets/53f8b354-be5c-4221-88b1-6c0316773a1d

robotboy655:
You should not be removing the default close button.

darkfated:
> You should not be removing the default close button.

All done

robotboy655 (approved):

PR #2263, opened 2025-04-02
[TTT] Misc. changes to ConVars

Tiagoquix:
This PR changes many misc. things about TTT ConVars.

## Misc. changes

- Changes some text used in the main menu "New Game" section.
- Adds `FCVAR_NOTIFY` to various ConVars.
- If voice drain is enabled, it will not be applied to admins anymore by default.

## Karma system

- Adds an option to disable the Karma system via the main menu "New Game" hosting.
- Increases max. Karma to 1050 by default, in order to give a little grace to players who do something stupid and not immediately drop below 1000.

## Default ConVar values changes

- Allows Innocents to pick-up dropped Detective hats by default. IDK why this isn't enabled by default, as it is harmless and just a cool little thing (`ttt_detective_hats_reclaim_any 1`).
- Makes the game mode print last words by default (`ttt_lastwords_chatprint 1`).
- Enables post-round deathmatch by default (`ttt_postround_dm 1`). As I see it, the post-round period is used to do dumb things, so allowing deathmatch without penalizing karma by default is the best option IMO. It's also one the things that IDK why it was the default already.


figardo:
Enabling ttt_lastwords_chatprint is a change for the worse imo.

It allows innocents to spam `ttt_radio traitor` commands on the person they're fighting, meaning their death will print a last words message that indisputably confirms that they have died and that the KOS is likely valid. Even if they don't have someone to fight, by typing something in their chatbox they can ensure that their death will be instantly confirmed without their body being found. Either way, it's a huge nerf to the traitor's ability to perform social deception.

Leaving the convar disabled doesn't render the last words functionality unused or anything, it still applies to fall deaths even when disabled. Since fall deaths don't leave DNA on the corpse, the instant confirmation that the victim fell to their death is a good downside to balance it out.

mgetJane:
last words should definitely not be enabled by default

svdm:
I'm fine with the New Game stuff, cvar flags and voice drain setting.

Otherwise, these convars have a default that was an intentional choice, and can be customised by servers that want to make a different one. I think the only reason to change them would be if people start a server and are incredibly confused with the default settings (realistically they are instead confused with the lack of extra roles, hats, weapons, and player models).

> Increases max. Karma to 1050 by default, in order to give a little grace to players who do something stupid and not immediately drop below 1000.

The opinion I've heard more over the years is that the Karma system is quite/too mild as-is.

> Allows Innocents to pick-up dropped Detective hats by default. IDK why this isn't enabled by default, as it is harmless and just a cool little thing (ttt_detective_hats_reclaim_any 1).

Makes those Innocents look like Detectives at a glance (esp. for newer players), which has gameplay impact. But I could live with changing this one.

> Makes the game mode print last words by default (ttt_lastwords_chatprint 1).

See other comments.

> Enables post-round deathmatch by default (ttt_postround_dm 1). As I see it, the post-round period is used to do dumb things, so allowing deathmatch without penalizing karma by default is the best option IMO. It's also one the things that IDK why it was the default already.

Server vibe choice. There's an argument that you should not randomly shoot people in the game and doing it after a round ends is out of place.

mgetJane:
i mostly had an issue with the lastwords part, the other changes seem fine, lastwords basically just confirms deaths for free which i think takes a lot more away from the game than it adds

detective hats being able to be picked up i think is ok and it's funny to use it when ppl are on the look out for you, letting you slip past in their peripheral vision or give you a split second of advantage when they see you shooting at someone, though i dont feel strongly at all about what the default should be

PR #2267, opened 2025-04-08
Functions `PANEL:Set/GetBlockScrolling`

Astralcircle:
Adds a simple function to `DVScrollBar` and `DHScrollBar` that allows you to block scrolling

### DIFF
--- a/garrysmod/lua/vgui/dhscrollbar.lua
+++ b/garrysmod/lua/vgui/dhscrollbar.lua
@@ -90,8 +90,6 @@ end
 
 function PANEL:AddScroll( dlta )
 
-	if ( !self:GetBlockScrolling() ) then return end
-
 	local OldScroll = self:GetScroll()
 
 	dlta = dlta * 25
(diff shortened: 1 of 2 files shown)


robotboy655:
Can you provide a use case for this?

robotboy655:
That didn't really make it easier to understand. What is the issue you are trying to solve? Scrolling in 1 panel scrolls the other or something?

Also, wouldn't `SetEnabled( false )` do this as well?

Astralcircle:
Yes, that's the problem. `SetEnabled( false )` does that, but it rolls back the scrolling and hides the scrollbar itself.

robotboy655:
That sounds like you want a workaround for your specific problem only useful for you to be included in the base game.

I'd say it would be more logical to fix the core issue of parent panel scrolling while a child panel being scrolled.
The only issue is that it appears to not be a problem by default already?

https://github.com/user-attachments/assets/23505279-653c-4973-8b18-5dd25918fe87

Could it be that you are doing something wrong on your end, like not returning true in `OnMouseWheeled` of the child panel? Or am I misunderstanding something? Can you provide a minimal reproduceable code to recreate the issue you are having?

Astralcircle:
oh really fixed, sorry :/

PR #2268, opened 2025-04-09
Add list.Remove helper function

blobles-dev:
Add list.Remove because I was too dumb to realise list.Set with nil as the value would do it

_Yes it's basically list.Set, but it's a logical addition, I think_


robotboy655 (approved):

robotboy655:
I will rename it to `list.RemoveEntry` to be more consistent with functions that use lists by key, not by value. `list.Remove` sounds like you are removing the entire list.

Let me know if you think otherwise.

PR #2269, opened 2025-04-10
Mipmaps for png icons in spawnmenu

dar-su:
Lots of custom weapons on workshop are using 256-512px icons for spawnmenu, but it creates materials for them without using any png parameters, which makes them very aliased.

I've compared all possible combinations of params and filtering options and found out "mips" without extra filtering produces best result. 
"mips smooth" makes icons very blurry, "smooth" keeps aliasing and filtering blurs out icons even more.

ezgif com-animated-gif-maker

Here's link to comparison between all combinations, taken on 1920x1080 screen - https://imgsli.com/MzY5MzE1/1/3

There's no visible changes to regular 128px icons like ones that included in the game.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/creationmenu/content/contenticon.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/creationmenu/content/contenticon.lua
@@ -69,7 +69,7 @@ function PANEL:SetMaterial( name )
 
 	self.m_MaterialName = name
 
-	local mat = Material( name )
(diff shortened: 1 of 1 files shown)


robotboy655:
So, you want to make all the icons blurrier, and increase memory usage of the game, because some addons ship incorrect icon sizes? No thank you.

Here's an example of "no visible changes" you claim, what the filters were added for: (https://github.com/Facepunch/garrysmod/commit/61cdbe9f8c5166cfccab790e661bda95027e27f6)
April10-1278-paintdotnet

PR #2272, opened 2025-04-16
Improve __index performance

darkjacky:
Entity.__index
Player.__index
Weapon.__index

Removed the if statements and  cached the entity.GetTable result to make it about 50% quicker.
There does not seem to be any negatives to this but we must be cautious. Please test it yourself because just because it works doesn't mean it can't cause issues.


Srlion:
Removing the if statements in here could break compatibility if someone ever stored a bool in entity/player metatables, revert them to if statements to check for nil.

PR #2277, opened 2025-04-18
Frontend impr

darkfated:
I switched the interface to an Inter font + fixed bugs with layot

### Screenshots
**Was**

**Become**

**Was**

**Become**

**Was**

**Become**


darkfated:
I'll note the disadvantage that the email was connected via a link. When there is no Internet connection, the font is not downloaded, and the default font is used

darkfated:
I'll note the disadvantage that the font was connected via a link. When there is no Internet connection, the font is not downloaded, and the default font is used

PR #2280, opened 2025-04-25
Add vector/angle 'reset' key to DAdjustableModelPanel

blobles-dev:
Pretty self explanatory. I kept moving around too quickly and losing where my models where, so this adds ability to use +reload / KEY_R to reset the panel to its base values

default values from DModelPanel init: https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/vgui/dmodelpanel.lua#L21

if anyone wants to chip in this could probably do with a way to set 'default' vector/angles and it use those instead, but wasnt sure of the best way to go about it tbh.

### DIFF
--- a/garrysmod/lua/vgui/dadjustablemodelpanel.lua
+++ b/garrysmod/lua/vgui/dadjustablemodelpanel.lua
@@ -99,6 +99,20 @@ function PANEL:FirstPersonControls()
 
 	local x, y = self:CaptureMouse()
 
+	if ( IsKeyBindDown( "+reload" ) or input.IsKeyDown( KEY_R ) ) then
+		self.vCamPos = Vector( 50, 50, 50 )
+		self.vLookatPos = Vector( 0, 0, 40 )
+
+		self.aLookAngle = ( self.vLookatPos - self.vCamPos ):Angle()
+
+		self.fFOV = 70
+
+		self.OrbitPoint = self.vLookatPos
+		self.OrbitDistance = ( self.OrbitPoint - self.vCamPos ):Length()
+
+		return
+	end
+
 	local scale = self:GetFOV() / 180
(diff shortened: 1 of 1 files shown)


blobles-dev:
I had thought of an alternative approach, which would be instead of 'reset', it's more of a 'focus' key, where it positions the camera to center the model, something like:

```
  if ( IsKeyBindDown( "+reload" ) or input.IsKeyDown( KEY_R ) ) then
  
          if IsValid( self.Entity ) then
  
              local mins, maxs = self.Entity:GetModelBounds()
              local center = ( mins + maxs ) / 2
  
              local modelSize = maxs - mins
              local maxDimension = math.max( modelSize.x, modelSize.y, modelSize.z )
  
              self.vCamPos = center + Vector( maxDimension, maxDimension, maxDimension )
              self.vLookatPos = center
  
              self.aLookAngle = ( self.vLookatPos - self.vCamPos ):Angle()
  
              self.OrbitPoint = self.vLookatPos
              self.OrbitDistance = ( self.OrbitPoint - self.vCamPos ):Length()
  
              self.fFOV = 70
          end
  
	  return
  end
```

which is a little bit different, i think first approach + some way to set defaults is best way, but open to any feedback/discussion

robotboy655:
I don't think this belongs in the base panel. For spawnicon editor you already have a few "reset"/preset buttons just below the model panel. If you have custom panel that uses DAdjustableModelPanel, you can just implement this for your own usecase yourself.

Just thinking in context of addons, having this available may not always be a good idea, and the hardcoded position will not always be the same as whatever position/angle the addon first set.

There's also the argument of accidentally pressing a key-bind that wasn't there before, or addons having something else bound to reload key in their custom versions of this panel.

PR #2281, opened 2025-04-30
Create Lua VMT proxies for getting full FB & depth

piqey:
# The Problem

As it stands, mappers have no way to access `_rt_ResolvedFullFrameDepth` without additional clientside Lua scripting (`GM:NeedsDepthPass`), limiting them to the `192`-unit depth buffer stored in the alpha channel of `_rt_FullFrameFB` and similar buffers. For mappers creating content for community servers that distribute maps via FastDL (and ship maps via singular BSP files), it's fairly limiting. Not to mention the hook-based implementation is limiting in its own right when in the majority of cases it's materials that end up using these textures anyways.

Problems also arise with `screenspace_general`. There _is_ a workaround with material flags and material proxies to force-assign the spritecard flag to `$flags2`, which can successfully generate an up-to-date framebuffer texture for use with pixel shaders. There is, however, no way to gain access to request an up-to-date `_rt_ResolvedFullFrameDepth` texture without clientside Lua scripting.

# The Pull

Included are two Lua material proxies: `GetScreenEffectTexture` and `GetResolvedFullFrameDepth`.
They are described in greater detail below:

## GetScreenEffectTexture

Returns the result of `render.GetScreenEffectTexture()` via `resultVar` (also see effects of `updateImmediately` under **Parameters**).

In most cases this is only when the material with the proxy is currently visible on screen (exceptions I've found include materials applied to entities such as the local player).

### Parameters
- `updateimmediately`
  - Set this to nothing (or `0`) to mark the screen effect texture as dirty, waiting for the next `GM:RenderScreenspaceEffects()` call to update it (asynchronous)
  - Set this to `1` to perform immediately screen effect texture updates for this material (synchronous)
    - This is necessary for some screenspace overlays, such as the `g_bokehblur` VMT example given below
  - Set this to `2` for the same behavior as `1`, but when your material can and/or will be used for rendering model halos (such as being grabbed by the physgun) to avoid turning the entire screen black.
- `resultvar`
  - The VMT parameter to which the screen effect texture should be supplied

#### In case it wasn't clear, the problem with model halos is as follows:

Model halos require the model to be drawn a second time. If a model has a material with the `GetScreenEffectTexture` on it and the proxy has `updateimmediately` enabled, that would cause a second screen effect texture update. It appears that the particular context in which that second drawing occurs results in a "blackout" of the entire screen, hence the implementation of mode `2`: do everything the same, unless we exist in that particular context.

Depending on when the material is getting rendered this may still cause problems, but that should be relatively easy to diagnose (model halos seem to be the only "breaking" issue worth a patch that goes beyond anything a developer might encounter and diagnose on an individual basis).

## GetResolvedFullFrameDepth

Returns `true` in `GM:NeedsDepthPass()` while any materials with this proxy are active (and returns `render.GetResolvedFullFrameDepth()` via `resultVar`).

In most cases this is only when the material with the proxy is currently visible on screen (the only exceptions I've found being those aforementioned).

### Parameters
- `resultvar`
  - The VMT parameter to which the screen effect texture should be supplied

## Example VMTs

`g_bokehblur` is a material shader traditionally implemented exclusively through Lua, in part due to the necessity of `GM:NeedsDepthPass`. With these material proxies, it could be implemented and shipped in a map as a standalone VMT file (using `env_screenoverlay` to draw it on screen):

```
"g_bokehblur"
{
	$basetexture	"_rt_FullFrameFB"
	$depthtexture	"_rt_ResolvedFullFrameDepth"

	$size		4.0
	$focus		.1
	$focusradius	3.0

	$ignorez		1

	"Proxies"
	{
		"GetScreenEffectTexture"
		{
			updateImmediately	1
			resultVar		$basetexture
		}
		"GetResolvedFullFrameDepth"
		{
			resultVar		$depthtexture
		}
		"Sine"
		{
			sineMin		.075
			sineMax		.8
			sinePeriod	5.5
			resultVar		$focus
		}
		"Sine"
		{
			sineMin		4.0
			sineMax		7.0
			sinePeriod	5.5
			resultVar		$size
		}
		"Sine"
		{
			sineMin		.5
			sineMax		3.0
			sinePeriod	8.0902
			resultVar		$focusradius
		}
	}
}
```

The `Sine` proxies were added in lieu of Derma sliders.
Here are some additional Lua `CreateMaterial` snippets for testing in case they're useful:

```lua
TESTMAT_SCREENEFFECT_0 = CreateMaterial("piqey_testmat_fb_0", "VertexLitGeneric", {
	["$basetexture"]	=	"models/props_junk/cardboard_boxes001a",
	["Proxies"] = {
		["GetScreenEffectTexture"] =
		{
			resultVar		= "$basetexture"
		}
	}
})
TESTMAT_SCREENEFFECT_1 = CreateMaterial("piqey_testmat_fb_1", "VertexLitGeneric", {
	["$basetexture"]	=	"models/props_junk/cardboard_boxes001a",
	["Proxies"] = {
		["GetScreenEffectTexture"] =
		{
			resultVar		= "$basetexture",
			updateImmediately = 1 -- This one is for breaking via physgun
		},
	}
})
TESTMAT_SCREENEFFECT_2 = CreateMaterial("piqey_testmat_fb_2", "VertexLitGeneric", {
	["$basetexture"]	=	"models/props_junk/cardboard_boxes001a",
	["Proxies"] = {
		["GetScreenEffectTexture"] =
		{
			resultVar		= "$basetexture",
			updateImmediately = 2 -- This one won't break via physgun
		}
	}
})
TESTMAT_DEPTH = CreateMaterial("piqey_testmat_depth", "VertexLitGeneric", {
	["$basetexture"]	=	"models/props_junk/cardboard_boxes001a",
	["Proxies"] = {
		["GetResolvedFullFrameDepth"] =
		{
			resultVar		= "$basetexture"
		}
	}
})
```

# Closing Notes

I wasn't sure about the naming conventions for hooks and variables in this context; not trying to be negligent, just uncertain.

## A Picture(?)
Very cool boxes


robotboy655:
Partially fulfilled, I think, with the addition of `UpdateScreenEffectTexture` matproxy for this issue: https://github.com/Facepunch/garrysmod-requests/issues/3383

image

Please let me know if that solves it for your use case(s). Its implemented in C++ for speed which is important for matproxies, and it also updates the texture once per frame (at the end of the frame, after rendering the world, but before HUD, so it sidesteps the halo issues), and without the $resultVar and input value shenanigans.

PR #2283, opened 2025-05-02
Add `player.GetByNick`

Astralcircle:
It will be a convenient addition for different admin mods so that, for example, when entering commands, you don’t have to write your own detection logic by nick

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -262,6 +262,17 @@ function meta:HasGodMode()
 end
 
 -- These are totally in the wrong place.
+function player.GetByNick( ID )
+	local players = player.GetAll()
+	for i = 1, #players do
+		if ( players[i]:Nick() == ID ) then
+			return players[i]
+		end
+	end
+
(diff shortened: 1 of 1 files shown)


Astralcircle:
> why not use player.Iterator()?

Because other `player.GetBy*` functions don't use it either and I don't know exactly how it can affect compatibility. If Rubat tells me to use `player.Iterator` as in my previous PR for optimizing the `team` library, I will do it.

robotboy655:
The thing is, the only use case I see for this is, as you said, plugging user input into this function. Now, I imagine most admin mods and the like use their own methods of achieving this not because this doesn't exist, but because this is simply not sufficient.

Many players use weird nicknames with non ASCII characters, long user names, etc, which make a function like this, in context of user input - absolutely useless. Adding "fuzzy search" introduces its own issues - matching multiple players.

Plus, many admin mods would implement special formatting to easily target players via SteamIDs and other metrics.

I think this is something that should stay mod territory - let people decide how exactly they want something like this to work. They will anyway, making this addition, again, useless in my opinion.

PR #2287, opened 2025-05-06
Minor improved cam.Start2D

FosterZRussian:


Astralcircle:
You should write why this change is needed. But since you did not give the reason, i will name it myself:
Reduces the overhead of creating a table and does not create garbage

robotboy655 (approved):

PR #2290, opened 2025-05-12
Localize Bodygroup Names if possible

IBRS-4Ever:
Also added a option to toggle localizations.

I suggest creating a 'bodygroups.properties' for bodygroup strings.


Satton2:
I don't like the option to disable certain localizations, but overall I like the idea of localizing body groups.

IBRS-4Ever:
Skin localized

PR #2292, opened 2025-05-18
Replaced UniqueID to AccountID

IP1A:


robotboy655:
This cannot be accepted due to backwards compatibility with addons.

The internal table is exposed to Lua, meaning that addons expecting UniqueID in the table would cease to work.

PR #2293, opened 2025-05-18
Replaced UniqueID to AccountID

IP1A:


robotboy655:
See https://github.com/Facepunch/garrysmod/pull/2292

PR #2294, opened 2025-05-20
Add COLOR:Copy() method

wrefgtzweve:
Adds a way to copy a color object over to a new object, useful when you want to keep the original color and you're working with functions that change the original like for example `color:AddHue()`

Thanks to https://github.com/TankNut for suggesting directly setting the metatable instead of using self:Unpack.

### DIFF
--- a/garrysmod/lua/includes/util/color.lua
+++ b/garrysmod/lua/includes/util/color.lua
@@ -182,7 +182,12 @@ end
 
 function COLOR:Copy()
 
-	return Color(self:Unpack())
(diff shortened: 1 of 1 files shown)


robotboy655:
I disagree on using `setmetatable` directly, it may be ever so slightly faster, but it makes the code ever so slightly worse/less maintainable.

By that logic, every call to `Color` should be replaced with a call to `setmetatable`, at least within that file.

robotboy655:
I think that would be best.

PR #2297, opened 2025-05-25
Add `PreRegisterPANEL` hook

Astralcircle:
Does the same thing as all other `PreRegister*` hooks, allowing addons to override panel functions without replacing the entire file/putting it into less-than-suitable hooks.

### DIFF
--- a/garrysmod/lua/includes/extensions/client/panel/scriptedpanels.lua
+++ b/garrysmod/lua/includes/extensions/client/panel/scriptedpanels.lua
@@ -76,6 +76,11 @@ function vgui.Register( classname, mtable, base )
 	-- Remove the global
 	PANEL = nil
 
+	-- Check if something wants to modify or prevent it
+	if ( hook.Run( "PreRegisterPANEL", classname, mtable, base ) == false ) then
+		return
+	end
+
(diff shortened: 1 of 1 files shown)


robotboy655:
In most cases you'd subclass the panel. What is the use case for this?

Astralcircle:
In most cases, it's for modding other addons/gamemodes without completely recreating the panel. Also sometimes this can't be done.

robotboy655:
Such as? Surely if you are proposing this change, you have an immediate use for it?

Astralcircle:
Anything related to spawnmenu in sandbox. To work properly, you'll either have to override the entire file (which could be problematic over time with updates) or modify the spawnmenu after it's already created, which is quite complicated.
I encountered this problem when i was making a mods for the spawnmenu, which is why i created this PR

Astralcircle:
Bumping this because it seems to have quite a few positive reviews. Maybe you'il change your mind.
https://discord.com/channels/565105920414318602/565108080300261398/1475213553781702877

garryspins:
Id appreciate this greatly as a monitor hook, the return value I'm not so sure about being very useful but it could be for very specific scenarios.

PR #2301, opened 2025-05-31
Allow players spawn duped weapons in multiplayer

Astralcircle:
Weapons are not in the list of script entities, so they require separate logic. This PR will allow players in multiplayer to easily spawn weapon dupes


kirillbrest123:
i think that allowing to dupe weapons is not the best idea right now, considering that there's an exploit that allows you to freely dupe weapons with edited stats (#2265)

Astralcircle:
Yes, but this problem is now plaguing all SENTs too

PR #2302, opened 2025-06-02
Update dmenubar.lua

Seevvveen:
allows the :AddMenu() and :AddOrGetMenu() functions to also return the DButton they create on the menubar as well as the actual DMenu like normal. 

Shouldn't break anything since the menu is still returned first and old code just wont utilize the additional button return.

### DIFF
--- a/garrysmod/lua/vgui/dmenubar.lua
+++ b/garrysmod/lua/vgui/dmenubar.lua
@@ -29,8 +29,8 @@ end
 function PANEL:AddOrGetMenu( label )
 
 	if ( self.Menus[ label ] ) then return self.Menus[ label ] end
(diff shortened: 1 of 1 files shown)


garrysmod/lua/vgui/dmenubar.lua
```diff
@@ -29,8 +29,8 @@ end
 function PANEL:AddOrGetMenu( label )
 
 	if ( self.Menus[ label ] ) then return self.Menus[ label ] end
-	return self:AddMenu( label )
-
+	local menu, btn = self:AddMenu( label )
```
Seevvveen, line 32: actually this is probably a pointless assignment, we can probably return the function returns without the need for the two local vars

PR #2304, opened 2025-06-06
Use player.Iterator instead of player.GetAll

aStonePenguin:
Follow up to #2073

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -263,10 +263,9 @@ end
 
 -- These are totally in the wrong place.
 function player.GetByAccountID( ID )
-	local players = player.GetAll()
-	for i = 1, #players do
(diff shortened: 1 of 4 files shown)


robotboy655 (approved):

garrysmod/lua/includes/extensions/player.lua
```diff
@@ -286,21 +284,20 @@ end
 
 function player.GetBySteamID( ID )
 	ID = string.upper( ID )
-	local players = player.GetAll()
-	for i = 1, #players do
-		if ( players[i]:SteamID() == ID ) then
-			return players[i]
+
+	for _, pl in player.Iterator() do
```
Astralcircle, line 288: Will this break addons that do checks in EntityRemoved/OnEntityCreated?

Astralcircle:
After the latest changes it is 100% stable so it can be merged.

PR #2305, opened 2025-06-06
Move player.GetAll to lua

aStonePenguin:
It's significantly faster to create the table in lua than it is to push it from C.

### Benhmarks:

With 1 player:
```
old
0.13022569999885
new
0.0056074999993143
```
With 128 players:
```
old
1.1069849000014
new
0.29961660000117
```

``` lua
function player.GetAll2()
	local players = {}

	for _, pl in player.Iterator() do
		table.insert(players, pl)
	end

	return players
end

print 'old'
local s = SysTime()

for i = 1, 100000 do
	player.GetAll()
end

print(SysTime() - s)

print 'new'
s = SysTime()
for i = 1, 100000 do
	player.GetAll2()
end

print(SysTime() - s)
```

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -307,12 +307,13 @@ function player.GetBySteamID64( ID )
 	return false
 end
 
+local player_GetAll = player.GetAll
 local inext = ipairs( {} )
 local PlayerCache = nil
 
 function player.Iterator()
 
-	if ( PlayerCache == nil ) then PlayerCache = player.GetAll() end
+	if ( PlayerCache == nil ) then PlayerCache = player_GetAll() end
 
 	return inext, PlayerCache, 0
 
@@ -326,3 +327,13 @@ end
 
 hook.Add( "OnEntityCreated", "player.Iterator", InvalidatePlayerCache )
 hook.Add( "EntityRemoved", "player.Iterator", InvalidatePlayerCache )
+
+function player.GetAll()
+	local players = {}
+
+	for _, pl in player.Iterator() do
+		table.insert( players, pl )
+	end
+
+	return players
+end
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -331,8 +331,8 @@ hook.Add( "EntityRemoved", "player.Iterator", InvalidatePlayerCache )
 function player.GetAll()
 	local players = {}
 
-	for _, pl in player.Iterator() do
-		table.insert( players, pl )
+	for i, pl in player.Iterator() do
+		players[i] = pl
 	end
 
 	return players


Vitroze:
Hello, 
You might as well use the player.Iterator to avoid making a loop for nothing.

Malivil:
> Hello, You might as well use the player.Iterator to avoid making a loop for nothing.

The table returned by `player.Iterator` is meant to be read-only (per the documentation wiki). Copying it into a fresh table protects it.

Astralcircle:
This can't be done until the hook library has priorities, addons can call `player.GetAll` in `EntityRemoved` or `OnEntityCreated` hooks

Srlion:
> This can't be done until the hook library has priorities, addons can call `player.GetAll` in `EntityRemoved` or `OnEntityCreated` hooks

Yes this is correct, I don't use player.Iterator because it's not safe.
We need priorities AND protected calls for this to work properly.

Astralcircle:
> > This can't be done until the hook library has priorities, addons can call `player.GetAll` in `EntityRemoved` or `OnEntityCreated` hooks
> 
> Yes this is correct, I don't use player.Iterator because it's not safe. We need priorities AND protected calls for this to work properly.

Personally I use it calmly just thanks to your hook library + PRE_HOOK

Srlion:
Wait, how can luajit be way faster than compiled code? Is the C code causing reallocations because it's not using https://www.lua.org/manual/5.1/manual.html#lua_createtable with supplying the size or is it doing something so stupid? The calling of a C function shouldn't be too slow?
@robotboy655 could you confirm?

robotboy655:
It does in fact use `lua_createtable` with provided pre-allocation size. I guess the slowdown comes from using the internal wrappers for Lua stuff, as well as perhaps some extra stuff the function does.

image

Removing the internal wrappers and some other unnecessary code, these are the results (new3 is a modified C function, new2 is the code above, 127 bots)

Srlion:
I guess ents.GetAll can benefit from same stuff and have like 2000x performance improvement? lol

robotboy655:
Not 2000x, but seems to be around 2x.

image

Srlion:
That number is so huge really, considering lots of servers have like 2k+ entities lol

aStonePenguin:
> I guess ents.GetAll can benefit from same stuff and have like 2000x performance improvement? lol

If https://github.com/Facepunch/garrysmod-issues/issues/5800 ever gets solved.

Astralcircle:
> > I guess ents.GetAll can benefit from same stuff and have like 2000x performance improvement? lol
> 
> If [Facepunch/garrysmod-issues#5800](https://github.com/Facepunch/garrysmod-issues/issues/5800) ever gets solved.

This is unaffected by this

robotboy655:
The benchmark needs retesting when recent changes on `dev` beta.

PR #2307, opened 2025-06-08
Add hook PlayerFullyConnected

Vitroze:
The PlayerFullyConnected hook has been added to enable developers to ensure that the client receives the net. It will run only when the client is loaded. 

Admittedly, on the wiki, there are solutions already proposed for PlayerInitialSpawn, but the addition of this hook will make it possible to use a single hook instead of each developer creating his own hook or function.

### DIFF
--- a/garrysmod/lua/includes/extensions/player_auth.lua
+++ b/garrysmod/lua/includes/extensions/player_auth.lua
@@ -47,12 +47,27 @@ function meta:GetUserGroup()
 
 end
 
+if CLIENT then
+	hook.Add( "InitPostEntity", "PlayerAuthInitPostEntity", function()
+		net.Start( "GMODPlayerFullyConnected" )
+		net.SendToServer()
+	end )
+end
 
 --[[---------------------------------------------------------
 	This is the meat and spunk of the player auth system
 -----------------------------------------------------------]]
 
 if ( not SERVER ) then return end
+
+util.AddNetworkString( "GMODPlayerFullyConnected" )
+net.Receive( "GMODPlayerFullyConnected", function( len, ply )
+	if not IsValid( ply ) then return end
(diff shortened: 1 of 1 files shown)


noaccessl:
Have you taken into consideration those who may forcefully attempt to remove that client-side hook in which you make the client send a net-message to the server that will mark that client as fully connected?

You can use https://wiki.facepunch.com/gmod/gameevent/OnRequestFullUpdate for the purpose you've intended. The `Early Networking Example` can be developed further in that direction.

Vitroze:
> Have you taken into consideration those who may forcefully attempt to remove that client-side hook in which you make the client send a net-message to the server that will mark that client as fully connected?
> 
> You can use https://wiki.facepunch.com/gmod/gameevent/OnRequestFullUpdate for the purpose you've intended. The `Early Networking Example` can be developed further in that direction.

Hello, 
Thank you for reporting the problem. Now it's fixed, I've removed all client-server relationships in my program to avoid this kind of problem.

Astralcircle:
Looks very hacky and like something that should rather be implemented by the engine

RaphaelIT7:
> The PlayerFullyConnected hook has been added to enable developers to ensure that the client receives the net

The client will already receive net messages reliably/net messages being lost in PlayerInitialSpawn doesn't seem to be a thing, though sending net messages in PlayerInitialSpawn can cause them to be received before the client has received any entities.
So, this instead would be useful to send net messages that depend on entities clientside/when the client needs to be fully loaded & had had received everything.

Currently `OnRequestFullUpdate` might be unreliable for servers because of https://github.com/Facepunch/garrysmod-issues/issues/6465

I'd say that right now the first implementation with `InitPostEntity` & the client -> server networking would **currently** be the most reliable method until the issue with `OnRequestFullUpdate` would be fixed

PR #2310, opened 2025-06-09
Make ``ShowHelp``, ``ShowTeam``, ``ShowSpare1`` and ``ShowSpare2`` shared.

Winkarst-cpu:
https://github.com/Facepunch/garrysmod-requests/issues/100


Vitroze (commented):
You'll need to add the player's argument to the hook, as mentioned on the wiki.

robotboy655:
I will include this on the C++ side so it would still work if the current gamemode overwrites the hook you are using.

PR #2313, opened 2025-06-14
Update base lights.rad

goodusername123:
Updates the lights.rad file to the one from pre-Anniversary base HL2/Half-Life-2 replacing the ancient one that's currently present. This should act as a nice base for future PRs related to adding lights or texture shadows (I plan to submit at least one other PR if this one gets accepted). The only light added in the Anniversary update for HL2 is "white002_soft" however GMod does not currently have this file/light and I felt like keeping this PR pretty bare bones hence why I have not included it here in this exact PR.

Additionally 3 of the defined lights (defined in both the new and old lights.rad) are missing and have seemingly never existed with those being:
- `white005_nochop`
- `white006`
- `white006_nochop`

attached below are materials for the missing lights setup exactly how they would be if they actually existed:
[missing base white lights.zip](https://github.com/user-attachments/files/20734527/missing.base.white.lights.zip)

### DIFF
--- a/garrysmod/lights.rad
+++ b/garrysmod/lights.rad
@@ -1,57 +1,182 @@
-lights/fluorescentcool001a	189 231 232 350
-lights/fluorescentcool001b	236 255 182 350
-lights/fluorescentcool002a	189 231 232 400
-lights/fluorescentcool002b	236 255 182 400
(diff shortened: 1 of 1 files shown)


robotboy655 (approved):
Looks to be identical to the one we have from HL2 repo, so I will allow it.

PR #2315, opened 2025-06-21
Make `MAX_PLAYER_BITS` global too

Astralcircle:
Would be as useful as `MAX_EDICT_BITS`


binbonn:
Using GMOD_MAX_EDICT_BITS instead of MAX_EDICT_BITS would be better

as it would prevent conflicts with other addon/script.

Astralcircle:
I don't think that even one addon declares this constant

robotboy655:
Will be done in-engine where it is more appropriate.

PR #2317, opened 2025-06-23
Show ENT.PrintName in property editor panel title

Grocel:
Make the entity property editor show `ENT.PrintName` in the title.

Before the change:
old property panel

After the change:
new property panel

Changes:
- Output translated ENT.PrintName
- Still output entity index, so windows can be distinguished.
- Falls back to old output (`tostring()`)

### DIFF
--- a/garrysmod/lua/autorun/properties/editentity.lua
+++ b/garrysmod/lua/autorun/properties/editentity.lua
@@ -19,9 +19,18 @@ properties.Add( "editentity", {
 
 	Action = function( self, ent )
 
+		local printName = ent.PrintName
+
+		if printName then
+			printName = language.GetPhrase( printName )
+			printName = string.format( "%s [%d]", printName, ent:EntIndex() )
+		else
+			printName = tostring( ent )
+		end
+
 		local window = g_ContextMenu:Add( "DFrame" )
 		window:SetSize( 320, 400 )
-		window:SetTitle( tostring( ent ) )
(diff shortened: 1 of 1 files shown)


robotboy655 (changes requested):
Please see https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md regarding code formatting.

garrysmod/lua/autorun/properties/editentity.lua
```diff
@@ -19,9 +19,18 @@ properties.Add( "editentity", {
 
 	Action = function( self, ent )
 
+		local printName = ent.PrintName
+
+		if printName then
+			printName = language.GetPhrase( printName )
+			printName = string.format( "%s [%d]", printName, ent:EntIndex() )
+		else
+			printName = tostring( ent )
```
robotboy655, line 28: This could go through `language.GetPhrase` as well, some entities add their class as the localization string.

Grocel: I changed that it falls back to the entity classname now. The both cases are GetPhrase'ed now.

robotboy655: I meant use the class instead of print name if there's no print name, sorry.

Grocel: Yeah, someone pointed that out on Discord already. 😅

PR #2319, opened 2025-07-02
lights.rad: episodic additions

goodusername123:
Texlights from EP1 and EP2 with values converted to the newer engine styling/formatting between/for LDR and HDR values and such.
Forced model texture shadows from EP1 and EP2.
Also includes single exclusive texlight from CSS that seems to come from HL:S and doesn't actually exist in CSS's assets.

NOTE: both EP1 and EP2 contain a defined texlight for `red001` with different values however it only ever ended up being used within retail EP1 and the values for EP2's version are odd and so because of such this commit contains the values present in EP1.

### DIFF
--- a/garrysmod/lights.rad
+++ b/garrysmod/lights.rad
@@ -1,3 +1,5 @@
+xeno/crys_4b			255 160 128 40000
+
 ldr: lights/fluorescentcool001a	189 231 232 350
 hdr: lights/fluorescentcool001a	189 231 232 175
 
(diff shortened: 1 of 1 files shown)


robotboy655:
>  and doesn't actually exist in CSS's assets.

So why are you including it then? Does it exist in HL:S?

goodusername123:
> > and doesn't actually exist in CSS's assets.
> 
> So why are you including it then? Does it exist in HL:S?

Yes the light itself (materials and all) does exist in HL:S which means if someone mounts HL:S and attempts to use it then it would come into play on map compile, of course every other HL:S light is missing however I figured it would be better for at least one light to work rather then none of them work.
Part of my reasoning for including it is I'm unsure of what the policy is for additions to the lights.rad file from games that are mountable by GMod and so as a result this PR only includes things that are present in the lights.rad files from EP1, EP2, and CSS since as far as I understand assets from these titles will be included in the base game next update

robotboy655 (approved):

PR #2320, opened 2025-07-03
Add `player_manager.GetStoredPlayerClass`

wrefgtzweve:
Currently the only way to directly get a player class is doing something hacky like
```lua
    local _, tbl = debug.getupvalue( player_manager.GetPlayerClasses, 2 )
    local defTbl = tbl.player_sandbox
```
It'd be great if there was a base game function to just get the direct table

### DIFF
--- a/garrysmod/lua/includes/modules/player_manager.lua
+++ b/garrysmod/lua/includes/modules/player_manager.lua
@@ -303,6 +303,12 @@ function GetPlayerClasses()
 
 end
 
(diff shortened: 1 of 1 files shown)


robotboy655:
Why? Why does it need to be not a copy?

wrefgtzweve:
Because modifications to the table would need to be done on the original
Some examples i need the original for are changing how model selection works on spawn and allowing bhop in certain scenarios, both cases are in sandbox

PR #2153, opened 2024-11-15
Added Missing Strings

NovaDiablox:
- Added Missing Strings
- Deleted Unused String "Creations"
- Added localization support for the "Other" category (for everything that doesn't have it's own category) in places like the new game menu or ents, sweps and similar tabs.
- Added localization support for categories in ents, sweps and similar tabs.
- Added localization support for weapons descriptions labels (not the description itself) in the weapon selector.
- Added localization support the "New category" option in spawn menu (new category also gets the same localization)

Huge thanks to @Satton2 for helping me improve this PR.

PR #2322, opened 2025-07-08
Localizationfix

Astralcircle:
After #2153 there are still many addons that, for example, set their entity category to `Fun + Games` instead of `#spawnmenu.category.fun_games`, causing a split of categories if the game language is not english

This PR solves it


robotboy655 (approved):

NovaDiablox:
Hi! Could you add "Options" string too?

robotboy655:
Not sure what you mean by "Options" string? None of the spawnmenu tabs has an "Options" category by default.

NovaDiablox:
> Not sure what you mean by "Options" string? None of the spawnmenu tabs has an "Options" category by default.

Some addon creators create an "Options" tab next to the "Tools" tab and they don't use the existing #options string. This results in 2 duplicate tabs. One of them is translated while the other one is not.

robotboy655:
The options tab is not part of the base game, so I don't see a reason to add hacks for 3rd party mods.

PR #2323, opened 2025-07-08
Re-add Episode 2 soundscapes to gm_construct

goodusername123:
Brings back the EP2 soundscapes that used to be present in gm_construct since the sounds for them should always be present now (as of the coming update).


robotboy655:
I don't think so. The bird one is basically indistinguishable from the birds that are already there, and the wind gusts just feel out of place and weird.

PR #2324, opened 2025-07-08
Prevent DAdjustableModelPanel from handling MMB presses

TheObtey:
This PR adds a guard clause to 'DAdjustableModelPanel:OnMousePressed' to return early if the pressed button is 'MOUSE_MIDDLE'.

- By default, the parent DoMiddleClick function is empty, so middle mouse input should have no effect.
- Curiously, pressing the middle mouse button on a DAdjustableModelPanel would still trigger camera logic capture in DAdjustableModelPanel, which lead to unintended behavior (FOV reseting).
- This fix aligns the panel's behavior with expectations and prevents accidental camera or FOV changes due to middle mouse clicks.

Middle mouse button now does nothing in DAdjustableModelPanel, as expected.

### DIFF
--- a/garrysmod/lua/vgui/dadjustablemodelpanel.lua
+++ b/garrysmod/lua/vgui/dadjustablemodelpanel.lua
@@ -17,8 +17,8 @@ function PANEL:OnMousePressed( mousecode )
 
 	-- input.SetCursorPos does not work while main menu is open
 	if ( !MENU_DLL and gui.IsGameUIVisible() ) then return end
-	-- Prevent the FOV to reset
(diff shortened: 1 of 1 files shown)


robotboy655:
Your reasoning is very weird, but since there's no reason to have all extra mouse buttons work like right click, I suppose it can be disabled.

TheObtey:
Yeah sorry about my reasoning, I was really tired.
Let's just say that the DAdjustableModelPanel was doing some weird stuff when we MMB or RMB click.
It is just safer to disable those unnecessary actions.

PR #2325, opened 2025-07-08
Add functions GetValue1(), GetValue2(), GetValues() - CtrlNumPad

Vitroze:
Add GetValue1(), GetValue2(), GetValues() functions to vgui: CtrlNumPad

GetValue1() : Returns the first DBinder value
GetValue2(): Returns the second DBinder value
GetValues() : Returns both DBinder values. If the second DBinder is not created, this will return nil for the second return.


robotboy655 (approved):

robotboy655:
I am not sure what the point of `GetValues()` is, I removed it. I am not sure what cases there are where it would be convenient.

I made both other functions return KEY_NONE with no convar/panel, i.e. no key bound, same as `DBinder`'s default.

PR #2328, opened 2025-07-13
Run Javascript functions on HTML, safely

DarthTealc:
I've written a function to let you run a Javascript function on an HTML panel from Lua without having to deal with complex messy injection issues. We have [AddFunction()](https://wiki.facepunch.com/gmod/DHTML:AddFunction) for js->lua which works great, but doing lua->js has generally been tricky to do cleanly. This should help solve that.

I based on this on the concept of the loading screen. It currently has this:
```lua
pnlLoading.JavascriptRun = string.format( [[if ( window.GameDetails ) GameDetails( "%s", "%s", "%s", %i, "%s", "%s", %.2f, "%s", "%s" );]],
	servername:JavascriptSafe(), serverurl:JavascriptSafe(), mapname:JavascriptSafe(), maxplayers, steamid:JavascriptSafe(), g_GameMode:JavascriptSafe(),
	GetConVarNumber( "snd_musicvolume" ), GetConVarString( "gmod_language" ), niceGamemode:JavascriptSafe() )
```
which is later run with `self.HTML:RunJavascript( str )`

Using the new function from this PR, you could instead do this:
```lua
self.HTML:RunFunction( "GameDetails", 
	servername, serverurl, mapname, maxplayers, steamid, g_GameMode, 
	GetConVarNumber( "snd_musicvolume" ), GetConVarString( "gmod_language" ), niceGamemode  )
```
I'm not suggesting the loading screen be updated to do this, I'm just using it as an example of how it's easier for developers.

The function is available as both `RunFunction()` and `QueueFunction()`, which internally use [RunJavascript()](https://wiki.facepunch.com/gmod/Panel:RunJavascript) or [QueueJavascript()](https://wiki.facepunch.com/gmod/DHTML:QueueJavascript) depending on the one you choose. That's the only difference between RunFunction and QueueFunction, and I picked those names to match the existing name style.

Syntax is `html:QueueFunction( javascript function name to call, varargs to pass to it )` eg running `html:QueueFunction("myfunc", "hello world", true, false, 10/2)` will execute `myfunc("hello world", true, false, 5);` in the javascript.

The Javascript function name is restricted to alphanumeric characters, `.`, and `_`. All other characters in the name are stripped.

All the variable values inserted are protected by [JavascriptSafe](https://wiki.facepunch.com/gmod/string.JavascriptSafe)

The following data types are handled: Booleans, Numbers, Colours, Tables, and Strings. Some additional notes:
- Colours are converted to CSS #rrggbb (via `:ToHex()`) or rgba() colour strings, eg if you provide `Color(255,0,255,255)` the JS function will receive string `"#ff00ff"`, or if you provide `Color(255,0,255,153)` the JS function will receive string `"rgba(255,0,255,0.6)"` -  the rgba() should be changed to #rrggbbaa (via `:ToHex()`) once Awesomium is removed. 
- Tables are converted to json objects with `util.TableToJSON`. They can be read at the other end eg `jsonobj[0]` `jsonobj["rank"]`.

Suggestions and improvements are very welcome. 

## Demo

I have created a test panel to demonstrate functionality. The lua file is attached,
[htmltest.lua.txt](https://github.com/user-attachments/files/21229074/htmltest.lua.txt)
 please try running it along with this PR to see functionality. You can open or reload the panel with autorefresh or running concommand `htmltest`.

Here are some examples from the test code:

### Works for native JS functions
```lua
html:QueueFunction("console.log", "'hello world!' is a mandatory reference to make in any demo.")
```
> image

### Works for custom JS functions, and multiple parameters
lua: 
```lua
html:QueueFunction("GameInfo", LocalPlayer():Name(), game.GetMap(), engine.ActiveGamemode())
```
Javascript: 
```js
function GameInfo(player, map, gamemode) {
	var div = document.createElement("div");
	div.innerHTML = "Player <span class='ply info'></span> is on map <span class='map info'></span> in gamemode <span class='gm info'></span>";
						
	div.querySelector(".ply").innerText = player;
	div.querySelector(".map").innerText = map;
	div.querySelector(".gm").innerText = gamemode;
					
	document.body.appendChild(div);
}
```
Output:
> image

 
### Tables /  Json
```lua
html:QueueFunction("ExplainJson", { ["hero"] = "freeman", ["villain"] = "combine", ["episode"] = 3 } )
```
> image

### Demo of various types
```lua
html:QueueFunction("ExplainType", "this is a string")
html:QueueFunction("ExplainType", true)
html:QueueFunction("ExplainType", false)
html:QueueFunction("ExplainType", 1234)
html:QueueFunction("ExplainType", -2/3)
html:QueueFunction("ExplainType", Color(255,0,255,255) )
html:QueueFunction("ExplainType", Color(255,0,255,153) )
```
> image
(note: code has been updated since screenshot taken, Color(255,0,255,255) will now return #ff00ff)


garrysmod/lua/vgui/dhtml.lua
```diff
@@ -63,6 +63,73 @@ function PANEL:Call( js )
...
+local function BuildFunction( func, ... )
+
+	-- Build a Javascript-safe JS function call.
+	-- To be run by either panel:RunFunction or panel:QueueFunction
+	--
+	-- First parameter is the JS function name,
+	-- additional parameters are the values to provide to the function call.
+	-- The values provided to the function call are treated as strings except where noted below.
+	
+	local args, amount = table.Pack( ... )
```
Grocel, line 75: wouldn't it not just be `local args = {...}` instead, as `amount` seems to be used.

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -63,6 +63,73 @@ function PANEL:Call( js )
...
+			
+		elseif istable( v ) then -- Tables, convert to json string
+		
+			formatArgs[ k ] = '"%s"'
+			safeArgs[ k ] = string.JavascriptSafe( util.TableToJSON( v ) )
+			
+		else -- Strings, and all else treated as strings
+		
+			formatArgs[ k ] = '"%s"'
+			safeArgs[ k ] = string.JavascriptSafe( v )
```
Grocel, line 110: You might want to add `tostring( v )`. So it looks like this. `string.JavascriptSafe( tostring( v ) )`. I am not sure if string.JavascriptSafe does it for you.

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -63,6 +63,73 @@ function PANEL:Call( js )
...
+		elseif isnumber( v ) then -- Numbers
+		
+			formatArgs[ k ] = '%s'
+			safeArgs[ k ] = v
+			
+		elseif IsColor( v ) then -- Colors, convert to CSS. Should be updated to use srlion's colour conversion PR when available.
+		
+			formatArgs[ k ] = '"%s"'
+			if v.a == 255 then -- don't include alpha if it isn't needed
+				safeArgs[ k ] = "rgb( " .. v.r .. ", " .. v.g .. ", " .. v.b .. " )"
```
Grocel, line 96: I would like to recommend to change it to `string.format("rgb(%d,%d,%d)", v.r, v.g, v.b)`. `%s` would work to, but I think numbers should be formatted as numbers.

garrysmod/lua/vgui/dhtml.lua
```diff
@@ -63,6 +63,73 @@ function PANEL:Call( js )
...
+			else
+				safeArgs[ k ] = "rgba( " .. v.r .. ", " .. v.g .. ", " .. v.b .. ", " .. ( v.a / 255 ) .. " )"
+			end
+			safeArgs[ k ] = string.JavascriptSafe( safeArgs[ k ] )
+			
+		elseif istable( v ) then -- Tables, convert to json string
+		
+			formatArgs[ k ] = '"%s"'
+			safeArgs[ k ] = string.JavascriptSafe( util.TableToJSON( v ) )
```
Grocel, line 105: So far I can tell this is correctly implemented. It will result in a valid json string. When the function runs on the JS side it will arrive as an unescaped ready to use json string. `string.JavascriptSafe` makes it save to pass in `RunJavascript/QueueFunction`.

Grocel (changes requested):
That's mosty fine, but I have some recommendations you may want to look at.

DarthTealc:
> That's mosty fine, but I have some recommendations you may want to look at.

Thanks for the recommendations! Made those changes with notes above.

DarthTealc:
Here's a draft for a wiki page, based on https://wiki.facepunch.com/gmod/DHTML:QueueJavascript and other similar pages. Does not include example section yet.

```html

<function name="QueueFunction" parent="DHTML" type="panelfunc">
	<description>
		Runs/Executes a JavaScript function with prepared parameters in a html panel.
		
		<note>This function does **NOT** return 'return' from JavaScript to Lua. If you wish to pass/return values from JS to Lua, you will need to have the JavaScript side call a function you've registered with <page>DHTML:AddFunction</page></note>
		<note>`QueueFunction` and <page>DHTML:RunFunction</page> are identical, except `QueueFunction` uses <page>DHTML:QueueJavascript</page> internally, and <page>DHTML:RunFunction</page> uses <page>panel:RunJavascript</page> internally. See those articles for information on the difference.</note>

	</description>
	<realm>Client and Menu</realm>
	<args>
		<arg name="function" type="string">Name of the JavaScript function to run. Restricted to letters/numbers, `.`, `_`.</arg>
		<arg name="args" type="vararg">Argument values to provide to the JavaScript function. Values will be converted to a format readable by JavaScript.
			Lua Type | JavaScript Type | Note
			----|----|----
			<page>string</page> | string | 
			<page>boolean</page> | boolean | 
			<page>number</page> | number | Decimal numbers may or may not have the same amount of decimal places. Consider converting to/from string yourself if precision is important.
			<page>color</page> | string, formatted as CSS color | Will be converted to `"#RRGGBB"` with <page>Color:ToHex</page> if alpha is 100% or not present, or `"rgba(255,255,255,1)"` if alpha is under 100% (for Awesomium compatibility). Note that in future `"rgba()"` strings will be replaced with `"#RRGGBBAA"` strings once Awesomium has been fully replaced with CEF.
			<page>table</page> | json string | <page>Table</page>s will be converted to a JSON object using <page>util.TableToJSON</page>. See TableToJSON page for notes. In JavaScript you can access the JSON object's properties directly or loop over it.
			All other types | string | Will attempt to convert to string with <page>Global.tostring</page>. You should convert to a suitable string format yourself to avoid unpredictable behaviour and unconvertible data.
		</arg>
	</args>
</function>

```

DarthTealc:
I've updated table handling, the json is now no longer a json string, it's now a json object. The original message and demo/test code (htmltest.lua) has been updated.
> image


The test code has also been updated to demonstrate how you might want to pass things in as a table for more programmatic access.
```lua
local myPos = LocalPlayer():GetPos()
html:QueueFunction("ExplainType", myPos )
html:QueueFunction("ExplainType", myPos:ToTable() )
```
> image

This includes colours, if you don't want a CSS colour and would prefer something where you can access the individual properties.
```lua
html:QueueFunction("ExplainType", Color(255,0,255,153) )
html:QueueFunction("ExplainType", Color(255,0,255,153):ToTable() )
```
> image

Grocel (approved):

WinterPhoenix:
This is pretty nice! Great work

DarthTealc:
I [updated colour handling](https://github.com/Facepunch/garrysmod/pull/2328/commits/dffbd7461eb74971a285a2503db84e288f6bc44b) to use the function now available from https://github.com/Facepunch/garrysmod/commit/a1a53cdaee546357e21bafb421e08395286f138b

I've somehow both rebased AND merged recent Gmod commits into this branch, meaning ctrlnumpad.lua and color.lua are both now listed, even though dhtml.lua is the only file that should be changed. If Rubat wants me to resubmit without that mess, I've created this branch https://github.com/Facepunch/garrysmod/compare/master...DarthTealc:garrysmod:htmlfunction2 with just the intended commits.

Grocel (approved):

DarthTealc:
Wanted to note that this works without any parameters, although there'd be no real point doing it.
eg ```html:QueueFunction("example")``` will call ```example();``` in the JS.

DarthTealc:
Re-based and re-submitted as #2344

PR #2335, opened 2025-07-23
Improve main menu's bottom buttons bar

DarthTealc:
I've updated the main menu's bottom buttons bar to use pre-standardised flex "-webkit-box" (so it works in Awesomium) instead of what it was doing before with floats and absolutes. The visual result is identical to the current game but is cleaner structurally.

The reason started doing this was because of suggestions the 'middle buttons' are weirdly positioned. I agree and I've made it very easy to change if wanted. 

image

(Top) - right aligned with 10% gap - this is how the game currently looks AND how this PR looks.
(middle) - center aligned - this can be done by changing `#NavBar .group.center`'s `-webkit-box-pack` to `center`
(bottom) - left aligned with 10% gap - this can be done by changing `#NavBar .group.center`'s `-webkit-box-pack` to `start`

My personal preference would be:
image
This can be done by changing `#NavBar .group.center`'s `-webkit-box-pack` to `start`, and `margin: 0 10%;` to `margin: 0;`

I didn't make those changes directly or any style changes as I expect some disagreement and thought it'd be best to just present the options for Rubat's decision. Even with no change, the layout handling is cleaner.

### DIFF
--- a/garrysmod/html/menu.html
+++ b/garrysmod/html/menu.html
@@ -58,8 +58,8 @@
 		<div class="button" ng-show="InGame" ng-click="Disconnect()"><a href="#/"><img src='img/disconnect.png'><span ng-Tranny="'disconnect'"></span></a></div>
 		-->
 		<div class="button bigicon" ng-show="InGame" ng-click="BackToGame()"><a href="#/"><img src='img/back_to_game.png' loading="lazy"><span ng-Tranny="'back_to_game'"></span></a></div>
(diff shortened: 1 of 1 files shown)


Seevvveen (approved):

Seevvveen:
so ignore me, apparently git targeted the official repo instead of my private one, my bad

robotboy655 (changes requested):
Seems to work fine except for this regression:

August08-1622-gmod

`dev` beta, joining a local dedicated server. It shouldn't show both favorite buttons.

DarthTealc:
Fixed! It should now behave correctly.

PR #2339, opened 2025-07-25
motor bug fix

AlexG345:
Fixes very annoying bug where rotating a motor constraint before having activated it at least once causes the torque to be applied on a wrong axis.


robotboy655:
Fixes https://github.com/Facepunch/garrysmod-issues/issues/6435

PR #2341, opened 2025-07-31
Server list: Added option to "hide outdated servers"

DarthTealc:
image

This option, when ticked, filters out any servers that are outdated and are therefore not able to be connected to. 

This is done by hiding any servers that are using a network version older than your client. Technically a more accurate label would be "hide servers using an older networking version than my game client" but I think that's a bit much.
This does mean if your game itself is outdated, you will still see servers that are technically outdated but compatible with your game, as your game doesn't consider them outdated. And you'll see servers marked "future" that have updated.


DarthTealc:
Filter option was requested by @NovaDiablox

robotboy655 (approved):

PR #2342, opened 2025-07-31
Add ents.FindInCube utility function

riggs9162:
This pull request introduces a new utility function to the `ents` library in `garrysmod/lua/includes/extensions/ents.lua`. The new function simplifies finding entities within a cube-shaped area.

### Additions to the `ents` library:

* `ents.FindInCube(center, radius)`: Added a new function that calculates a cube based on a center point and radius, and uses `ents.FindInBox` to return entities within that cube. This provides a more intuitive way to search for entities in a cubic area.

### DIFF
--- a/garrysmod/lua/includes/extensions/ents.lua
+++ b/garrysmod/lua/includes/extensions/ents.lua
@@ -18,3 +18,11 @@ function ents.FindByClassAndParent( classname, entity )
 	return out
 
 end
+
+function ents.FindInCube( center, radius )
+
+	local rvec = Vector( radius, radius, radius )
+
+	return ents.FindInBox( center - rvec, center + rvec )
+
+end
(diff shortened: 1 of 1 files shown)


bloodycop6385 (approved):

Be1zebub (commented):
Kinda pointless, just micro wrapper around `ents.FindInBox` - no real value.

riggs9162:
> Kinda pointless, just micro wrapper around `ents.FindInBox` - no real value.

It is a simple convenience purpose function meant to replace FindInSphere for some developers.

FindInSphere is generally from my tests back then more expensive than a cubic based implementation like done here.

Yes it is a micro wrapper, but with good intentions, it uses the same parameters for FindInSphere, but it does the calculations in a cubic matter using FindInBox.

robotboy655:
I don't think this is necessary in the base game.

PR #2343, opened 2025-08-02
Add spawnmenu icon to npc_tf2_ghost

thecraftianman:
This just adds a spawnmenu icon to the example NPC so that it isn't totally blank.

image


DarthTealc:
I like the idea of adding an icon instead of being blank. But I don't think it should be the same as Mossman's spawn icon since it doesn't really act like her. Might be better taking a screenshot of the bot itself, maybe when it's sitting down since it does that repeatedly?

thecraftianman:
That could also be an option. I just reused the Mossman icon because it's already in the game and the bot just uses her model. I'll wait for Rubat's feedback on what icon should be used before I change anything here.

robotboy655:
I think the icon should be unique, and also I really hate the classname of that NPC, so I will probably end up renaming it.

PR #2348, opened 2025-08-10
Menu: Minor code optimisation

DarthTealc:
- Replaced `ToggleGamemodes`/`ToggleLanguage`/`ToggleGames` with the existing `TogglePopup` as they're all essentially the same thing
- Reduced duplication in the translation code. Also stopped adding a translated `placeholder` attribute if the element doesn't actually support the `placeholder` attribute.
- Replaced `jquery-3.7.1.min.js` with `jquery-3.7.1.slim.min.js` - the difference is the slim version doesn't contain ajax and 'effects', which aren't being used by Gmod anyway. This reduces the file size and load of jquery slightly. Source available here <https://code.jquery.com/jquery-3.7.1.slim.min.js> to verify what I've included is authentic and unmodified.


DarthTealc:
Rebased for clean merge with recent changes

DarthTealc:
Working on some better and more effective optimisation, so I'm closing this to include as part of that.

PR #2349, opened 2025-08-12
Add new table extensions

Bagellll:
I've made 2 simple table lib extensions.

`table.Swap` simply swaps 2 key-value pairs.

`table.SwapRemove` behaves like `table.remove` except when removing a value it puts the last value of the table where the removed value used to be. The last value is replaced with `nil`. In tables where you do not care about the order of elements after removal, this should be useful and even preferred.

### DIFF
--- a/garrysmod/lua/includes/extensions/table.lua
+++ b/garrysmod/lua/includes/extensions/table.lua
@@ -778,6 +778,27 @@ function table.Flip( tab )
 
 end
 
+function table.SwapRemove( tab, index )
+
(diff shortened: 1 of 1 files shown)


CornerPin:
Does `table.Swap` really needs to exist when Lua supports multiple assignment? You can simply do `tbl[key1], tbl[key2] = tbl[key2], tbl[key1]`.

I think `table.SwapRemove` looks useful.

bloodycop6385 (approved):

PR #2350, opened 2025-08-18
Update color.lua

InfernoLua1337:
a faster and easier HexToColor rework than srlion's


Astralcircle:
Srilion's version is fully jited and therefore fast, yours is not. 
+Doesn't support alpha

PR #2351, opened 2025-08-18
Create surface.lua

InfernoLua1337:
Cached surface.GetTextSize for performance


robotboy655:
We do not need to cache every single "get" function in the whole game. (At expense of memory usage)

PR #2352, opened 2025-08-22
Add CapabilitiesHas method to NPC metatable

bloodycop6385:
Introduces a CapabilitiesHas function for NPCs, allowing capability checks using bitwise operations. This extends the NPC metatable for easier capability management.


DrVrej:
Helpful function 👍 however it would be better if it was done in the engine like rest of the NPC functions, plus we are dipping into C++ by calling CapabilitiesGet, might as well just do the whole thing in C++

bloodycop6385:
Well, whatever @robotboy655 decides

robotboy655:
I do think that this should be added in-engine.

PR #2353, opened 2025-08-30
TTT: Raise passive equipment item limit from 16 to 32

figardo:
Also includes net message optimizations to save 12 bits on TTT_Equipment and 14 bits on TTT_BoughtItem by default.

If the bitwise library is ever updated to support numbers greater than a 32-bit signed int in the future, the limit could in theory be raised to 53. However, 32 items is already way more than plenty.


robotboy655:
Code looks more or less fine to me, but feels like also a balancing change as well, so would like @svdm's input on this.

figardo:
> feels like also a balancing change as well

The limit being raised will only affect servers with custom passive equipment addons, at which point the game balance is being modified by the server anyway.

bloodycop6385:
Why don't we get that `util.` function as a PR?

PR #2356, opened 2025-09-05
TTT: Russian localization update

Satton2:
- Localized strings that were added in: https://github.com/Facepunch/garrysmod/pull/2354

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/lang/russian.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/lang/russian.lua
@@ -138,6 +138,10 @@ L.radio_button_huge    = "H.U.G.E"
 L.radio_button_c4      = "Пиканье С4"
 L.radio_button_burn    = "Горение"
 L.radio_button_steps   = "Шаги"
+L.radio_button_glock   = "Glock"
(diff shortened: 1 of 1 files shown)


robotboy655:
image

I have made some changes to specify that these are gun shot sounds, instead of just saying the name of the weapon.

Let me know if you think that's wrong in any way.

Satton2:
> I have made some changes to specify that these are gun shot sounds, instead of just saying the name of the weapon.
> 
> Let me know if you think that's wrong in any way.

I would like to have a full phrase, instead of shortened ones, but not having enough space is the exact reason why I shortened it in the past to just weapon names.

If we really are into fitting a full phrase, let's make a new line then. Or make buttons a bit wider (there is plenty of space).
image

figardo:
> If we really are into fitting a full phrase, let's make a new line then. Or make buttons a bit wider (there is plenty of space).

It seems like the buttons were made 100px wide when the equipment panel as a whole was thinner, so I don't see any issue with increasing the button width nowadays.

For reference, this is how it used to look with the thinner panel (in GM12):
image

robotboy655 (approved):

PR #2357, opened 2025-09-06
Fix bug with "P2P - Friends Only" toggle not being saved on New Game

DarthTealc:
Currently when starting a new game with P2P's "Friends Only" toggle, the moment you start the server and look back on the New Game page, it has already been unchecked. 

This PR fixes that so the setting is retained, in line with how Local and P2P are saved. ~Additionally the "Friends Only" toggle is hidden when P2P is disabled.~

image

Video: (visuals slightly outdated, functionality is accurate) 

https://github.com/user-attachments/assets/695ab6dd-1000-4933-b815-0fdb6e0a8ffd

### DIFF
--- a/garrysmod/resource/localization/en/main_menu.properties
+++ b/garrysmod/resource/localization/en/main_menu.properties
@@ -63,8 +63,8 @@ gmod_suit=Enable HEV Suit
 realistic_falldamage=Enable realistic fall damage
 
 lan_server=Local Server
-p2p_server=Peer-To-Peer\: Friends Only
-p2p_server_friendsonly=Friends Only
+p2p_server=Peer-To-Peer
+p2p_server_friendsonly=Peer-To-Peer\: Friends Only
 server_name=Server Name\:
 server_password=Server Password\:
 maxplayers_1=Single Player


robotboy655:
I don't think the option should be hidden with P2P disabled. In might make sense on paper, but I think it would just hide the option existing for no gain.

DarthTealc:
The reason for that is otherwise the option is set to a 'disabled' state, which is fine, but if it was checked it remains checked which might confuse some users who don't realise the disabled state means it's being ignored. That's why I made it hidden instead. Alternatively perhaps it could use a strikethrough when disabled?

DarthTealc:
I've updated it so the "Friends Only" item is no longer hidden. It's now indented so that it's clearly a sub-setting of the Peer-to-Peer setting, which makes it clearer that when p2p is off the setting is being ignored.

https://github.com/user-attachments/assets/695ab6dd-1000-4933-b815-0fdb6e0a8ffd

Also fixed the alignment of other fields in relation to this. Only noticeable one for users is demonstrated below with an orange line.
Left = current version of game. Right = this PR
image

robotboy655:
The alignment changes look weird when there's no scrollbar:
image

(The uneven padding on left and right)

It should done by moving the small text boxes, not the other way around I think.

DarthTealc:
Here's the fixed version + how it looks when scrollbar isn't present + with alignment lines.
image

robotboy655 (approved):

robotboy655:
I have decided to go for a different approach with the scrollbar margin on the right, like so:
September19-1747-gmod

This has added benefit of the settings staying the same size with and without scrollbar, and also technically consistent with the map list scrollbar.

PR #2359, opened 2025-09-11
TTT: Added ability to use parameters when defining radio sound names

Malivil:
Reviewed with @figardo prior to submitting

The updated translation logic preserves the fallback to the untranslated name when the translation string does not exist

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_radio.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_radio.lua
@@ -1,6 +1,7 @@
 --- Traitor radio controls
 
 local GetTranslation = LANG.GetTranslation
+local GetPTranslation = LANG.GetParamTranslation
 local TryTranslation = LANG.TryTranslation
 
 TRADIO.SoundOrder = {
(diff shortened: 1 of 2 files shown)


robotboy655:
What is this meant to achieve?

Malivil:
> What is this meant to achieve?

It allows addons that add custom sounds to have placeholders in their translation strings.
In my particular case, I use it to replace role names (which are configurable by the user) within the translation for a radio sound related to that role.

robotboy655:
The sound would be the exact same every time though, no? Can you give a direct example of how the text would change?

Malivil:
In my specific case, the role name is configured by convar. If they changed the convar, the string on the UI would update without having to go in and change the translation string as well.

robotboy655:
Sounds like a solution to a problem only you have though.

I will let @svdm decide.

@svdm if you do not wish to be bothered about these PRs (in general), do let me know.

Malivil:
Yea, that's fair.
I just figured since I needed it, there may be other addon authors with a similar need so I submitted it as a PR.

svdm:
I can't entirely envision the use case but the change is small and seems harmless.

> @svdm if you do not wish to be bothered about these PRs (in general), do let me know.

I don't mind, there's not that many.

PR #2362, opened 2025-09-22
Added a variable that allows hiding help for specific tools

Vitroze:
I noticed that in GMod, it is not possible to hide the help for certain specific tools natively.

I decided to add a HideHelp variable to the tool that allows you to hide (if it is true) the HUD.

```lua
TOOL.HideHelp = true
```


robotboy655:
What is this for? If the user wants to hide the help text, they can via the convar/option in menu bar.

Why would tool want to hide the help text itself?

Vitroze:
> What is this for? If the user wants to hide the help text, they can via the convar/option in menu bar.
> 
> Why would tool want to hide the help text itself?

If a developer wants to modify the HUD help to adapt it to the tool design. The main problem with the menu bar is that it hides all HUD Help (tools) and cannot be limited to some tools natively.

PR #2370, opened 2025-10-09
Don't call MsgAll in the PlayerDeath hook if SendDeathNotice returned anything 

Winkarst-cpu:
Also replaces all the ``GM:SendDeathNotice`` calls with the ``hook.Run``.


bloodycop6385 (approved):

bloodycop6385 (approved):
LGTM

CornerPin:
Is introducing a new hook necessary? You could replace direct `GM.SendDeathNotice` calls with `hook.Run` and then print using `MsgAll` if the returned value is `nil`.

CornerPin (approved):
Looks good to me.

Winkarst-cpu:
Bump

PR #2375, opened 2025-10-19
Change LocalPlayer function to cache local player instance

Srlion:
It's not a huge difference and does not allow it to be jitted, but any improvement for a function that is used either in Think/Paint hooks is a win, benchmark result:
```
New: 4.50590 ms (Average: 0.45059 ns)
Old: 317.08870 ms (Average: 31.70887 ns)
Fastest vs slowest: 6937.2% slower (70.37x difference)
```

### DIFF
--- a/garrysmod/lua/includes/extensions/client/globals.lua
+++ b/garrysmod/lua/includes/extensions/client/globals.lua
@@ -12,3 +12,13 @@ function ScreenScaleH( height )
 end
 
 SScale = ScreenScale
+
+local localPly = nil
+C_LocalPlayer = LocalPlayer
+function LocalPlayer()
+	if ( not localPly ) then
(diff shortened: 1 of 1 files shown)


noaccessl:
But does allow to be JIT-compiled once stops jumping to a (`Lua -> C -> Lua`)-call?

Or it's not how it works?

Srlion:
> But does allow to be JIT-compiled once stops jumping to a (`Lua -> C -> Lua`)-call?
> 
> Or it's not how it works?

Eh, on luajit 2.1 it cane be jitted, but I think it insta gets blacklisted on 2.0.4, but I benchmarked on beta branch so I don't know how faster it can be.

bloodycop6385:
bump

PR #2387, opened 2025-11-11
Add Category Headers for Other Content Types

Zaurzo:
This PR adds the option to give headers (seen for props in the spawnlists) for other content types. This is implemented for NPCs, Weapons, Entities, Vehicles, and even Post Processing. This allows for much more organized categories, especially for ones with a lot of content icons with a wide variety.

### Demonstration: (this is not a change in the PR)
demonstration

To set a header the content icon should be under, use the `ContentHeader` field. Just like categories, this is auto-translated. For Post Processing, the field is called `contentHeader` instead, to keep it consistent with the camelCase style of the `postprocess` content type. If there is an entry with a header present, all other entries that do not have a header set will be put at the very top of the list.

### Note
- I noticed that a lot of code is copy-pasted for how these creation tabs are populated, so this PR also adds an internal helper function `PopulateFromList` to `SpawnmenuContentPanel` that does it all. This cleans up a lot of code, and makes it easier to change how these creation tabs are populated across multiple different content types.
- This PR also fixes a small issue with ordering. Since default weapons have their print names set to translatable strings, it would sort them based off that, and not the translated versions. It will now sort them based off the translated names; however, I can revert this change if the current behavior is wanted.


robotboy655:
While I do agree about your code duplication point, I do not think the new function belongs in the spawnmenu library. It should stay within the Sandbox gamemde, probably on whatever panel type all the creation tabs are based on.

Zaurzo:
> While I do agree about your code duplication point, I do not think the new function belongs in the spawnmenu library. It should stay within the Sandbox gamemde, probably on whatever panel type all the creation tabs are based on.

Got it. I have moved it to `SpawnmenuContentPanel`, is that better?

Grocel:
I was looking for that kind of functionally in the past. I actually managed to implement these my self in an addon base. But it was quite an endeavor to get it done. I would like to see the implementation in the base game.

robotboy655 (approved):

robotboy655:
I have reworked the internals a bit, taking it further with reducing code duplication, and making the auto refresh stuff easier to reuse for other entity types later.

It works exactly as it did before, but I also renamed `CategoryHeader` to `SubCategory` and removed the lower case variant for Post Processing effects, because screw it.

I have put it into `dev` beta with todays update, but I am not merging this yet to let you provide feedback if you so choose. @Zaurzo 

I think `SubCategory` is better if we choose to add sub categories into the tree structure at some later point. 

I have also used this for the base game Entities tab now, as an example, and just because it makes sense to use it there:
image

robotboy655:
One thing I am noticing with fixed alphabetical sorting is that certain entities that used to be grouped are no longer, such as the combine soldiers here:
image

Zaurzo:
> One thing I am noticing with fixed alphabetical sorting is that certain entities that used to be grouped are no longer, such as the combine soldiers here: image

Ah, yeah. Things like grouping and putting important characters first was why I included that sort order field in my previous PR. That old PR was really messy and weird, and should've been split into multiple instead of just one trying to add three new features at once.

Zaurzo:
> I have reworked the internals a bit, taking it further with reducing code duplication, and making the auto refresh stuff easier to reuse for other entity types later.
> 
> It works exactly as it did before, but I also renamed `CategoryHeader` to `SubCategory` and removed the lower case variant for Post Processing effects, because screw it.
> 
> I have put it into `dev` beta with todays update, but I am not merging this yet to let you provide feedback if you so choose. @Zaurzo
> 
> I think `SubCategory` is better if we choose to add sub categories into the tree structure at some later point.
> 
> I have also used this for the base game Entities tab now, as an example, and just because it makes sense to use it there: image

Thanks for the extra rework and cleanups. I briefly tested it out and all seems to be working just fine

PR #2391, opened 2025-11-14
Adding a cooldown to the camera

Vitroze:
**_Hello,_**
I added a cooldown to the camera to avoid sound and effect spam. Some users who don't have powerful PCs find that players spam the camera a lot, which can cause their PCs to **crash** or **freeze**. 

## Why a cooldown of 0.75 seconds?
I decided to set a ‘high’ cooldown to avoid disturbing players during their gaming sessions. Anything less than this value is too fast and the effect will not end with a ‘rest’ period.

Nevertheless, I propose two solutions for the system: 
## Solution 1
Either we stick with the current code, which limits the total use of the camera with a cooldown of 0.75 seconds.

## Solution 2
Or I create a cooldown only on the DoShootEffect function, which allows players to spam photos but greatly limits the spamming of effects and sound, which can be annoying for the jo

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/weapons/gmod_camera.lua
+++ b/garrysmod/gamemodes/sandbox/entities/weapons/gmod_camera.lua
@@ -84,7 +84,6 @@ end
 function SWEP:PrimaryAttack()
 
 	self:DoShootEffect()
-	self:SetNextPrimaryFire( CurTime() + 0.75 )
 
 	-- If we're multiplayer this can be done totally clientside
 	if ( !game.SinglePlayer() && SERVER ) then return end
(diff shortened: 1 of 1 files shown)


Jakob-bb:
solution 2 sounds better to me simply because it's the least intrusive of the two.

robotboy655:
I think applying the cooldown only for the effect and sound is the better solution here. I really doubt the ridiculous claim that it can crash or freeze PCs, but I suppose you shouldn't be able to spam it as much as you can right now.

Jakob-bb:
yeah I've never heard of it causing crashes whatsoever.

robotboy655 (approved):

PR #2400, opened 2025-11-28
Add `math.Cot` function

Akabenko:
Returns cotangent of number. Useful for calculations of scaling factor using FOV.
Usage example:
```lua
local viewSetup = render.GetViewSetup()
local fov = viewSetup.fov
local factor = math.Cot( math.rad( fov * 0.5 ) )
```


Grocel (approved):

bloodycop6385 (approved):

robotboy655 (approved):

robotboy655:
This is very borderline useless addition, but since we already have `math.tau`, might as well I guess.

PR #2407, opened 2025-12-12
Added the GetModel function (SpawnIcon)

Vitroze:
Hello,
I just added an alias for the GetModelName function.

### DIFF
--- a/garrysmod/lua/vgui/spawnicon.lua
+++ b/garrysmod/lua/vgui/spawnicon.lua
@@ -120,6 +120,10 @@ function PANEL:SetModel( mdl, iSkin, BodyGroups )
 
 end
 
+function PANEL:GetModel()
+	return self:GetModelName()
+end
+
 function PANEL:RebuildSpawnIcon()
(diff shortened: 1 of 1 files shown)


robotboy655:
Why?

Vitroze:
Based on the logic of the code, without necessarily looking at the wiki, I thought that GetModel could work on a SpawnIcon (given that the function for changing the model is SetModel). From a logical point of view, I thought it would be better to implement an alias.

After that, it's just for the logic of the code.

Astralcircle:
I hate aliases... A striking example is the `:Name` and `:GetName` functions, which are aliases of the `:Nick` function, providing nothing but a little code confusion

Vitroze:
Okay, but in terms of code logic, it would be better to implement a GetModel alias because if the developer thinks it works like DModelPanel:GetModel(), they will be mistaken. It's just to follow the logic of the vgui DModelPanel. After that, I agree that adding 40 aliases could be problematic.

PR #2408, opened 2025-12-13
Added GM:OnEntityPasted(ent, entDupeTable)

Grocel:
Works like `ENTITY:OnDuplicated(ent, entDupeTable)` but doesn't require adding/detouring `ent.OnDuplicated` on `OnEntityCreated` calls for addon-foreign entities.

This adds a easy way to add generic logic early to the duplication life cycle. It is called a way before `duplicator.RegisterEntityModifier()` added callers are being iterated. The new hook also reduces the risk of conflicts caused by detours of pre-existing `ent.OnDuplicated` code.

Test script:
```lua
if SERVER then
    hook.Add("OnEntityPasted", "Test", function(ent, entTable)
        print(ent)
        PrintTable(entTable)
    end)
end
```


marchc1:
Would be very useful to have this hook coming from the perspective of a developer on ACF-3. We've wanted something like this for a while, iirc

thegrb93:
The name should be something more like PostDuplicatorPaste which is more consistent with the PostEntityPaste hook.

Grocel:
> The name should be something more like PostDuplicatorPaste which is more consistent with the PostEntityPaste hook.

Don't think it would reflect what it does. It is called in the beginning if the duplication process (even before ent:Spawn()), not the end of it. The name is being picked as an analog to of the entity's hook name.

However what do you think of naming it `OnEntityDuplicated` instead?

thegrb93:
Oh I thought you were adding a post paste hook, which gmod also needs. Nvm

Grocel:
> Oh I thought you were adding a post paste hook, which gmod also needs. Nvm

If this one gets of the ground, I can submit PRs adding those for both projects as well.

robotboy655:
My concern with this is cost of execution. [This is already a problem for large dupes.](https://github.com/Facepunch/garrysmod/commit/ecd97269fab43a4f3524009180f9fe2b84fa758f)

A post-duplication hook would be a safer bet, even if it would be more annoying to work with to modify a single entity entry.

Grocel:
> My concern with this is cost of execution. [This is already a problem for large dupes.](https://github.com/Facepunch/garrysmod/commit/ecd97269fab43a4f3524009180f9fe2b84fa758f)
> 
> A post-duplication hook would be a safer bet, even if it would be more annoying to work with to modify a single entity entry.

1. It should be O(n) for entity count of dupe, or O(1) for a single entity. Similar how `ent.OnDuplicated()` already is.
2. Adding a post hook after dupe would not do this justice as it is specifically asked to have a hook that runs early on the pipeline and per entity.
3. A hook running in post is something I would like to make in an separate PR.

Can we ensure that there is no misunderstanding between your and my idea?

marchc1:
> My concern with this is cost of execution. [This is already a problem for large dupes.](https://github.com/Facepunch/garrysmod/commit/ecd97269fab43a4f3524009180f9fe2b84fa758f)
> 
> A post-duplication hook would be a safer bet, even if it would be more annoying to work with to modify a single entity entry.

I think it would be worth actually profiling the cost vs a hypothetical before coming to this conclusion.

DarthTealc (approved):

robotboy655:
I have done some tests, and with a dupe of 500-800 entitles it seems to add virtually no impact on dupe paste times, so that's good.

I am assuming your use case here (since you did not say what you want this for) is to track which entities are pasted by the duplicator system. A whole hook for this purpose seems overkill, but whatever.

I do have some further feedback - the hook name, it's too generic. "on duplicated" can mean many things:
* The duplicator library just finished copying all entities it was told to
* The duplicator library just finished pasting all entities it was told to
* The duplicator library just finished copying a single entity
* The duplicator library just finished pasting a single entity

I initially assumed, from the name, that this is about copying entities, not pasting them.

I think would be better to include the word "Entity" in the hook name to signify that the hook relates to a specific entity, and to specify that it is about pasting specifically. So something like "OnEntityPasted" would be a better fit in my opinion.

Furthermore, I want to question the time of execution of this hook. 

Why does it need to be called "early to the duplication life cycle"?
Why does it need to be called way before `duplicator.RegisterEntityModifier` added callers are being iterated? In its current form it is in fact called so soon, that the entity modifier data is not even copied over to the entity.
My concern there is forcing people to throw on more 0 second timers here because the entity is not fully copied yet so they gotta wait until it is.

For your use case, I also do think that a single "OnDuplicatorPaste" hook call with a list of pasted entities would work just as well as the hook for each entity individually.

I invite any of the people who upvoted this PR to let me know what their use cases for this, so we can decide on the best solution for everyone who wants this.

Grocel:
Thanks for your reply, Rubat. I was not exactly sure if the hook name would be a particular good, so I am open to feedback. I already changed it to `OnEntityPasted` for the time being.

### Most concrete
To be honest: To answer your questions I reviewed my latest project where I still had the `GM:OnEntityPasted()` requirement. It turned out that it might have gone obsolete in the meanwhile, but I need to have a closer look into it this weekend to be sure. I even got a bit unsure about the duplication mark in that particular project. If I still need it I will formulate a much more concrete reason for this addition.

### Theory
However I remember to briefly needing such a hook once in a while.
I had a few cases where my goal was to prepare entities, before they where handled by the entity modifiers.
Basically it was about modifying certain entity modifier data before they can get used. Usually that of other addons or addon parts outside my control. It even helps catching entities with no pre set modifiers.

Other reason I can think off is preparation logic that was outside of any entity modifier logic as some kind of modifier logic dependency, maybe even across addons. This is very vague, but I think Wiremod could use such a thing.

Not a project of mine, but one thing I can think off: Admin mods and prop protections could use this for filtering/removing/changing dupe modifiers or modifier parameters to prevent abuse or apply other kind of restrictions without detouring the modifier functions.

### About a ent:IsDuplicated() mark
Independently from all this I still think it is good idea to have some kind of `ent:IsDuplicated()` marker. It would set as early as possible, so it is available in all the other duplication and spawn hooks. And also being available in a `PostEntityCreated` hook as suggested on another PR:
- https://github.com/Facepunch/garrysmod/pull/2447

### Other hooks
I also think that other hooks later in the live cycle are useful as well. My proposal was a per entity approach, but I am not (no longer) against a entity list variant as suggested earlier as long they have useful live cycle spots.

### This PR
I will change the PR accordingly if it turns out that other discussed hooks/changes are needed/wanted.

DarthTealc:
> I invite any of the people who upvoted this PR to let me know what their use cases for this, so we can decide on the best solution for everyone who wants this.

It looked useful for admin/permissions tools, and gamemodes/mods that want to treat certain things differently depending on if they were pasted or spawned normally. Without resorting to more complex or potentially conflicting detours.

SligWolf:
@robotboy655: To answer your question as one of up voters: Today @Grocel and me reviewed the usefulness of this PR in context with our addon code. These are our results:


### GMod API

We searched for alternative ways in the GMod API, but we could not get around the function as requested in this PR. In our case we want to add extra value to duped entities after the dupe was pasted, so we can distinguish between duplicated and unduplicated entities for further logic. We investigated these other ideas and found them to not be useful to the full extent: 

1. `duplicator.RegisterEntityModifier` and `duplicator.StoreEntityModifier`: The Register function could work, but these require the targeted entity to be have called with the `duplicator.StoreEntityModifier` function first. We do not always have a way to control the entity's StoreEntityModifier behavior, especially not for older dupes.

2. `ent.PostEntityPaste`: Using `ent.PostEntityPaste` has similar issues to `ent.OnDuplicated` and would require risky detours. Both function can not be used directly, as our code does not always contain scripted entities, most of them are Source Engine based vehicles. Additionally it would have been called later than `ent.OnDuplicated`.

3. Detouring `duplicator.Paste`: This has way to many downsides and also does not work well with other duplicator addons. It would been the most risky and dirty solution to our problem. And it would be a way beyond the scope of our addons too.


### Our specific use case

Many of our projects involves around trains and vehicles. We added a slight spawn height offset to our trains in case they are duplicated. We do this to avoid dupes getting stuck into the railway / ground. This is especially useful in saves.

For this we need the `IsDuplicated` info relatively early during the spawn live cycle. I have talked to Grocel about this and he thought that adding `GM:OnEntityPasted()` via this PR would have been to most flexible way to integrate our and other's possible needs into the API. However we are open to alternatives!


### Our workaround

To accomplish the above named functionality we used a code similar to the one blow:

```lua
local function OnDuplicatedDetour(ent, ...)
    if not IsValid(ent) then return end

    local oldOnDuplicated = ent._oldOnDuplicated

    ent.isDuped = true -- The parameter we need, example name
    print("onDuplicated detour run") -- test output

    if isfunction(oldOnDuplicated) then
        oldOnDuplicated(ent, ...)
    end
end

local function AddOnDuplicatedDetour(ent)
    if not IsValid(ent) then return end

    ent._oldOnDuplicated = ent._oldOnDuplicated or ent.OnDuplicated
    ent.OnDuplicated = OnDuplicatedDetour
end

hook.Add("OnEntityCreated", "AddOnDuplicatedDetour", AddOnDuplicatedDetour)
```

This has two downsides:
1. The detour is risky and can cause addon conflicts.
2. The detour requires an `OnEntityCreated` hook that would even be ran on unrelated entities.


### My personal opinion

For me as an addon creator, I would prefer to use already built-in API functions. I see programming detours or other workarounds to get a good result as tedious at best. With additions like this we can avoid creating such extra code that could break something or have a longer than needed runtime.

dvdvideo1234:
Oh man, that thing is pretty neat! Dupes debug will be easier, dupe paste arguments can be validated before even sending them to the duplicator. Imagine if someone tries to change the dupe values. You can handle this in the hook also. I can think of so many applications of this right now, but the one that stands out for me is a server man-in-the-middle attack, forcing personal values on the dupe created.

Imagine the following situation: Johnny has a cool GMod server...

An addon uses the hook, and the author knows the angular force on a wheel on this ( or any other ) `dupe` should not exceed a magnitude of 1000, for example.

The little Bobby says: `Check out this cool hack, guys! Let's crash this server`, and he forces the angular velocity parameter to 10k by editing the dupe's guts, for example ( or some way that does not matter ). Suddenly, the pasted dupe spazzes out. The server generates physics collisions, and boom, it starts to lag, then crashes ;)

Johnny decides to use the hook:
```lua
if SERVER then
    hook.Add("OnEntityPasted", "ForceSaneAngVelocity", function(ent, data)
        data.ang_velocity = math.max(-1000, math.min(1000, data.ang_velocity))
    end)
end
```
Boom! Magic the dupe will not spazz out when it's pasted! I'll give it 7 out of 5 horses because it is so awesome!

🐴 🐴 🐴 🐴 🐴 🐴 🐴

Grocel:
> ### Most concrete
> 
> To be honest: To answer your questions I reviewed my latest project where I still had the `GM:OnEntityPasted()` requirement. It turned out that it might have gone obsolete in the meanwhile, but I need to have a closer look into it this weekend to be sure. I even got a bit unsure about the duplication mark in that particular project. If I still need it I will formulate a much more concrete reason for this addition.

I had a talk to @SligWolf last week end and we have found that it is still pretty much needed. SligWolf already posted the details why. This is the most concrete use case in terms of an actual project I/we can come up with.

I am still open for alternative ideas to this, though. So far I can tell, solution with respect to our and the other posts could look like this:

1. Add `ent:IsDuplicated()`, active as soon as the entity is created by the duplicator code
2. Add `GM:OnEntityPasted` or `GM:OnEntitiesPasted`, called early, at best it should allow pre-modifying stored modifiers
3. Add `GM:OnPostEntityPasted` or `GM:OnPostEntitiesPasted`, called late, at best called after the entire dupe finished pasting

Please keep in mind that only `GM:OnEntityPasted` in the current form could premodify stored modifiers, as explained by @dvdvideo1234. Useful for admin or other security mods.

Another point is: Adding (multiple) entity list based hooks requires larger rearrangement of the duplicator code and that of third party addons (Adv Dupe 2, @thegrb93) to stay compliment to the base game. I am not sure Rubat, if this would be something you would care about, though.

My original proposal was designed to minimize all that additional friction and also to avoid having to write up large PRs around uncertain development environments/circumstances.

Grocel:
Closing this as a lack of interest by the developer (Rubat). I am not going to invest any more time in this. If someone else need this, good luck with this.

PR #2417, opened 2026-01-10
Globalize InternalSpawnNPC + Rename it

DrVrej:
This commit includes 2 changes:
- Globalizes `InternalSpawnNPC` to cut down on redundant code in SNPC mods since it's necessary for proper duplicator and save support.
- Renames `InternalSpawnNPC` to `Spawn_NPC_Internal` which better matches the other associated global function, `Spawn_NPC`. Feel free to revert or change it to whatever.

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/commands.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/commands.lua
@@ -394,7 +394,7 @@ function DoPlayerEntitySpawn( ply, entity_name, model, iSkin, strBody )
 
 end
 
-function InternalSpawnNPC( NPCData, ply, Position, Normal, Class, Equipment, SpawnFlagsSaved, NoDropToFloor )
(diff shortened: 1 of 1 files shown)


robotboy655:
Wouldn't it make more sense to do this to `GenericNPCDuplicator` instead? (Renaming it something like `MakeGenericNPC`)
Or maybe even `AddNPCToDuplicator`?
Since that's what is actually used for duplicator support by regular NPCs, and handles all the Sandbox hook calling.

robotboy655:
Also shouldn't this really be handled by the same system as normal scripted entities get saved?

DrVrej:
I recommend making `GenericNPCDuplicator` accessible alongside `ADD_ITEM_DUPEFUNC` and `ADD_WEAPON_DUPEFUNC`. Possibly renaming all 3 while at it, this would also cut down on redundant code in many mods. This was going to be my next PR after this one.

However for this issue, `InternalSpawnNPC` needs to be accessed directly as many mods copy `GenericNPCDuplicator` with changes. For example to add extra spawn options or in my case I use it to add compatibility for old saves and dupes made over 10 years ago by editing the NPC's saved table.

PR #2418, opened 2026-01-10
Restore cursor position in Spawnmenu on Linux

acnbhop:
<hr></hr>
Adds a file to `/lua/autorun/client` for restoring the cursors position in the spawnmenu only on Linux.
<hr></hr>
Opening this since I think it would be useful. On Windows, the cursor's position is saved when you re-open the spawnmenu when you opened it before, but on Linux, this behavior doesn't apply.

I would have to guess it's something with the actual C++ code (_yk, the code we can't touch..._), and not the Lua code.

I tired multiple ways and it seems that this works the best from what I can figure out by myself, since VGUI seems to reset the cursor to the center, when, we don't want that, so I had to add a temp frame-override, I don't know of any other ways to do it lol. If you have something better please let me know.

I think I got formatting correct and line endings correct? lol.

### DIFF
--- a/garrysmod/lua/autorun/client/linux_cursor.lua
+++ b/garrysmod/lua/autorun/client/linux_cursor.lua
@@ -1,0 +1,40 @@
+if ( SERVER ) then return end
+
+if ( not system.IsLinux() ) then return end
+
+local lastX = ScrW() / 2
+local lastY = ScrH() / 2
+local forceFrames = 3
+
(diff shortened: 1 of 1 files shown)


robotboy655:
Sorry, but hacky PRs like that is not something I am willing to accept. It should really be fixed properly in-engine.

This also only addresses one case of the Spawnmenu, and would not affect every single other usage of `RememberCursorPosition`/`RestoreCursorPosition` including addons and the context menu.

PR #2434, opened 2026-02-16
Add `notification.Add`

Zaurzo:
This adds a new function `Add` to the notification library. This serves as a more flexible version of `AddLegacy`.
```lua
number notification.Add( string text, number length = 5, string icon = "vgui/notices/generic", Color color = nil )
```
Returns the ID of the added notice to use with `notification.Kill` (if you need to)

Unlike `AddLegacy`, this isn't restricted to a preset group of icons, and takes a material path, allowing you to give the notice any icon that you want. A `color` argument is added to change the text color for more customizability. Additionally, the name is shorter, and removes any potential confusion with the name `AddLegacy`.

I also cleaned up the code a bit and reduced some code copy-paste.

I originally made a PR (https://github.com/Facepunch/garrysmod/pull/1967) that added a function that would add custom types to use with `AddLegacy`, but in hindsight, I think it wasn't a very good idea. This PR serves as an improved version of my old one.

**Demonstration**
```lua
local color = Color(200, 255, 200) 

notification.Add('You captured the green flag!', 5, 'icon16/flag_green.png', color)
```
notification


Be1zebub:
IMHO `notification.Add('You captured the green flag!', 5, 'icon16/flag_green.png', color)` would be better.
you need length often than icon, lot of cases you need just `notification.Add('You captured the green flag!', 5)`

PR #2441, opened 2026-02-23
Clean up the net.lua file and add TYPE_PLAYER

Vitroze:
Hello everyone,
I cleaned up the file to remove unnecessary lines and maintain a “global standard” for the file.

I also moved the `TYPE_COLOR` variable to its correct location as indicated in the TODO.
In addition, I redefined the variable from **255** to **45** because, from what I could see in the code, this variable is only used in net.ReadType and net.WriteType, and there is no reason for the value to be so high.

Thanks to this change, I reduced the number of bits from **8** to **6**.

I also created a `TYPE_PLAYER` (Value : **46**) variable so that WriteType and ReadType can distinguish between an entity and a player, which saves on the number of bits sent. _I put `TYPE_PLAYER` in `net.lua` because I didn't know where else to put this variable._
I replaced maxplayers_bits with `MAX_PLAYER_BITS` as indicated in the TODO.


garrysmod/lua/includes/util/color.lua
```diff
@@ -1,6 +1,7 @@
 
 local COLOR = {}
 COLOR.__index = COLOR
+TYPE_COLOR = 45
```
robotboy655: It should be set to `COLOR.MetaID` (After `RegisterMetaTable` call).

If we are going to set a different value for it, it might as well be the actually correct one.

Adding a comment that this is meant for the net library wouldn't hurt either.

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -101,10 +97,11 @@ function net.WriteColor( col, writeAlpha )
 	if ( writeAlpha ) then
 		net.WriteUInt( a, 8 )
 	end
+
 end
 
 function net.ReadColor( readAlpha )
-	if ( readAlpha == nil ) then readAlpha = true end
+	readAlpha = readAlpha == nil or readAlpha
```
robotboy655: This could just be moved to the `if` block, the only place where it is actually used.

```
if ( readAlpha or readAlpha == nil ) then
```

garrysmod/lua/includes/extensions/net.lua
```diff
@@ -1,6 +1,6 @@
 
--- TODO: Hack. Move to where color is defined?
-TYPE_COLOR = 255
+-- TODO: Move to where player is defined?
+TYPE_PLAYER = 46
```
robotboy655: I am not sure this needs to be a global. Again, it should be at the end of the possible range, not in the middle.

robotboy655 (changes requested):

garrysmod/lua/includes/util/color.lua
Vitroze: If we take the example of the second add-on, it still risks being broken if I set TYPE_PLAYER to the value 254, as I can see that it is used in its program. But I now understand why it is set so high.

Vitroze:
Is everything okay, or do I need to change anything else?

Vitroze:
Hey @robotboy655,
I've updated my PR. I added back the TYPE_COLOR that had been missing and changed the value of TYPE_PLAYER.

I think it would be a good idea to change TYPE_COUNT to 46. (Make TYPE_COLOR (COLOR.MetaID) compatible with 44, leaving TYPE_PLAYER at 45)

PR #2446, opened 2026-03-02
Allow tools to not appear in the spawn menu

Vitroze:
Hello,
I made sure that via the `SpawnMenuShowTool` hook, the tool does not appear if you return false.

In addition, I added a `ReloadTools` function to reload the tools (and not the entire spawn menu) if the tools need to disappear or reappear. 
I decided to make this PR to improve the user interface and prevent users from having too many tools that they don't need. 

Example : 
```lua
hook.Add("SpawnMenuShowTool", "SpawnMenuShowTool::Test", function( sCategory, tTool )
    if tTool.ItemName == "axis" then
        return false
    end
end)
```


robotboy655:
I am not sure I agree with this implementation. As with the previous attempt for this PR - this does not update live.

Plus, we already have a system in place to disable any specific tools via convars:
`toolmode_allow_axis 0`
image

This **does** update live. Why not just use that?

Vitroze:
> I am not sure I agree with this implementation. As with the previous attempt for this PR - this does not update live.
> 
> Plus, we already have a system in place to disable any specific tools via convars: `toolmode_allow_axis 0` image
> 
> This **does** update live. Why not just use that?

The problem is that the allow is global and cannot be assigned to specific users. If we need to create a permission system that hides the tools, this will not be possible.

Vitroze:
Regarding the update, I can arrange for it to be updated live if you wish.

robotboy655:
I see.

I kinda dislike the idea of adding another hook for what `CanTool` already does. I definitely do not think adding a UI specific hook is a good idea, it should be more generic.
And also I dislike the idea of completely hiding the tools, which I think is bad user experience and could lead to confusion about where the tools went, especially in singleplayer should a hook like that be officially included.

I think calling `CanTool` from `PANEL:UpdateToolDisabledStatus()` (`toolpanel.lua`) might be a decent solution, as it would clearly show the user that the tool is disabled, can be per-client, and doesn't add new hooks, so it should retroactively work for existing admin mods or whatever else, although it is not a perfect solution as we basically have to "fake" the trace argument of `CanTool`.

Anyway, I will add this idea to `dev` and later `x86-64` and we can see if this solution would work well or not.

Vitroze:
> I see.
> 
> I kinda dislike the idea of adding another hook for what `CanTool` already does. I definitely do not think adding a UI specific hook is a good idea, it should be more generic. And also I dislike the idea of completely hiding the tools, which I think is bad user experience and could lead to confusion about where the tools went, especially in singleplayer should a hook like that be officially included.
> 
> I think calling `CanTool` from `PANEL:UpdateToolDisabledStatus()` (toolpanel.lua`) might be a decent solution, as it would clearly show the user that the tool is disabled, can be per-client, and doesn't add new hooks, so it should retroactively work for existing admin mods or whatever else, although it is not a perfect solution as we basically have to "fake" the trace argument of `CanTool`.
> 
> Anyway, I will add this idea to `dev` and later `x86-64` and we can see if this solution would work well or not.



> I see.
> 
> I kinda dislike the idea of adding another hook for what `CanTool` already does. I definitely do not think adding a UI specific hook is a good idea, it should be more generic. And also I dislike the idea of completely hiding the tools, which I think is bad user experience and could lead to confusion about where the tools went, especially in singleplayer should a hook like that be officially included.
> 
> I think calling `CanTool` from `PANEL:UpdateToolDisabledStatus()` (toolpanel.lua`) might be a decent solution, as it would clearly show the user that the tool is disabled, can be per-client, and doesn't add new hooks, so it should retroactively work for existing admin mods or whatever else, although it is not a perfect solution as we basically have to "fake" the trace argument of `CanTool`.
> 
> Anyway, I will add this idea to `dev` and later `x86-64` and we can see if this solution would work well or not.

What if we add a client convar that allows users to see disabled tools? This convar is **enabled** by default and will improve the user experience for those who do not want to see unnecessary tools.

robotboy655:
Maybe a checkbox near the searchbar to hide disabled entries.

Vitroze:
Hello,
I added the checkbox as you said next to the searchbar.

In the fifth argument, I decided to put -1 to indicate that we are on a menu and that we are not interacting with the toolgun.

image

Vitroze:
Hi Rubat,
I modified my code so that game.GetWorld is not redefined each time and indicated that the hit is false. In addition, I modified the fifth argument of CanTool to be consistent with the wiki.

robotboy655:
Fixes https://github.com/Facepunch/garrysmod-requests/issues/612

Although I definitely think it should be "tick to hide", not "tick to unhide".

Vitroze:
> Fixes [Facepunch/garrysmod-requests#612](https://github.com/Facepunch/garrysmod-requests/issues/612)
> 
> Although I definitely think it should be "tick to hide", not "tick to unhide".

Good

robotboy655:
Had to rework the internals entirely so it works better, including tool search.

I also increased the default update rate to 1.5 because there's really no need for it to be more frequent. Maybe it even should use convar change callbacks instead of 1.5 interval timer.


https://github.com/user-attachments/assets/c771984d-cc67-4eba-9753-72c0d19b9c4d

robotboy655 (approved):

Grocel:
@robotboy655: Your change to put the checkbox into the Text field covers the Keyboard language UI.

grafik

The OG contribution had the language UI in mind. I would like to recommend to disable the language UI for that field or move the checkbox next to the field instead of inside it.

https://wiki.facepunch.com/gmod/Panel:SetDrawLanguageID

robotboy655:
I tried to be cute (and save an extra panel), but alas...

PR #2447, opened 2026-03-04
Add PostEntityCreated hook

StrawWagen:
Adds a hook that runs after entities are setup, their models are set, ETC

Many, many addons create their own timer.Simple inside of OnEntityCreated. (feels like I've done it in almost a dozen places!)
Will just be able to use this hook when this is merged.

*billions of timers must be removed from garry's mod*

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -727,4 +727,3 @@ function vehicle:GetCameraDistance()
 	return self:GetDTFloat( 3 )
 
 end
-
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -165,6 +165,23 @@ end
 
 hook.Add( "EntityRemoved", "DoDieFunction", DoDieFunction )
 
+--[[---------------------------------------------------------
+	PostEntityCreated, 
+	the entity will likely have it's model set, etc, by now
+-----------------------------------------------------------]]
+local function DoPostEntityCreated( ent )
+
+    timer.Simple( 0, function()
+
+        if not IsValid( ent ) then return end
+        hook.Run( "PostEntityCreated", ent )
+
+    end )
+
+end
+
+hook.Add( "OnEntityCreated", "DoPostEntityCreated", DoPostEntityCreated )
+
 function meta:PhysWake()
 
 	local phys = self:GetPhysicsObject()
@@ -710,3 +727,4 @@ function vehicle:GetCameraDistance()
 	return self:GetDTFloat( 3 )
 
 end
+


Astralcircle:
2x more hook calls when creating an entity, how much can this affect performance?

garryspins:
1000x less `timer.Simple(0)` jank, how much does that effect performance?

Zvbhrf:
I think adding ENT:PostInitialize() or something like that will be a good addition too.

Instead of doing:
```lua
function ENT:Initialize()
    -- something
    -- old: timer.Simple(0, ...)
    -- with hook:
    hook.Add("PostEntityCreated", self, function(ent)
        if self ~= ent then return end
        hook.Remove("PostEntityCreated", self)
        
        -- ~~rn we dunno how this hook would be implemented (maybe better on C++ side?)~~
        -- ~~so maybe "self" can be removed somewhere between tick(s)~~
        -- if not IsValid(self) then return end
        -- EDIT: NVM, not needed
        
        -- ... something
    end)
end
```

we can do:
```lua
function ENT:Initialize()
    -- init something that will be ready only next tick
    -- or we need to apply some changes to entity's meta 
    -- only AFTER Initialize(), etc.
end

-- called 100% for an existing entity
-- called 100% for "self"
function ENT:PostEntityCreated()
    -- called right before GM:PostEntityCreated()
    -- do some logic with something you have initialized in Initialize()
    -- or apply some changes/overrides
end
```

which really fits great with gmod's ENT structure.
It REALLY fixes the problem of gmod's default BaseClasses - they do not inherit keyvalues stored in ENT 
during ent's lua startup (see [this](https://github.com/Facepunch/garrysmod/blob/d900857953043d88b8d016144fa98ecb8585fef8/garrysmod/lua/includes/modules/scripted_ents.lua#L153), sometimes it makes headaches with bases)

SuperTemich2005:
Here's my opinion on this that I expressed in discord. **tldr I fully support this**
image

WilliamVenner:
The implementation itself seems very hacky

But the concept of a PostEntityCreated hook is totally fine and could be useful

> @Astralcircle 2x more hook calls when creating an entity, how much can this affect performance?

Why are you spawning so many entities

StrawWagen:
The implementation could be much better, should probably be engine level haha.
I just wanted to make something, this issue's been tearing at me for years

ValsdalV:
This hook would be extremely useful, no doubt about it.

> 2x more hook calls when creating an entity

> The implementation itself seems very hacky

I'm with you guys on this, we are just hiding the `timer.Simple(0)`, starting a new one each `OnEntityCreated` hook call, instead of only when it's needed.

My concern is just based on principle (the timer jank is not solved, just swept under the rug). Is the performance hit even noticeable, anyway?

SuperTemich2005:
Okay, so, I gave it a thought and I realized what's the problem. The timer method introduces an issue that entities that would initialize in one tick would all call PostInitialize all at once in the same order as they initialized. If a complex SENT like an airplane creates wheels and seats by calling ents.Create in Initialize, those sub-entities will call PostInitialize after the airplane itself calls that hook, which for my personal purposes is a bit meh. This is a great idea but it should be engine-level.

Grocel:
> Okay, so, I gave it a thought and I realized what's the problem. The timer method introduces an issue that entities that would initialize in one tick would all call PostInitialize all at once in the same order as they initialized. If a complex SENT like an airplane creates wheels and seats by calling ents.Create in Initialize, those sub-entities will call PostInitialize after the airplane itself calls that hook, which for my personal purposes is a bit meh. This is a great idea but it should be engine-level.

I actually also need this for that reason. Many of my projects have that entity + sub entity structure. Be careful about `timer.Simple` though. It is not quarantined to be always called in the next tick. Sometimes it can be called same tick, making it useless. See this:
- https://github.com/Facepunch/garrysmod-issues/issues/6668#issuecomment-3725044829
- https://commits.facepunch.com/575132

grafik

https://wiki.facepunch.com/gmod/timer.Simple

Better use timer.Create, which is guaranteed to be called in the next tick.

StrawWagen:
Really hope that a solution to this is merged, I know I've shipped timer.Simple inside of OnEntityCreated at least half a dozen times, would hate to ship yet another lmao

StrawWagen:
Shipping yet another....

Grocel:
There had been some discussion going on on Discord. It was about live cycle and timing of entities. Contrary to some claims a single tick of delay is sufficient for the use case of this PR. All stuff that the entity has been initialized with exists on the next tick. This includes: Model, KeyValues, HammerIO, Networking, Spawn/Activate call, position/angle, etc.

You might want to change the PR away from a timer to a tick-hook/array combo. Something like this:

```lua
	PostEntityCreated,
	the entity will likely have it's model set, etc, by now
local postEntityCreatedQueue = {}

local function PostEntityCreatedTick()
	for _, ent in ipairs( postEntityCreatedQueue ) do
		if IsValid( ent ) then
			hook.Run( "PostEntityCreated", ent )
		end
	end

	postEntityCreatedQueue = {} -- or use table.Empty()

	hook.Remove( "Tick", "PostEntityCreatedTick" )
end

local function DoPostEntityCreated( ent )
	table.insert( postEntityCreatedQueue, ent )

	if #postEntityCreatedQueue == 1 then
		hook.Add( "Tick", "PostEntityCreatedTick", PostEntityCreatedTick )
	end
end

hook.Add( "OnEntityCreated", "DoPostEntityCreated", DoPostEntityCreated )
```

This will completely avoid spamming timers. It is also guaranteed to be called in the next tick, unlike [timer.Simple](https://wiki.facepunch.com/gmod/timer.Simple).


## Advantages of the tick+array approach over timer.Simple:

1. Batching: Processes all queued entities in a single frame instead of creating individual timers per entity, reducing overhead when many entities spawn at once (e.g., map loads, dupe pastes).
2. Lower memory/CPU usage: Avoids creating many short-lived timer objects; uses a single reusable array and hook. Low garbage collection overhead.
3. Deterministic timing: All PostEntityCreated calls occur in the same tick (next tick), improving consistency. Avoids inconsistent 0-timer behavior.
4. Automatic cleanup: The hook removes itself when the queue is empty, preventing unneeded calls.
5. Better performance under load: Scales efficiently with bursts of entity creation, whereas timer.Simple per-entity adds linear overhead.

Astralcircle:
Why....
>table.Empty( postEntityCreatedQueue )

You should use `postEntityCreatedQueue = {}` for performance, using `table.Empty` is generally slower

Grocel:
> Why....
> 
> > table.Empty( postEntityCreatedQueue )
> 
> You should use `postEntityCreatedQueue = {}` for performance, using `table.Empty` is generally slower

The function `table.Empty` helps to avoid hammering GC? Wouldn't GC also have to empty the table (during delete) later down the line?

Astralcircle:
Entities in the table will already be efficiently garbage collected, but with table.Empty you'll simply incur a huge overhead of iterating through the table to clean it up

Grocel:
> Entities in the table will already be efficiently garbage collected, but with table.Empty you'll simply incur a huge overhead of iterating through the table to clean it up

Ok then. I changed it. 🤷‍♂️

PR #2455, opened 2026-03-20
TTT: Use replicated convars instead of global nwvars

figardo:
While using global nwvars to network replicated cvars was necessary in 2010, convar replication has been working normally in GMod for a while now.

The main benefit is that changes to these convars will apply instantly instead of requiring a round restart to take effect.
I've also updated the behaviour of some of these convars to better take advantage of this:

- Idle checking can be disabled by setting ttt_idle_limit to 0 or less.
- Disabling ttt_voice_drain hides the voice drain HUD for clients, preventing the bar from appearing stuck if disabled mid-round.
- Toggling ttt_karma will also toggle the karma column of the scoreboard.

Most of the nwvars have been left alone for backwards compatibility, but I've removed the ones where a shared function was already used to access them (e.g. DetectiveMode, HasteMode, KARMA.IsEnabled), which should free up a bit of networking.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/admin.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/admin.lua
@@ -92,7 +92,7 @@ end
 concommand.Add("ttt_print_karma", PrintKarma)
 
 
-CreateConVar("ttt_highlight_admins", "1")
+CreateConVar("ttt_highlight_admins", "1", FCVAR_REPLICATED)
 local function ApplyHighlightAdmins(cv, old, new)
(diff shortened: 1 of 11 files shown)


Malivil:
imo it would be cleaner to create an init_shd.lua or sh_init.lua to house the replicated convars rather than creating them in two different files, though I'm not sure what the precedent is outside of TTT.

figardo:
> imo it would be cleaner to create an init_shd.lua or sh_init.lua to house the replicated convars rather than creating them in two different files, though I'm not sure what the precedent is outside of TTT.

This probably wouldn't work for the two karma convars, but I'm down to move the others to shared.lua if that is preferred.

PR #2456, opened 2026-03-21
Add check for checking weapon GetHeadshotMultiplier

PeteSavarese:
Covers odd edge case on weapons that don't have a GetHeadshotMultiplier

```
[ERROR] gamemodes/terrortown/gamemode/player.lua:772: attempt to call method 'GetHeadshotMultiplier' (a nil value)
  1. unknown - gamemodes/terrortown/lua/terrortown/gamemode/player.lua:772
```

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/player.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player.lua
@@ -768,7 +768,7 @@ function GM:ScalePlayerDamage(ply, hitgroup, dmginfo)
 
       local wep = util.WeaponFromDamage(dmginfo)
 
-      if IsValid(wep) then
+      if IsValid(wep) and wep.GetHeadshotMultiplier then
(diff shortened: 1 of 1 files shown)


robotboy655 (changes requested):
I think the check should be moved inside the if-block, so that missing `GetHeadshotMultiplier` would default to 2x damage scale, just like `GetHeadshotMultiplier` returning nil already does, instead of not applying any scaling at all for headshots like your current proposition.

PeteSavarese:
Seen. Will update and reply back when commit is pushed with review.

PeteSavarese:
> I think the check should be moved inside the if-block, so that missing `GetHeadshotMultiplier` would default to 2x damage scale, just like `GetHeadshotMultiplier` returning nil already does, instead of not applying any scaling at all for headshots like your current proposition.

Latest commit will call `GetHeadshotMultiplier` if not nil. If its nil, it'll fall back to a base value of 2

robotboy655 (approved):

PR #2458, opened 2026-03-25
Additions to default non-game spawnlist(s)

goodusername123:
With the default inclusion of CS:S, EP1, and EP2 assets into base GMod I figured it would maybe make sense attempting to update the default non-game specific spawnlist(s) to incorporate some of these newly added props in ways that hopefully aren't disruptive.

This PR aims to just extend the default non-game spawnlist(s) in ways that feel the most natural without removing or reordering anything as even though there are a few props present within the base lists that somewhat suck to actually make use of, trying to do anything about it would be out of scope for this. Also as a note most of the props in this PR aren't from base HL2 but a few are.

Below is an image showcasing what has been added since the Git diff view isn't too useful in this case (warning: icons/thumbnails may not be representative of what the game generates by default and will probably have to be adjusted manually). Green is stuff I added while Purple is stuff that was removed in 2014 for requiring CS:S be mounted which I have restored in their original positions:
default gmod spawnlist PR 1

## Construction Props:

- (CS:S) Radioactive barrel next to Oil barrel. has the same model size and also a multitude of skins like the base HL2 Oil barrel.
- (CS:S) Breakable wooden barrel.
- (EP1) Extra/Missing Modular/Customizable door alongside the other doors from base HL2 that serve the same function.
- 2 extra wooden furniture props. NOTE: only the first one is breakable. I'm unsure if this is a concern or not.
- (CS:S) Dryer added right next to base HL2 washing machine (they are a matching set intended for each other).
- (CS:S) Restored kinda fugly elevator prop. I wasn't sure if I should have restored this or not since it's a bit janky but then I realized a bunch of the other default props are as well so eh idk.
- (CS:S) Long version/variation of lockers.
- (base HL2) Final damage state for single create (no clue why this was missing in the first place).
- (base HL2) Alternative cinder brick/block.
- (base HL2) Other color skins for soda pop can added to match/mirror what's being done for the plastic crate prop already.
- (EP2/CS:S) Alternative pushcart added next to base HL2 pushcart (EP2 version used).
- (EP2 and CS:S) Traffic warning props added next to traffic cone.
- (EP1) Extra dumpster prop added in-between other ones. **NOTE:** The position of is this prop between the others is very intentional as it makes the most sense due to being a more natural/linear size and purpose progression.
- (CS:S) Alternative breakable wooden pallet (smaller and different texture).
- Missing pieces for modular locker that matches how the modular fridge set is complete instead of incomplete.
- (CS:S/EP2) Extra tires (improved EP2 version used).

## Comic Props:

- (EP2) Strider brain ragdoll added next to pre-existing strider weapon ragdoll (also is next to the only other ragdolls in this section).
- (EP2) Antlion worker head added next to regular base HL2 antlion head.
- (CS:S) Alternative clock added.
- (base HL2) Missing pliers tool added next to matching wrench tool.
- (CS:S) Restored section of Pots in original position.
- (base HL2) Extra combine control panels and monitor.
- (base HL2) Added pile of trash next to all the garbage composites.
- (CS:S and EP2) Various breakable foods added next to base HL2's sole breakable watermelon prop. I wasn't fully confident in the order(ing) I laid these out in so feel free to reorder.
- (CS:S) Restored food pile effects in their original position.
- (EP2) Added gnome next to other non-natural manufactured bipedal(?) prop (hula doll). also adjacent to the blood stained kennel subtly implying he is a violent animal.
- (EP2) Extra old computer monitor added alongside others. **NOTE:** This is not a carbon copy of the first monitor, it's smaller, has a differently shaped bezel, and different shading.
- (EP2?) Extra servers and alternative workstation.

As a final note there was more I could have tried adding but I figured I would just submit this PR as is and leave other additions to future PRs rather then endlessly trying to work on this one.

### DIFF
--- a/garrysmod/settings/spawnlist_default/001-construction props.txt
+++ b/garrysmod/settings/spawnlist_default/001-construction props.txt
@@ -53,1329 +53,1443 @@
 		"10"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/canister_propane01a.mdl"
+			"model"		"models/props/de_train/barrel.mdl"
 		}
 		"11"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/bench01a.mdl"
+			"model"		"models/props_c17/canister_propane01a.mdl"
 		}
 		"12"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/column02a.mdl"
+			"model"		"models/props_c17/woodbarrel001.mdl"
 		}
 		"13"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/concrete_barrier001a.mdl"
+			"model"		"models/props_c17/bench01a.mdl"
 		}
 		"14"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/display_cooler01a.mdl"
+			"model"		"models/props_c17/column02a.mdl"
 		}
 		"15"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/fence01a.mdl"
+			"model"		"models/props_c17/concrete_barrier001a.mdl"
 		}
 		"16"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/fence01b.mdl"
+			"model"		"models/props_c17/display_cooler01a.mdl"
 		}
 		"17"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/fence02a.mdl"
+			"model"		"models/props_c17/fence01a.mdl"
 		}
 		"18"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/fence02b.mdl"
+			"model"		"models/props_c17/fence01b.mdl"
 		}
 		"19"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/fence03a.mdl"
+			"model"		"models/props_c17/fence02a.mdl"
 		}
 		"20"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/fence04a.mdl"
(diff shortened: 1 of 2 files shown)


robotboy655:
Alright, I will allow it this once.

I have also reworked the default CS:S spawnlist to include all props from CS:S, similarly to HL2 and its episodes.

goodusername123:
If I had the chance to do this PR again I would have likely included a few more entries which I had planned to split out for "future PRs" and also tried to enforce a more unified ordering for new entries (such as if **all** new entries should have been ordered against pre-existing entries based on size or some other criteria rather then the current mixed approach), Though overall I think this came out pretty well in the end ***however*** something that's been really irking me is that I found a noticeably superior wooden barrel prop after submitting this PR:
<video alt="gm_construct 2026-4-18 16-41-26.webm" src="https://github.com/user-attachments/assets/15dffa9c-8257-43d3-a290-04ffeb293be9" />
```patch
--- garrysmod/settings/spawnlist_default/001-construction props.txt
+++ garrysmod/settings/spawnlist_default/001-construction props.txt
@@ -61,11 +61,11 @@
 		}
 		"12"
 		{
 			"type"		"model"
-			"model"		"models/props_c17/woodbarrel001.mdl"
+			"model"		"models/props/de_inferno/wine_barrel.mdl"
 		}
 		"13"
 		{
 			"type"		"model"
```
@robotboy655 If this is an acceptable change should I submit a PR just for it, try pushing the change to the branch for this PR, or would you be fine just pushing the one line change on your end?

robotboy655:
Alright, I will replace the model.

PR #2459, opened 2026-03-30
Refactor cleanup tool UI to work with custom spawnmenu

darkfated:
The cleanup tool UI currently depends on `controlpanel.Get`, which only works with panels created by the default spawnmenu. As a result, the Cleanup tabs appear empty when using a custom spawnmenu with its own controlpanel.

By replacing this dependency with `spawnmenu.AddToolMenuOption`, the panel is now built directly from the provided instance, making the cleanup UI compatible with both default and custom spawnmenu implementations.

## Screenshots

{3F800928-FBE6-4A63-9A60-F5EFFBDB5510}

{5A4D7CA2-029E-4573-9652-2359A6686F53}

### DIFF
--- a/garrysmod/lua/autorun/utilities_menu.lua
+++ b/garrysmod/lua/autorun/utilities_menu.lua
@@ -248,13 +248,11 @@ end
 -- Tool Menu
 hook.Add( "PopulateToolMenu", "PopulateUtilityMenus", function()
 
-	spawnmenu.AddToolMenuOption( "Utilities", "User", "User_Cleanup", "#spawnmenu.utilities.cleanup", "", "", User_Cleanup )
 	spawnmenu.AddToolMenuOption( "Utilities", "User", "Undo", "#spawnmenu.utilities.undo", "", "", Undo )
 	spawnmenu.AddToolMenuOption( "Utilities", "User", "PhysgunSettings", "#spawnmenu.utilities.physgunsettings", "", "", PhysgunSettings )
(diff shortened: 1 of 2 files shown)


robotboy655:
I take issue with this. Why are we changing the base game internal workings to support a **single** mod? Are you gonna ask  every mod developer who uses  `controlpanel.Get` to adapt to your needs? Why doesn't the custom spawnmenu just support the core feature?

Edit: Not to mention the fact that mods would expect `controlpanel.Get` to work when accessing existing maps for whatever reason.

darkfated:
> I take issue with this. Why are we changing the base game internal workings to support a **single** mod? Are you gonna ask every mod developer who uses `controlpanel.Get` to adapt to your needs? Why doesn't the custom spawnmenu just support the core feature?
> 
> Edit: Not to mention the fact that mods would expect `controlpanel.Get` to work when accessing existing maps for whatever reason.

Because the key issue is: if you use the default controlpanel, it will stand out from the overall visual style of a custom spawnmenu - even if you apply your own DermaSkin to it. This change does not alter the game's core logic, and other addons continue to work. The only thing that changes is that the logic for adding tools is rewritten to use a more modern api `spawnmenu.AddToolMenuOption`.

I understand there are valid scenarios where `controlpanel.Get` is appropriate - for example, when building a custom context menu - but this is specifically not one of those cases.

darkfated:
@robotboy655 what do i need to improve so the pull request gets accepted? about whether i'm going to ask other mod developers to rewrite their tools to use `spawnmenu.AddToolMenuOption` - no, i'm not. but at least in the default garry's mod code, the modern api will be used, which is the right thing to do from an overall architecture perspective anyway

robotboy655 (approved):

PR #2460, opened 2026-04-04
[TTT] Fixed errors when a weapon works in TTT but doesn't use weapon_tttbase

Malivil:
Example weapon: https://steamcommunity.com/sharedfiles/filedetails/?id=2884784564

Adds the following to all weapons using the "weapon_base" base:
- SWEP:IsEquipment
- SWEP.HeadshotMultiplier
- SWEP:GetHeadshotMultiplier

Supersedes #2456

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_equip.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_equip.lua
@@ -176,7 +176,7 @@ local function GetOwnedEquipment()
    -- Determine if we already have equipment
    local owned_ids = {}
    for _, wep in ipairs(LocalPlayer():GetWeapons()) do
-      if IsValid(wep) and wep:IsEquipment() then
+      if IsValid(wep) and wep.IsEquipment and wep:IsEquipment() then
          table.insert(owned_ids, wep:GetClass())
       end
    end
--- a/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/player_ext_shd.lua
@@ -54,7 +54,7 @@ function plymeta:GetBaseKarma() return self:GetNWFloat("karma", 1000) end
 
 function plymeta:HasEquipmentWeapon()
    for _, wep in ipairs(self:GetWeapons()) do
-      if IsValid(wep) and wep:IsEquipment() then
+      if IsValid(wep) and wep.IsEquipment and wep:IsEquipment() then
          return true
       end
    end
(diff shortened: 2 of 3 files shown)


robotboy655:
Wouldn't it be better to use the default implementation of `SWEP:IsEquipment` (which is to call `WEPS.IsEquipment( weaponEnt )`) if one is not provided by the weapon?

Malivil:
I would agree exce

> Wouldn't it be better to use the default implementation of `SWEP:IsEquipment` (which is to call `WEPS.IsEquipment( weaponEnt )`) if one is not provided by the weapon?

I would agree except `WEPS.IsEquipment` just checks `SWEP.Kind` which is established in `weapon_tttbase` so that would be missing too. It would always return `false`.

robotboy655:
Alright, fair enough, but this request is in context of adding new weapons to the TTT equipment tab, correct? (That are not necessarily based on the TTT weapon base)

I think having those weapons be assumed to not be equipment by the gamemode (like you propose) is wrong. (Unless I am misremembering what "Equipment" is in TTT?)

In my opinion the best way to workaround the issue is to add the necessary fields/functions onto the weapons being added to the equipment tab via the https://wiki.facepunch.com/gmod/GM:PreRegisterSWEP hook, or ideally subclass the weapon (i.e. make a new SWEP file that uses SWEP.Base of the one you want to add) and implement the necessary stuff in it, and add that instead.

Malivil:
In this case "IsEquipment" is just a "Kind" of a weapon, it's not an "Equipment" like the shop menu is talking about where they are passive items that don't show as weapons in the weapon selector HUD.

I can adjust this PR to use PreRegisterSWEP to add `IsEquipment` (and `GetHeadshotMultiplier` for #2456)

robotboy655:
You shouldn't just change the entire PR like that, but whatever. (Should open new one) I will leave the final decision to @svdm.

From what I understand this would allow people to bypass "1 weapon perslot" limitation, which I don't think TTT's weapon selector supports.

svdm:
I think it's a bad idea to make any weapon "kind of work" in TTT without specific adjustments. I also can't really validate that just the headshot multiplier stuff is enough to do even that, just at a glance it seems like ammo will not behave correctly for one.

Maybe for some weapons on some servers this change makes sense to get a specific weapon to work. But at that point, given that all this does is add 1 hook, perhaps it should just be an addon?

Malivil:
> I think it's a bad idea to make any weapon "kind of work" in TTT without specific adjustments. I also can't really validate that just the headshot multiplier stuff is enough to do even that, just at a glance it seems like ammo will not behave correctly for one.
> 
> Maybe for some weapons on some servers this change makes sense to get a specific weapon to work. But at that point, given that all this does is add 1 hook, perhaps it should just be an addon?

That's fair enough

PR #2461, opened 2026-04-07
Add hook PlayerStartVoice/PlayerEndVoice Serverside

Vitroze:
Hello,
I've added the [PlayerStartVoice](https://wiki.facepunch.com/gmod/GM:PlayerStartVoice) and [PlayerEndVoice](https://wiki.facepunch.com/gmod/GM:PlayerEndVoice) hooks on the server side.
It's best if this is integrated into the engine.

### DIFF
--- a/garrysmod/lua/includes/extensions/player.lua
+++ b/garrysmod/lua/includes/extensions/player.lua
@@ -239,6 +239,37 @@ function meta:GodDisable()
 
 end
 
+timer.Create( "PlayerVoice", 0.1, 0, function()
+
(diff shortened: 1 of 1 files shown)


darkfated:
And frankly, I don't see much need to add this, given the added timer. The additional burden outweighs the "convenience." And if someone needs to check for this event, they can add it themselves

Vitroze:
That could always come in handy

And as I said, it would be better if it were built directly into the engine, just like what rubat did for the Player:IsSpeaking function.

PR #2462, opened 2026-04-20
`string.ToColor` `.rgb` support without alpha input

Akabenko:
`string.ToColor` by default returns `Color(255, 255, 255, 255)` as fallback, if argument to input has no alpha value. This pull request allow to input string of color without alpha value. For example string `"128 0 0"` as argument will return same Color.


robotboy655 (approved):

Akabenko:
Sorry. Miss click on re-request button

robotboy655 (approved):

PR #2463, opened 2026-04-25
`base.fgd` Static prop scale & Render color

Akabenko:
Added `uniformscale` and `rendercolor` from CS:GO `base.fgd`:

```cpp
uniformscale(float) : "Uniform Scale Override" : 1 : "Resize the static prop."
rendercolor(color255) : "Color (R G B)" : "255 255 255"
```

Using Render color in Gmod:
image-1012

We can compile maps with scaled static props for Gmod, using VBSP++ with `-staticpropformat 11` and VRAD++ by ficool2.


robotboy655:
It does not work without map compiler & map format changes.

PR #2464, opened 2026-04-26
Avoid to draw `RenderOverlay` on `propspawn` in Resolved depth buffer

Akabenko:
Existing check by `isDepthPass` helps for prop spawn, but it not enough for NPCs spawn. Thats why i added check by name of `render.GetRenderTarget()` for avoid calling `RenderOverlay` to `_rt_resolvedfullframedepth`.

As alternative: `if ( !isDepthPass and !rt ) then` check works too. 

Before (`RenderOverlay` force writting color `.rgb` values to resolved depth buffer):
image

After:
image

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/effects/propspawn.lua
+++ b/garrysmod/gamemodes/sandbox/entities/effects/propspawn.lua
@@ -115,7 +115,9 @@ function EFFECT:RenderParent( flags )
 	render.PopCustomClipPlane()
 	render.EnableClipping( bClipping )
 
-	if ( !isDepthPass ) then
+	local rt = render.GetRenderTarget()
+
+	if ( !isDepthPass and ( !rt or rt:GetName() != "_rt_resolvedfullframedepth" ) ) then
 		self.SpawnEffect:RenderOverlay( self )
 	end
 


robotboy655:
I think it should be testing for [STUDIO_SSAODEPTHTEXTURE ](https://wiki.facepunch.com/gmod/Enums/STUDIO#STUDIO_SSAODEPTHTEXTURE) and [STUDIO_SHADOWDEPTHTEXTURE](https://wiki.facepunch.com/gmod/Enums/STUDIO#STUDIO_SHADOWDEPTHTEXTURE) instead of testing the render target name.

robotboy655:
Wait, it already does, so what's the issue then? Maybe the flags are not being passed to DrawModel calls?

Akabenko:
@robotboy655 It works well with prop spawns. But it doesn't work with NPC spawns. That's weird.

Test by spawn `nb_example` NPC to avoid spawning of weapon for testing:
image

```lua
print(flags, isDepthPass)
```
Output:
```LUA
1	false
1	false
1	false
... etc, only 1	false
```

robotboy655:
Ok, I have looked into it, and it's an issue with calling `render.MaterialOverride()` during the depth passes. There's a weird quirk where the entity render code will call `DrawModel` with `0` for flags to "recompute master entity bone structure", seemingly related to bonemerged/parented models, which for NPCs would be the the weapon.

So, the simple fix is to not call `MaterialOverride` functions for `flags == 0` as well. I will include this fix for the propspawn effect today.

PR #2467, opened 2026-05-16
Remove unused NW vars in gmod_thruster code + switch them to datatables

Astralcircle:
They are not used anywhere, what is their purpose?

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_thruster.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_thruster.lua
@@ -16,8 +16,10 @@ function ENT:SetupDataTables()
 
 end
 
--- Backwards compact
-ENT.IsOn = ENT.GetOn
+-- Backwards compact, use :GetOn instead
+function ENT:IsOn()
+	return self:GetOn()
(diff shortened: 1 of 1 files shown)


robotboy655:
I smell addon breaking change, ~~especially with that IsOn function being removed.~~ Oh I am blind.

Astralcircle:
> I smell addon breaking change, ~especially with that IsOn function being removed.~ Oh I am blind.

It won't break anything really, no methods was removed

robotboy655:
It's not about the methods but about removal of NWVars being set on those entities. Addons could be trying to access them directly, and removing the NWVars would break said addons.

That being said, it is a theoretical issue and after doing a search on GitHub I don't see any such cases, although it definitely could've missed some. The setter/getter methods also exist since before 2013 so any competent addon maker should've used those and not the NWVars directly.

This doesn't apply to the weird "1" and "2" NWVars though.

Astralcircle:
So there is any reason to not merge it? I don't think any addons have ever used these undocumented NW vars

robotboy655 (approved):
Alright, lets see if this breaks anything.

PR #2468, opened 2026-05-18
Add utf8.upper / utf8.lower from UnicodeData.txt (alternative to the PR #2164)

Ethjuro:
Hi! Adds `utf8.upper` and `utf8.lower` so non-ascii characters can be cased correctly in the utf8 libraries.

Alternative to PR #2164 , but the upper/lower tables are generated independently from UnicodeData.txt columns 12 and 13, rather than inverting one into the other. That sidesteps the ambiguous reverse-mapping issue CornerPin raised on PR #2164 (Σ -> σ vs ς, S -> s vs ſ, etc.), since each direction uses its canonical simple mapping straight from Unicode.

Simple case mappings only ([UnicodeData.txt ](https://www.unicode.org/Public/17.0.0/ucd/UnicodeData.txt)17.0.0).


Ethjuro:
Up

Ethjuro:
@robotboy655 Hey, is this PR going to be merged for the next GMod update?

robotboy655:
No, it will not be merged for tomorrow's update.

Astralcircle:
I think these big tables should be moved to separate file, they bloat utf8 lib code too much

PR #2470, opened 2026-05-25
Move `DrawSprite` of `prop_effect` to `DrawTranslucent`

Akabenko:
Here i draw effect of AO in `PostDrawOpaqueRenderables` hook.

Before fix:
image

After fix:
image


robotboy655 (approved):

robotboy655:
I feel like this is going to stir up some shit, but we will see I guess,

PR #2473, opened 2026-05-30
gmod_light.lua micro optimization

nnk000:
1 Color() call per frame instead of 3

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_light.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_light.lua
@@ -98,20 +98,20 @@ function ENT:GetOverlayText()
 end
 
 local matLight = Material( "sprites/light_ignorez" )
+local tubeCol = Color( 255, 255, 255, 255 )
 function ENT:DrawEffects()
 
 	if ( !self:GetOn() ) then return end
 
 	local LightPos = self:GetPos()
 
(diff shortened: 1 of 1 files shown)


GrayWolf64:
the remaining one can be moved out of the func, 0 new Color per frame? just update alpha before the render calls

ceifa (approved):

garrysmod/gamemodes/sandbox/entities/entities/gmod_light.lua
```diff
@@ -111,10 +111,12 @@ function ENT:DrawEffects()
 	local Alpha = 255 * Visibile
```
KrzOkun: Shouldn't be Visible?

GrayWolf64: It doesn't error because all visible here are visibile.

NovaDiablox: https://github.com/Facepunch/garrysmod/blob/6d0b1b8937ae743147ddbb9744082dbdb1c35b17/garrysmod/gamemodes/sandbox/gamemode/spawnmenu/creationmenu/content/contenttypes/custom.lua#L186 in the meanwhile

KrzOkun: Maybe not an error but we could fix typo 😄

robotboy655 (approved):

PR #2474, opened 2026-05-31
Try to fix rare prop_effect error

Astralcircle:
Sometimes i see errors where AttachedEntity is nil for some players. This is probably because prop_effect was networked, but its children isn't

Simply adding `:SetTransmitWithParent(true)` should fix this and make it more reliable


robotboy655:
Assuming this experience of errors is from your own server, have you applied the change on the server to actually test if the change fixes the issue?

Astralcircle:
I applied this fix and it doesn't seem to have fixed anything. This is probably because when initializing on the client, AttachedEntity still doesn't exist for some reason
```
- addons/server-scripts/gamemodes/base/entities/entities/prop_effect.lua:77: attempt to index field 'AttachedEntity' (a nil value)
1. <unknown> - addons/server-scripts/gamemodes/base/entities/entities/prop_effect.lua:77
```

PR #2476, opened 2026-06-02
Add dormancy check to `DynamicLight` creation

wrefgtzweve:
Currently `DynamicLight` is always created even if the entity is dormany, taking up scarce dlight slots which breaks other lamps. Only creating them when in PVS makes it so we're not hitting the dlight limit as much.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_light.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_light.lua
@@ -57,7 +57,7 @@ function ENT:Think()
 	self.BaseClass.Think( self )
 
 	if ( CLIENT ) then
-
+		if ( self:IsDormant() ) then return end
 		if ( !self:GetOn() ) then return end
 
 		local noworld = self:GetLightWorld()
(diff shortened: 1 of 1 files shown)


Grocel (approved):

bloodycop6385 (approved):

robotboy655 (approved):

robotboy655:
Now that I think more about it, this does create issues in certain situations like this, because the glow can be very large:

https://github.com/user-attachments/assets/6a8b9476-ce84-460d-a975-ee03ea43c951

Maybe the benefit of closer ones working when at limit outweigh this negative?

Same thing really applies to projected textures in a bigger way, since the distances they can shine around a corner are usually decently high.

Grocel:
Aggressive dormant culling seems to be the new addon dev meta now. Looks like it is being pushed to GMod as well. Like even if it causes issues/regressions.

Maybe it helps to increase the render bounds of the entity to reduce the likely hood of it being culled too early. @robotboy655

robotboy655:
Render bounds have no effect on entity dormancy, since it's a serverside mechanism.

DarthTealc:
If the issue is the dynamiclight limit being hit when there are other lights that are likely irrelevant to the player, couldn't the dormancy check be used when that limit is actually hit, but not used when it hasn't been hit? That way lights around corners in other PVSs will still be visible 99% of the time but if the dynamiclight limit has been hit it'll prioritise in-PVS lights over non-PVS lights

wrefgtzweve:
maybe using a elight when it's dormant instead of a dlight could help to alleviate this issue, you'd still be hitting performance issues though.

PR #2479, opened 2026-06-15
Add `:Left()` angle/vector functions

DarthTealc:
### Summary

Adds `:Left()` variants of `:Right()` angle/vector functions.

_Optional extra:_ For consistency, the second commit adds `:Backward()` and `:Down()` variants of `:Forward()` and `:Up()`.

---
### Background

The [Source Engine's 3D coordinate system](https://developer.valvesoftware.com/wiki/Coordinates) uses the 'Right-hand rule'.
Coordinates are (X,Y,Z), where X is forward, Y is left, and Z is up.
[x/y axis graph] [right-hand rule] <details><summary><mark>(Click to view more images)</mark> </summary>
[ingame image]
[ingame image]
[ingame image]
[ingame image]
[x/y axis graph]
[right-hand rule] </details>
#### Gmod's functions

Gmod angles/vectors have functions for `:Forward()`, `:Right()`, `:Up()`. 
`Forward` is X in the positive direction
`Right` is Y in the negative direction
`Up` is Z in the positive direction

Image

#### Problem
Developers need to use `(+Forward, -Right, +Up)` as there is no `:Left()`.
Gmod providing `Forward`/`Right`/`Up` and not providing their opposites implies they are the positive directions, and developers may not realise that isn't the case until they encounter unexpected issues or research it.
Even once they are aware, because there is no `:Left()` they must continue to use `(+,-,+)` which is unintuitive. 

---

### PR Details - `:Left()`

Adding a `:Left()` allows developers to use `(+Forward, +Left, +Up)` when working with angles/vectors. This is more intuitive for use with Source's coordinate system.
_Optional extra:_ For consistency, `:Backward()` and `:Down()` variants are included in the second commit.

image

Developers can still use `:Right()`. It hasn't been removed or changed. There are no changes to Gmod's coordinates system.

#### Function list
"`Left`" functions as the positive direction opposite the existing negative direction "`Right`" functions.
- `Angle:Left()`, as related to [`Angle:Right()`](https://wiki.facepunch.com/gmod/Angle:Right) 
- `Entity:GetLeft()`, as related to [`Entity:GetRight()`](https://wiki.facepunch.com/gmod/Entity:GetRight)
- `VMatrix:GetLeft()`, as related to [`VMatrix:GetRight()`](https://wiki.facepunch.com/gmod/VMatrix:GetRight)
- `VMatrix:SetLeft()`, as related to [`VMatrix:SetRight()`](https://wiki.facepunch.com/gmod/VMatrix:SetRight)

As per the existing positive direction Forward/Up versions of the above, the second commit includes negative direction Backward/Down versions of above too.

#### Demo
Demo of `Entity:Get*()`. The red/green/blue lines correspond to the X/Y/Z axis and are drawn in the positive direction.

https://github.com/user-attachments/assets/81166f56-252f-4628-be46-395e95fe9ea9

Credit to @marchc1 and @Grocel for code collaboration
Original suggestion https://github.com/Facepunch/garrysmod-requests/issues/3168

### DIFF
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -30,6 +30,18 @@ function meta:GetLeft()
 	return vec
 end
 
+function meta:GetBackward()
+	local vec = self:GetForward()
+	vec:Mul( -1 )
+	return vec
+end
+
+function meta:GetDown()
+	local vec = self:GetUp()
+	vec:Mul( -1 )
+	return vec
+end
+
 --
 -- Entity index accessor. This used to be done in engine, but it's done in Lua now because it's faster
 --
--- a/garrysmod/lua/includes/extensions/entity.lua
+++ b/garrysmod/lua/includes/extensions/entity.lua
@@ -24,6 +24,12 @@ function meta:SetShouldPlayPickupSound( bPlaySound )
 	self.m_bPlayPickupSound = tobool( bPlaySound ) or false
 end
 
+function meta:GetLeft()
+	local vec = self:GetRight()
+	vec:Mul( -1 )
+	return vec
+end
+
 --
 -- Entity index accessor. This used to be done in engine, but it's done in Lua now because it's faster
 --
--- a/garrysmod/lua/includes/extensions/vmatrix.lua
+++ b/garrysmod/lua/includes/extensions/vmatrix.lua
@@ -1,0 +1,12 @@
+local meta = FindMetaTable( "VMatrix" )
+
+function meta:GetLeft()
+	local vec = self:GetRight()
+	vec:Mul( -1 )
+	return vec
+end
+
+function meta:SetLeft( vec )
+	if ( !isvector( vec ) ) then error( "bad argument #1 to 'SetLeft' (Vector expected, got " .. type( vec ) .. ")", 2 ) end
+	self:SetRight( vec * -1 )
+end
--- a/garrysmod/lua/includes/extensions/vmatrix.lua
+++ b/garrysmod/lua/includes/extensions/vmatrix.lua
@@ -6,7 +6,29 @@ function meta:GetLeft()
 	return vec
 end
 
+function meta:GetBackward()
+	local vec = self:GetForward()
+	vec:Mul( -1 )
+	return vec
+end
+
+function meta:GetDown()
+	local vec = self:GetUp()
+	vec:Mul( -1 )
+	return vec
+end
+
 function meta:SetLeft( vec )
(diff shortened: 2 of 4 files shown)


garrysmod/lua/includes/extensions/entity.lua
```diff
@@ -24,6 +24,24 @@ function meta:SetShouldPlayPickupSound( bPlaySound )
...
+	vec:Mul( -1 )
+	return vec
+end
+
+function meta:GetDown()
+	local vec = self:GetUp()
+	vec:Mul( -1 )
+	return vec
+end
+
```
Grocel, line 44: As far as I can tell, all the getters in this PR feature the best possible implementation for Lua, while also harmonizing with the rest of the codebase. No additional temporary vector objects are created. This saves on GC load.

garrysmod/lua/includes/extensions/vmatrix.lua
```diff
@@ -0,0 +1,34 @@
...
+
+function meta:SetBackward( vec )
+	if ( !isvector( vec ) ) then error( "bad argument #1 to 'SetBackward' (Vector expected, got " .. type( vec ) .. ")", 2 ) end
+	self:SetForward( vec * -1 )
+end
+
+function meta:SetDown( vec )
+	if ( !isvector( vec ) ) then error( "bad argument #1 to 'SetDown' (Vector expected, got " .. type( vec ) .. ")", 2 ) end
+	self:SetUp( vec * -1 )
+end
```
Grocel, line 34: This is the best possible implementation in Lua that has no side effects on the passed vector `vec`. Calling these functions creates an additional temporary vector object. That's something that cannot be avoided in Lua. This impacts GC when the functions are called frequently.

To solve this issue, all new functions (at least the setters) would need to be implemented on the C side, as discussed in the original GitHub issue (https://github.com/Facepunch/garrysmod-requests/issues/3168).

Grocel (approved):

robotboy655:
I find usefulness of these dubious at best. In most cases you are not directly negating the values of GetRight/Forward/Up, in most use cases you'd be multiplying the return value and adding or subtracting it from some other vector, so at that point it's a difference of swapping `+` for a `-` and vice versa as needed.

And regardless of that, it makes little sense to implement it in Lua, which really doesn't achieve anything outside of just to have these functions for the sake of having them.

DarthTealc:
- https://github.com/Facepunch/garrysmod/pull/2400#issuecomment-5096423540
> This is very borderline useless addition, but since we already have `math.tau`, might as well I guess.

Couldn't that apply here too? Gmod already has $${\color{green}+}$$`Forward`, $${\color{red}-}$$`Right`, $${\color{green}+}$$`Up`, shouldn't it also have $${\color{red}-}$$`Backward`, $${\color{green}+}$$`Left`, $${\color{red}-}$$`Down`?

PR #2480, opened 2026-06-18
Refactor ConVar Access

Vitroze:
Hello,
In the Garry's Mod code, there were still some deprecated functions. I've decided to modify this and implement a caching system as soon as possible. Otherwise, I'll assign them via the functions in the cvars table, which are not deprecated.

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/prop_effect.lua
+++ b/garrysmod/gamemodes/base/entities/entities/prop_effect.lua
@@ -74,9 +74,10 @@ function ENT:Draw( flags )
 
 end
 
+local cl_draweffectrings = GetConVar( "cl_draweffectrings" )
 function ENT:DrawTranslucent( flags )
 
-	if ( GetConVarNumber( "cl_draweffectrings" ) == 0 ) then return end
+	if ( cl_draweffectrings:GetInt() == 0 ) then return end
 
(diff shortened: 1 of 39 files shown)


robotboy655:
This is mostly an unnecessary change. Maybe we should un deprecate these to discourage PRs like this. Not the first PR to propose this.

The only place it could make sense to cache the convars is in draw/think hooks, but even then, both `cvars.*` and `GetConVar*` functions use a convar object cache internally, so at most you'd be doing is saving a function call and an if-check. The problem is that this could introduce race condition bugs where the convars are not created at the time you try to cache them.

You also cannot blindly replace `GetConVar*` with `cvars.*` calls, not only it achieves nothing (actually probably is slower), their behavior is not fully identical and may lead to errors, because `cvars.*` functions can return `nil` if not providing the second argument.

robotboy655:
I have updated the wiki pages with useful info on differences and specific behavior, as well as removing the deprecated notice.

Vitroze:
Okay, thank you. At first, I didn't understand why it was obsolete, hence the PR to remove obsolete code.

robotboy655:
They were marked as deprecated when they did not use to cache the convars internally IIRC, so when used in render/think hooks it would spam a lot of useless objects that then get GCd or something like that. Similarly to Material() function and why it is not advised to use them in the same scenarios.

PR #2481, opened 2026-06-18
Change the function from ents.GetAll to ents.Iterator

Vitroze:
Hello,
Micro-optimization: in some cases, it's better to use `ents.Iterator` than `ents.GetAll()`.


robotboy655 (approved):
This is most definitely achieves basically nothing, but whatever I guess.

PR #2483, opened 2026-06-18
[Terrotown] Optimize distance checks by replacing Distance with DistToSqr

Vitroze:
Hello,
Only in the Terrotown gamemode, I modified the ``Distance`` using ``DistToSqr``. In some cases, this is useful when dealing with convar spam, general spam or in a hook/function because it is much more expensive. The only instance where I'm unsure of its 100% effectiveness is in ``gamemodes/terrotown/gamemode/cl_init.lua L.338``, since it's executed only once, I included it just in case.

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/cl_radar.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/cl_radar.lua
@@ -115,6 +115,7 @@ local GetPTranslation = LANG.GetParamTranslation
 local FormatTime = util.SimpleTime
 
 local near_cursor_dist = 180
+local near_cursor_distsqrt = near_cursor_dist ^ 2
 
 function RADAR:Draw(client)
    if not client then return end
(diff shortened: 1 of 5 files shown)


garrysmod/gamemodes/terrortown/gamemode/cl_radar.lua
```diff
@@ -172,8 +173,8 @@ function RADAR:Draw(client)
       if not scrpos.visible then
          continue
       end
-      md = mpos:Distance(Vector(scrpos.x, scrpos.y, 0))
-      if md < near_cursor_dist then
+      md = mpos:DistToSqr(Vector(scrpos.x, scrpos.y, 0))
+      if md < near_cursor_distsqrt then
          alpha = math.Clamp(alpha * (md / near_cursor_dist), 40, 230)
```
robotboy655: This is wrong. `md` in this case is now the squared distance, but is divided by non squared distance.

No, you can't divide it by the squared distance either, it will not be equal to previous behavior.

robotboy655 (changes requested):

garrysmod/gamemodes/terrortown/gamemode/cl_radar.lua
Vitroze: Hello,
It's normally fixed; it will send the same value as if it were a Distance.

robotboy655 (approved):
LGTM (I didn't actually look)

PR #2484, opened 2026-06-18
Micro-optimization + Fixed bugs on searches in the spawnmenu

Vitroze:
Hello,
I've optimized the search function in the spawn menu.
I've also fixed an issue where if you typed the exact name of the spawn menu and it was under its language ID, it wouldn't be found.

With PR :
image

Without PR : 
image


garrysmod/gamemodes/sandbox/gamemode/cl_search_models.lua
```diff
@@ -113,38 +113,27 @@ local function AddSearchProvider( listname, ctype, stype )
...
+			local name = language.GetPhrase(v.PrintName or v.Name)
 			if ( !isstring( name ) and !isstring( name_c ) ) then continue end
 
 			if ( ( isstring( name ) and name:lower():find( str, nil, true ) ) or ( isstring( name_c ) and name_c:lower():find( str, nil, true ) ) ) then
 
 				local contentIconData = {
-					nicename = v.PrintName or v.ClassName,
-					spawnname = v.ClassName,
-					material = "entities/" .. v.ClassName .. ".png",
+					nicename = name or name_c,
```
robotboy655, line 128: This should not use `name` here. These content icons can be saved into spawnlists. If they are saved with this change, I think they would also save the localized name into the spawnlist instead of using the "#localization_string"

Vitroze: Good

robotboy655: What do you mean "Good"? This would be a regression that does not need to exist. The name on the icon is automatically localized and has nothing to do with search.

Vitroze: If I understand the problem correctly, it should normally be resolved. I tried to reproduce the problem described in-game and I couldn't even before this change, and I wasn't successful.

robotboy655 (approved):

PR #2485, opened 2026-06-19
Fix cookie library ignoring call order

RaphaelIT7:
As of right now the cookie library ignored any Set that was made immediately (made in a 0.1s time window) after a Delete as it uses two seperate queues.
Soo this PR simplifies the Queues merging them into one- and ensures that cookies will preserve the right value for the call order.

Example:
`lua_run cookie.Delete("Hello") cookie.Set("Hello", "World") print(cookie.GetString("Hello"))`

I encountered this issue in https://github.com/Mantibro/SlashCo/commit/6ea795c6f7ca36fba582bff4f8e8fe5bef5a4f2c as we set the cookie instantly after having called `Delete`

This also adds a guard to `cookie.Set` as previously it accepted any value and silently depended on later SQLStr to use `tostring` for the value which in a really obscure case could fail when a developer messed up a metatable's `__tostring` method to return `nil`
Example of this:
`lua_run local mt = { __tostring = function() return nil end } local ud = newproxy() debug.setmetatable(ud, mt) print(tostring(ud))`

### DIFF
--- a/garrysmod/lua/includes/modules/cookie.lua
+++ b/garrysmod/lua/includes/modules/cookie.lua
@@ -7,13 +7,22 @@ end
 
 module( "cookie", package.seeall )
 
-local CachedEntries = {}
-local BufferedWrites = {}
-local BufferedDeletes = {}
+local CachedEntries = {
+	--[[
+		key:
+			1 = number expireTime - if the SysTime exired then the cookie is fetched from SQL
+			2 = string cookieValue
+	]]
+}
+local BufferedQueue = {
+	--[[
+		key = value (false for delete else for insert it's a string value)
+	]]
+}
 
 local function GetCache( key )
 
-	if ( BufferedDeletes[ key ] ) then return nil end
+	if ( BufferedQueue[ key ] == false ) then return nil end
 
 	local entry = CachedEntries[ key ]
 
@@ -35,8 +44,7 @@ end
 local function FlushCache()
 
 	CachedEntries = {}
-	BufferedWrites = {}
-	BufferedDeletes = {}
+	BufferedQueue = {}
 
 end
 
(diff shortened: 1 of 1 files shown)


robotboy655 (changes requested):
Seems that the entire reworking of the queue is entirely an unnecessary change?

Image

This simple change should have the exact same effect.

Also the error you put in `SetCache` is inconsistent with other type errors, it should say `bad argument #2 to 'cookie.Set' ...` to be consistent with other type errors.

It can also be as simple as this:
```lua
-- It is unlikely that this could ever happen but with a invalid __tostring method tostring may silently fail
if ( tostring( value ) == nil ) then
	error( "bad argument #2 to 'value' (string expected, got " .. type( value ) .. ")", 2 )
end
```

RaphaelIT7:
> Seems that the entire reworking of the queue is entirely an unnecessary change?

I don't think keeping track of two different tables is necessary when one works just as well and simplifies it in my opinion.
This issue exists after all because, for setting a value it was forgotten to account for the second table.
**Though if requested I can change that back too.**

> It can also be as simple as this
> if ( tostring( value ) == nil ) then

For values like `true` I think it would change the behavior, as then the value stored would still be the original type, causing issues with the later assumption that all values are strings.
And if we were to use `isstring` then of course we would break the silent behavior of it accepting anything that was able to be turned into a string.

robotboy655:
> For values like true I think it would change the behavior, as then the value stored would still be the original type, causing issues with the later assumption that all values are strings.

Your proposed changes already cause this change in behavior, because you are now storing and then writing the tostring()'d value, not the raw value like it does currently. So I am not really sure what your point here is.

```lua
print("before setting:", cookie.GetString("test2"), type(cookie.GetString("test2")) )
cookie.Set("test2", true)
print("after setting:", cookie.GetString("test2"), type(cookie.GetString("test2")) )
```

The output of this without this PR:
```
before setting:	nil	nil
after setting:	true	boolean
```

The output with the PR:
```
before setting:	nil	nil
after setting:	true	string
```

Although you could argue that the latter is the correct-er behavior.

I am just not sure it's a good idea to touch things unnecessarily because time and time again this has proven to cause unforeseen breakages later on.

RaphaelIT7:
> Your proposed changes already cause this change in behavior

Yep soo that's something I did not notice/wasn't aware of changing xD

The question is- when we know that after a map switch / once it loads the value from SQL anyways- would it be an issue if we forced it to be a string even now beforehand?
After all, no addon should rely on GetString returning a non string and if they did, then probably after a map change, they would've thrown an error?

Personally I'd think this is an acceptable change as it prevents "magic" errors (after when it was loaded from SQL) when a developer unknowingly relied on that behavior.

PR #2486, opened 2026-06-21
Add `SELF` to `lua_run` globals

piqey:
As `ACTIVATOR` and `CALLER` are already present and resemble the incumbent entity I/O system, I believe `SELF` should be available within `lua_run` as well. For instance, it would be handy in situations that demand the location of the `lua_run` entity. Having access to the `lua_run` entity from within itself opens up more possibilities with entity I/O and Lua for mapping.


piqey:
I'm working on some other additions to `lua_run`, not sure if I should push them here or make a second pull request(?)

robotboy655:
Do not bundle extra changes to existing PRs.

PR #2491, opened 2026-06-30
Optimize favorite map checks and background scanning in main menu

InfernoLua1337:
### Description
This PR introduces performance optimizations to the main menu's core Lua logic, focusing on reducing garbage collection (GC) overhead during game startup and eliminating UI micro-freezes when opening the map selection screen.

### Changes
1. **`lua/menu/getmaps.lua`**
   * Extracted favorite map checks from the nested loop in `RefreshMaps` and built a single `fav_lookup` dictionary before starting the map iteration.
   * Replaced the expensive $O(N \cdot M)$ `table.HasValue` search with a fast $O(1)$ key lookup. This completely eliminates noticeable menu lag/micro-freezes for players who have large map collections combined with several favorited maps.
2. **`lua/menu/mainmenu.lua`**
   * Replaced the `RandomPairs` iterator inside `PANEL:ScreenshotScan` with `table.Shuffle` and a standard `ipairs` loop.
   * Since `RandomPairs` generates temporary table copies of keys and iterator closures, replacing it prevents unnecessary memory allocation during the game startup sequence when loading background images.

### DIFF
--- a/garrysmod/lua/menu/mainmenu.lua
+++ b/garrysmod/lua/menu/mainmenu.lua
@@ -62,21 +62,19 @@ function PANEL:Init()
 	end
 
 end
-
 function PANEL:ScreenshotScan( folder )
-
 	local bReturn = false
-
 	local Screenshots = file.Find( folder .. "*.*", "GAME" )
-	for k, v in RandomPairs( Screenshots ) do
 
+	-- Shuffle the table in-place to randomize backgrounds without generating table garbage
+	table.Shuffle( Screenshots )
+
+	for _, v in ipairs( Screenshots ) do
 		AddBackgroundImage( folder .. v )
 		bReturn = true
-
 	end
 
 	return bReturn
-
 end
 
 function PANEL:Paint()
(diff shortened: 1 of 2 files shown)


garrysmod/lua/menu/mainmenu.lua
```diff
@@ -62,21 +62,19 @@ function PANEL:Init()
...
 end
-
 function PANEL:ScreenshotScan( folder )
-
 	local bReturn = false
-
 	local Screenshots = file.Find( folder .. "*.*", "GAME" )
-	for k, v in RandomPairs( Screenshots ) do
 
+	-- Shuffle the table in-place to randomize backgrounds without generating table garbage
```
DarthTealc: Could probably skip shuffling here entirely, since the background table gets randomly selected from each time it picks an image to display https://github.com/Facepunch/garrysmod/blob/fa5c215ad0e96ac7627a0633e9cf5eee62ce407e/garrysmod/lua/menu/background.lua#L116

InfernoLua1337: Yeah, that makes sense, thanks! Updated the PR.

PR #2499, opened 2026-07-09
[TTT] Fix italic fonts on Linux

Tiagoquix:
This creates a new gloval variable that sets the font to DejaVu Sans instead of Tahoma if it detects that the system is Linux.

The current Tahoma definition makes the fallback font render as italic, somehow. This fixes the italic font problem in many places (scoreboard, HUD, C4 menu, weapon menu etc.).

### DIFF
--- a/garrysmod/gamemodes/terrortown/gamemode/shared.lua
+++ b/garrysmod/gamemodes/terrortown/gamemode/shared.lua
@@ -6,6 +6,14 @@ GM.Version = "shrug emoji"
 
 GM.Customized = false
 
+-- Font definiton
+GAMEMODE_DEFAULT_UI_FONT = "Tahoma"
+
+if ( system.IsLinux() ) then
+   GAMEMODE_DEFAULT_UI_FONT = "DejaVu Sans"
+end
+
+
 -- Round status consts
 ROUND_WAIT   = 1
 ROUND_PREP   = 2
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/cl_init.lua
@@ -282,7 +282,7 @@ vgui.Register( "DisarmPanel", PANEL, "DPanel" )
 
 
 surface.CreateFont("C4Timer", {
-                      font = "Tahoma",
+                      font = GAMEMODE_DEFAULT_UI_FONT,
                       size = 30,
                       weight = 750
                    })
--- a/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
+++ b/garrysmod/gamemodes/terrortown/entities/entities/ttt_c4/shared.lua
@@ -586,7 +586,7 @@ end
 
 if CLIENT then
    surface.CreateFont("C4ModelTimer", {
-                         font = "Tahoma",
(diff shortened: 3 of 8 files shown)


Tiagoquix:
@robotboy655 Here is a before and after:

# Before
image

# After
image-1

garrysmod/gamemodes/terrortown/gamemode/shared.lua
```diff
@@ -6,6 +6,14 @@ GM.Version = "shrug emoji"
 
 GM.Customized = false
 
+-- Font definiton
+GAMEMODE_DEFAULT_UI_FONT = "Tahoma"
+
+if ( system.IsLinux() ) then
+   GAMEMODE_DEFAULT_UI_FONT = "DejaVu Sans"
+end
```
Astralcircle, line 14: Don't need to create global for this

Tiagoquix: I am just following the format of the other definitions in the same file. Is there a problem creating a global for that?

razorzinho: > Don't need to create global for this

I mean… you either do it that way or you if else system.IsLinux in every instance of CreateFont.

Astralcircle (commented):
As i said better replace `font = GAMEMODE_DEFAULT_UI_FONT` with `font = system.IsLinux() and "DejaVu Sans" or "Tahoma"`

Tiagoquix:
> As i said better replace `font = GAMEMODE_DEFAULT_UI_FONT` with `font = system.IsLinux() and "DejaVu Sans" or "Tahoma"`

But then you would have to do that statement across all the files. I think that is worse. A global variable makes more sense, kind like of how Derma does: https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/derma/init.lua

Astralcircle:
But this global is useless after init

Tiagoquix:
Well, add-ons could be updated to use it. It may be "useless" now, but it would serve as a base for the future.

razorzinho:
> But this global is useless after init

I don’t understand. What’s wrong with our global approach? We centralized the font definition which is clean for a string that is used across 7 different files and more than 10 usages. What you’re suggesting does not improve the solution in performance nor organization. It will only make code redundant and harder to refactor if needed. 
It’s just one global with one single if statement versus TWO logic validations and 2 different string literals repeating across 7 files more than 10 times.

With all due respect, is that just your opinion or our approach goes against an undocumented standard?
If not, I see no reason to object the implementation.

EDIT: even if we declare a local variable for that in each of the 7 affected files, it’s still 7 different definitions versus one definition which can also be used by custom community content as mentioned by @Tiagoquix
I do know some addons use TTT’s font, and they just blatantly redeclare “Tahoma” as a global themselves in their addons. This way they get consistent font with the gamemode and Linux users won’t have to deal with the hassle

DarthTealc:
Having a global isn't going to hurt performance.

As a related idea, I've posted a request that would allow addons to create fonts based on existing fonts, similar to entities and weapons.  eg custom fonts specifying TTT fonts as their base fonts would inherit improved Linux updates from the TTT fonts - https://github.com/Facepunch/garrysmod-requests/issues/3282

PR #2500, opened 2026-07-09
Fix several base and sandbox logic errors

ShiroKSH:
Summary
- Preserve friendly NPC attacker flags in player death notices while requiring an allied disposition.
- Stop generic and editable entity spawning when entity creation fails.

Validation
- Checked current upstream and open pull requests for matching or overlapping fixes.
- Reviewed Lua 5.1 loop semantics and dropped two false-positive callback changes.
- Ran static invariants and a CRLF-aware whitespace check on the final three-file diff.

### DIFF
--- a/garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua
+++ b/garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua
@@ -356,7 +356,8 @@ end
 --
 function ENT:PlaySequenceAndWait( name, speed )
 
-	local len = self:SetSequence( name )
+	self:SetSequence( name )
+	local len = self:SequenceDuration()
 	speed = speed or 1
 
 	self:ResetSequenceInfo()
--- a/garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua
+++ b/garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua
@@ -356,8 +356,7 @@ end
 --
 function ENT:PlaySequenceAndWait( name, speed )
 
-	self:SetSequence( name )
-	local len = self:SequenceDuration()
+	local len = self:SetSequence( name )
 	speed = speed or 1
 
 	self:ResetSequenceInfo()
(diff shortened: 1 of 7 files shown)


garrysmod/gamemodes/sandbox/gamemode/save_load.lua
```diff
@@ -119,7 +119,7 @@ else
 			return
 		end
 
-		engine.WriteSave( buffer, game.GetMap() .. " " .. util.DateStamp() )
+		engine.WriteSave( uncompressed, game.GetMap() .. " " .. util.DateStamp() )
```
robotboy655: This is not correct. The data going to `engine.WriteSave` must be compressed, the uncompress call is simply a test if the data is intact after receiving it over the network.

robotboy655 (changes requested):

garrysmod/gamemodes/sandbox/gamemode/save_load.lua
ShiroKSH: Okay, no problem. Reverted it.

garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua
```diff
@@ -356,7 +356,8 @@ end
 --
 function ENT:PlaySequenceAndWait( name, speed )
 
-	local len = self:SetSequence( name )
+	self:SetSequence( name )
+	local len = self:SequenceDuration()
```
robotboy655: This is also wrong, there's a previously undocumented behavior where `SetSequence` returns the sequence duration.

robotboy655 (changes requested):

garrysmod/gamemodes/base/entities/entities/base_nextbot/sv_nextbot.lua
ShiroKSH: > This is also wrong, there's a previously undocumented behavior where `SetSequence` returns the sequence duration.

Okay, reverted.

robotboy655 (approved):

PR #2503, opened 2026-07-12
Edict Limit

Vitroze:
Hello,
I only added checks to verify if the entity actually exists before a Lua error occurs.

I also re-indented a piece of code that lacked consistent indentation and removed an unused variable.
I ensured that the effects appear before the explosion is created.

### DIFF
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
@@ -115,6 +115,7 @@ if ( SERVER ) then
 		local lightInfo = self:GetLightInfo()
 
 		self.flashlight = ents.Create( "env_projectedtexture" )
+		if ( !IsValid( self.flashlight ) ) then self.flashlight = nil return end
 		self.flashlight:SetParent( self )
 
 		-- The local positions are the offsets from parent..
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
@@ -116,16 +116,30 @@ if ( SERVER ) then
 
 		self.flashlight = ents.Create( "env_projectedtexture" )
 		if ( !IsValid( self.flashlight ) ) then
-
 			local ply = self:GetPlayer()
 			if ( IsValid( ply ) ) then
 				ply:PrintMessage( HUD_PRINTTALK, "Failed to create flashlight entity." )
 			end
 
-			ErrorNoHalt( "Failed to create flashlight entity for " .. self:GetClass() .. " (" .. self:EntIndex() .. ")!\n" )
 			self.flashlight = nil
+
+			-- Turn off, since that's what we look like now
+			-- And play a cute effect
+			timer.Simple( 0, function()
+				self:SetOn( false )
+
+				local effectdata = EffectData()
+				effectdata:SetOrigin( self:GetPos() )
+				effectdata:SetNormal( self:GetForward() )
+				effectdata:SetMagnitude( 1 )
+				effectdata:SetScale( 1 )
+				effectdata:SetRadius( 16 )
+				util.Effect( "Sparks", effectdata, false, true )
+			end )
+
 			return
 		end
+
 		self.flashlight:SetParent( self )
 
 		-- The local positions are the offsets from parent..
--- a/garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
+++ b/garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
@@ -115,7 +115,17 @@ if ( SERVER ) then
 		local lightInfo = self:GetLightInfo()
 
 		self.flashlight = ents.Create( "env_projectedtexture" )
-		if ( !IsValid( self.flashlight ) ) then self.flashlight = nil return end
+		if ( !IsValid( self.flashlight ) ) then
+
+			local ply = self:GetPlayer()
+			if ( IsValid( ply ) ) then
+				ply:PrintMessage( HUD_PRINTTALK, "Failed to create flashlight entity." )
+			end
+
+			ErrorNoHalt( "Failed to create flashlight entity for " .. self:GetClass() .. " (" .. self:EntIndex() .. ")!\n" )
+			self.flashlight = nil
+			return
+		end
 		self.flashlight:SetParent( self )
 
 		-- The local positions are the offsets from parent..
--- a/garrysmod/lua/autorun/properties/bone_manipulate.lua
+++ b/garrysmod/lua/autorun/properties/bone_manipulate.lua
@@ -36,7 +36,11 @@ properties.Add( "bone_manipulate", {
 		if ( !self:Filter( ent, ply ) ) then return end
 
 		ent.widget = ents.Create( "widget_bones" )
-		if !IsValid(ent.widget) then ent.widget = nil return end
+		if ( !IsValid( ent.widget ) ) then
+			if ( IsValid( ply ) ) then ply:PrintMessage( HUD_PRINTTALK, "Failed to create bone manipulation widget." ) end
+			ent.widget = nil
+			return
+		end
 
 		ent.widget:Setup( ent )
 		ent.widget:Spawn()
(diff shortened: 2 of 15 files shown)


garrysmod/lua/includes/modules/constraint.lua
```diff
@@ -581,6 +584,8 @@ function Elastic( Ent1, Ent2, Bone1, Bone2, LPos1, LPos2, constant, damping, rda
 		onStartConstraint( Ent1, Ent2 )
 
 			Constraint = ents.Create( "phys_spring" )
+			if ( !IsValid(Constraint) ) then return NULL, NULL end
+
```
robotboy655: Why are some changes/additions not using consistent style within at least the same file?

garrysmod/lua/includes/modules/constraint.lua
```diff
@@ -1078,6 +1090,8 @@ function Motor( Ent1, Ent2, Bone1, Bone2, LPos1, LPos2, friction, torque, forcet
 	onStartConstraint( Ent1, Ent2 )
 
 		local Constraint = ents.Create( "phys_torque" )
+		if ( !IsValid(Constraint) ) then return NULL end
+
```
robotboy655: All of these early returns mean that `onFinishConstraint` is never called in all these cases.

For certain cases like here with the Motor constraint, this means that you are basically not only "leaking" entities here (`axis` variable in this case), you are also leaving half-creating bogus constrains in place.

While before it would just error and leave the same mess, if we are to try to handle these edict limit errors nicely, we need to make an effort of actually handling them nicely, not just suppressing Lua errors and leaving a mess behind.

The return value you chose is also inconsistent with other failure returns in these fuctions, which is `false`, not `NULL`.

garrysmod/lua/entities/widget_bones.lua
```diff
@@ -153,6 +153,8 @@ function ENT:Setup( ent )
 		if ( !ent:BoneHasFlag( k, BONE_USED_BY_VERTEX_LOD0 ) ) then continue end
 
 		local btn = ents.Create( "widget_bone" )
+		if ( !IsValid( btn ) ) then continue end
+
```
robotboy655: Again, leaves the half created state. If you want to do it properly, it probably should cleanup all previously created bone widgets and return a failure so the code calling Setup() can detect the failure and let the user know that it has, in fact, failed.

garrysmod/lua/entities/widget_axis.lua
```diff
@@ -88,6 +92,8 @@ function ENT:Setup( ent, boneid, rotate )
 	self.ArrowY:SetAxisIndex( 2 )
 
 	self.ArrowZ = ents.Create( EntName )
+	if ( !IsValid( self.ArrowZ ) ) then return end
+
```
robotboy655: Again, each of these failures leaves a half created mess, which is now silent (bad).

garrysmod/lua/autorun/properties/kinect_controller.lua
```diff
@@ -64,6 +64,7 @@ properties.Add( "motioncontrol_ragdoll", {
 		end
 
 		local ragdoll_motion = ents.Create( "ragdoll_motion" )
+		if ( !IsValid( ragdoll_motion ) ) then return end
 		ragdoll_motion:SetPos( player:EyePos() + player:EyeAngles():Forward() * 10 )
```
robotboy655: Should probably tell the user why nothing appeared, like it does above.

garrysmod/lua/autorun/properties/bone_manipulate.lua
```diff
@@ -36,6 +36,8 @@ properties.Add( "bone_manipulate", {
 		if ( !self:Filter( ent, ply ) ) then return end
 
 		ent.widget = ents.Create( "widget_bones" )
+		if !IsValid(ent.widget) then ent.widget = nil return end
+
```
robotboy655: Again, code style, half created silent mess.

garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
```diff
@@ -115,6 +115,7 @@ if ( SERVER ) then
 		local lightInfo = self:GetLightInfo()
 
 		self.flashlight = ents.Create( "env_projectedtexture" )
+		if ( !IsValid( self.flashlight ) ) then self.flashlight = nil return end
```
robotboy655: This again, creates a silent issue of the flashlight not working. Yeah, you "fixed" the Lua error by hiding it, but the user sees a lamp not working without any feedback now.

Honestly not sure what the best approach here would be.

robotboy655 (changes requested):
How many times do I have to say to not include unnecessary whitespace changes in your PRs and to not mix and match TTT and non TTT changes? And to keep coding style consistent?

Please go to https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md and actually read it.

Code style, including the indentations, is up to the author of TTT to choose for that project. Which is the reason for not mixing TTT changes with others.

I have not looked through how sensible are the TTT NULL entity error suppressions are, but given the issues I found with the base game stuff, I imagine no thought was given to those cases either.

I am confident these changes are inspired by this commit: https://github.com/Facepunch/garrysmod/pull/2500/changes
The difference there is that the function that calls `SpawnFunction` already handles NULL entity returns decently enough (although it could definitely be improved there as well) and these functions are not doing anything else except spawn entities.

Vitroze:
> How many times do I have to say to not include unnecessary whitespace changes in your PRs and to not mix and match TTT and non TTT changes? And to keep coding style consistent?
> 
> Please go to https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md and actually read it.
> 
> Code style, including the indentations, is up to the author of TTT to choose for that project. Which is the reason for not mixing TTT changes with others.

Oh, I'm sorry—I was sure I'd followed the style. I admit I didn't pay close enough attention to the TTT game mode at the time. I'll look into your requested changes.

Vitroze:
> I am confident these changes are inspired by this commit: https://github.com/Facepunch/garrysmod/pull/2500/changes The difference there is that the function that calls `SpawnFunction` already handles NULL entity returns decently enough (although it could definitely be improved there as well) and these functions are not doing anything else except spawn entities.

I don't recall if it was this PR, another one, or simply something I noticed when looking at the Garry's Mod code seeing that errors were prevented in some cases but not others, that led me to assume it was an oversight (aside from the creation of base entities).

garrysmod/lua/includes/modules/constraint.lua
Vitroze: I noticed that the return statement involved two values, so I decided to return two null values; however, the second value isn't actually needed, given that we'll always be checking the first one.

garrysmod/lua/includes/modules/constraint.lua
Vitroze: Indeed, I hadn't noticed what onStartConstraint and onFinishConstraint were doing. That has been fixed.

garrysmod/lua/entities/widget_bones.lua
Vitroze: I have to admit I'm not convinced by this method, but the problem is fixed (https://github.com/Facepunch/garrysmod/pull/2503/commits/907e68fcbb0d256a66706efff43a68a5d9d3f0d1)

garrysmod/lua/entities/widget_axis.lua
Vitroze: Good (https://github.com/Facepunch/garrysmod/pull/2503/changes/a7c8402b67ea9be9031feeadfc09768af9684d72)

garrysmod/lua/autorun/properties/kinect_controller.lua
Vitroze: Good (https://github.com/Facepunch/garrysmod/pull/2503/changes/7c2b70f3bb1db121795d459d5c30d7fa6c5fe273)

garrysmod/lua/autorun/properties/bone_manipulate.lua
Vitroze: Good (https://github.com/Facepunch/garrysmod/pull/2503/changes/66e99725709b4836392bb690296b8220b046b432)

garrysmod/gamemodes/sandbox/entities/entities/gmod_lamp.lua
Vitroze: I made sure it triggers a Lua error and a visual message for the player.

Vitroze:
This evening, I’m going to separate the TTT and this PR to make things easier for you.

robotboy655:
I guess it's up to me to clean up the mess.. Again.

robotboy655:
I have looked through and merged some of these changes to `dev`, I will finish the rest next time, probably next week.

robotboy655 (approved):

PR #2504, opened 2026-07-13
Fix IsValid error when passed a boolean

03fj:
Prevent `IsValid` from erroring on invalid input types.

### DIFF
--- a/garrysmod/lua/includes/util.lua
+++ b/garrysmod/lua/includes/util.lua
@@ -315,8 +315,11 @@ function IsValid( object )
 
 	if ( !object ) then return false end
 
-	local isvalid = object.IsValid
-	if ( !isvalid ) then return false end
+	local ok, isvalid = pcall( function()
+		return object.IsValid
+	end )
+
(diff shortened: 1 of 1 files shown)


Astralcircle:
Will slow down IsValid a lot, as it can be called hundreds of thousands of times every frame, so just don't use it on invalid types

robotboy655:
> so just don't use it on invalid types

I mean, its kind of counter intuitive to the function's name, but I agree, you shouldn't be using `IsValid` on booleans. It already handles `false` with `if ( !object ) then return false end`.

The proposed implementation just to test for `true` is absurd in my mind. `if ( object == true ) then return true end` would be a lot more sane, but again, you shouldn't be using `IsValid` on variables you know to be booleans. That's not really what its for. It's for custom objects that have `IsValid` methods, but it also tests for `nil` (and `false` I guess) just so you can have `if IsValid( obj ) then ` instead of having to do `if obj and obj:IsValid() then` everywhere.

PR #2505, opened 2026-07-14
Add missing soundscripts for doors and L4D world

goodusername123:
Should fix quite a bit of the missing sounds present when playing on L4D maps (example: `CSoundEmitterSystemBase::GetParametersForSound:  No such sound Weather.thunder_far_away_1`). should also presumably fix some missing door sounds on L4D(2) and CS:GO maps (I've seen a player report `CSoundEmitterSystemBase::GetParametersForSound: No such sound Doors.Wood.FullOpen1` before).

doors soundscript seemingly only exists in both L4D(2) and CS:GO and differs slightly between them (one has some sounds that the other doesn't and vice versa) so for now it's added to the non-prefixed section in the manifest.
This PR does not contain sounds exclusive to L4D2, only contains sounds present in L4D1 (same as the pre-existing L4D weapons soundscript).

### DIFF
--- a/garrysmod/scripts/game_sounds_manifest.txt
+++ b/garrysmod/scripts/game_sounds_manifest.txt
@@ -173,6 +173,7 @@ game_sounds_manifest
 
 	// Left 4 Dead
 	"precache_file"		"scripts/sounds/l4d_game_sounds_weapons.txt"
+	"precache_file"		"scripts/sounds/l4d_game_sounds_world.txt"
 
 	// Stanley Parable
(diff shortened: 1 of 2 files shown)


goodusername123:
I should probably mention that `l4d_game_sounds_world.txt` in this PR is based on a combination of `game_sounds_world.txt`, `game_sounds_world_dlc3.txt`, and `game_sounds_world_dlc4.txt` from within L4D1, with default HL2/CSS entries removed/excluded.

PR #2508, opened 2026-07-19
Optimize `Vector:ToColor`

Astralcircle:
1 C call vs 3 C calls

Should give a good boost


Bagellll:
At this point you may as well replace `self:Unpack` with `meta.Unpack(self)` too.

Astralcircle:
> At this point you may as well replace `self:Unpack` with `meta.Unpack(self)` too.

I thinked about that but it's kinda inconsistent and i don't think rubat will like it

robotboy655 (approved):

PR #2509, opened 2026-07-21
Added language system for achievements

Vitroze:
Hello,
I noticed that the achievements aren't translated and can't be, so I added that.


CornerPin:
Shouldn't have spaces around strings in localization files, instead add them where the string is used.

Vitroze:
> Shouldn't have spaces around strings in localization files, instead add them where the string is used.

Good

NovaDiablox:
I would prefer control over variables otherwise its a headache.
Example: "%p earned the achievement %a"

CornerPin:
> I would prefer control over variables otherwise its a headache. Example: "%p earned the achievement %a"

Oh yeah, that would be better and would also make it easy to localize for languages where the word order might be different.

Vitroze:
> I would prefer control over variables otherwise its a headache.
> Example: "%p earned the achievement %a"

I didn't do it at first because of the color system, and honestly, I didn't see how to do it cleanly while maintaining the same result.

CornerPin:
Perhaps something akin to `string.Interpolate`?

```lua
local function colorize( str, lookup )
    local elements = {}
    local chunkStart = 1

    for from, key, to in string.gmatch( str, "(){([_%a][_%w]*)}()" ) do
        local value = lookup[key]

        if ( value != nil ) then
            if ( chunkStart < from ) then
                elements[#elements + 1] = string.sub( str, chunkStart, from - 1 )
            end

            elements[#elements + 1] = value
            chunkStart = to
        end
    end

    if ( chunkStart - 1 < #str ) then
        elements[#elements + 1] = string.sub( str, chunkStart )
    end

    return unpack( elements )
end

language.Add("achievement.unlocked", "{player}{normal_color}が実績「{achievement_color}{achievement}{normal_color}」を獲得しました")

concommand.Add("test_colorize", function()
    chat.AddText(colorize(language.GetPhrase("achievement.unlocked"), {
        player = LocalPlayer(),
        achievement = achievements.GetName(15),
        achievement_color = Color( 255, 200, 0 ),
        normal_color = Color( 230, 230, 230 )
    }))
end)
```

image

Vitroze:
> Perhaps something akin to `string.Interpolate`?
> 
> ```lua
> local function colorize(str, lookup)
>     local elements = {}
>     local chunk_start = 1
> 
>     for from, key, to in string.gmatch(str, "(){([_%a][_%w]*)}()") do
>         local value = lookup[key]
> 
>         if value ~= nil then
>             if chunk_start < from then
>                 elements[#elements + 1] = string.sub(str, chunk_start, from - 1)
>             end
> 
>             elements[#elements + 1] = value
>             chunk_start = to
>         else
>             elements[#elements + 1] = "{" .. key .. "}"
>         end
>     end
> 
>     if chunk_start - 1 < #str then
>         elements[#elements + 1] = string.sub(str, chunk_start)
>     end
> 
>     return unpack(elements)
> end
> 
> language.Add("achievement.unlocked", "{player}{normal_color}が実績「{achievement_color}{achievement}{normal_color}」を獲得しました")
> 
> concommand.Add("test_colorize", function()
>     chat.AddText(colorize(language.GetPhrase("achievement.unlocked"), {
>         player = LocalPlayer(),
>         achievement = achievements.GetName(15),
>         achievement_color = Color( 255, 200, 0 ),
>         normal_color = Color( 230, 230, 230 )
>     }))
> end)
> ```
> 
> image

Fair enough, but adding all that code just for a color feature seems a bit excessive to me.

Vitroze:
I added a system to apply custom colors and correctly translate the phrase. I think this is the most optimized approach, but I'm open to any suggestions.

PR #2510, opened 2026-07-21
Removal of unnecessary multiplications, modification of the "style"

Vitroze:
Removal of unnecessary multiplications (Unnecessary if you multiply by 1)
Transformation of 
- `&&` to `and`, 
- `||` to `or`, 
- `not` to `!`

### DIFF
--- a/garrysmod/gamemodes/base/gamemode/cl_init.lua
+++ b/garrysmod/gamemodes/base/gamemode/cl_init.lua
@@ -200,7 +200,7 @@ function GM:OnChatTab( str )
 
 		local nickname = v:Nick()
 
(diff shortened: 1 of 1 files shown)


robotboy655:
The multiplication is there to make a copy of the object.

Do not submit PRs that modify coding style. See https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md

Vitroze:
> The multiplication is there to make a copy of the object.
> 
> Do not submit PRs that modify coding style. See https://github.com/Facepunch/garrysmod/blob/master/CONTRIBUTING.md

Yes, I realized that; that’s why I reverted the previous change. You can't just merge changes to create a consistent style (adding spaces, using only 'and' or 'or', etc.). Does that actually result in better readability?

PR #2514, opened 2026-07-24
Fix Typos In Source Strings For Consistency

NovaDiablox:
Initially I was going to change "gamemode" to "game mode" too but in the end I concluded that players may not like this and may complain. I would like to hear your thoughts. Some may argue that the changes made are game specific terms. Such as "spawnmenu" & "spawnlist" but I consider them as typos. "spawnmenu" has been not consistent in the game. We can see that in hints.properties its "SPAWN MENU". Other examples include but not limited to: spawnmenu.lua, dtextentry.lua spawnmenu.properties. As for "Tool Gun" I think its pretty clear. I only mentioned things that are important in the description, please provide feedback accordingly.

I might change "disk drive" to "storage device". Since SSDs don't have "disks" inside but the term stuck for so long it sounds natural.


robotboy655:
I really don't think these changes, especially from the last commit, are necessary.

For the last commit, these strings can be reused by addons in a non checkbox context.
They might also be reused by the game  itself in a dropdown menu, the one that appears above spawnmenu/context menu.

There's just literally no reason to change those, it's just asking for trouble, or changing them for the sake of changing them, not for an actual improvement.

robotboy655 (approved):

PR #2517, opened 2026-07-30
Added a Lua error when the "Model Detail" setting is set to Low.

Vitroze:
Hello,
Many players unfamiliar with Garry's Mod (especially new ones) complain about display issues when the model rendering setting is set to "Low." It would be a good idea to implement an error notification for this, similar to how it is already handled for HDR.

**I set it to "0" by default because some of the most popular Workshop vehicles absolutely require the "Model Detail" setting to be on "High."**

# Display issue [(Example)](https://steamcommunity.com/workshop/filedetails/?id=120766823)
<details>
image
</details>

<details>
image
</details>

### DIFF
--- a/garrysmod/lua/menu/problems/problems.lua
+++ b/garrysmod/lua/menu/problems/problems.lua
@@ -206,6 +206,12 @@ end
 
 timer.Create( "menu_check_for_problems", 1, 0, function()
 
+	if ( math.floor( GetConVarNumber( "r_rootlod" ) ) == 2 ) then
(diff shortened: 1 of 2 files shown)


robotboy655:
Low model quality is not an error. 

To be clear, this is a model authoring issue, not a game or user issue.

Vitroze:
> Low model quality is not an error.

It can be reclassified as severity 0, like multi-core, which is not a problem in itself.

PR #2518, opened 2026-07-30
Disable leafblower by default

Astralcircle:
This tool is hidden and seems far from finished? Players can still use it via `gmod_tool leafblower`, but i don't think it was ever intended for use at all?


robotboy655:
It's just a hidden feature, I don't know why it exists, most likely as an example, so there's no need to just nuke it.

tschumann:
It was added because JBMod had a leaf blower, but when it was added to
Garry's Mod it was hidden for some reason (as a joke I think).

PR #2522, opened 2026-08-07
Prevent negative player count in server list

DarthTealc:
### Summary
If a server is faking data to give a negative player count, don't add the server to the server list.

### Background
There are servers intentionally messing with the server list's gamemode player count. They are broadcasting that they have 127 slots, with 0 players, but 255 bots. Gmod interprets this as -128 human players currently connected. There are 20 identical servers doing this in DarkRP which means DarkRP starts at -2560 players. 
Actual player counts bring it back up closer to zero but it is still negative.
image
In the case of this screenshot there are 1229 players, minus the -2560 players = a listing of -1331.
See https://discord.com/channels/565105920414318602/565108080300261398/1535122344803303515 for more information.

### Solution
This PR looks at the human player count when adding a server. If it is below zero, the `AddServer` function is aborted for that server. The server doesn't get added to the gamemode's total player count and therefore can't add negative playercount.

### DIFF
--- a/garrysmod/html/js/menu/control.Servers.js
+++ b/garrysmod/html/js/menu/control.Servers.js
@@ -504,6 +504,8 @@ function AddServer( type, id, ping, name, desc, map, players, maxplayers, botpla
 		favorite:		isFav == "true"
 	};
 
+	if ( data.players < 0 || data.maxplayers < 1 ) return;
+
 	if ( data.flag == "eu" ) data.flag = "europeanunion";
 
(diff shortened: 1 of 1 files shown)


robotboy655:
Not a good idea to outright hide those servers, and I also implemented this in C++ alongside the other limits.

DarthTealc:
The servers were already hidden because the player count was -128 which is way outside the range of the default playercount filter
image

PR #2524, opened 2026-08-10
Make VehicleMemDupe entity modifier more strict

Astralcircle:
Currently it can modify any part of the table of any entity, even if it's not a vehicle
I added checks to only do this for vehicles and not rely on information from the modifier (which could be harmful)

To completely prevent this, you also need to merge https://github.com/Facepunch/garrysmod/pull/2265


robotboy655:
I think the entire `VehicleMemDupe` thing can just be deleted.

As far as I can tell, its meant to store the "HandleAnimation" field, but it's a function so it cannot even be stored in duplications. Maybe it worked back then for "live" dupes that are not loaded from a file but simply copied and immediately pasted.

Anyway, the custom animations post duplication work just fine without this entity modifier, so it should be fine to just remove it.

robotboy655:
I have removed it entirely on `dev`, we will see if that causes any issues. It shouldn't though.

PR #2526, opened 2026-08-14
Fix: Rounding issue with the GetTextValue function and the Player Editor

Vitroze:
Hello,
This PR fixes the issue where the bodygroup changes when close to the next bodygroup. Currently, the system rounds up, which doesn't seem very logical to me; it would be better if the system simply dropped the decimal part.

I modified the `PANEL:GetTextValue` function of the `DNumberScratch` vgui element and the `UpdateBodyGroups` local function of the Player Editor.

I chose not to modify the value received from `PANEL:OnValueChanged` directly at the source, this was done for compatibility reasons and to maintain the logic required by certain addons that need to know the value's precise position on the slider.

# Result
## Before 
<details>

image

</details>

## After
<details>

image

</details>

### DIFF
--- a/garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
+++ b/garrysmod/gamemodes/sandbox/gamemode/editor_player.lua
@@ -136,17 +136,17 @@ list.Set( "DesktopWindows", "PlayerEditor", {
 		local function UpdateBodyGroups( pnl, val )
 			if ( pnl.type == "bgroup" ) then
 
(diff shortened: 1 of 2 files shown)


robotboy655:
What are the changes to `DNumberScratch` for? In my tests it works correctly as is?
image

Vitroze:
> What are the changes to `DNumberScratch` for? In my tests it works correctly as is? image

I just tested it, and it does indeed work. It wasn't working during my tests, likely due to an addons. I'm going to push a commit to revert this change.

robotboy655 (approved):

PR #2530, opened 2026-08-23
Change console font size to 12px

darkfated:
### 1920x1080

gmod_841L79loKG

### 1280x720

Zed_zqdlzsjXeC

### DIFF
--- a/garrysmod/resource/SourceScheme.res
+++ b/garrysmod/resource/SourceScheme.res
@@ -487,7 +487,7 @@ Scheme
 			{
 				"name"		"Lucida Console" [$WINDOWS]
 				"name"		"Verdana" [!$WINDOWS]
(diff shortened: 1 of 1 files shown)


robotboy655:
And this is meant to achieve what?

darkfated:
> And this is meant to achieve what?

For better readability. A lot of people I know use 14px also, but 12 is a balance.

## Default

{0149723F-6719-4372-A869-64B5F368647C}

DarthTealc:
This would result in less lines visible, and less text in a line causing wrapping. If the console text size was to be increased it should be optional or connected to DPI.

robotboy655:
The console is perfectly readable as is. You can apply your desired changes to your own game client if you so desire.

PR #2545, opened 2026-09-13
Validate title and model to player_manager.AddValidModel calls

EricZhang456:
Currently the [John Cena player model](https://steamcommunity.com/sharedfiles/filedetails/?id=546377707) will pass a boolean value to the title parameter of `player_manager.AddValidModel` and will break the player model selector later on when it tries to use `language.GetPhrase` on the boolean value. This patch adds validation to the title parameter as well as the model parameter.


robotboy655 (approved):
Whoops, I meant it to check `title` and `category` parameters, but I guess checking the others is also a good idea.

